PR #2, opened 2016-03-17
Fix import path

kchodorow:
Fixes CI, see http://ci.bazel.io/job/rules_go/1/BAZEL_VERSION=HEAD,PLATFORM_NAME=linux-x86_64/console.


kchodorow:
Also: any opinions on who should own the go rules?

damienmg:
LGTM

PR #9, opened 2016-03-29
This change makes go_library_impl include a test's library's transitive dependencies.

carednd:
Before, if you had //:lib depending on //:lib2, a test with library = "//:lib" would not pull in "//:lib2" for linking.


damienmg:
Sorry if you don't sign the CLA we cannot accept your contribution. Closing.

PR #10, opened 2016-04-03
cgo support

yugui:
Adds a new macro `cgo_library`, which wraps build steps with cgo.
See the following thread in bazel-dev for the overall design.

* https://groups.google.com/d/msg/bazel-dev/xFcxQyaxQec/YqIYjqA2CAAJ

Supported:
* Both Linux and Darwin
* Import types and functions from C/C++ to Go sources
* Export Go funtions to C/C++ sources within a package
* Add dependency to `cc_library` rules
* Build a package which contains both cgo-enabled sources and pure-Go
  sources.

Unsupported:
* Export Go packages to C binaries, a.k.a `-buildmode=c-archive` or
  `-buildmode=c-shared`.
* Directly or indirectly depend on prebuilt dynamic libraries on Darwin.
  i.e. `cgo_library` cannot depend on cc_library rules which contain *.so
  files in their srcs.
* Something like `cc_binary.linkstatic=0`, which is most similar to the
  default behavior of "go build".
  Instead, go rules always try to mostly statically link dependencies
  unless the dependency has only *.so.


go/def.bzl
```diff
@@ -385,6 +430,127 @@ go_test = rule(
...
+          "go_tool_dir=$$(cd $$(dirname $(location {go_tool})) && pwd)",
+          "go_tool=$$go_tool_dir/$$(basename $(location {go_tool}))", 
+          "objdir=$$(cd $(@D)/{out_dir} && pwd)",
+          "export GOROOT=$$go_tool_dir/..",
+          "export CGO_LDFLAGS='" + " ".join(linkopts) + "'",
+          "mkdir -p $$objdir",
+          "cd {pkgname}",
+          ("$$go_tool tool cgo -objdir $$objdir " + " ".join(go_srcs)), 
+          "rm -f $$objdir/_cgo_.o $$objdir/_cgo_flags",
+      ]).format(go_tool=go_tool, out_dir=out_dir, pkgname=PACKAGE_NAME),
```
hanwen: pkgname=pkgname, no? 

PACKAGE_NAME is the directory leading up to the BUILD file.

go/def.bzl
```diff
@@ -385,6 +430,127 @@ go_test = rule(
...
+  )
+  native.genrule(
+      name = "%s.cgo.importgen" % name,
+      srcs = ["%s.cgo.dir/_cgo_.o" % name],
+      outs = ["%s.cgo.dir/_cgo_import.go" % name],
+      cmd = " && ".join([
+          "export GOROOT=$$(dirname $(location {go_tool}))/..",
+          ("$(location {go_tool}) tool cgo -dynout $@" +
+           " -dynimport $<" +
+           " -dynpackage " + pkgname),
```
hanwen: ah, it's this one.

yugui:
@hanwen, could you take another look?

I believe I have almost completed my implementation. Now I have two remaining tasks:
* Hard-coed path to ld(1). I should read the configuration in CROSSTOOL.
* How should I document the new macro?  I have added a Skydoc string, but should I document it in `README.md` for consistency to other rules?

go/def.bzl
```diff
@@ -385,6 +452,371 @@ go_test = rule(
...
+)
+
+def _cgo_codegen(name, srcs, c_hdrs=[], deps=[], linkopts=[],
+                 go_tool=None, toolchain=None):
+  """Generates glue codes for interop between C and Go
+
+  Args:
+    name: A unique name of the rule
+    srcs: list of Go source files.
+
```
hanwen: nit: drop blank lines here.

go/tools/extract_package/extract.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func main() {
+	args := os.Args
+	if len(args) != 2 {
+		log.Fatalln("Usage: extract_package GO_FILE")
+	}
+	name, err := extract(args[1])
+	if err != nil {
+		log.Fatalln(err)
```
hanwen: (log.Fatal is the same and 2 characters less to type :)

go/tools/extract_package/extract_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"testing"
+)
+
+func TestExtract(t *testing.T) {
+	for _, fname := range []string{
+		"extract.go",
+		"extract_test.go",
+		"ignored.go",
+	} {
+		name, err := extract(fname)
```
hanwen: this is a bit iffy, because it depends on the cwd in which the test executes. This should probably use runfiles if it is to run in bazel. You can work around this by inlining the input source, and writing it to a ioutil.TempFile and running on that.

go/tools/extract_package/extract_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+)
+
+func TestExtract(t *testing.T) {
+	for _, fname := range []string{
+		"extract.go",
+		"extract_test.go",
+		"ignored.go",
```
hanwen: what's the function of igore.go ?

hanwen:
David created a new tool for documenting Skylark, he'll be able to explain
how it works. The README predates that, so I'm not sure what should happen.

I looked your code over, and it looks good to me.

Many, many thanks for taking this on! I've been daunted by this ever since
I put together the rules.


On Sun, Apr 10, 2016 at 9:27 AM, Yuki Yugui Sonoda <notifications@github.com
> wrote:

> @hanwen <https://github.com/hanwen>, could you take another look?
>
> I believe I have almost completed my implementation. Now I have two
> remaining tasks:
>
>    - Hard-coed path to ld(1). I should read the configuration in
>    CROSSTOOL.
>    - How should I document the new macro? I have added a Skydoc string,
>    but should I document it in README.md for consistency to other rules?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/bazelbuild/rules_go/pull/10#issuecomment-207939428>
>



-- 
Han-Wen Nienhuys
Google Munich
hanwen@google.com

go/tools/extract_package/extract_test.go
yugui: I tried to say that the function can extract a package name even if the file contains build tags before the `package` clause.  It didn't need to be `+build ignored` but it just came to my mind first.

examples/cgo/cc_dependency/BUILD
```diff
@@ -0,0 +1,36 @@
...
+    name = "darwin",
+    values = {"cpu": "darwin"},
+)
+
+cc_library(
+    name = "version",
+    srcs = ["cxx_version.cc"],
+    hdrs = ["version.h"],
+    # TODO(yugui) Support darwin too and remove this workaround.
+    # See also comments in cxx_version.cc.
```
damienmg: you can use [@bazel_tools//tools/cpp:osx_cc_wrapper.sh](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/osx_cc_wrapper.sh) for wrapping calls to gcc for the link phase and it will replace the -Wl,-rpath command with the correct calls to otool.

yugui: No, as I described in `cxx_version.cc`, `osx_cc_wrapper.sh` is the source of the problem here.

Without this `select`, bazel would add several instances of `-Wl,-rpath` into the command line. But `osx_cc_wrapper.sh` expects only one `-Wl,-rpath`.  It is the problem.

go/tools/extract_package/extract_test.go
yugui: done with the latter approach with `ioutil.TempFile`.

go/tools/extract_package/extract.go
yugui: done.

go/def.bzl

examples/cgo/cc_dependency/BUILD
damienmg: Oh I see we should fix osx_cc_wrapper.sh then.

yugui:
@hanwen I have reflected your comments, and resolved the first TODO.

I added several new commits on top of the first commit for ease of code review, but it might be better to squash them once I get LGTMs.

damienmg:
By the way skydoc is what we will use in the future but it is not yet integrated, pinging @davidzchen for more comment on it.

examples/cgo/cc_dependency/BUILD
damienmg: There is actually an issue in fixing it on OSX: you don't see a rpath option, you said where to find the library. I guess I need to spend more than 5 minutes to make it work because we have to figure out which is the good rpath for every passed library.

/cc @lberki that's another hell in our CROSSTOOL file.

go/def.bzl
```diff
@@ -643,18 +658,18 @@ _cgo_import = rule(
...
   sample_go_src: A go source which is compiled together with the generated file.
-
     The generated file will have the same Go package name as this file.
   out: Destination of the generated codes.
 """
 
 def _cgo_object_impl(ctx):
-  arguments = ctx.fragments.cpp.c_options + [
+  arguments = _c_linker_options(ctx, blacklist=[
+      # never link any dendency libraries
```
hanwen: typo. dependency

go/tools/extract_package/extract_test.go
```diff
@@ -1,21 +1,81 @@
...
+
+package example
+`,
+			want: "example",
+		},
 	} {
-		name, err := extract(fname)
-		if err != nil {
-			t.Errorf("extract(%q) fails with %v; want success", fname, err)
+		withTempFile(t, spec.src, func(fname string) {
```
hanwen: this looks a little bit like python (or ruby?).


  t, err :=  ioutil.TempFile()
  if err != nil  { t.Fatal(err) }
  if _, err :=  ioutil.WriteFile(.., 0644); err != nil { t.Fatal(err) }
  ..


you don't have to add so many cases because you only add 3 lines on top of go/parser.

In fact, I'd even consider dropping the test entirely. If you break this, you should have an end2end test that catches it anyway.

Come to think of it: you do have an example, but do you have an e2e test that builds a binary and checks that it works as expected? There used to be one, but not sure what happened after the rules got moved out of bazel.

yugui: > ```go
> t, err := ioutil.TempFile()
> if err != nil { t.Fatal(err) }
> if _, err := ioutil.WriteFile(.., 0644); err != nil { t.Fatal(err) }
> ```

done.  Now it looks less scary.

> In fact, I'd even consider dropping the test entirely. If you break this, you should have an end2end test that catches it anyway.

You are right.  I reduced the number of the examples. The example could be even one `package main" but I still kept another because I wanted to illustrate the reason why we need this helper program.

go/def.bzl

go/tools/extract_package/extract_test.go
```diff
@@ -1,21 +1,45 @@
...
-		name, err := extract(fname)
+		f, err := ioutil.TempFile(os.Getenv("TEST_TMPDIR"), "example-go-src")
 		if err != nil {
-			t.Errorf("extract(%q) fails with %v; want success", fname, err)
+			t.Fatal(err)
 		}
-		if got, want := name, "main"; got != want {
-			t.Errorf("extract(%q) = %q; want %q", fname, got, want)
+		defer os.Remove(f.Name())
+		if err := f.Close(); err != nil {
```
hanwen: since you don't write through f, you don't have to close the file.

davidzchen:
Thanks for working on this, @yugui!

[Skydoc](https://github.com/bazelbuild/skydoc) is the documentation generation tool for Skylark rules (think Doxygen or epydoc for Skylark). See the [Skydoc documentation](https://github.com/bazelbuild/skydoc/blob/master/README.md) and [`skylark.bzl`](https://github.com/bazelbuild/skydoc/blob/master/skylark/skylark.bzl) for examples.

As @damienmg mentioned, we are still in the process of integrating Skydoc with the documentation pipeline. In the meantime, can you add the documentation for the new rules to the README in addition to the docstrings? That way, we can directly switch over once Skydoc is integrated, but people can still read the documentation in the README in the meantime.

go/tools/extract_package/extract_test.go
yugui: removed.

yugui: removed.

yugui:
Thank you for your reviews, @hanwen, @damienmg,@davidzchen
Could you take another look?

> can you add the documentation for the new rules to the README in addition to the docstrings? 

done.

yugui:
@hanwen, @damienmg,@davidzchen

Ping.  Do I need to do anything else to get this PR merged?

damienmg:
Just talked with @hanwen and he is fine with the PR so LGTM

damienmg:
Oops it brokes (probably due to sandboxing): http://ci.bazel.io/job/rules_go/BAZEL_VERSION=HEAD,PLATFORM_NAME=ubuntu_15.10-x86_64/39/console

PR #13, opened 2016-04-19
Fix CI build failures on Ubuntu 15.04 and Mac OS X

### LINKED ISSUE #12: cgo tests fails with sandboxing
See http://ci.bazel.io/job/rules_go/BAZEL_VERSION=HEAD,PLATFORM_NAME=ubuntu_15.10-x86_64/39/console

This is after merging #10 

/cc @yugui

yugui:
Fixes #12, which was caused by #10.

* Adds a workaround for a bug of `ld` in XCode 7.2 or earlier.
* Avoid depending on a behavior of old GNU `ld` so that it works on Ubuntu 15.04.
  Make the link process more similar to `cc_binary` to avoid the issue.


damienmg:
LGTM

PR #22, opened 2016-05-14
Add a build rule go_asm_library to enable building Assembly programs …

linuxerwang:
…in Go; add examples for go_asm_library; change go_library to take multiple cgo_objects; update README for go_asm_library and go_library.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -867,14 +867,14 @@ filegroup(
 def go_repositories():
   native.new_http_archive(
     name=  "golang_linux_amd64",
-    url = "https://storage.googleapis.com/golang/go1.6.linux-amd64.tar.gz",
+    url = "https://storage.googleapis.com/golang/go1.6.2.linux-amd64.tar.gz",
     build_file_content = GO_TOOLCHAIN_BUILD_FILE,
-    sha256 = "5470eac05d273c74ff8bac7bef5bad0b5abbd1c4052efbdbc8db45332e836b0b"
+    sha256 = "e40c36ae71756198478624ed1bb4ce17597b3c19d243f3f0899bb5740d56212a"
   )
 
   native.new_http_archive(
     name=  "golang_darwin_amd64",
-    url = "https://storage.googleapis.com/golang/go1.6.darwin-amd64.tar.gz",
+    url = "https://storage.googleapis.com/golang/go1.6.2.darwin-amd64.tar.gz",
     build_file_content = GO_TOOLCHAIN_BUILD_FILE,
-    sha256 = "8b686ace24c0166738fd9f6003503f9d55ce03b7f24c963b043ba7bb56f43000"
+    sha256 = "6ebbafcac53bbbf8c4105fa84b63cca3d6ce04370f5a04ac2ac065782397fc26"
   )


bazel-io:

linuxerwang:
This change is to solve the problem compiling golang.org/x/sys/unix with bazel. I created a post in mailing list to ask but nobody replied, so I had to do it myself. The problem is the asm files under x/sys/unix, and with this change the asm files can be put in their own targets. Added examples to verify.

Thanks.

README.md
```diff
@@ -197,6 +198,20 @@ go_library(name, srcs, deps, data)
         <p>List of files needed by this rule at runtime.</p>
       </td>
     </tr>
+    <tr>
```
yugui: I think you can keep the rule API simpler.

* I didn't document `cgo_object` by intention because it was an implementation detail to propagate dependency to native c compiler objects.  It is better to keep this configuration hidden from users if possible.
* Similarly, it is better to avoid having a new rule for go assembler for users to keep consistency to mental model of go tool and `cc_library`.
  * We defined a new rule `cgo_library` because there was no clear way for Bazel to distinguish cgo-enabled sources from pure-go sources at runtime. On the other hand, in this case, you can distinguish assembler inputs from go sources by file extension.

So, we should be able to do something like the following if you call `emit_go_asm_action` from `go_library_impl`.
```bzl
go_library(
  name = "go_default_library",
  srcs = ["asm_go.go", "add.s", "sub.s"],
  library = ":foo",
)

cgo_library(
  name = "foo",
  srcs = ["a.go", "b.c", "b.h", "c.cc", "c.h"],
)
```

examples/cgo/asm_dependency/asm_go.go
```diff
@@ -0,0 +1,14 @@
+package asm_dependency
```
yugui: golint error:
```
examples/cgo/asm_dependency/asm_go.go:1:1: don't use an underscore in package name
```

examples/cgo/asm_dependency/asm_go.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "fmt"
+
+func add(x, y int64) int64
+func sub(x, y int64) int64
+
+func AddTwoNumbers() {
```
yugui: golint errors
```
xamples/cgo/asm_dependency/asm_go.go:8:1: exported function AddTwoNumbers should have comment or be unexported
examples/cgo/asm_dependency/asm_go.go:12:1: exported function SubTwoNumbers should have comment or be unexported
```

README.md
linuxerwang: One question: different from ".c" or ".h", the "go tool asm" command has to be run separately for each ".s" file which generates multiple ".o" files. I had the problem to define a proper outputs attr for these actions.

Thanks.

damienmg:
test this please

damienmg:
retest this please (just for fun)

damienmg:
retest this please (previous build got stalled)

linuxerwang:
I am learning more about skylark and had some new thoughts on improving the implementation. Will figure it out today.

damienmg:
Sure. Thank you very much for your help on making the go rules nice. (sorry I am also using your PR to test some things on Jenkins).

retest this please.

damienmg:
retest this please

bazel-io:

linuxerwang:
Sorry, I messed up this pull request. Please review on pull request #24 which reimplemented the function with Yugui's suggestions.

Issue #16, opened 2016-04-26
Support specifying custom package name


damienmg:
See bazelbuild/bazel#828

yugui:
IIUC, the main problem described in bazelbuild/bazel#828 has gone because now label names can contain slash. 
However there are other several patterns which potentially require custom package names.

I'll summarize the patterns and remaining problems tomorrow or day after tomorrow.

ivucica:
Preliminary work in f32a39c. It suits my immediate hacking needs, but I'll try to come up with something that's actually properly committable (not to mention with the correct git author field).

@damienmg I didn't check yet, but I'll assume this is maintained primarily externally? And also, given that there's no project on Gerrit, I'll assume that a Github PR is the correct way to contribute?

yugui:
It'd be better to discuss use cases of customization.

In my understanding, basically users need to customize import paths when they import third-party libraries. So before discussing solutions to customize import paths, I'd like to summarize several possible patterns of importing third-party libraries.
Next, I'd like to propose a solution.

# Patterns of Bazelization of third-party libraries
TL;TR. It shows that the main use case of import path customization is the case that the user uses `new_git_repository` to import third-party libraries.

## Pattern 1: Add a `BUILD` file into each third-party library.
This is the primary way which rules_go assumes.

example: https://github.com/yugui/rules_go-vendor-example/tree/sutree-merge

### Pro
* It works perfectly when users manage vendored subtree with `git merge -s subtree` or something.

### Cons
* It requires users to manually add a file into the library tree.
  So it does not work when they use `git submodule`, `git subtree` or some library management tools like [Godep](https://github.com/tools/godep), [Gom](https://github.com/mattn/gom) or [Glide](https://github.com/Masterminds/glide).
* It is hard to manage the third-party library if it has its own `WORKSPACE` and `BUILD` files.
   e.g. yugui/rules_go-vendor-example@fdf5a74e0251ba24e2fa91b2e3e7439de403cd94
   
   The user have to modify the existing `BUILD` files in the third-party library to adjust references to package names whenever they upgrade the library.  It often causes merge-conflict, so it is hard to upgrade the library.
   
## Pattern 2: Add a single `//vendor:BUILD`.
This was the main problem in bazelbuild/bazel#828.
It did not work fine when bazelbuild/bazel#828 was discussed but now it works because target labels can contain slash.

example:
* https://github.com/yugui/rules_go-vendor-example/tree/submodule
[...]

PR #30, opened 2016-05-30
 Respect go_prefix attr of dependencies

yugui:
Uses go_prefix attr of individual dependencies to calculate their
importpaths.  This allows users to use Bazel's external dependency
resolution -- {,new_}git_repository rules -- to manage Go libraries.

c.f. https://github.com/bazelbuild/rules_go/issues/16#issuecomment-216010843

Supports case 1 and 3 in #16.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,3 +2,25 @@ workspace(name = "io_bazel_rules_go")
 
 load("//go:def.bzl", "go_repositories")
 go_repositories()
+
+GLOG_BUILD = """
+load("@//go:def.bzl", "go_prefix", "go_library")
+
+go_prefix("github.com/golang/glog")
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "glog.go",
+        "glog_file.go",
+    ],
+    visibility = ["//visibility:public"],
+)
+"""
+
+new_git_repository(
+    name = "com_github_golang_glog",
+    remote = "https://github.com/golang/glog.git",
+    commit = "23def4e6c14b4da8ac2ed8007337bc5eb5007998",
+    build_file_content = GLOG_BUILD,
+)
--- a/examples/external/BUILD
+++ b/examples/external/BUILD
@@ -1,0 +1,11 @@
+load("//go:def.bzl", "go_binary")
+
+go_binary(
+    name = "record_log",
+    srcs = [
+        "main.go",
+    ],
+    deps = [
+        "@com_github_golang_glog//:go_default_library",
+    ],
+)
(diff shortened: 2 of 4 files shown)


bazel-io:

yugui:
Jenkins: test this please.

yugui:
Hmm, "test this please" does not work.
Does it mean I need to be a member of bazel organization but not just a collaborator of rules_go to trigger CI?

damienmg:
Yes, you need to be member of bazelbuild and make it public (https://github.com/orgs/bazelbuild/people/yugui)

Jenkins test this please.

WORKSPACE
```diff
@@ -2,3 +2,25 @@ workspace(name = "io_bazel_rules_go")
...
+    visibility = ["//visibility:public"],
+)
+"""
+
+new_git_repository(
+    name = "com_github_golang_glog",
+    remote = "https://github.com/golang/glog.git",
+    commit = "23def4e6c14b4da8ac2ed8007337bc5eb5007998",
+    build_file_content = GLOG_BUILD,
+)
```
damienmg: This seems like a pretty easy way to do a go_vendor macro rule:

```python
TEMPLATE = """
load("@io_bazel_rules_go//go:def.bzl", "go_prefix", "go_library")

go_prefix("github.com/%s/%s")

go_library(name = "go_default_library", srcs = glob(["**/*.go"]), visibility = ["//visibility:public"])
"""

def go_vendor(org, project, commit):
  new_git_repository(name = "com_github_%s_%s" % (org, project), remote = "https://github.com/%s/%s.git" % (org, project), commit = commit, build_file_content = TEMPLATE % (org, project))
```

damienmg: Actually io_bazel_rules_go won't work currently. You have to use a trick with str(Label(

yugui: @damienmg 
I would like to keep `go_vendor` separated from this PR because other go libraries requires much more work.  It it acceptable for you?

The simple macro works fine with `glog` but other common go libraries are more complex. e.g. https://github.com/golang/oauth2

yugui:
@damienmg 
The URL is invalid because I'm not a member of bazel organization.
Could you add me to the organization again?

WORKSPACE
damienmg: Yes sounds reasonable.

I don't think you need to handle all the cases for an initial version but sure you might want to handle some more than a basic build.

damienmg:
LGTM

damienmg:
For the org: you have a pending invitation, you should be able to accept it.

yugui:
I found the invitation mail. Thank you!

PR #47, opened 2016-07-17
Add .gitignore file

jart:
I noticed you were missing it.

PR #48, opened 2016-07-18
Test that AUTHORS and CONTRIBUTORS are sorted

jart:
@pmbethe09 said in #47 that these files are to be sorted. So I've added a regression test to help contributors know that.

I've also removed Googlers from the AUTHORS file, as this file is intended for copyright holders. Google Inc. owns the code. See the comments in the AUTHORS and CONTRIBUTORS files to learn more about this policy.

**Recommended TODOs**

1. Consider auditing other AUTHORS to find out if their employer is actually the copyright holder. For instance, I noticed one contributor is employed by Fitbit Inc. Perhaps that's what should go in the AUTHORS file.

2. Consider changing copyright headers on source files to read "The Bazel Go Rules Authors" rather than the Bazel Authors, since this project has an independent AUTHORS file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,14 +6,9 @@
 # Name or Organization <email address>
 # The email address is not required for organizations.
 
-Google Inc.
 Benjamin Staffin <benley@gmail.com>
 Brian Silverman <bsilver16384@gmail.com>
-Damien Martin-Guillerez <dmarting@google.com>
-David Chen <dzc@google.com>
 David Santiago <david.santiago@gmail.com>
-Han-Wen Nienhuys <hanwen@google.com>
+Google Inc.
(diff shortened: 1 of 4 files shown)


yugui:
LGTM, but I want to wait for @pmbethe09 because it is originally coming from his request.

yugui:
@pmbethe09 
BTW, Could you use "Confirm squash and merge" instead of "Create a merge commit" next time you merge because `rules_go` seems to have been using that style?

pmbethe09:
Good catch.  The @google.com folks should be only in CONTRIBUTORS as acknowledgement of work without copyright.  Some other folks are in CONTRIBUTORS and AUTHORS, and presumably should only be in one or the other?

Issue #15, opened 2016-04-26
Create "glaze" BUILD file generator for Go


damienmg:
See bazelbuild/bazel#758 for more context

kamalmarhubi:
@damienmg Issue title needs a small fix :-)

Is the intention that Google will adapt and release the internal tool, or will this be an entirely new tool for the open source Bazel?

damienmg:
oops done. The idea is to adapt the internal tools AFAIU.

kamalmarhubi:
Will the tool allow generating BUILD files for a project with vendored libraries? This is the issue I'm currently coming up against. (See also [my mailing list posting](https://groups.google.com/d/topic/bazel-discuss/TX1iCmvG_RA/discussion).)

justinsb:
Is there any timeline on when this might happen?  I'm new to bazel, but if there are problems with open sourcing the internal tool, I wonder if maybe we could recreate it as open source (the go parsing libraries are pretty good and have lots of examples because of generators now).  I even wonder if we could write it as a go generator...

What I would find very helpful are some example outputs of glaze so that we know what we're aiming for.  For example, I think having the glaze-output for https://github.com/golang/glog would be a nice trivial example, and https://github.com/golang/protobuf would be a good non-trivial example (and both libraries are fairly widely used, I presume also inside google...)

yugui:
> Is there any timeline on when this might happen

No, but I'm trying to show some limited version of glaze.

I don't think there's a timeline for opensourcing `glaze` itself for now.
This is because currently we have none inside Google who maintain rules_go -- while I am maintaining it as an external contributor.  It means none who is working on rules_go has access to their internal source codes.
So we have to wait for someone inside Google volunteering in order to get glaze opensourced.

To unblock this situation, I have just started to prototype a limited version of glaze which generates `BUILD` files for external dependencies. It would generate something similar to [`glog.BUILD`](https://github.com/gengo/grpc-gateway/blob/7f5ef70504b9720b5824223cb705f81ca21ae7f8/glog.BUILD) or [`goprotobuf.BUILD`](https://github.com/gengo/grpc-gateway/blob/7f5ef70504b9720b5824223cb705f81ca21ae7f8/goprotobuf.BUILD).
I think the prototype will be also a good start point of generating `BUILD` files for your project.

damienmg:
[...]

PR #51, opened 2016-07-20
Import Gazel, BUILD file generator

yugui:
Supported:
* `load` golang build rules
* Generate `go_prefix` if necessary
* Generate `go_library`, `go_binary`, `go_test` rules
* Refer dependencies between rules within a repository
* Write `BUILD` files, show diff or just prints the result into stdout

To be supported later:
* cgo
* Refer dependencies to external repository
* The pattern 2 in https://github.com/bazelbuild/rules_go/issues/16#issuecomment-216010843
* `visibility`
* respect manual modifications on fixing existing `BUILD` files
* `go_vendor` rule which automatically applies Gazel to external repositories, as discussed in https://github.com/bazelbuild/rules_go/issues/16#issuecomment-216218578

Gazel should be considered experimental. It is possible that this is replaced with glaze when Google opensources it.

c.f. #15

### DIFF
--- a/go/tools/gazel/gazel/diff.go
+++ b/go/tools/gazel/gazel/diff.go
@@ -1,0 +1,58 @@
+/* Copyright 2016 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"io/ioutil"
+	"os"
+
+	bzl "github.com/bazelbuild/buildifier/core"
+	"github.com/bazelbuild/buildifier/differ"
+)
+
+func diffFile(file *bzl.File) (err error) {
+	f, err := ioutil.TempFile("", "BUILD")
+	if err != nil {
+		return err
+	}
+	defer func() {
+		merr := os.Remove(f.Name())
+		if err == nil {
+			err = merr
+		}
+	}()
+	err = func() (err error) {
+		defer func() {
+			cerr := f.Close()
+			if err == nil {
+				err = cerr
+			}
+		}()
+		_, err = f.Write(bzl.Format(file))
+		return err
+	}()
+	if err != nil {
+		return err
+	}
+
+	diff := differ.Find()
+	if _, err := os.Stat(file.Path); os.IsNotExist(err) {
+		diff.Show(os.DevNull, f.Name())
+		return nil
+	}
+	diff.Show(file.Path, f.Name())
+	return nil
+}
--- a/go/tools/gazel/gazel/fix.go
+++ b/go/tools/gazel/gazel/fix.go
@@ -1,0 +1,27 @@
+/* Copyright 2016 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"io/ioutil"
+
+	bzl "github.com/bazelbuild/buildifier/core"
+)
+
+func fixFile(file *bzl.File) error {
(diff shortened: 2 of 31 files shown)


yugui:
http://ci.bazel.io/job/PR-rules_go/41/BAZEL_VERSION=latest,PLATFORM_NAME=darwin-x86_64/console

Looks to be a disk quota issue in Jenkins?

go/tools/gazel/gazel/diff.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		merr := os.Remove(f.Name())
+		if err == nil {
+			err = merr
+		}
+	}()
+	err = func() (err error) {
```
pmbethe09: In my experience, worrying about the error from the defer is overkill, and creates complexity without much gain.
I would have written (instead of line 31-49):
defer os.Remove(f.Name())
defer f.Close()
if _, err := f.Write(bzl.Format(file)); err != nil {
  return err
}

go/tools/gazel/gazel/fix.go
```diff
@@ -0,0 +1,27 @@
...
+package main
+
+import (
+	"io/ioutil"
+
+	bzl "github.com/bazelbuild/buildifier/core"
+)
+
+func fixFile(file *bzl.File) error {
+	// TODO(yugui) Respect exisiting manual configurations as well as possible
```
pmbethe09: go-style uses a ':'
// TODO(yugui): Respect ...

pmbethe09:
This is a long review so it will take some time.  FYI all TODO should have
a ':' after the paren and before the text.
e.g.

// TODO(yugui): Respect exisiting manual configurations as well as possible

go/tools/gazel/rules/resolve_structured_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestStructuredResolverError(t *testing.T) {
+	r := structuredResolver{goPrefix: "example.com/repo"}
+
+	for _, importpath := range []string{
+		"example.com/another",
+		"example.com/another/sub",
+		"example.com/repo_suffix",
+	} {
+		l, err := r.resolve(importpath, "")
```
pmbethe09: combine with if stmt

go/tools/gazel/generator/BUILD
```diff
@@ -0,0 +1,22 @@
...
+
+go_library(
+    name = "go_default_library",
+    srcs = ["generator.go"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "@io_bazel_buildifier//core:go_default_library",
+        "//go/tools/gazel/packages:go_default_library",
+        "//go/tools/gazel/rules:go_default_library",
+        "//go/tools/gazel/testdata:go_default_library",
```
pmbethe09: I think this dep is unused, and also, shouldn't this fail as testdata is 'testonly = 1' ?

davidzchen:
Thanks for creating this tool!

I have one particular request: can we name this something other than "gazel?" I have outlined a few reasons [here](https://github.com/bazelbuild/rules_rust/issues/13#issuecomment-231951026).

pmbethe09:
Would you consider splitting the PR into 2-3 smaller ones, making it easier on me to review?

yugui:
@pmbethe09 
Sure. I'll close this PR and create other smaller ones.

@davidzchen 
I don't stick with a specific name, but I also don't have a good idea about other names. So I need your help to name it.
Did you have any recommendation in the discussion?  gaze?

FYI, when I named it "gazel" I thought about the following things:
* Something different from "glaze" so that it is not confusing with Google's internal one.
* Something reminds "bazel"
* Something starts with "g" for "generate"
* Something pronounceable

damienmg:
Side note for the name gazel: https://en.wikipedia.org/wiki/Gazelle

go/tools/gazel/generator/BUILD
yugui: done in #57

go/tools/gazel/rules/resolve_structured_test.go
yugui: done in #58

go/tools/gazel/gazel/diff.go
yugui: done in #60

go/tools/gazel/gazel/fix.go
yugui: done in #60

davidzchen:
@yugui, thanks for the explanation, and @damienmg, that is a great suggestion!

How about if we name the tool Gazelle? The main thing that makes me hesitant about "Gazel" is that it doesn't really resemble (or sound like) a real word in any language that I am aware of. On the other hand, Gazelle would not only fulfill the criteria that @yugui listed above but it is also a real word and adds the impression that it is fast. :)

Another point that I just realized: while gazel is, in fact, [the archaic form of gazelle](https://en.wiktionary.org/wiki/gazelle), I think that if we want to go this route, it would be better to go with the name "gazelle" since the old form, "gazel," would likely introduce more confusion about how "gazel" and "bazel" are pronounced (since "bazel" is meant to be pronounced like "basil").

yugui:
@davidzchen Sounds good. Let's go with "Gazelle".

hanwen:
thanks for doing this, and Gazelle is a lovely name :)

PR #59, opened 2016-08-03
Support more rules in Gazel

yugui:
* Extracted from #51 
* Depends on #57 and #58 

* Supports generation of `go_test` rules for internal and external tests
* Supports generation of `go_prefix` rules
* Supports generation of `load` of the rules.
* Not yet support generation of `cgo_library`.

PR #58, opened 2016-08-03
Support dependencies to go_library rules in the same repository

yugui:
* Extracted from #51 
* Depends on #57

PR #57, opened 2016-08-03
Import minimal version of gazelle, a BUILD file generator

yugui:
* Extracted from #51
* c.f. #15 

- [x] Decide the name of the tool: https://github.com/bazelbuild/rules_go/pull/51#issuecomment-234122016

PR #60, opened 2016-08-03
Make output modes configurable

yugui:
* Extracted from #51 
* Depends on #57, #58 and #59 
* c.f. #15

### DIFF
--- a/go/tools/gazelle/gazelle/diff.go
+++ b/go/tools/gazelle/gazelle/diff.go
@@ -1,0 +1,47 @@
+/* Copyright 2016 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"io/ioutil"
+	"os"
+
+	bzl "github.com/bazelbuild/buildifier/core"
+	"github.com/bazelbuild/buildifier/differ"
+)
+
+func diffFile(file *bzl.File) error {
+	f, err := ioutil.TempFile("", "BUILD")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(f.Name())
+	defer f.Close()
+	if _, err := f.Write(bzl.Format(file)); err != nil {
+		return err
+	}
+	if err := f.Sync(); err != nil {
+		return err
+	}
+
+	diff := differ.Find()
+	if _, err := os.Stat(file.Path); os.IsNotExist(err) {
+		diff.Show(os.DevNull, f.Name())
+		return nil
+	}
+	diff.Show(file.Path, f.Name())
+	return nil
+}
--- a/go/tools/gazelle/gazelle/print.go
+++ b/go/tools/gazelle/gazelle/print.go
@@ -1,0 +1,27 @@
+/* Copyright 2016 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


go/tools/gazel/gazel/diff.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		merr := os.Remove(f.Name())
+		if err == nil {
+			err = merr
+		}
+	}()
+	defer f.Close()
```
yugui: > In my experience, worrying about the error from the defer is overkill, and creates complexity without much gain.

I agree.

In my experience it was necessary to correctly deal with disk full in the target file system or something. But as you said, worrying about that does not add much value in this specific case of short-living process unlike my experience.

> I would have written (instead of line 31-49):
> ```go
> defer os.Remove(f.Name())
> defer f.Close()
> if _, err := f.Write(bzl.Format(file)); err != nil {
> return err
> }
> ```

done. But I added `Sync` because `Close` used to also imply `Sync` in the original code.

go/tools/gazelle/gazelle/diff.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		merr := os.Remove(f.Name())
+		if err == nil {
+			err = merr
+		}
+	}()
+	defer f.Close()
```
yugui: > In my experience, worrying about the error from the defer is overkill, and creates complexity without much gain.

I agree.

In my experience it was necessary to correctly deal with disk full in the target file system or something. But as you said, worrying about that does not add much value in this specific case of short-living process unlike my experience.

> I would have written (instead of line 31-49):
> ```go
> defer os.Remove(f.Name())
> defer f.Close()
> if _, err := f.Write(bzl.Format(file)); err != nil {
> return err
> }
> ```

done. But I added `Sync` because `Close` used to also imply `Sync` in the original code.

go/tools/gazelle/gazelle/diff.go
```diff
@@ -0,0 +1,47 @@
...
+	bzl "github.com/bazelbuild/buildifier/core"
+	"github.com/bazelbuild/buildifier/differ"
+)
+
+func diffFile(file *bzl.File) (err error) {
+	f, err := ioutil.TempFile("", "BUILD")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(f.Name())
```
yugui: > In my experience, worrying about the error from the defer is overkill, and creates complexity without much gain.

I agree.

Actually in my experience it was necessary to correctly deal with disk full in the target file system, temporary network issue or something. But as you said, worrying about that does not add much value in this specific case of short-living process unlike my experience.

> I would have written (instead of line 31-49):
> ```go
> defer os.Remove(f.Name())
> defer f.Close()
> if _, err := f.Write(bzl.Format(file)); err != nil {
> return err
> }
> ```

Thank you for your suggestion. Fixed.
But I added `Sync` because `Close` used to also imply `Sync` in the original code.

yugui:
@pmbethe09 Could you take another look at this PR?

go/tools/gazelle/gazelle/diff.go
```diff
@@ -0,0 +1,47 @@
...
+	defer f.Close()
+	if _, err = f.Write(bzl.Format(file)); err != nil {
```
pmbethe09: use := instead of = whenever possible, like here.

go/tools/gazelle/gazelle/print.go
```diff
@@ -0,0 +1,27 @@
...
+package main
+
+import (
+	"os"
+
+	bzl "github.com/bazelbuild/buildifier/core"
+)
+
+func printFile(f *bzl.File) (err error) {
+	_, err = os.Stdout.Write(bzl.Format(f))
```
pmbethe09: instead of naming the return error, just use :=

pmbethe09:
LGTM with a couple of nitpicks.

go/tools/gazelle/gazelle/diff.go
yugui: done

go/tools/gazelle/gazelle/print.go
yugui: done

yugui:
Thank you for your review.

Issue #46, opened 2016-07-15
Download all the go packages each time


Mistobaan:
I find it annoying that the rule has to download the whole go package each time (~100MB+).

Can we assume instead that the HOST has go installed and maybe just have a rule to enforce the version ?

yugui:
Right.  I feel the same issue.

Unfortunately we cannot simply assume that because I have no plan to force users install anything other than Bazel and it prerequisites.
But there are several other solutions come to my mind.

1. Cache `http_xxx` rules independently from workspace when it has sha256 sum specified.
    * Pro: solves the same problem not only in rule_go but also in other language supports
    * Cons: requires a change in Bazel itself.  I am not sure if it is feasible.
2. Have another version of `go_repositories()` rule and detect the go runtime installed in the system as [cc_configure.bzl](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl) does.
   * Pro: No extra download is even better than download once.
   * Con: Makes the implementation of `go.bzl` even more complex.

BTW, even today it should not be a big problem for you because Bazel is good for building a larger, complex and potentially polyglot repositories fast and correctly.  So you usually don't need to work on a different repository than the one you worked on yesterday.
So the downloading the runtime should be mostly a one-off overhead.

pmbethe09:
What about adding a config_setting "use_local_go" which uses a
new_local_repository looking for go in /usr/local/go?

yugui:
@pmbethe09 Right.  That's a better alternative to the option 2.

jayconrod:
Closing some old issues. There are a couple ways to avoid the large download now.

Bazel has an experimental flag, --experimental_repository_cache, which specifies a directory for caching downloaded archives. You can add this to a bazelrc file. This is probably the most portable mechanism.

```
echo build --experimental_repository_cache ~/.bazel_downloads >>~/.bazelrc
```

When #321 lands, you'll also be able to use your local Go installation instead of downloading a Go toolchain. This can result in non-hermetic builds though, and it requires that all developers have Go installed in the same place.

```bzl
new_local_repository(
    name = "local_go",
    path = "/opt/go-1.8",
    build_file_content = "",
)

go_repositories(go_linux = "@local_go")
```

PR #72, opened 2016-08-20
Avoid involving a toolchain for the other build platform

yugui:
Resolve a go toolchain for the build platform earlier because it does not make sense to download toolchains for other build platforms.

* this makes it easier to support even more platform
* this makes it easier to support locally installed toolchain, c.f. #46
* this prevents "bazel query" from forcing users to download a toolchain for the other platform

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -422,7 +422,7 @@ def go_binary_impl(ctx):
   lib_out = ctx.outputs.lib
 
   emit_go_link_action(
-    ctx, 
+    ctx,
     transitive_libs=lib_result.transitive_go_library_object,
     importmap=lib_result.transitive_go_importmap,
     cgo_deps=lib_result.transitive_cgo_deps,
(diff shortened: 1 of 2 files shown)


go/def.bzl
```diff
@@ -960,6 +960,22 @@ def cgo_library(name, srcs,
       **kwargs
   )
 
+
+################
+
+_toolchain_map = {
+    'linux': struct(
+        url = "https://storage.googleapis.com/golang/go1.7.linux-amd64.tar.gz",
+        sha256 = ("702ad90f705365227e902b42d91dd1a4" +
```
achew22: Can you put the whole sha256 on line line? I don't think there is a strict 80 char line limit on .bzl files and the ability to grep by the sha256 without having to remember that it is just the first 32 characters is something I personally enjoy having.

achew22:
Nice patch, LGTM

go/def.bzl

go/def.bzl
```diff
@@ -960,37 +960,54 @@ def cgo_library(name, srcs,
...
 
+
+################
+
+_toolchain_map = {
+    'linux': struct(
+        url = "https://storage.googleapis.com/golang/go1.7.linux-amd64.tar.gz",
+        sha256 = "702ad90f705365227e902b42d91dd1a40e48ca7f67a2f4b2fd052aaa4295cd95",
+    ),
+    'mac os x': struct(
```
pmbethe09: The bazel-standard name is "darwin", so maybe just use that?

pmbethe09:
LGTM

pmbethe09:
Jenkins, test this please.

go/def.bzl
yugui: It must be consistent to `ctx.os.name` and it is actually "mac os x".

PR #73, opened 2016-08-23
Support using locally installed go toolchain as Bazel does for C++

### LINKED ISSUE #46: Download all the go packages each time

yugui:
Added an attribute "local" to go_repositories rules.
Fixes #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 <div class="toc">
   <h2>Rules</h2>
   <ul>
+    <li><a href="#go_repositories">go_repositories</a></li>
     <li><a href="#go_prefix">go_prefix</a></li>
     <li><a href="#go_library">go_library</a></li>
     <li><a href="#cgo_library">cgo_library</a></li>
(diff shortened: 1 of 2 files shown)


pmbethe09:
I don't love the part where I have to edit my WORKSPACE file to get this to work.

Is there a way you can think of to add a 'config_setting(name="local_go", ...)'

So that I can build something via:
bazel build --define=local_go=1 some_lib/...

damienmg:
@pmbethe09 : in any case you have to have the local go toolchain repository defined in your WORKSPACE file. I cannot see any way out of that.

pmbethe09:
sorry I lost track of this one.

@damienmg I assumed that in the 'go_repositories' function there would be a passthrough 'local_goroot' which defaulted to /usr/local/go.
The user could enable it via 'select' by --define=local_go=1
In that case most linux/Mac users wouldn't have to edit their WORKSPACE file to be different than what is checked in.

Does that make sense?

damienmg:
Yes you could use an environment variable to make it configurable from the command line. Like export BAZEL_GO_TOOLCHAIN=local; bazel build //...

pmbethe09:
My preference is not environment, but a bazel --define flag

damienmg:
well that is not possible although the original design doc spoke about having configuration variable definable from the command line we decided to use the environment as it was sufficient and needed anyway.

Maybe we can make access to the --define variable happen in the future (and it seems reasonable) but I'd rather avoid it for now.

pmbethe09:
I see, I was thinking that you could pass 'go_toolchain' from a Macro into
the rule via a select statement, but you are saying that wouldn't work?
That is a shame.

damienmg:
Yes select are not working in the WORKSPACE, we are thinking of providing something similar though.

yugui:
This feature is now discussed in #91.

PR #74, opened 2016-08-23
Support external dependency in gazelle

yugui:
c.f. #15

### DIFF
--- a/go/tools/gazelle/rules/resolve_external.go
+++ b/go/tools/gazelle/rules/resolve_external.go
@@ -1,0 +1,63 @@
+/* Copyright 2016 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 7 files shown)


yugui:
@pmbethe09 This is another step toward #15.  Could you take a look?

go/tools/gazelle/rules/resolve_external.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	prefix := r.Root
+	var pkg string
+	if importpath != prefix {
+		pkg = strings.TrimPrefix(importpath, prefix+"/")
+	}
+
+	components := strings.Split(prefix, "/")
+	labels := strings.Split(components[0], ".")
```
pmbethe09: why not just call sort.Reverse(labels)

pmbethe09:
LGTM with one comment.

go/tools/gazelle/rules/resolve_external.go
yugui: `sort.Reverse` is not what we want.  It wraps an ascending comparator and returns a descending comparator.

Issue #42, opened 2016-06-29
/bin/bash: Argument list too long


stephanheinze:
Trying to compile a library with several thousand (generated) go-files will cause failure because of a too long argument list passed to bash.

yugui:
Is it possible to build the go package with "go build"?
If it is possible, I'll treat this issue as a bug.

damienmg:
For most of the rule we use a param files that we generate to avoid hiting
the bash command line limit.

stephanheinze:
I've published the repository: github.com/fgrid/iso20022
The master branch ist go-getable.
I've setup a ```bazel-build``` branch that's NOT buildable by bazel with rules_go 0.0.3.

Hope that helps.

Warning: loads of files :-)
Compiling takes some CPU time.

PR #77, opened 2016-08-28
add executable and c opts

nathantchan:
Go compile command now written to executable file to avoid arg list being too long. Also, add ability for c compile options to be passed to _cgo_codegen function. @bsilver8192 @AustinSchuh #42

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -133,6 +133,14 @@ def go_environment_vars(ctx):
                                    {"GOOS": "linux",
                                     "GOARCH": "amd64"})
 
+def _tmp_cmd_file(cmds, ctx, fn):
+  cmds_all = ["set -e"]
+  cmds_all += cmds
+  cmds_all_str = "\n".join(cmds_all)
+  f = ctx.new_file(ctx.configuration.bin_dir, fn)
+  ctx.file_action( output = f, content = cmds_all_str, executable = True)
+  return f
+
 def emit_go_asm_action(ctx, source, out_obj):
   """Construct the command line for compiling Go Assembly code.
   Constructs a symlink tree to accomodate for workspace name.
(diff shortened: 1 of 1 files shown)


bazel-io:

nathantchan:
I signed it!

yugui:
@nathantchan Why do we need to have different sets of options for code generation and actual compilation?

brian-peloton:
One major limitation is that the actual compilation takes place via `cc_library`, which has very different escaping rules than the existing copts attr of `_cgo_codegn_rule`. In particular, `cc_library` does make variable and location expansion which means while `_cgo_codegn_rule` sets some environment variables as alternatives. I can't think of any way to create a single string which indicates a path relative to the root of the execdir for both of them, and it seems like adding make variable or location expansion to the Go rules will also add a lot of unnecessary complication.

Also, even with allow location expansion, I don't see any way to get directory names, which are important for things like `-I` flags.

yugui:
@brian-peloton 
I am trying to make `_cgo_codegen_rule` as consistent to `cc_library` as possible. As you said, make variable and location expansion are what are missing in the current implementation.  So it is a bug we can fix.

Could you show me concrete use cases about other directory things?

@nathantchan I'd like to discuss this topic about C options separately.  Could you exclude that part from this PR?

go/def.bzl
```diff
@@ -153,11 +161,13 @@ def emit_go_asm_action(ctx, source, out_obj):
       " ".join(args),
   ]
 
+  f = _tmp_cmd_file(cmds, ctx, out_obj + "GoAsmCompileFile")
```
yugui: Could you name the temp file `out_obj + ".GoAsmCompile.params"` to be consistent to follow convention in Bazel?

go/def.bzl
```diff
@@ -133,6 +133,14 @@ def go_environment_vars(ctx):
                                    {"GOOS": "linux",
                                     "GOARCH": "amd64"})
 
+def _tmp_cmd_file(cmds, ctx, fn):
```
yugui: Can we call this function `emit_generate_params_action` or something?
I prefer the suffix `_action` to be consistent to other actions, and I prefer the name `params` to follow a convention in Bazel.

nathantchan:
I was not quite sure how to incorporate all the changes suggested into this PR, so I created a new one: https://github.com/bazelbuild/rules_go/pull/87

yugui:
Can I close this PR for now because #87 was merged?

PR #84, opened 2016-08-30
Add go_repository and new_go_repository rules

yugui:
c.f. #15 and https://github.com/bazelbuild/rules_go/issues/16#issuecomment-216218578

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,9 @@
 <div class="toc">
   <h2>Rules</h2>
   <ul>
+    <li><a href="#go_repositories">go_repositories</a></li>
+    <li><a href="#go_repository">go_repository</a></li>
+    <li><a href="#new_go_repository">new_go_repository</a></li>
     <li><a href="#go_prefix">go_prefix</a></li>
     <li><a href="#go_library">go_library</a></li>
     <li><a href="#cgo_library">cgo_library</a></li>
@@ -119,6 +122,132 @@ and it should work.
 ## Disclaimer
 
 These rules are not supported by Google's Go team.
+
+<a name="go_repositories"></a>
+## go\_repositories
+
+```bzl
+go_repositories()
+```
+
+Instantiates external dependencies to Go toolchain in a WORKSPACE.
+All the other workspace rules and build rules assume that this rule is
+placed in the WORKSPACE.
+
+
+<a name="go_repository"></a>
+## go\_repository
+
+```bzl
+go_repository(name, importpath, commit, tag)
+```
+
+Fetches a remote repository of a Go project, expecting it contains `BUILD`
+files. It is an analogy to `git_repository` but it recognizes importpath
+redirection of Go.
+
+<table class="table table-condensed table-bordered table-params">
+  <colgroup>
+    <col class="col-param" />
+    <col class="param-description" />
+  </colgroup>
+  <thead>
+    <tr>
+      <th colspan="2">Attributes</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><code>name</code></td>
+      <td>
+        <code>String, required</code>
+        <p>A unique name for this external dependency.</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>importpath</code></td>
+      <td>
+        <code>String, required</code>
+        <p>An import path in Go, which corresponds to the root of the target
+        remote repository</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>commit</code></td>
+      <td>
+        <code>String, coptional</code>
+        <p>The commit hash to checkout in the repository.</p>
+        <p>Note that one of either <code>commit</code> or <code>tag</code> must be defined.</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>tag</code></td>
+      <td>
+        <code>String, coptional</code>
+        <p>The tag to checkout in the repository.</p>
+        <p>Note that one of either <code>commit</code> or <code>tag</code> must be defined.</p>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+
+<a name="new_go_repository"></a>
+## new\_go\_repository
+
+```bzl
+new_go_repository(name, importpath, commit, tag)
+```
+
+Fetches a remote repository of a Go project and automatically generates
+`BUILD` files in it.  It is an analogy to `new_git_repository` but it recognizes
+importpath redirection of Go.
+
+<table class="table table-condensed table-bordered table-params">
+  <colgroup>
+    <col class="col-param" />
+    <col class="param-description" />
+  </colgroup>
+  <thead>
+    <tr>
+      <th colspan="2">Attributes</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><code>name</code></td>
+      <td>
+        <code>String, required</code>
+        <p>A unique name for this external dependency.</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>importpath</code></td>
+      <td>
+        <code>String, required</code>
+        <p>An import path in Go, which corresponds to the root of the target
+        remote repository</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>commit</code></td>
+      <td>
+        <code>String, coptional</code>
+        <p>The commit hash to checkout in the repository.</p>
+        <p>Note that one of either <code>commit</code> or <code>tag</code> must be defined.</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>tag</code></td>
+      <td>
+        <code>String, coptional</code>
+        <p>The tag to checkout in the repository.</p>
+        <p>Note that one of either <code>commit</code> or <code>tag</code> must be defined.</p>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
 
 <a name="go_prefix"></a>
 ## go\_prefix
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,29 +1,13 @@
 workspace(name = "io_bazel_rules_go")
 
-load("//go:def.bzl", "go_repositories", "go_internal_tools_deps")
+load("//go:def.bzl", "go_repositories", "go_internal_tools_deps", "new_go_repository")
 
 go_repositories()
 
-GLOG_BUILD = """
-load("@//go:def.bzl", "go_prefix", "go_library")
-
-go_prefix("github.com/golang/glog")
-
-go_library(
-    name = "go_default_library",
-    srcs = [
-        "glog.go",
-        "glog_file.go",
-    ],
-    visibility = ["//visibility:public"],
-)
-"""
-
-new_git_repository(
+new_go_repository(
     name = "com_github_golang_glog",
-    build_file_content = GLOG_BUILD,
     commit = "23def4e6c14b4da8ac2ed8007337bc5eb5007998",
-    remote = "https://github.com/golang/glog.git",
+    importpath = "github.com/golang/glog",
 )
 
 go_internal_tools_deps()
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,6 @@
 workspace(name = "io_bazel_rules_go")
 
-load("//go:def.bzl", "go_repositories")
+load("//go:def.bzl", "go_repositories", "go_internal_tools_deps")
 
 go_repositories()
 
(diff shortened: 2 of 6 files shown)


pcj:
I like the code!  Good the see this landing in rules_go.

While I understand the naming of the repository rules as `go_vendor` and `new_go_vendor`, the names `go_repository` and `new_go_repository` seem more bazel-like and less cognitive burden for people to learn given that the attributes are very similar to `git_repository`.

Same with attribute naming`revision` vs. `commit`:  While this is the nouns used by the implementation in `x/go/vcs`, is the value of a revision not a commit 99% of the time?

yugui:
> the names go_repository and new_go_repository seem more bazel-like and less cognitive burden for people to learn given that the attributes are very similar to git_repository.

Good point. I'll rename them.

> Same with attribute naming revision vs. commit: While this is the noun used by the implementation in x/go/vcs, is the value of a revision not a commit 99% of the time?

In my experience, it is a tag in 50% of use cases and it is a commit in the other 50%.
So I have avoided both the words and used "revision" instead.

pmbethe09:
Is this ready for review?  you didn't assign it to me, so I haven't looked
at it.

On Tue, Aug 30, 2016 at 8:13 PM, Yuki Yugui Sonoda <notifications@github.com

sayrer:
This looks good! What are the next steps?

yugui:
@pmbethe09, @sayrer 
No, it's not yet completely ready but the overall design won't change later.

In the next step, I'd like to rename rules as @pcj suggested and I'd like to confirm this rule works fine with larger example in a private project.

yugui:
Request for comments.

I still think `go_repository` is a good name because it is actually a repository rule similar to `git_repository` and it is not confusing to `vendor` directory in Go.
But I noticed that `go_repository` is confusing with `go_repositories`.
So I am thinking of renaming `go_repositories` to `go_tool_repositories` to be clearer.

Do you have any other better ideas?

yugui:
Not this PR is ready for review.

* I have renamed the rules to `go_repository` and `new_go_repository`.
* Rebased onto master
* Confirmed that it works fine with a larger workspace.

yugui:
Jenkins, retest this please.

yugui:
The test is failing because of a mismatch of SHA256 of https://storage.googleapis.com/golang/go1.7.1.darwin-amd64.tar.gz.
But the SHA256 sum in `go/def.bzl` matches to the one described at https://golang.org/dl/ and it also matches to the SHA256 sum of the file which I downloaded from the URL.

Jenkins looks to be fetching something wrong from the same URL.

README.md
```diff
@@ -120,6 +123,113 @@ and it should work.
...
+Instantiates external dependencies to Go toolchain in a WORKSPACE.
+All the other workspace rules and build rules assume that this rule is
+placed in the WORKSPACE.
+
+
+<a name="go_repository"></a>
+## go\_repository
+
+```bzl
+go_repository(name, importpath, revision)
```
pmbethe09: Lets get bazel-core opinion on the use of 'revision' as opposed to having both a 'tag' and 'commit' field where the user picks one.

WORKSPACE
```diff
@@ -1,58 +1,16 @@
...
-)
-"""
-
-new_git_repository(
-    name = "org_golang_x_tools",
-    build_file_content = X_TOOLS_BUILD,
-    commit = "2bbdb4568e161d12394da43e88b384c6be63928b",
-    remote = "https://go.googlesource.com/tools",
-)
+go_internal_tools_deps()
```
pmbethe09: This implies that users will have to add this to their WORKSPACE?

README.md
pcj: +1 on tag and commit as opposed to revision (matches existing convention)

WORKSPACE
yugui: No. This is only necessary to test `gazelle` and `fetch_repo` in this repository.

README.md
yugui: As an interface, it sounds good.
But as implementation `fetch_repo` cannot distinguish tags from commits because it relies on [`Cmd.CreateAtRev`](https://godoc.org/golang.org/x/tools/go/vcs#Cmd.CreateAtRev).

Is this acceptable for you?

yugui:
Rebased and renamed the `revision` attr to `commit` and `tag`.

@pmbethe09 Could you take another look?

pcj:
I tested new_go_repository on https://github.com/gogo/protobuf and it worked well, saved me from having to write the BUILD files by hand.  A few comments and questions:

1. Unfortunately there is no progress indicator mechanism for repository rules to hook into.  The user ends up blindly waiting for fetch-repo to finish.  That might be something to bring up to the main bazel team.  

2. The user has to figure out what the targets are once gazelle has done its job.  I use `bazel query @WORKSPACE//... --output label_kind` for that purpose.  At a minimum, you might  make such a recommendation in the documentation.  Alternatively, a verbose option could be considered.

3. I somewhat expected there to be a `remote` attribute for the github URL.  Instead, the `importpath

go/def.bzl
```diff
@@ -999,9 +999,172 @@ def cgo_library(name, srcs,
...
+  if ctx.attr.commit and ctx.attr.tag:
+    fail("cannot specify both of commit and tag", "commit")
+  if ctx.attr.commit:
+    rev = ctx.attr.commit
+  elif ctx.attr.tag:
+    rev = ctx.attr.tag
+  else:
+    fail("neither commit or tag is specified", "commit")
+
+  result = ctx.execute([
```
pmbethe09: Perhaps add a TODO to think about submodules?
https://www.bazel.io/versions/master/docs/be/workspace.html#git_repository.init_submodules

go/def.bzl
```diff
@@ -999,9 +999,172 @@ def cgo_library(name, srcs,
...
+
+    "_fetch_repo": attr.label(
+        default = Label("@io_bazel_rules_go_repository_tools//:bin/fetch_repo"),
+        allow_files = True,
+        single_file = True,
+    ),
+}
+
+go_repository = repository_rule(
+    _go_repository_impl,
```
pmbethe09: its verbose, but I like using "implementation = _go_repository_impl,"

go/def.bzl
```diff
@@ -999,9 +999,172 @@ def cgo_library(name, srcs,
...
+    attrs = _go_repository_attrs,
+)
+
+new_go_repository = repository_rule(
+    _new_go_repository_impl,
+    attrs = _go_repository_attrs + {
+        "_gazelle": attr.label(
+            default = Label("@io_bazel_rules_go_repository_tools//:bin/gazelle"),
+            allow_files = True,
+            single_file = True,
```
pmbethe09: Also,
executable = True
or is that not used in WORKSPACE rules the same as BUILD rules?

go/def.bzl
```diff
@@ -999,9 +999,172 @@ def cgo_library(name, srcs,
...
+    result = ctx.execute(['mkdir', '-p', ctx.path('src/' + dep.importpath)])
+    if result.return_code:
+      fail('failed to create directory: %s' % result.stderr)
+    ctx.download_and_extract(
+        '%s/archive/%s.zip' % (dep.repo, dep.commit),
+        'src/%s' % dep.importpath, '', 'zip', '%s-%s' % (name, dep.commit))
+
+  result = ctx.execute([
+      'env', 'GOROOT=%s' % goroot, 'GOPATH=%s' % gopath, 'PATH=%s/bin' % goroot,
+      'go', 'generate', 'github.com/bazelbuild/buildifier/core'])
```
pmbethe09: Is this not already taken care of by the BUILD file in that directory?

README.md
```diff
@@ -120,6 +123,131 @@ and it should work.
...
+
+<a name="go_repository"></a>
+## go\_repository
+
+```bzl
+go_repository(name, importpath, commit, tag)
+```
+
+Fetches a remote repository of a Go project. It is an analogy to
+`git_repository` but it recognizes importpath redirection of Go.
```
pmbethe09: ... and expects BUILD files to be present.  ?

WORKSPACE
```diff
@@ -1,58 +1,16 @@
...
-    build_file_content = GLOG_BUILD,
     commit = "23def4e6c14b4da8ac2ed8007337bc5eb5007998",
-    remote = "https://github.com/golang/glog.git",
-)
-
-git_repository(
-    name = "io_bazel_buildifier",
-    commit = "0ca1d7991357ae7a7555589af88930d82cf07c0a",
-    remote = "https://github.com/bazelbuild/buildifier.git",
+    importpath = "github.com/golang/glog",
```
pmbethe09: I presume this becomes the 'go_prefix' of this repository?  If so, should it be called 'go_prefix', and then should it be optional, where it is inferred from the 'remote'?

sayrer: I'm not sure I understand this comment. Using an older version of this patch I have something like this working:

new_go_repository(
   name = "com_github_PagerDuty_godspeed",
   importpath = "github.com/PagerDuty/godspeed",
   revision = "ef757b820a7d6760a89641ac29541967eb6d9f05",
 )

I'd like "importpath"/"go_prefix" to make "name" optional.

yugui:
@pcj Thank you for your feedback.

> Unfortunately there is no progress indicator mechanism for repository rule implementations to hook into. This isn't rules_go's fault, but the user ends up blindly hoping and praying for fetch-repo to finish. It might be something to bring up to the main bazel team.

Right.
Unfortunately there looks to be no way to "stream" messages from subprocesses in repository rule.  But I felt the same necessity several times. I'll file a feature request to Bazel.

Another issue I found was that `gazelle` takes more than 5 minutes for `google.golang.org/api`. So probably it is better to process several packages in parallel with goroutine.

> The user has to figure out what the targets are once gazelle has done its job. I use bazel query @WORKSPACE//... --output label_kind for that purpose but I don't know if that is well-known. At a minimum, you might make such a recommendation in the documentation. Alternatively, a verbose option that reports this information could be considered.

Sure. Let me work on the documentation in another PR.

> I somewhat expected there to be a remote attribute for the github URL. Instead, the importpath attribute fills the dual role of both remote and go_prefix. Do we anticipate this to always be the case? What if the user wants to reference an alternative git server? Curious what others think about this.

Personally I don't want to recommend that "flexibility" because importpath is always the actual URL to fetch from in the way of Golang. And users can still combine `new_git_repository` with Gazelle-generated `BUILD` file as a workaround in the worst case.

So I'd like to revisit this possibility once we have a concrete use case with a valid reason.

WORKSPACE
yugui: @pmbethe09 
Personally I don't want to allow users to decouple `go_prefix` from remote URL because they must match in the way of Go.
I don't care whether we call it `importpath` or `go_prefix`, but I'd like to have only one of them for now.

And `importpath` was natural for me because it is the Go importpath of the repository root.
The name `go_prefix` is coming from that the root importpath will be a prefix for subpackages but it is till a importpath.

What do you think?  Should we rename it `go_prefix`?

README.md
yugui: `git_repository` also expects `BUILD` files to be present. So there's no difference.
But I have explicitly mentioned the expectation.

go/def.bzl
yugui: done

go/def.bzl

go/def.bzl
yugui: done

go/def.bzl
yugui: No. Here we need to build `gazelle` and `fetch_repo` without any help by Bazel build rules because the tools are built and used in the loading phase.
This is a variation of the typical issues of bootstrapping.

pmbethe09:
LGTM.

I am happy with where this is -- I suggest a merge and then we can iterate if we find new issues.

I thought about it, and I am happy with 'importpath' in new_go_repository for now, we can think about also having a remote for intermediate hosting later.

powelljo:
I see a couple comments here asking why there's no `remote` attribute to allow a separation of the dual roles that `importpath` plays.  I'd like to echo that question too.  Many companies encourage their engineers to cache third party dependencies in a local repository and pull from there.  For example, a user may need to pull from `git+ssh://megacorp.biz/repos/local_copy_of_protobuf`, which can cause problems when importing a Go package unless this separation can be made.  I have a tiny patch to optionally add this functionality ready to go if there's an appetite for it.

pmbethe09:
There has been a similar request here, and I was planning on looking at this in the next few weeks.
However if you have a patch ready, by all means please send out a PR

powelljo:
@pmbethe09 Great, I'm ready to send the PR.  Please forgive the newbie question:  is the Gerrit URL provided in the link on the (contributing how-to page)[https://github.com/bazelbuild/rules_go/blob/master/CONTRIBUTING.md] still correct?  It looks to me like it's carried over from before pulling this repository out from Bazel itself.

benley:
@powelljo it looks like this particular repo is maintained entirely on github, so opening a PR here should do the trick.

pmbethe09:
Yes, just a regular PR here, as this is the source-of-truth repo

On Tue, Nov 29, 2016 at 10:23 PM, Benjamin Staffin <notifications@github.com

PR #85, opened 2016-08-30
Avoid downloading go toolchain whenever go/def.bzl changes

yugui:
The issue was introduced at #72. This commit reverts a part of #72
to let Bazel manage go toolchain archives, but it still resolves a tool
chain for the current build platform in the analysis phase.


pmbethe09:
LGTM

PR #86, opened 2016-08-30
Fix build failures of cgo targets in OSX sandbox

### LINKED ISSUE #82: Missing dependency causes breakage on Mac
Breakage: http://ci.bazel.io/job/rules_go/BAZEL_VERSION=HEAD,PLATFORM_NAME=darwin-x86_64/281/console

Reason: we now have sandbox on Mac, and the missing dependency `cc_wrapper` is not in the sandbox so we can't find it.

Suggested fix: add dependency `//tools/defaults:crosstool` to every rule that need gcc.

yugui:
Adds dependency to crosstool

Fixes #82

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -408,8 +408,8 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
   ]
 
   ctx.action(
-      inputs = (list(transitive_libs) + [lib] +
-                list(cgo_deps) + ctx.files.toolchain),
+      inputs = (list(transitive_libs) + [lib] + list(cgo_deps) +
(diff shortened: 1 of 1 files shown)


go/def.bzl
```diff
@@ -529,6 +529,13 @@ go_library_attrs = go_env_attrs + {
     ),
 }
 
+_crosstool_attrs = {
+    "_crosstool": attr.label(
+        default = Label("//tools/defaults:crosstool"),
+        cfg = HOST_CFG,
```
damienmg: Technically you should ask for the target configuration but that's does not really matter now.

yugui: Thanks. I confused something.
Fixed.

pmbethe09:
LGTM

PR #90, opened 2016-09-06
Symlink local go toolchain if 'goroot' is specified

### LINKED ISSUE #46: Download all the go packages each time

pcj:
Adds the 'goroot' string attribute to the go_repositories rule.  If
present, the value of the goroot attribute will be used to symlink the
locally installed go toolchain rather than download it.

Closes #46.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -994,6 +994,36 @@ filegroup(
 )
 """
 
+def _go_toolchain_download(ctx, toolchain):
+  """Download the go toolchain into the external workspace."""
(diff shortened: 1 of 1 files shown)


bazel-io:

pmbethe09:
There appear to be merge conflicts -- can you fix?

pcj:
I wrote this primarily to be able to develop on rules_go without having to re-install the go toolchain every time def.bzl changes.  I consider it best practice not to use this option unless necessary, so I have left it undocumented in the README.md.

PR #91, opened 2016-09-06
Symlink local go toolchain if 'goroot' is specified

### LINKED ISSUE #46: Download all the go packages each time

pcj:
Adds the 'goroot' string attribute to the go_repositories rule.  
If present, the value of the goroot attribute will be used to symlink the
locally installed go toolchain rather than download it.

Renames '_go_repository_select' to '_go_toolchain' to match the
workspace name.

Avoids symlinking the entire directory to avoid the warning 'Workspace
name in .../external/io_bazel_rules_go_toolchain/WORKSPACE
(@golang_darwin_amd64) does not match the name given in the
repository's definition (@io_bazel_rules_go_toolchain); this will
cause a build error in future versions.'

Closes #46.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -990,22 +990,56 @@ filegroup(
 )
 """
 
-def _go_repository_select_impl(ctx):
-  os = ctx.os.name
-  # NOTE: This mapping cannot be table-driven to prevent
-  # Bazel from downloading the other archive.
-  if os == 'linux':
-    goroot = ctx.path(ctx.attr._linux).dirname
-  elif os == 'mac os x':
-    goroot = ctx.path(ctx.attr._darwin).dirname
+def _go_toolchain_impl(ctx):
+  """Symlinks the correct go toolchain."""
+
+  # 1. Configure the goroot path
+  if ctx.attr.goroot:
+    goroot = ctx.path(ctx.attr.goroot)
   else:
-    fail("unsupported operating system: " + os)
+    os = ctx.os.name
+    # NOTE: This mapping cannot be table-driven to prevent
+    # Bazel from downloading the other archive.
+    if os == 'linux':
+      goroot = ctx.path(ctx.attr._linux).dirname
+    elif os == 'mac os x':
+      goroot = ctx.path(ctx.attr._darwin).dirname
+    else:
+      fail("unsupported operating system: " + os)
 
-  ctx.symlink(goroot, ctx.path(''))
+  # 2. Create the symlinks and write the BUILD file.
+  gobin = goroot.get_child("bin")
+  gopkg = goroot.get_child("pkg")
+  ctx.symlink(gobin, "bin")
+  ctx.symlink(gopkg, "pkg")
+  ctx.file("BUILD", GO_TOOLCHAIN_BUILD_FILE, False)
 
-_go_repository_select = repository_rule(
-    _go_repository_select_impl,
-    attrs = {
+  # 3. If the user has specified the goroot explicitly, confirm a
+  # working installation by checking the output of 'go version'.
+  if ctx.attr.goroot:
+    go = gobin.get_child("go")
+    result = ctx.execute([go, "version"])
+    if result.return_code:
+      fail("""
+Something's not right.  Are you sure '%s' points to a functional GOROOT?
+--> %s
+""" % (go, result.stderr))
+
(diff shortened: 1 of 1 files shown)


bazel-io:

pcj:
The version check may be too fragile however.  For people building from source, it may look like `go version devel +2e4b659 Wed Jul 15 06:04:51 2016 +0000 darwin/amd64`.  Perhaps just failing on a non-zero return_code is enough.

yugui:
LGTM, but I want to hear @pmbethe09's opinion because this PR is basically a reimplementation of #73.

Jenkins, test this please.

yugui:
BTW, @pcj, what do you think about querying `GOROOT` with `go env`, which is implemented in #73?

pcj:
Somehow I completely missed #73 when I started this.  You are right, they are quite similar.  I'm fine with either one.  

In regards to querying `GOROOT`, I have noticed that querying `go env` has some interesting behavior.  If I query my locally homebrew installed go1.6.2, I get the following:

```sh
$ which go && go env GOROOT
/usr/local/bin/go
/usr/local/Cellar/go/1.6.2/libexec
```

If the user has `GOROOT` in `.bash_profile` (they probably shouldn't) or in the shell environment, then querying `go env` will report that value.  

```sh
$ export GOROOT=$(go env GOROOT)
$ which external/golang_darwin_amd64/bin/go && external/golang_darwin_amd64/bin/go env GOROOT
external/golang_darwin_amd64/bin/go
/usr/local/Cellar/go/1.6.2/libexec
```

However, if GOROOT is not set, then `external/golang_darwin_amd64/bin/go` will return exit code 2 (with any go command), because it cannot find the hardcoded fallback GOROOT in `/usr/local/go`.  Since all rules_go ctx.action's explicitly set GOROOT, this does not come up.

```sh
$ external/golang_darwin_amd64/bin/go env; echo $?
go: cannot find GOROOT directory: /usr/local/go
2
```

pmbethe09:
Not sure I have a preference between this one and #73 (which I lost, sorry
yugui@)

But in both cases, I don't like that if you set 'goroot' in the WORKSPACE,
it always uses your local go.
That means that if you have a Jenkins that you want to keep "pure" with no
entry in your WORKSPACE, then you have to make local edits and remember not
to commit them -- I am against that.

Whereas a '--define' flag can be put into your local .bazelrc without
polluting the .git repo

On Thu, Sep 8, 2016 at 9:36 AM, Paul Cody Johnston <notifications@github.com

pcj:
It's a good point about not having to remember local edits.  Here's an alternative implementation that reads an environment variable and seems to work:

```python
  bazel_goroot = ctx.os.environ.get("BAZEL_GOROOT")

  # 1. Configure the goroot path
  if bazel_goroot:
    goroot = ctx.path(bazel_goroot)
  else:
    os_name = ctx.os.name
    # NOTE: This mapping cannot be table-driven to prevent
    # Bazel from downloading the other archive.
    if os_name == 'linux':
      goroot = ctx.path(ctx.attr._linux).dirname
    elif os_name == 'mac os x':
      goroot = ctx.path(ctx.attr._darwin).dirname
    else:
      fail("unsupported operating system: " + os_name)
```

I tried a different impl. that attempted to use a `config_setting()` + `select()` on a `--define` but was not able to get it to work.  

Anyone have a preference as to the env variable name?  I don't think it should be just `GOROOT` as that can have negative consequences outside bazel.

yugui:
@pcj 
> However, if GOROOT is not set, then external/golang_darwin_amd64/bin/go will return exit code 2 (with any go command), because it cannot find the hardcoded fallback GOROOT in /usr/local/go. Since all rules_go ctx.action's explicitly set GOROOT, this does not come up.

Right. So the querying makes sense only if the properly configured installation of go tool chain is available and the user explicitly wants the rule to query, as you see in #73.

@pmbethe09 
> I don't like that if you set 'goroot' in the WORKSPACE, it always uses your local go.

I suspect if we are talking about different use cases.
For me, the restriction is natural and acceptable because this local link feature is an analogy to `new_local_repository`. i.e. if the user can guarantee that the target repository is available at the same path in all the possible environments, they can refer the local path without downloading it.

I wonder if you are talking about allowing users to overwrite remote dependency with local installation based on runtime configuration.
If so, I don't like the idea because it reduces reproducibility of builds.

pcj:
No longer relevant, closing this.

PR #96, opened 2016-09-13
Add an error message to gazelle when no package directories are given

kchodorow:
Otherwise, if no packages are passed, the script appears to work but
does nothing.

CC @pmbethe09


pmbethe09:
LGTM

Issue #70, opened 2016-08-20
go_binary cross-compilation


pcj:
What's the correct way to build a go_binary for a linux platform on osx?  For example, `bazel build //my/server` where `":server"` is a go_binary rule works on my osx machine but fails with `"can't find import: html/template"` when invoked with `--cpu armeabi_v7a`.  I was expecting a toolchain for k8 but the only available --cpu options are:

```
ERROR: No toolchain found for cpu 'k8'. Valid cpus are: [
  darwin,
  armeabi-v7a,
  x64_windows_msvc,
].
```

pcj:
Here's the error while attempting to compile the most trivial HelloWorld program: 

```sh
ERROR: /Users/pcj/foo/go/BUILD:6:1: error executing shell command: 'rm -rf bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir && mkdir -p bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir && mkdir -p bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir/g...' failed: bash failed: error executing command 
  (cd /private/var/tmp/_bazel_pcj/75329c14c71b627aaf1fecd401bf1588/execroot/foo && \
  exec env - \
    GOARCH=arm \
    GOOS=linux \
  /bin/bash -c 'rm -rf bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir \
&& mkdir -p bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir \
&& mkdir -p bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir/foo/go \
&& ln -s ../../../../../../../../bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.a bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir/foo/go/server.a \
&& export PATH=$PATH:/usr/bin \
&& export GOROOT=$(pwd)/external/golang_darwin_amd64/go/bin/.. \
&& cd bazel-out/stub_armeabi-v7a-fastbuild/bin/go/server.dir \
&& ../../../../../external/golang_darwin_amd64/go/bin/go tool link -L . -o foo/go/server -s -extld /bin/false -extldflags '\''-Wl,-S -Wl,-rpath,$ORIGIN/../ -Lfoo/'\'' foo/go/server.a \
[...]

PR #98, opened 2016-09-13
Adds cross-compilation support to rules_go.

pcj:
- New rule 'xgo_binary' takes a list of go_binary dependencies and an
  'os_arch' string attribute.  Build output 'foo' is namespaced in
  bazel-bin/PACKAGE_NAME/_xgo/GOOS_GOARCH/foo.  Cross-compilation is
  achieved via a parameterized aspect implementation that propogates
  down the 'deps' attribute; platform target is informed by the
  'os_arch' attribute whose values are enumerated in
  //go/private/goos_goarch.bzl.
- New rule 'go_root' captures the absolute path of GOROOT at the time
  of toolchain installation.  This rule is only used internally.  For
  some reason cross-compilation requires an absolute path to GOROOT.
- Repository rule 'go_repositories' takes an optional environment
  variable BAZEL_GOROOT to use local go toolchain if desired.
- Partitions def.bzl file into individual bzl files in go/private
  where possible.  Each file implements a single rule.  These are
  loaded by def.bzl, so end-user is unaffected.
- Reorganization of def.bzl into sections: constants, utility
  functions, action functions, aspects, and rules.
- Reengineered go_library_impl and go_binary_impl into functions
  go_library_build and go_binary_build that act as common entrypoints
  for the reworked __impl functions as well the new aspect
  implementations _xgo___impl.  These functions are agnostic to the
  layout of the ctx argument (different for normal rules vs. aspect
  rules).  The parameter lists are longer but declare their
  requirements more explicitly.
- Expanded documentation of most functions with skydoc docstrings and
  type annotations.
- Cross-compilation of cgo targets is not supported.  Assembly files
  are permitted but must be consistent with the target architecture.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -19,4 +19,5 @@ Jake Voytko <jake@reviewninja.com>
 Justine Alexandra Roberts Tunney <jart@google.com>
 Kristina Chodorow <kchodorow@google.com>
 Lukacs Berki <lberki@google.com>
+Paul Johnston <pcj@pubref.org>
 Yuki Yugui Sonoda <yugui@yugui.jp>
--- a/README.md
+++ b/README.md
@@ -1,15 +1,13 @@
 # Go rules
 
-<div class="toc">
-  <h2>Rules</h2>
-  <ul>
-    <li><a href="#go_prefix">go_prefix</a></li>
-    <li><a href="#go_library">go_library</a></li>
-    <li><a href="#cgo_library">cgo_library</a></li>
-    <li><a href="#go_binary">go_binary</a></li>
-    <li><a href="#go_test">go_test</a></li>
-  </ul>
-</div>
+| ---: | :--- |
+| [go_prefix](#go_prefix) | Declare an import prefix. |
+| [go_library](#go_library) | Compile a pure go library. |
+| [cgo_library](#cgo_library) | Build a cgo library. |
+| [go_binary](#go_binary) | Build an executable binary file. |
+| [xgo_binary](#xgo_binary) | Build a cross-compiled binary file. |
+| [go_test](#go_test) | Generate a test. |
+
 
 ## Overview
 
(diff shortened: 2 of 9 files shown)


bazel-io:

pcj:
I know, it's a big PR.  As I alluded, hard to do incrementally.  Clearly, it's a fair amount of work to break it down into smaller steps.  Before doing that, can I get some confirmation that:

1. It's OK to partition def.bzl into smaller parts.  

2. Support for cross-compilation is desirable.  Obviously I think so (particularly cross-compile + bazel's deterministic docker images = holy grail), but I'd hate to do all this work and realize nobody else wants it.

I would proceed by:

1. Partitioning def.bzl.
2. Adding go_root.bzl.
3. Refactor environment var handling.
4. Reorganizing def.bzl.
5. Refactoring go_binary_impl and go_library_impl.
6. Adding xgo_binary.

pmbethe09:
Everything fine by me, and I would happily review the smaller PRs, but lets
wait for @yugui to opine on the def.bzl refactor

pcj:
OK thanks Paul.  Those PR's won't exactly be tiny, but thank you for offering to review.

yugui:
1. I agree about splitting the `bzl` file. 
   Actually I have been also thinking about it. So this is a good opportunity to split.
2. I am still not confident about if it is the right solution to have another rule for cross compilation.
   I understand the necessity of cross compilation and the use case described in [a comment](https://github.com/bazelbuild/rules_go/issues/70#issuecomment-246700574) by @pcj.
   But for me it looks to be similar to a `cc_library` dependency in `android_binary.deps` attr.
   So I still suspect if the right solution is to combine `--cpu` flag and `config_setting` to build a docker binary.
  
  I know the right solution might not work as we expected right now as @pcj reported in #70.
  And I am happy to accept a workaround solution for short term.  But I need to know what is the right solution which matches the other rules in Bazel ecosystem best.
  
  @damienmg Can I hear your opinion?  What is the right solution to cross compile a Linux binary to build a docker image?  Is it valid to have a special cross compilation rule which always emits a binary for a specific platform?

damienmg:
I don't think this is a right solution in the long term but this might be the only option right now.

/cc @katre @lberki for platform / cross compilation

damienmg:
(Sorry for the delay, I totally missed that a long time ago)

apesternikov:
Is there any chance to see it in? This is exactly what we need to build docker images as a part of a large deployment in a single bazel run.

jayconrod:
@apesternikov We're waiting for Bazel to complete support for [platforms and toolchains](https://docs.google.com/document/d/1-G-VPRLEj9VyfC6VrQBiR8To-dZjnBSQS66Y4nargGM/edit) before proceeding with cross-platform support. I expect we'll have some level support in late Q2 or early Q3.

You might want to follow #70 for more information.

ianthehat:
Cross compilation is now working using the new bazel toolchains in #625, and def.bzl is now broken up.
I think that's everything that was in this pull request that we did not want to lose track of.

Issue #97, opened 2016-09-13
cgo_library does not respect `includes` of cc_library dependencies


shahms:
We're trying to build `levigo` (Go bindings for LevelDB) as a `cgo_library()` rule.  The issue is that the LevelDB `cc_library` rule needs to have: `includes = ["include"],` for it and its dependents to build successfully.  This (and other) target-specific flags are not being passed to the `go tool cgo` invocation.  As a result, building the levigo rule is failing with:

go/batch.go:4:11: fatal error: 'leveldb/c.h' file not found

Given the symlink-forest shenanigans done by the go rules, I suspect it is far from straightforward to just slap the `cc.compiler_flags` into the cgo command line.

pcj:
I'm no cgo expert, but I was looking at this code recently so I'm interested in clarifying the requirement.  

cgo gathers headers and linkopts by iterating the transitive set of cc dependencies like so:

```python
  srcs = ctx.files.srcs + ctx.files.c_hdrs
  linkopts = ctx.attr.linkopts
  deps = set([], order="link")
  for d in ctx.attr.deps:
    srcs += list(d.cc.transitive_headers)
    deps += d.cc.libs
    for lib in d.cc.libs:
      if lib.basename.startswith('lib') and lib.basename.endswith('.so'):
        dirname = _short_path(lib)[:-len(lib.basename)]
        linkopts += ['-L', dirname, '-l', lib.basename[3:-3]]
      else:
        linkopts += [_short_path(lib)]
```

Headers get symlinked as you say with:

```python
  tree_layout = {}
  for s in srcs:
    tree_layout[s.path] = _short_path(s)
```

Symlinking includes would be tricky as they are strings and not file labels.

The linkopts go in an environment variable `"CGO_LDFLAGS": " ".join(linkopts),` 

So do you need the transitive set of imports added?  What would an example 'go tool cgo' command look like?  

Also, why is leveldb/c.h not being picked up by `ctx.files.c_hdrs`?  This ultimately comes from `cgo_library.srcs`, filtered by `c_hdrs = [s for s in srcs if any([s.endswith(ext) for ext in hdr_exts])]`.

shahms:
[...]

PR #99, opened 2016-09-13
Include execroot-relative include paths in cgo command line

shahms:
I'm not sure this is the *correct* fix for #97 , but it resolves the issue for us.


bazel-io:

yugui:
Jenkins, test this please

PR #101, opened 2016-09-14
Handle relative import paths.

melindalu:


bazel-io:

PR #102, opened 2016-09-15
Handle relative import paths.

melindalu:
We ran into a problem with imports using relative paths to reference packages in the same repository of the form "../othersubpackage". @yugui @pmbethe09 does this look reasonable?


bazel-io:

melindalu:
`I signed it!`

yugui:
@melindalu 
LGTM, but could you name the importpath `lib/relativeimporter` or just `lib/relative` instead of`lib/relative_importer` to avoid underscore character?

Jenkins, test this please.

melindalu:
@yugui good point, thanks — changed.

pmbethe09:
LGTM

PR #108, opened 2016-09-16
Gazelle: add filegroup for .proto whenever .pb.go found

pmbethe09:
this is particularly useful for github.com/golang/protobuf
which has things like //ptypes/any/any.pb.go and //ptypes/any/any.proto

This change detects the .pb.go, and then adds a filegroup called "go_default_library_protos" for the .proto files.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -1046,7 +1046,7 @@ _go_repository_attrs = {
         allow_files = True,
         single_file = True,
         executable = True,
-	cfg = "host",
+        cfg = "host",
     ),
 }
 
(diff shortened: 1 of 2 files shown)


pmbethe09:
This is a small hack, but allows new_go_repository on 
github.com/golang/protobuf
which has been resistant to:
https://github.com/golang/protobuf/pull/224

yugui:
LGTM.

But this issue with `.proto` can be a part of a larger issue:
* sometimes the remote repository contains non-go targets but there's no generic way to generate or manually provide such targets unlike [reconcilation](https://github.com/bazelbuild/rules_go/commit/7142fa959373a237a964016ed04ef721eb2bd0d4) of local repositories.

So let's revisit more generic solutions if we see similar issues in other types of build targets in future.

pmbethe09:
SGTM.

PR #110, opened 2016-09-20
[WIP] Resolves a warning about the workspace name:

yugui:
"Workspace name in .../external/io_bazel_rules_go_toolchain/WORKSPACE
(@golang_darwin_amd64) does not match the name given in the
repository's definition (@io_bazel_rules_go_toolchain); this will
cause a build error in future versions."

The change was extracted from #91 by Paul Cody Johnston.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -1001,7 +1001,11 @@ def _go_repository_select_impl(ctx):
   else:
     fail("unsupported operating system: " + os)
 
-  ctx.symlink(goroot, ctx.path(''))
+  gobin = goroot.get_child("bin")
+  gopkg = goroot.get_child("pkg")
+  ctx.symlink(gobin, "bin")
+  ctx.symlink(gopkg, "pkg")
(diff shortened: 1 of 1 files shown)


yugui:
@pcj -- I have extracted your fix of the warning from #91 so that it gets merged earlier.

Hmm, cla/google requires the committer and the author of the commit match?

pcj:
OK by me.

pmbethe09:
Something broke downloading Go in the tests

pcj:
Confirmed that this works on osx and breaks on linux.  The difference is that the GO_TOOLCHAIN build file get written to the symlinked directory in master but not in this PR.  I think it may be due to variable behavior of glob() not traversing symlinks across different packages https://www.bazel.io/versions/master/docs/be/functions.html#glob.  Looking into it...

pcj:
I'm have not resolved why this behavior differs between linux and osx, but consider PR #112 instead of this one.

yugui:
@pcj Thank you.

PR #116, opened 2016-09-26
Gazelle: search for repoRoot when flag not used.

pmbethe09:


pmbethe09:
@yugui after this I have follow up PRs ready for:
* read go_prefix if not set and available in the root BUILD file
* merge BUILD file updates with existing files
* [WIP] update WORKSPACE with any newly found 'new_go_repository' dependencies.

Also I think the default -mode should be fix, so that you can just run:
gazelle
and have that just work.

go/tools/gazelle/wspace/finder_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"testing"
+)
+
+type testCase struct {
+	dir  string
+	want string // "" means should fail
+}
+
+func TestFind(t *testing.T) {
+	tmp, err := ioutil.TempDir("", "")
```
yugui: Could you use `os.Getenv("TEST_TMPDIR")` as the first argument?

go/tools/gazelle/wspace/finder_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp)
+	if err := os.MkdirAll(filepath.Join(tmp, "base", "sub"), 0755); err != nil {
+		t.Fatal(err)
+	}
+	if err := ioutil.WriteFile(filepath.Join(tmp, "base", workspaceFile), []byte{}, 0755); err != nil {
```
yugui: You can safely pass `nil` instead of `[]byte{}`.
You don't need to instantiate an empty underlying array.

go/tools/gazelle/wspace/finder_test.go
pmbethe09: done.

go/tools/gazelle/wspace/finder_test.go
pmbethe09: done.

damienmg:
This PR broke all tests on ci.bazel.io, see http://ci.bazel.io/job/rules_go/355/

pmbethe09:
I don't think this PR was the cause:
ERROR: /home/ci/workspace/rules_go/BAZEL_VERSION/HEAD/PLATFORM_NAME/linux-x86_64/examples/cgo/BUILD:7:1: in copts attribute of cc_library rule //examples/cgo:cgo_lib.cgo.dir/gen/_cgo_lib: unterminated quotation. Since this rule was created by the macro 'cgo_library', the error might have been caused by the macro implementation in /home/ci/workspace/rules_go/BAZEL_VERSION/HEAD/PLATFORM_NAME/linux-x86_64/go/def.bzl:927:15.

I think it was the previous one.

PR #121, opened 2016-09-28
add go_proto_library skylark rule and examples

pmbethe09:
This is a basic go_proto_library with gRPC support.

Caveat: only supports the bazel-style naming convention "foo_proto", and does not yet handle the Go-style of directory=package that we use "go_default_library" for.

### DIFF
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -64,9 +64,12 @@ def _proto_import(deps):
 
 def _go_proto_library_gen_impl(ctx):
   """Rule implementation that generates Go using protoc."""
+  bazel_style = ctx.label.name != "go_default_library_protos"
   protos = list(ctx.files.src)
-  go_package_name = ctx.label.name[:-len(_PROTOS_SUFFIX)]
-  m_import_path = ",".join(["M%s=%s%s/%s" % (f.path, _go_prefix(ctx), 
+  go_package_name = "/" + ctx.label.name[:-len(_PROTOS_SUFFIX)]
+  if not bazel_style:
+    go_package_name = ""
+  m_import_path = ",".join(["M%s=%s%s%s" % (f.path, _go_prefix(ctx), 
                                            ctx.label.package, go_package_name)
                           for f in ctx.files.src])
   for d in ctx.attr.deps:
@@ -80,10 +83,14 @@ def _go_proto_library_gen_impl(ctx):
   if ctx.attr.grpc:
     use_grpc = "plugins=grpc,"
 
+  offset = 0
+  proto_out = ctx.outputs.out
+  if bazel_style:
+    offset = -1  # extra directory added, need to remove
+    proto_out = ctx.new_file(
+        ctx.configuration.bin_dir, ctx.files.src[0].basename[:-5] + "pb.go")
   outdir = "/".join(
-      ctx.outputs.out.dirname.split("/")[:-1-len(ctx.label.package.split("/"))])
-  proto_out = ctx.new_file(
-      ctx.configuration.bin_dir, ctx.files.src[0].basename[:-5] + "pb.go")
+      ctx.outputs.out.dirname.split("/")[:offset-len(ctx.label.package.split("/"))])
   ctx.action(
       inputs=protos + ctx.files.protoc_gen_go,
       outputs=[proto_out],
@@ -99,14 +106,15 @@ def _go_proto_library_gen_impl(ctx):
       executable=ctx.executable.protoc)
   # This is the current hack for files without 'option go_package'
   # Generate into .pb.go, then cp into "real location"
-  ctx.action(
-      inputs=[proto_out],
-      outputs=[ctx.outputs.out],
-      command="cp %s %s" % (proto_out.path, ctx.outputs.out.path),
-      mnemonic="GoProtocGenCp")
+  if bazel_style:
+    ctx.action(
+        inputs=[proto_out],
+    	outputs=[ctx.outputs.out],
+      	command="cp %s %s" % (proto_out.path, ctx.outputs.out.path),
+      	mnemonic="GoProtocGenCp")
   return struct(_protos=protos,
                 _m_import_path=m_import_path,
-                name=(ctx.label.package + "/" + go_package_name))
+                name=(ctx.label.package + go_package_name))
 
 go_proto_library_gen = rule(
     attrs = {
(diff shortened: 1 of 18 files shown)


yugui:
Do we need to have `go_proto_library` in this repository even though we also have https://github.com/pubref/rules_protobuf/tree/master/go ?

hochhaus:
> Do we need to have go_proto_library in this repository even though we also have https://github.com/pubref/rules_protobuf/tree/master/go?

From my perspective, yes. Having a single canonical set of bazel go rules that the community can coalesce behind is tremendously valuable. The most logical location for `go_proto_library()` is `rules_go` as this repository is maintained as part of the bazel project.

pmbethe09:
@yugui The discussion among my colleagues was that we wanted a "pure" go_proto_library that we controlled and thus could easily use internally, and as @hochhaus wrote was just an add-on to rules_go.

proto/go_proto_library.bzl
```diff
@@ -0,0 +1,238 @@
...
+    deps: a mixed list of either go_proto_libraries, or
+          any go_library which has a companion
+          filegroup(name=name+"_protos",...)
+          which contains the protos which were used
+    has_services: indicates the proto has gRPC services and deps
+    testonly: mark as testonly
+    visibility: visibility to use on underlying go_library
+  """
+  if not name:
+    fail("name is required")
```
yugui: Then, why do you provide the default value of `name`?

proto/go_proto_library.bzl
```diff
@@ -0,0 +1,238 @@
...
+  if not srcs or len(srcs) != 1:
+    fail("exactly 1 src is required")
```
yugui: Use the second parameter of `fail` to report the source attribute name of the error.

proto/go_proto_library.bzl
```diff
@@ -0,0 +1,238 @@
...
+    ctx.action(
+        inputs=[proto_out],
+      outputs=[ctx.outputs.out],
+        command="cp %s %s" % (proto_out.path, ctx.outputs.out.path),
+        mnemonic="GoProtocGenCp")
+  return struct(_protos=protos,
+                _m_import_path=m_import_path,
+                name=(ctx.label.package + go_package_name))
+
+go_proto_library_gen = rule(
```
yugui: Can we make this rule private?

proto/go_proto_library.bzl
```diff
@@ -0,0 +1,238 @@
...
+    attrs = {
+        "deps": attr.label_list(),
+        "src": attr.label(
+            mandatory = True,
+            single_file = True,
```
yugui: Let me confirm.
IIRC, you cannot apply `protoc` code generation one-by-one to `.proto` files. Is it the reason why you decided to accept only one `srcs` item in `go_proto_library`?

When `foo.proto` imports `bar.proto` in the same Go package, `protoc-gen-go` generates a wrong import of the Go package itself for `foo.proto` unless you pass the two `.proto` files together.

yugui:
@hochhaus , @pmbethe09 
I see.  Then, it is OK for this specific PR.

But I want Bazel team to show their plan/direction about supporting protobuf in other languages and keeping their integrity. What I want to avoid is that individual language supports have their own incompatible ways of bundling `.proto` files, which makes it hard to refer `.proto` files in external repositories.

damienmg:
@pmbethe09: I am a bit overloaded right now and I have very few expertise to share for this PR, sorry :(

I would suggest adding @cgrushko for protobuf and maybe also @lberki

cgrushko:
Sorry about the uncertainty; I'll try to describe our plans re protos.

In short, we'll use language-agnostic `proto_library` rules to describe what a `.proto` file depends on, and language-specific rules ("LANG_proto_library") that depend on `proto_library`-trees.

Bazel currently has native `proto_library`, `java_proto_library` and `java_lite_proto_library` rules, but they're still experimental and so I didn't announce them to avoid stable users relying on them.
Their source-code shows how to use them.
(There are several bugs/blockers that prevent me from launching them, one of them is https://github.com/bazelbuild/bazel/issues/1718)

The new rules use Aspects (https://bazel.io/docs/skylark/aspects.html) in order to be incremental and cacheable, so they're in general superior to `genrule()`-based solutions.

~~~

As for gRPC:

I strongly recommend to separate gRPC codegen from proto-buffer codegen (in the "message" sense). We do so internally, and currently working on a Go gRPC rule. 
@MakMukhi leads that effort.

~~~

As for this PR - @damienmg, I don't know anything on Go compilation (yet), so might not be a useful reviewer.

proto/go_proto_library.bzl
pmbethe09: a srcs list matches the style of other *_proto_library -- for now I only accept 1 src, but I can imagine changing/fixing that.
The reason is one only of current needs.

As to the other thing, yes you need all transitive protos in the action at generation time and their import mapping, but that is handled via the '_protos' rule provider.

proto/go_proto_library.bzl
pmbethe09: habit.  Note that the user could pass name = "" and we still need to fail on that.

proto/go_proto_library.bzl

pmbethe09:
@cgrushko interesting.  
As opposed to other languages, in the case of go_proto_library, gRPC adds about 3 lines of skylark code.  This seems much simpler than duplicating or figuring out a factoring.

No aspects used, but also no genrule.  For generation there is a rule which has '_protos' and  '_m_import_path' providers to scoop up the transitive closure of protos which seems sufficient.

I wasn't aware of the desire to have the LANG_proto_library depend on the .proto tree of proto_library, but could make a future change to support that.

kchodorow:
@cgrushko <https://github.com/cgrushko>, you know that proto_library is
documented, right?  (http://bazel.io/docs/be/protocol-buffer.html)  Do you
want that documentation removed?

cgrushko:
@kchodorow `BazelProtoLibraryRule` calls `setUndocumented()`, so I'm surprised it is documented.
Yes, we should make it undocumented, as it's not ready for wide consumption yet.

proto/go_proto_library.bzl
```diff
@@ -0,0 +1,239 @@
...
+          any go_library which has a companion
+          filegroup(name=name+"_protos",...)
+          which contains the protos which were used
+    has_services: indicates the proto has gRPC services and deps
+    testonly: mark as testonly
+    visibility: visibility to use on underlying go_library
+    **kwargs: any other args which are passed through to the underlying go_library
+  """
+  if not name:
+    fail("name is required", name)
```
yugui: nit.
`fail("name is required", "name")`

proto/go_proto_library.bzl
```diff
@@ -0,0 +1,239 @@
...
+  if not srcs or len(srcs) != 1:
+    fail("exactly 1 src is required", srcs)
```
yugui: `fail("exactly 1 src is required", "srcs")`

Issue #124, opened 2016-10-03
cgo example has invalid quotes


kchodorow:
22ababf573115a40fbfc79b83f264668185f2d42 seems to have broken `bazel build //...`.  See http://ci.bazel.io/job/rules_go/BAZEL_VERSION=HEAD,PLATFORM_NAME=linux-x86_64/351/console:

```
WARNING: /home/ci/.cache/bazel/_bazel_ci/18356ce76ec27b0c4bd1f77ab8d37827/external/golang_linux_amd64/WORKSPACE:1: Workspace name in /home/ci/.cache/bazel/_bazel_ci/18356ce76ec27b0c4bd1f77ab8d37827/external/golang_linux_amd64/WORKSPACE (@io_bazel_rules_go_toolchain) does not match the name given in the repository's definition (@golang_linux_amd64); this will cause a build error in future versions.
ERROR: /home/ci/workspace/rules_go/BAZEL_VERSION/HEAD/PLATFORM_NAME/linux-x86_64/examples/cgo/BUILD:7:1: in copts attribute of cc_library rule //examples/cgo:cgo_lib.cgo.dir/gen/_cgo_lib: unterminated quotation. Since this rule was created by the macro 'cgo_library', the error might have been caused by the macro implementation in /home/ci/workspace/rules_go/BAZEL_VERSION/HEAD/PLATFORM_NAME/linux-x86_64/go/def.bzl:927:15.
ERROR: Analysis of target '//examples/cgo:cgo_lib_test' failed; build aborted.
____Elapsed time: 2.682s
ERROR: Couldn't start the build. Unable to run tests.
/home/ci/workspace/rules_go/BAZEL_VERSION/HEAD/PLATFORM_NAME/linux-x86_64/.bin/bazel: line 9: ((: != 0 : syntax error: operand expected (error token is "!= 0 ")
```

(Ignore that the CI thinks that the test is passing, filed https://github.com/bazelbuild/continuous-integration/issues/33 about that.)

jayconrod:
Closing old issues. Looks like this one is fixed.

PR #123, opened 2016-09-30
Apply the given copts to cgo code generation

yugui:
Fixes #119

Issue #119, opened 2016-09-28
cgo_library doesn't pass copts down to cgo


ecaraszi:
Hello,

Thanks for making rules_go!   Here's a small bug report & potential fix:  

Bug: cgo_binary doesn't pass copts down to the compiler flags in the cgo invocation.

Fix:  github.com/ecaraszi/rules_go on branch cflags 

I used this patch to successfully build the github.com/go-gl libraries (glfw 3.2, gl 4.1-core, and mathgl) and link a working example.

pmbethe09:
We welcome contributions, so please make a PR

jmhodges:
(Sorry, just poking around in other cgo tickets. This maybe can be closed again now?)

jayconrod:
Closing old issues. Seems like this one is fixed.

PR #126, opened 2016-10-04
Fixes a build error introduced in #123

### LINKED ISSUE #125: rules_go fails on ci.bazel.io
After PR #116. See http://ci.bazel.io/job/rules_go/355/

yugui:
Fixes #125 
c.f. #124


yugui:
This is a quick fix of #125.
The root cause of the issue is probably an inconsistency of `copt` tokenization between `CGoCodeGen` and `cc_library`.
I'll revisit the root cause later together with other inconsistency of `$(location)` and Make variables in `copt` and other options.

Reopens #119 because #125 shows that there's no way to define string macro.

PR #130, opened 2016-10-07
allow go_proto_library to accept multiple proto srcs

pmbethe09:


examples/proto/lib/lib2.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package examples.lib;
+
+import "examples/proto/dep/useful.proto";
+import "ptypes/wrappers/wrappers.proto";
+
```
yugui: Could you import and use `lib.proto`?
IIRC, one-by-one compilation of proto would also work fine unless one .proto imports another .proto in the same package.

pmbethe09: done

PR #133, opened 2016-10-11
support example only test packages

mikedanese:
Without this change, you would get an unused import error. It's common to have empty test packages with only GoDoc Example functions.


pmbethe09:
Jenkins, test this please.

PR #128, opened 2016-10-05
Gazelle: read go_prefix from top-level BUILD if unset and available.  Merge BUILD content with existing

pmbethe09:

PR #139, opened 2016-10-12
Gazelle: make the default mode 'fix'

pmbethe09:
After this and the go_prefix PR https://github.com/bazelbuild/rules_go/pull/128 in most cases after the initial setup, you can just run 'gazelle' with no arguments.

### DIFF
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -32,7 +32,7 @@ import (
 var (
 	goPrefix = flag.String("go_prefix", "", "go_prefix of the target workspace")
 	repoRoot = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
-	mode     = flag.String("mode", "print", "print, fix or diff")
+	mode     = flag.String("mode", "fix", "print, fix or diff")
 )
 
 var modeFromName = map[string]func(*bzl.File) error{
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -32,7 +32,7 @@ import (
 var (
 	goPrefix = flag.String("go_prefix", "", "go_prefix of the target workspace")
 	repoRoot = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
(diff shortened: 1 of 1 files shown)


go/tools/gazelle/gazelle/main.go
```diff
@@ -32,7 +32,7 @@ import (
 var (
 	goPrefix = flag.String("go_prefix", "", "go_prefix of the target workspace")
 	repoRoot = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
-	mode     = flag.String("mode", "print", "print, fix or diff")
+	mode     = flag.String("mode", "fix", "print, fix or diff")
```
damienmg: Could you also expand the "print, fix or diff" to explain a bit more what each one does?

I guess fix fix the exiting BUILD file, print output the generated build file and diff shows a diff between the generated and the existing (or is it a diff corresponding to what fix do?)

pmbethe09: yes-ish.
before https://github.com/bazelbuild/rules_go/pull/128
print prints what would be written, fix actually overwrites, diff shows the diff of the overwrite to current.

after https://github.com/bazelbuild/rules_go/pull/128
print prints what was generated, 
diff shows the diff of the overwrite to current.
while fix, will do a "merge" in place.

So I suppose 128 should be fixed to do the merge regardless, and then do one of the three operations.

pmbethe09: Fixed https://github.com/bazelbuild/rules_go/pull/128 to merge in all cases.

Added more info on each mode.

go/tools/gazelle/gazelle/main.go
```diff
@@ -32,7 +32,7 @@ import (
 var (
 	goPrefix = flag.String("go_prefix", "", "go_prefix of the target workspace")
 	repoRoot = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
-	mode     = flag.String("mode", "fix", "print, fix or diff")
+	mode     = flag.String("mode", "fix", "print: which prints all of the updated BUILD files\n\tfix: which rewrites all of the BUILD files in place\n\tdiff: which computes the rewrite but then just does a diff")
```
damienmg: nit: remove which for all sentence :)

pmbethe09: done

PR #145, opened 2016-10-14
Add design document for cross-compile support.

pcj:


bazel-io:

pcj:
I thought this would be appropriate to write up to be able to see where this is going and gather feedback.  If others think a different forum such as a google doc would be better let me know.

yugui:
@damienmg 
Would you mind taking a look at the "questions" section and give him some suggestions?
https://github.com/bazelbuild/rules_go/pull/145/files#diff-bda770a36978661048770947bad7eb6cR74

I want to go along with the design of cross-compile in Bazel. But I also don't know much about how we can combine `--cpu` or related flags with cross-compile supports in Skylark rules.
Once we agree on cross-compile support in Skylark rules in general, I can comment on Go-specific designs.

damienmg:
@lberki is probably a better person to review that.

go/cross-compilation.md
```diff
@@ -0,0 +1,92 @@
...
+  inherent to go itself, cross-compilation does not seem to work when
+  invoked via a non-absolute path (see
+  https://github.com/golang/go/issues/17017).
+
+* In `def.bzl`, `GOOS` and `GOARCH` are already included in the
+  environment of ctx.actions.  Add `GOROOT` into that dict, and change
+  all invocations of `go tool CMD` to use `$GOROOT/bin/go` rather than
+  `('../' * out_depth) + ctx.file.go_tool.path`.
+
+* Add a non-mandatory string attribute `os_arch` to the go_binary
```
lberki: Would `os_arch` designate the target architecture?

Then this would mean that the Go rules would work differently from any other rule (which always target the target platform).

ianthehat:
Cross compilation with toolchains in #625 should meet all the design requirements listed in this doc.
Being explicit about a target platform for a specific binary is not ready yet, at the moment toolchains are picked for the entire build, but it's planned and not a go specific feature, and the current design and implementation should just work when that feature arrives.

PR #150, opened 2016-10-17
update README.md for tag 0.3.0

pmbethe09:


yugui:
The change itself LGTM.

Would you mind also describing `go_root`?

pcj:
As `go_root` is primarily for internal use I don't think it warrants public documentation at the moment.

PR #154, opened 2016-10-19
Add well-known special case paths to externalResolver

pmbethe09:
This avoids a lot of network calls in vcs.RepoRootForImportPath.

Testing in repos like google.golang.org/genproto took
runtime from ~9s on average to ~0.01s

### DIFF
--- a/go/tools/gazelle/rules/resolve_external.go
+++ b/go/tools/gazelle/rules/resolve_external.go
@@ -34,9 +34,12 @@ type externalResolver struct{}
 // recommended reverse-DNS form of workspace name as described in
 // http://bazel.io/docs/be/functions.html#workspace.
 func (e externalResolver) resolve(importpath, dir string) (label, error) {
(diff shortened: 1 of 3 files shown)


go/tools/gazelle/rules/resolve_external.go
```diff
@@ -61,3 +64,23 @@ func (e externalResolver) resolve(importpath, dir string) (label, error) {
...
+var knownImports = []string{"golang.org/x/", "google.golang.org/", "cloud.google.com/"}
+
+// specialCases looks for matches in knownImports to avoid making a network call.
+func specialCases(importpath string) string {
+	for _, known := range knownImports {
+		if !strings.HasPrefix(importpath, known) {
+			continue
+		}
+		l := len(known)
+		if idx := strings.Index(importpath[l:], "/"); idx != -1 {
```
pmbethe09: I did...

PR #158, opened 2016-10-21
A simple script for adding new_go_repository entries to your WORKSPACE

pmbethe09:
e.g.
wtool com_github_golang_glog 
will add

new_go_repository(
    name = "com_github_golang_glog",
    commit = <latest commit>,
    importpath = "github.com/golang/glog",
)

using the latest commit.


go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+		X: &bzl.LiteralExpr{Token: "new_go_repository"},
+		List: []bzl.Expr{
+			binExpr("name", name),
+			binExpr("importpath", importpath),
+			binExpr("commit", commit),
+		},
+	}, nil
+}
+
+func binExpr(key, val string) *bzl.BinaryExpr {
```
yugui: Can you name the function `attr` or something because actually it returns something more specific than just a binary expression?

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
+/*wtool augments your bazel WORKSPACE file with new_go_repository entries
+
+Example Usage: wtool com_github_golang_glog com_google_cloud_go
+will add 2 new_go_repository to your WORKSPACE
+by converting com_github_golang_glog -> github.com/golang/glog
+and so forth and then doing a 'git ls-remote' to get
+the latest commit.
+
+if wtool cannot figure out the bazel -> Go mapping, try
+Other Usage: wtool -asis github.com/golang/glog
```
yugui: I don't think it would not be a problem for users even if `wtool` always behave `-asis`.
Also if it did so, we could avoid assuming the first two components of the given repository name were host names.

Why is the default mode better than `-asis`?

pmbethe09: The way I use it, I add some dependency in my Go code, say on google.golang.org/genproto/protobuf

I run gazelle, then build and bazel dies saying
.../BUILD:3:1: no such package '@org_golang_google_genproto//protobuf': error loading package 'external': The repository named 'org_golang_google_genproto' could not be resolved and referenced by ...

So I now just type:
 wtool org_golang_google_genproto

Without having to reverse engineer the importpath myself.

go/tools/wtool/main.go
pmbethe09: done.

PR #164, opened 2016-10-23
perform the main sandwich for go_binary

### LINKED ISSUE #157: Teach gazelle the "main sandwich"
e.g. https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/gazelle/BUILD

gazelle should be able to generate this BUILD file, but it doesn't know how to make the go_library for testing, and then use 'library =' to create the go_binary rule.

jmhodges:
When gazelle runs on a Go main package, it now creates a private
go_library for the package so that any tests in that main package can be
run on it.

Previously, gazelle was creating only a go_binary for the go_test to
depend on which fails at `bazel test` time.

Fixes #157

### DIFF
--- a/go/tools/gazelle/generator/generator_test.go
+++ b/go/tools/gazelle/generator/generator_test.go
@@ -65,7 +65,29 @@ func TestGenerator(t *testing.T) {
 			"bin": {
 				{
 					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
 						X: &bzl.LiteralExpr{Token: "go_binary"},
+					},
+				},
+			},
+			"bin_with_tests": {
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_binary"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_test"},
 					},
 				},
 			},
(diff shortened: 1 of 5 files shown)


bazel-io:

jmhodges:
(I'm porting over a monorepo to bazel, and this cropped up in a bunch of places, so I wrote the code.)

yugui:
LGTM.
Jenkins, test this please.

pmbethe09:
Sorry but this doesn't work.  For some reason the go_binary needs the same dependencies as the go_library at link time, otherwise the compilation fails.

Please add a test for this next time.  The simplest might be calling a go_binary from a new_go_repository that is known to have dependencies.

jmhodges:
Could you give me an example of what broke? I've been using this code all weekend without running into a problem. I've building binaries in external repos and I'm not clear when else you would do with them that would highlight this problem.

jmhodges:
On my Mac, I've had no issues with go_binary from inside a new_go_repository. For instance, cmd/goose in bitbucket.org/liamstask/goose builds fine and it has a number of dependencies. Is there some difference in how rules_go is doing linking on Linux or something?

jmhodges:
(Note: goose is broken on linux for other reasons. Specifically, it uses build tags in cgo and we don't support that.)

pmbethe09:
I reverted the revert, and sent out
https://github.com/bazelbuild/rules_go/pull/170

which fixes the merge issue.

pmbethe09:
I see what happened.  The fix did not merge well with existing BUILD files
with the old-style go_binary

For an existing
go_binary(
    name = "main",
    srcs = ["main.go"],
    deps = some_deps,
)

gazelle didn't scrub the srcs in the go_binary, and for some reason it
didn't set 'library' so this is what was re-generated.

go_binary(
    name = "main",
    srcs = ["main.go"],
    deps = ["go_default_library"],
)

go_library(
    name = "go_default_library",
    srcs = ["main.go"],
    deps = some_deps,
)

when what should have been generated for the binary was:

go_binary(
    name = "api",
)

Note: if I delete the file and regenerate, it does populate 'library', but
it also incorrectly populates deps = ["go_default_library"],
The new code failed to deal with the migration.

PR #166, opened 2016-10-23
add cgo support 

### LINKED ISSUE #156: force cgo in new_go_repository?
First up, thanks so much for all this work. It's been a real joy to use.

Okay, so when I use `new_go_repository` on a project requiring cgo, it seems to miss that there are c files it needs.

A relatively common example of this is

    new_go_repository(
        name = "com_github_mattn_go_sqlite3",
        importpath = "github.com/mattn/go-sqlite3",
        commit = "7204887cf3a42df1cfaa5505dc3a3427f6dded8b",
    )

and running a build command on its targets.

I'm not quite sure what would be the best way to solve it. Looking for cgo code every time that bazel hits this repo seems like a likely non-starter. A param to specify a list of cgo packages inside that repo that we pass into new_go_repository? Or maybe the long term solution is to just not work on cgo repositories and have people specify them by hand?

jmhodges:
This adds cgo support to gazelle.

A cgo_library rule called cgo_default_library is created if there are
any CFiles, HFiles, or CgoFiles in the package. If there are no GoFiles,
a go_library without srcs is added but with library set to
cgo_default_library.

If a cgo_library was created, we move the assembly SFiles over to the
cgo_library to make sure they are in scope for that code.
   
This change also makes sure we include the Go-expected capital S
extension of assembly files.

This change depends on the main sandwich PR #164 and the cgo_library in
root of repo workaround #160.

Fixes #156

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -626,9 +626,11 @@ def _cgo_codegen_impl(ctx):
   for s in srcs:
     tree_layout[s.path] = _short_path(s)
 
+  p = _pkg_dir(ctx.label.workspace_root, ctx.label.package) + "/"
+  if p == "./":
+    p = "" # workaround when cgo_library in repository root
   out_dir = (ctx.configuration.genfiles_dir.path + '/' +
-             _pkg_dir(ctx.label.workspace_root, ctx.label.package) + "/" +
-             ctx.attr.outdir)
+             p + ctx.attr.outdir)
   cc = ctx.fragments.cpp.compiler_executable
   cmds = symlink_tree_commands(out_dir + "/src", tree_layout) + [
       "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -937,6 +937,7 @@ def cgo_library(name, srcs,
       copts = copts + [
           "-I", pkg_dir,
           "-I", "$(GENDIR)/" + pkg_dir + "/" + cgogen.outdir,
+          "-pthread",
           # The generated thunks often contain unused variables.
           "-Wno-unused-variable",
       ],
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -934,6 +934,7 @@ def cgo_library(name, srcs,
       name = cgogen.outdir + "/_cgo_lib",
       srcs = cgogen.c_thunks + cgogen.c_exports + c_srcs + c_hdrs,
       deps = cdeps,
+      # TODO(bazel-team): use -mthreads when windows support is added
       copts = copts + [
           "-I", pkg_dir,
           "-I", "$(GENDIR)/" + pkg_dir + "/" + cgogen.outdir,
--- a/go/tools/gazelle/generator/generator_test.go
+++ b/go/tools/gazelle/generator/generator_test.go
@@ -65,7 +65,29 @@ func TestGenerator(t *testing.T) {
 			"bin": {
 				{
 					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
 						X: &bzl.LiteralExpr{Token: "go_binary"},
+					},
+				},
+			},
+			"bin_with_tests": {
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_binary"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_test"},
 					},
 				},
 			},
@@ -125,8 +147,18 @@ func TestGenerator(t *testing.T) {
 		{
 			Path: "bin/BUILD",
 			Stmt: []bzl.Expr{
-				loadExpr("go_binary"),
+				loadExpr("go_library", "go_binary"),
 				stub.fixtures["bin"][0].Call,
+				stub.fixtures["bin"][1].Call,
+			},
+		},
+		{
+			Path: "bin_with_tests/BUILD",
+			Stmt: []bzl.Expr{
+				loadExpr("go_library", "go_binary", "go_test"),
+				stub.fixtures["bin_with_tests"][0].Call,
+				stub.fixtures["bin_with_tests"][1].Call,
+				stub.fixtures["bin_with_tests"][2].Call,
 			},
 		},
 	}
--- a/go/tools/gazelle/generator/generator_test.go
+++ b/go/tools/gazelle/generator/generator_test.go
@@ -125,23 +125,6 @@ func TestGenerator(t *testing.T) {
 					},
 				},
 			},
-			"bin_with_tests": {
-				{
-					Call: &bzl.CallExpr{
-						X: &bzl.LiteralExpr{Token: "go_library"},
-					},
-				},
-				{
-					Call: &bzl.CallExpr{
-						X: &bzl.LiteralExpr{Token: "go_binary"},
-					},
-				},
-				{
-					Call: &bzl.CallExpr{
-						X: &bzl.LiteralExpr{Token: "go_test"},
-					},
-				},
-			},
 		},
 	}
 
--- a/go/tools/gazelle/generator/generator_test.go
+++ b/go/tools/gazelle/generator/generator_test.go
@@ -125,6 +125,23 @@ func TestGenerator(t *testing.T) {
 					},
 				},
 			},
+			"bin_with_tests": {
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_binary"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_test"},
+					},
+				},
+			},
 		},
 	}
 
--- a/go/tools/gazelle/generator/generator_test.go
+++ b/go/tools/gazelle/generator/generator_test.go
@@ -217,7 +217,7 @@ func TestGenerator(t *testing.T) {
 	sort.Sort(fileSlice(want))
 
 	if !reflect.DeepEqual(got, want) {
-		t.Errorf("g.Generate(%q) = %s;\n\nwant:\n%s", repo, prettyFiles(got), prettyFiles(want))
+		t.Errorf("g.Generate(%q) = %s; want: %s", repo, prettyFiles(got), prettyFiles(want))
 	}
 }
 
@@ -226,9 +226,9 @@ type prettyFiles []*bzl.File
 func (p prettyFiles) String() string {
 	var items []string
 	for _, f := range p {
-		items = append(items, fmt.Sprintf("\t{Path: %q, Stmt: %s}", f.Path, string(bzl.Format(f))))
+		items = append(items, fmt.Sprintf("{Path: %q, Stmt: %s}", f.Path, string(bzl.Format(f))))
 	}
-	return fmt.Sprintf("[\n%s\n]", strings.Join(items, ",\n"))
+	return fmt.Sprintf("[%s]", strings.Join(items, ","))
 }
 
 type fileSlice []*bzl.File
--- a/go/tools/gazelle/generator/generator_test.go
+++ b/go/tools/gazelle/generator/generator_test.go
@@ -91,6 +91,40 @@ func TestGenerator(t *testing.T) {
 					},
 				},
 			},
+			"cgolib": {
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "cgo_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_test"},
+					},
+				},
+			},
+			"allcgolib": {
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "cgo_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_library"},
+					},
+				},
+				{
+					Call: &bzl.CallExpr{
+						X: &bzl.LiteralExpr{Token: "go_test"},
+					},
+				},
+			},
 		},
 	}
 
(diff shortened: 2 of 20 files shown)


bazel-io:

jmhodges:
I had need of this during my build tooling port to bazel, and it was easy enough to get done.

Thanks for the nice, hackable codebase.

yugui:
Thank you for your contribution.

Jenkins, test this please.

yugui:
Looks to be a temporary error?  http://ci.bazel.io/job/PR-rules_go/175/BAZEL_VERSION=latest,PLATFORM_NAME=ubuntu_15.10-x86_64/

Jenkins, retest this please.

yugui:
@jmhodges 
Could you add the license header comments to the example files as other examples have?

I thought about if we can skip generating an empty `go_default_library` for `allcgolib` and name the `cgo_library` `go_default_library` instead.  But it would make the implementation much harder to read and the current output is simple enough. So 👍  for the current design.

LGTM except the license header. I'll merge this PR as soon as the headers are added.

jmhodges:
I've added the license headers! That was a good catch.

I've also added basic handling of in-file cgo CFLAGS and LDFLAGS support. Sorry to put it in after your review, but I totally blanked on it.

To build things like `go-sqlite3` reliably across platforms, we'll still need tag support because they have directives like `#cgo linux LDFLAGS: -ldl` but this gives us a good base for that work.

go/tools/gazelle/rules/generator.go
```diff
@@ -148,26 +158,81 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
+	attrs = append(attrs, keyvalue{key: "srcs", value: srcs})
+
+	copts := append([]string{}, pkg.CgoCFLAGS...)
+	copts = append(copts, pkg.CgoCPPFLAGS...)
+	copts = append(copts, pkg.CgoCXXFLAGS...)
+	if len(copts) > 0 {
+		attrs = append(attrs, keyvalue{key: "copts", value: copts})
+	}
+	if len(pkg.CgoLDFLAGS) > 0 {
+		attrs = append(attrs, keyvalue{key: "linkopts", value: pkg.CgoLDFLAGS})
```
pmbethe09: this should use "clinkopts"  -- perhaps add a test that catches this, then fix?

go/tools/gazelle/rules/generator.go
```diff
@@ -148,26 +158,81 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
+	}
+
+	visibility := "//visibility:private"
+	visibility = checkInternalVisibility(rel, visibility)
```
pmbethe09: merge into 1 line instead

visibility := checkInternalVisibility(rel, "//visibility:private")

go/tools/gazelle/rules/generator_test.go
```diff
@@ -146,6 +146,69 @@ func TestGenerator(t *testing.T) {
...
+				cgo_library(
+					name = "cgo_default_library",
+					srcs = [
+						"foo.go",
+						"foo.c",
+						"foo.h",
+						"asm.S",
+					],
+					copts = ["-I/weird/path"],
+					linkopts = ["-lweird"],
```
pmbethe09: clinkopts

pmbethe09:
Also, when I run this against github.com/mattn/go-sqlite3
I get a number of pthread errors like:
sqlite3-binding.c:(.text+0x40f4): undefined reference to `pthread_mutexattr_init'

There is not mention of '-lpthread' in any .go files, yet 'go build' works fine.
This leads me to think that on linux '-lpthread' should be added by default?

go/tools/gazelle/rules/generator.go
jmhodges: You mean a test different than the ones in generator_test.go?

Is there a high-level "build this repo" test I can steal ideas from in here?

jmhodges:
Handled the easy feedback. Off to lunch before the complicated one. Hoping y'all have a bead on where a high-level build repo test might be at that I can use.

jmhodges:
Thanks. Any idea how this integration test would work? Is there something in the repo already I've not seen?

jmhodges:
A couple of new directories in the examples directory where one depends on the other as a new_go_repository?

jmhodges:
Pushed up the typo fix.

Hm, assuming you wanted a real test, doing the remote but BUILD-less cgo example will require a dependency on an actual external repo because new_go_repository checks to make sure that the importpath given it is at the root and checks the repo name which would have to be the same as the one the examples are running in. Plus, you know, that'd be network IO in the tests

Otherwise, we'd have to run gazelle by hand in some directory.

Neither of these are ideal. Do you have a preference?

jmhodges:
Okay, I figured the network IO wouldn't be a problem since it's just a dependency.

I have a branch at jmhodges/rules_go/cgo_examples with the example needed but it's failing because the gazelle-generated BUILD file references "io_bazel_rules_go" which doesn't exist in the scope we need it. I'm not sure how to fix that. Breaking the new `examples/buildless_external` into its own WORKSPACE seems not desirable. Options?

I've pushed up a commit for the `-pthread` bit.  That just seems like it got missed in the def.bzl implementation of cgo. The `go build` equivalent handles pthread always. See this section of code: https://github.com/golang/go/blob/75fef5a0f60bcace88a8d7470df2d85d8eee048f/src/cmd/go/build.go#L3084-L3093

That section, however, also checks if it's running on windows. The version I pushed up always adds `-pthread`, but I'm not sure how to check for the OS in Skylark. Do y'all know how to do that?

yugui:
@pmbethe09 Thank you for your review.  Could you merge this PR once it looks good to you?

go/def.bzl
```diff
@@ -937,6 +937,7 @@ def cgo_library(name, srcs,
       copts = copts + [
           "-I", pkg_dir,
           "-I", "$(GENDIR)/" + pkg_dir + "/" + cgogen.outdir,
+          "-pthread",
```
pmbethe09: This appears to work for both linux and OSX, and rules_go doesn't have Windows support anyway.  Maybe add a TODO to fix if/when Windows support is added?

go/tools/gazelle/generator/generator_test.go
```diff
@@ -161,11 +195,29 @@ func TestGenerator(t *testing.T) {
...
+				stub.fixtures["cgolib"][1].Call,
+				stub.fixtures["cgolib"][2].Call,
+			},
+		},
 	}
 	sort.Sort(fileSlice(want))
 
 	if !reflect.DeepEqual(got, want) {
-		t.Errorf("g.Generate(%q) = %s; want %s", repo, prettyFiles(got), prettyFiles(want))
+		t.Errorf("g.Generate(%q) = %s;\n\nwant:\n%s", repo, prettyFiles(got), prettyFiles(want))
```
pmbethe09: remove these newlines.

go/tools/gazelle/generator/generator_test.go
```diff
@@ -174,9 +226,9 @@ type prettyFiles []*bzl.File
 func (p prettyFiles) String() string {
 	var items []string
 	for _, f := range p {
-		items = append(items, fmt.Sprintf("{Path: %q, Stmt: %s}", f.Path, string(bzl.Format(f))))
+		items = append(items, fmt.Sprintf("\t{Path: %q, Stmt: %s}", f.Path, string(bzl.Format(f))))
 	}
-	return fmt.Sprintf("[%s]", strings.Join(items, ","))
+	return fmt.Sprintf("[\n%s\n]", strings.Join(items, ",\n"))
```
pmbethe09: remove these newlines.

go/tools/gazelle/generator/generator_test.go
```diff
@@ -174,9 +226,9 @@ type prettyFiles []*bzl.File
 func (p prettyFiles) String() string {
 	var items []string
 	for _, f := range p {
-		items = append(items, fmt.Sprintf("{Path: %q, Stmt: %s}", f.Path, string(bzl.Format(f))))
+		items = append(items, fmt.Sprintf("\t{Path: %q, Stmt: %s}", f.Path, string(bzl.Format(f))))
```
pmbethe09: remove the \t

go/tools/gazelle/rules/generator.go
```diff
@@ -91,15 +91,25 @@ func (g *generator) Generate(rel string, pkg *build.Package) ([]*bzl.Rule, error
 		rules = append(rules, p)
 	}
 
-	libName := defaultLibName
-	r, err := g.generateLib(rel, libName, pkg)
+	cgoLibrary := ""
+	if len(pkg.CgoFiles) != 0 || len(pkg.CFiles) != 0 || len(pkg.HFiles) != 0 {
```
pmbethe09: does it ever make sense to have a cgo_library with no CgoFiles but CFiles?
Also, what about CXXFiles? (which you already appear to handle in generateCgoCLib)

e.g. if len(CgoFiles) == 0 but any of C/CXX/H are >0, should gazelle just print a warning and skip cgo?

go/tools/gazelle/rules/generator.go
```diff
@@ -91,15 +91,25 @@ func (g *generator) Generate(rel string, pkg *build.Package) ([]*bzl.Rule, error
...
+		cgoLibrary = "cgo_default_library"
```
pmbethe09: move "cgo_default_library" to be a file-level constant next to the others for clarity.

go/tools/gazelle/rules/generator.go
```diff
@@ -148,26 +158,80 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
 	// Libraries made for a go_binary should not be exposed to the public.
 	if pkg.IsCommand() {
 		visibility = "//visibility:private"
 	}
+
+	attrs := []keyvalue{
+		{key: "name", value: name},
+	}
+
+	if len(cgoName) == 0 {
```
pmbethe09: Go-style is to do:
cgoName == ""

go/tools/gazelle/rules/generator.go
```diff
@@ -148,26 +158,80 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
+
 	attrs := []keyvalue{
 		{key: "name", value: name},
-		{key: "srcs", value: append(pkg.GoFiles, pkg.SFiles...)},
-		{key: "visibility", value: []string{visibility}},
 	}
 
+	srcs := append([]string{}, pkg.CgoFiles...)
+	srcs = append(srcs, pkg.CFiles...)
+	srcs = append(srcs, pkg.CXXFiles...)
```
pmbethe09: I wonder if we should add a "warning ignored" if SwigFiles, SwigCXXFiles or MFiles has elements?

go/tools/gazelle/rules/generator.go
```diff
@@ -148,26 +158,80 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
+	}
+
+	visibility := checkInternalVisibility(rel, "//visibility:private")
+	attrs = append(attrs, keyvalue{key: "visibility", value: []string{visibility}})
+
 	deps, err := g.dependencies(pkg.Imports, rel)
 	if err != nil {
 		return nil, err
 	}
+
```
pmbethe09: remove blankline

go/tools/gazelle/rules/generator.go
```diff
@@ -148,26 +158,80 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
+	deps, err := g.dependencies(pkg.Imports, rel)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(deps) > 0 {
+		attrs = append(attrs, keyvalue{key: "deps", value: deps})
+	}
+
```
pmbethe09: feel like too many blanklines.  Either remove this one, or the one just above (or both).

go/tools/gazelle/rules/generator.go
jmhodges: You're right, it doesn't make sense. Thanks.

jmhodges:
Handled the feedback except the warning. Lmk if I missed one of the newline bits and what kind of logging (if any) you'd like in specific.

go/tools/gazelle/rules/generator.go
pmbethe09: standard log.Printf should be fine.  I don't expect this to happen often, but it is nice not to silently ignore things.

go/tools/gazelle/rules/generator.go
```diff
@@ -91,15 +94,25 @@ func (g *generator) Generate(rel string, pkg *build.Package) ([]*bzl.Rule, error
...
-	r, err := g.generateLib(rel, libName, pkg)
+	cgoLibrary := ""
+	if len(pkg.CgoFiles) != 0 {
+		cgoLibrary = defaultCgoLibName
+		r, err := g.generateCgoCLib(rel, cgoLibrary, pkg)
+		if err != nil {
+			return nil, err
+		}
+		rules = append(rules, r)
+	}
```
pmbethe09: perhaps a warning also when CgoFiles is empty, but any of C/CXX/M files are not?

jmhodges:
Okay, pushed up the logging. I broke out a logging statement for each file category to be a little more explicit.

There seems to be a bug in the travisci config now, unfortunately. I merged in master but the linux V=HEAD version is failing after it hits a 404 downloading a shell script. https://travis-ci.org/bazelbuild/rules_go/jobs/170928220

jmhodges:
Ah, that's #172. It would be nice to get that in

go/tools/gazelle/rules/generator.go
```diff
@@ -146,22 +160,84 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
 	}
 
+	if len(pkg.MFiles) != 0 {
+		log.Printf("warning: %s has Objective-C files but rules_go does not yet support Objective-C", rel)
+	}
+	if len(pkg.FFiles) != 0 {
+		log.Printf("warning: %s has Fortran files but rules_go does not yet support Fortran", rel)
+	}
+	if len(pkg.SwigFiles) != 0 || len(pkg.SwigCXXFiles) != 0 {
+		log.Printf("warning: %s has SWIG files but rules_go does not yet support Fortran", rel)
```
pmbethe09: s/Fortran/Swig

go/tools/gazelle/rules/generator.go
```diff
@@ -39,6 +40,9 @@ const (
 	// defaultProtosName is the name of a filegroup created
 	// whenever the library contains .pb.go files
 	defaultProtosName = "go_default_library_protos"
+	// defaultCgoName is the naem of the default cgo_library rule in a GO
```
pmbethe09: s/naem/name/

go/tools/gazelle/rules/generator.go
```diff
@@ -39,6 +40,9 @@ const (
...
+	// package directory.
```
pmbethe09: leave on last line, no need to linebreak here.

go/tools/gazelle/rules/generator.go
```diff
@@ -146,22 +160,84 @@ func (g *generator) generateBin(rel, library string, pkg *build.Package) (*bzl.R
...
+		return nil, err
+	}
+	if len(deps) > 0 {
+		attrs = append(attrs, keyvalue{key: "deps", value: deps})
+	}
+
+	return newRule(kind, nil, attrs)
+}
+
+// Generates a cgo_library rule for C/C++ code.
```
pmbethe09: style: lead the comment with the func name
so
// generateCgoCLib generates a cgo_library rule for C/C++ code.

jmhodges:
Okay, here we go.

jmhodges:
Also: can't tell from this side of the PR, but if there's no squash-and-merge button here, I can do some squashing myself if you'd prefer.

pmbethe09:
Jenkins, retest this please.

PR #172, opened 2016-10-24
Update Jenkins build artifact name change

pcj:
The bazel installer shell script build output from jenkins seems to have changed from `bazel--installer.sh` to `bazel-0.4.0rc1-installer.sh`.  This update tests for a version string matching a release candidate and uses ci.bazel.io as the base url for that artifact.


bazel-io:

pcj:
@pmbethe09 Thanks for assigning this as it may be a suboptimal fix.  Now it seems like Jenkins is back to the original behavior?

pcj:
Force pushed an updated version that is more general by parsing the name of installer.

pmbethe09:
Jenkins test this please

PR #176, opened 2016-10-28
add go_srcs to toolchain

mikedanese:
Tools that generate go code an use [go/build](http://godoc.org/go/build) often require all source to be present. This allows rules that use go_env_attrs to declare GOROOT/src as input.


bazel-io:

pmbethe09:
Jenkins, test this please.

pmbethe09:
LGTM

PR #177, opened 2016-10-31
add wtool for simple adding of new_go_repository to the WORKSPACE file

pmbethe09:
retry/rebased.

### DIFF
--- a/go/tools/wtool/main.go
+++ b/go/tools/wtool/main.go
@@ -135,7 +135,7 @@ func attr(key, val string) *bzl.BinaryExpr {
 }
 
 func lsRemote(repo string) (string, error) {
-	cmd := exec.Command("git", "ls-remote", repo)
+	cmd := exec.Command("git", "ls-remote", repo, "HEAD")
 	r, err := cmd.StdoutPipe()
 	if err != nil {
 		return "", err
--- a/go/tools/wtool/main.go
+++ b/go/tools/wtool/main.go
@@ -1,0 +1,144 @@
+/*wtool augments your bazel WORKSPACE file with new_go_repository entries
+
+Example Usage: wtool com_github_golang_glog com_google_cloud_go
+will add 2 new_go_repository to your WORKSPACE
+by converting com_github_golang_glog -> github.com/golang/glog
+and so forth and then doing a 'git ls-remote' to get
+the latest commit.
+
+if wtool cannot figure out the bazel -> Go mapping, try
+Other Usage: wtool -asis github.com/golang/glog
+which takes an importpath, and computes the bazel name + ls-remote as above.
+*/
+package main
+
+import (
+	"bufio"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	bzl "github.com/bazelbuild/buildifier/core"
+	"github.com/bazelbuild/rules_go/go/tools/gazelle/rules"
+	"github.com/bazelbuild/rules_go/go/tools/gazelle/wspace"
+	"golang.org/x/tools/go/vcs"
+)
+
+var (
+	asis = flag.Bool("asis", false, "if true, leave the import names as-is (by default they are treated as bazel converted names like org_golang_x_net")
+)
+
+func main() {
+	flag.Parse()
+	if err := run(flag.Args()); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run(args []string) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	w, err := wspace.Find(cwd)
+	if err != nil {
+		return err
+	}
+	p := filepath.Join(w, "WORKSPACE")
+	b, err := ioutil.ReadFile(p)
+	if err != nil {
+		return err
+	}
+	f, err := bzl.Parse(p, b)
+	if err != nil {
+		return err
+	}
+	for _, arg := range args {
+		imp, err := findImport(arg)
+		if err != nil {
+			return err
+		}
+		f.Stmt = append(f.Stmt, imp)
+	}
+	bzl.Rewrite(f, nil)
+	return ioutil.WriteFile(f.Path, bzl.Format(f), 0644)
+}
+
+func nameAndImportpath(name string) (string, string, error) {
+	if *asis {
+		return rules.ImportPathToBazelRepoName(name), name, nil
+	}
+	s := strings.Split(name, "_")
+	if len(s) < 4 {
+		return "", "", fmt.Errorf("only 4-part strings supported: %q", name)
+	}
+	rest := strings.Join(s[3:], "-")
+	if strings.HasPrefix(name, "org_golang_google") {
+		return name, "google.golang.org/" + rest, nil
+	}
+	if strings.HasPrefix(name, "com_google_cloud") {
+		return name, "cloud.google.com/" + rest, nil
+	}
+	return name, strings.Join([]string{s[1] + "." + s[0], s[2], rest}, "/"), nil
+}
+
+func findImport(nameIn string) (bzl.Expr, error) {
+	name, importpath, err := nameAndImportpath(nameIn)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf(importpath)
+	r, err := vcs.RepoRootForImportPath(importpath, false)
+	if err != nil {
+		return nil, err
+	}
+	if r.VCS.Cmd != "git" {
+		return nil, fmt.Errorf("only git supported, not %q", r.VCS.Cmd)
+	}
+	commit, err := lsRemote(r.Repo)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
+/*wtool augments your bazel WORKSPACE file with new_go_repository entries
```
wlynch: /*
wtool

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+Example Usage: wtool com_github_golang_glog com_google_cloud_go
```
wlynch: If you put the example command on a new line and indent it, godoc will put it in a box when rendered to HTML.

e.g. https://godoc.org/golang.org/x/tools/cmd/godoc https://github.com/golang/tools/blob/master/cmd/godoc/main.go

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+will add 2 new_go_repository to your WORKSPACE
+by converting com_github_golang_glog -> github.com/golang/glog
+and so forth and then doing a 'git ls-remote' to get
+the latest commit.
+
+if wtool cannot figure out the bazel -> Go mapping, try
```
wlynch: s/if/If

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+		return err
+	}
+	for _, arg := range args {
+		imp, err := findImport(arg)
+		if err != nil {
+			return err
+		}
+		f.Stmt = append(f.Stmt, imp)
+	}
+	bzl.Rewrite(f, nil)
```
wlynch: Future work: We probably want to add a mode that is append only. e.g. only append new imports that do not exist in the workspace file, but leave the existing imports pinned to their current SHAs.

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+	return ioutil.WriteFile(f.Path, bzl.Format(f), 0644)
+}
+
+func nameAndImportpath(name string) (string, string, error) {
+	if *asis {
+		return rules.ImportPathToBazelRepoName(name), name, nil
+	}
+	s := strings.Split(name, "_")
+	if len(s) < 4 {
+		return "", "", fmt.Errorf("only 4-part strings supported: %q", name)
```
wlynch: This error doesn't accurately reflect the error. This should probably be something like "workspace names must be at least 4 parts"

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	rest := strings.Join(s[3:], "-")
```
wlynch: Will this work for all repos? I don't think there is anything preventing users from using '_' in their repo/package names.

We should probably just take the substring of the name.

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
+/*wtool augments your bazel WORKSPACE file with new_go_repository entries
+
+Example Usage: wtool com_github_golang_glog com_google_cloud_go
+will add 2 new_go_repository to your WORKSPACE
+by converting com_github_golang_glog -> github.com/golang/glog
+and so forth and then doing a 'git ls-remote' to get
+the latest commit.
+
+if wtool cannot figure out the bazel -> Go mapping, try
+Other Usage: wtool -asis github.com/golang/glog
```
wlynch: Perhaps we should rewrite the docs to just list every flag (e.g. https://godoc.org/golang.org/x/tools/cmd/godoc). This will make it easier to add docs for new flags later.

However since there is only 1 flag at the moment, feel free to punt on this.

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+		return "", "", fmt.Errorf("only 4-part strings supported: %q", name)
+	}
+	rest := strings.Join(s[3:], "-")
+	if strings.HasPrefix(name, "org_golang_google") {
+		return name, "google.golang.org/" + rest, nil
+	}
+	if strings.HasPrefix(name, "com_google_cloud") {
+		return name, "cloud.google.com/" + rest, nil
+	}
+	return name, strings.Join([]string{s[1] + "." + s[0], s[2], rest}, "/"), nil
```
wlynch: There's no way to distinguish foo.golang.org vs golang.org/foo is there? =[

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func findImport(nameIn string) (bzl.Expr, error) {
+	name, importpath, err := nameAndImportpath(nameIn)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf(importpath)
```
wlynch: I'm fine with giving some feedback to the user about what actions were taken, but this type of log should probably happen by the `f.Stmt = append(f.Stmt, imp)`

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+		return "", err
+	}
+	b := bufio.NewScanner(r)
+	if !b.Scan() {
+		if err := b.Err(); err != nil {
+			return "", err
+		}
+		return "", fmt.Errorf("nothing returned from ls-remote %q", repo)
+	}
+	log.Printf(b.Text())
```
wlynch: Remove extra log (or stick it behind a verbosity flag).

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+	cmd := exec.Command("git", "ls-remote", repo)
+	r, err := cmd.StdoutPipe()
+	if err != nil {
+		return "", err
+	}
+	if err := cmd.Start(); err != nil {
+		return "", err
+	}
+	b := bufio.NewScanner(r)
+	if !b.Scan() {
```
wlynch: Could this be racy with cmd.Start()? e.g. if we start scanning before we get the first byte from the ls-remote, will we get an EOF?

Perhaps we should wait with some timeout before scanning instead.

go/tools/wtool/main.go
```diff
@@ -0,0 +1,144 @@
...
+		if err := b.Err(); err != nil {
+			return "", err
+		}
+		return "", fmt.Errorf("nothing returned from ls-remote %q", repo)
+	}
+	log.Printf(b.Text())
+	go cmd.Wait()
+	return strings.Split(b.Text(), "\t")[0], nil
```
wlynch: I don't believe ls-remote makes any guarantees of ref ordering. We should probably default to master (`git ls-remote $repo master`).

Perhaps we should provide a means to specify the desired ref in the args? (e.g. com_github_golang_glog:mybranch)

go/tools/wtool/main.go
pmbethe09: done

go/tools/wtool/main.go
pmbethe09: fair question.  I am guessing at best here, the ones I have seen have all been '-', and that is why I added the -asis mode when our guesses are not good enough.

go/tools/wtool/main.go
pmbethe09: I special cased google.golang.org and cloud.google.com, we could make that configurable, or again, allow the user to use -asis

go/tools/wtool/main.go
pmbethe09: AFAIK, it will block, not EOF

pmbethe09:
PTAL

go/tools/wtool/main.go
```diff
@@ -0,0 +1,158 @@
...
+	f, err := bzl.Parse(p, b)
+	if err != nil {
+		return err
+	}
+	for _, arg := range args {
+		imp, err := findImport(arg)
+		if err != nil {
+			return err
+		}
+		// TODO(pmbethe09): ignore or maybe update sha1 if already defined in workspace.
```
wlynch: For any TODOs, let's make a new issue to track so it is not forgotten.

go/tools/wtool/main.go
```diff
@@ -0,0 +1,158 @@
...
+		log.Printf(importpath)
+	}
+	r, err := vcs.RepoRootForImportPath(importpath, false)
+	if err != nil {
+		return nil, err
+	}
+	if r.VCS.Cmd != "git" {
+		return nil, fmt.Errorf("only git supported, not %q", r.VCS.Cmd)
+	}
+	// TODO(pmbethe09): allow tag to be provided, e.g. com_github_golang_glog:mybranch
```
wlynch: s/tag/ref

wlynch:
Make sure travis is green before submitting, and that "squash and merge" is used for the the submission.

go/tools/wtool/main.go
pmbethe09: done

go/tools/wtool/main.go
pmbethe09: done

Issue #178, opened 2016-10-31
Gazelle creates empty srcs


sethkoehler:
I've noticed Gazelle generates an empty-source library for folders that only have test code in them (e.g., for an integration test):

go_library(
    name = "go_default_library",
    srcs = [],
    visibility = ["//visibility:public"],
)

This causes a problem with "bazel build //..." in that folder...presumably not generating this would be better?

PR #179, opened 2016-10-31
do not create a go_library if there are no srcs

pmbethe09:
Fixes https://github.com/bazelbuild/rules_go/issues/178

### DIFF
--- a/go/tools/gazelle/rules/generator.go
+++ b/go/tools/gazelle/rules/generator.go
@@ -105,11 +105,13 @@ func (g *generator) Generate(rel string, pkg *build.Package) ([]*bzl.Rule, error
 	}
 
 	library := defaultLibName
-	r, err := g.generateLib(rel, library, pkg, cgoLibrary)
+	libRule, err := g.generateLib(rel, library, pkg, cgoLibrary)
 	if err != nil {
 		return nil, err
 	}
-	rules = append(rules, r)
+	if libRule != nil {
+		rules = append(rules, libRule)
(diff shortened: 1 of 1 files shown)


go/tools/gazelle/rules/generator.go
```diff
@@ -105,11 +105,13 @@ func (g *generator) Generate(rel string, pkg *build.Package) ([]*bzl.Rule, error
...
 
 	library := defaultLibName
-	r, err := g.generateLib(rel, library, pkg, cgoLibrary)
+	libRule, err := g.generateLib(rel, library, pkg, cgoLibrary)
 	if err != nil {
 		return nil, err
 	}
-	rules = append(rules, r)
+	if libRule != nil {
+		rules = append(rules, libRule)
```
yugui: Does the `go_binary` rule depend on an undefined library rule if `libRule` is `nil`?
IIUC, it can happen when the command package consists of only cgo-enabled sources.

pmbethe09: I am not sure I follow the question

yugui: IIUC,
when `len(pkg.CgoFiles) > 0 && len(pkg.GoFiles) == 0 && len(pkg.SFiles) == 0` `generateLib` returns `nil`
In this case, and if `pkg.IsCommand()` is `true`. the generated `go_binary` target depends on a `go_library` rule named `go_default_library` but no such target is generated.

PR #182, opened 2016-11-02
go_proto_library: attempt to make the well known protos Just Work

pmbethe09:
This method requires a one-time "go_google_protobuf()" in your root BUILD file, but after that, engineers don't have to worry about setting their import paths/maps.

### DIFF
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -52,18 +52,37 @@ def _go_prefix(ctx):
     prefix = prefix + "/"
   return prefix
 
+def _external_dir(file):
+  """Returns the include dir for external files or empty.
+
+  file: a bazel File object
+
+  Returns:
+    a string path or "" if nothing to do
+
+  Examples to look for
+    'external/some_repo'
+    'bazel-$project/external/some_repo'
+    'bazel-out/local-fastbuild/genfiles/external/some_repo'
+    'bazel-out/local-fastbuild/genfiles/'
+"""
+  toks = file.dirname.split("/")
+  if toks[0] == "external":
+    return "/".join(toks[:2])
+  if not toks[0].startswith("bazel-"):
+    return ""
+  for idx in range(1, len(toks)):
+    if toks[idx] == "external":
+      return "/".join(toks[:idx+2])
+  return "/".join(toks[:3])
+
 def _external_dirs(files):
   """Compute any needed -I options to protoc from external filegroups."""
   res = []
   for f in files:
-    toks = f.dirname.split("/")
-    if toks[0] == "external":
-      res.append("/".join(toks[:2]))
-    elif len(toks) > 4 and toks[3] == "external":
-      res.append("/".join(toks[:5]))
-    elif toks[1] == "external" or toks[0][:6] == "bazel-":
-      res.append("/".join(toks[:3]))
-    
+    e = _external_dir(f)
+    if e:
+      res.append(e)    
   return set(res)
 
 def _go_proto_library_gen_impl(ctx):
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -161,15 +161,16 @@ def _add_target_suffix(target, suffix):
 
 _well_known_repo = "@com_github_golang_protobuf//ptypes/"
 
-def _well_known_proto_deps(deps):
+def _well_known_proto_deps(deps, repo):
   for d in deps:
     if d[:len(_well_known_repo)] == _well_known_repo:
-      return ["//:" + _GO_GOOGLE_PROTOBUF]
+      return [repo + "//:" + _GO_GOOGLE_PROTOBUF]
   return []
 
 def go_proto_library(name, srcs = None, deps = None,
                      has_services = 0,
                      testonly = 0, visibility = None,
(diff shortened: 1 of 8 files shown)


pmbethe09:
hold on this -- I have a better idea

pmbethe09:
Ready for review.  Fixed it so now you don't have to do anything special for this to Just Work

proto/go_proto_library.bzl
```diff
@@ -213,6 +235,26 @@ def go_proto_library(name, srcs = None, deps = None,
...
+def go_google_protobuf(name = _GO_GOOGLE_PROTOBUF):
+  deps = [_add_target_suffix(_well_known_import_key(wk), _PROTOS_SUFFIX)
+          for wk in _well_known_imports]
+  outs = [wk + ".proto" for wk in _well_known_imports]
+
+  native.genrule(
+      name = name,
+      srcs = deps,
+      outs = ["google/protobuf/"+o for o in outs],
+      cmd = "cp $(SRCS) $(@D)/google/protobuf/",
```
wlynch: What will happen if the user depends on both google/protobuf and golang/protobuf? Will these files conflict?

proto/go_proto_library.bzl
```diff
@@ -54,8 +54,17 @@ def _go_prefix(ctx):
...
 def _external_dirs(files):
   """Compute any needed -I options to protoc from external filegroups."""
-  return set(["/".join(f.dirname.split("/")[:2])
-              for f in files if f.dirname[:9] == "external/"])
+  res = []
+  for f in files:
+    toks = f.dirname.split("/")
+    if toks[0] == "external":
+      res.append("/".join(toks[:2]))
+    elif len(toks) > 4 and toks[3] == "external":
```
pmbethe09: Attempted to clean up and add explanations.

proto/go_proto_library.bzl
```diff
@@ -157,9 +166,18 @@ def _add_target_suffix(target, suffix):
   toks = target.split("/")
   return target + ":" + toks[-1] + suffix
 
+_well_known_repo = "@com_github_golang_protobuf//ptypes/"
+
+def _well_known_proto_deps(deps, repo):
+  for d in deps:
+    if d[:len(_well_known_repo)] == _well_known_repo:
```

proto/go_proto_library.bzl
```diff
@@ -157,9 +166,18 @@ def _add_target_suffix(target, suffix):
...
+      return [repo + "//:" + _GO_GOOGLE_PROTOBUF]
+  return []
+
 def go_proto_library(name, srcs = None, deps = None,
                      has_services = 0,
                      testonly = 0, visibility = None,
+                     rules_go_repo_only_for_internal_use = "@io_bazel_rules_go",
```

proto/go_proto_library.bzl
pmbethe09: No, because in  google/protobuf these files are in external/com_github_google_protobuf/src/google/protobuf

pmbethe09:
PTAL

proto/go_proto_library.bzl
```diff
@@ -52,10 +53,38 @@ def _go_prefix(ctx):
...
+    'bazel-out/local-fastbuild/genfiles/'
+"""
+  toks = file.dirname.split("/")
+  if toks[0] == "external":
+    return "/".join(toks[:2])
+  if not toks[0].startswith("bazel-"):
+    return ""
+  for idx in range(1, len(toks)):
+    if toks[idx] == "external":
+      return "/".join(toks[:idx+2])
```
pmbethe09: After external is the directory named for the repo
e.g. bazel-proj/external/com_google_foo/...
or bazel-out/local-fastbuild/host/external/com_google_foo/
etc.

So when idx is at "external" you always want to grab 1 more directory to construct the import path.
Following bazel-style for proto and C++, users must still do "root-relative" imports.

So 'import "any.proto";' will not work, it has to be:
import "ptypes/any/any.proto";
or
import "google/protobuf/any.proto";

hochhaus:
@pmbethe09 Thanks for landing support for well known protos!

Does this include `google/protobuf/descriptor.proto` support? If so, should I be adding a dependency to https://github.com/golang/protobuf/blob/master/protoc-gen-go/descriptor/descriptor.pb.go or https://github.com/google/go-genproto/blob/master/protobuf/descriptor.pb.go?

pmbethe09:
Sorry for the delay.  No this just handles the protos found under
@com_github_golang_protobuf//ptypes/...

I don't have a good solution for descriptor.proto, as only the .pb.go is
checked into the repo.

hochhaus:
Thanks @pmbethe09. I'm sure you are already aware, but I just thought I would mention that both the `*.proto` and `*.pb.go` are checked into the [go-genproto](https://github.com/google/go-genproto/) repo. This isn't ideal as it introduces another dependency.

https://github.com/google/go-genproto/blob/master/protobuf/descriptor.pb.go
https://github.com/google/go-genproto/blob/master/protobuf/descriptor.proto

Either way, I've got a work around for the time being. Thanks for your pointers.

pmbethe09:
That is sort of normal for those repos.  new_go_repository will run
gazelle, and that will create both a go_library and a filegroup, the latter
will be used by go_proto_library's generation step.

PR #183, opened 2016-11-03
compile stdlib for the (GOOS, GOARCH)'s of --cpus we support

mikedanese:
Without this, the build complains about stdlib packages missing when using --cpu.

```console
$ bazel build --cpu armeabi-v7a //cmd/kubectl                   
INFO: Found 1 target...
ERROR: /usr/local/google/home/mikedanese/go/src/k8s.io/kubernetes/vendor/BUILD:180:1: null failed: linux-sandbox failed: error executing command /usr/local/google/home/mikedanese/.cache/bazel/_bazel_mikedanese/6c65a147ac17188f348dcacc7c554ab3/execroot/kubernetes/_bin/linux-sandbox ... (remaining 3 argument(s) skipped).
k8s.io/kubernetes/vendor/github.com/MakeNowJust/heredoc/heredoc.go:26: can't find import: "fmt"
```

This works especially well when not depending on cgo libraries.

### DIFF
--- a/go/private/go_repositories.bzl
+++ b/go/private/go_repositories.bzl
@@ -157,6 +157,22 @@ def _go_repository_select_impl(ctx):
   ctx.symlink(gopkg, "pkg")
   ctx.symlink(gosrc, "src")
 
+  cross_targets = (
+      ('linux', 'amd64'),
+      ('linux', 'arm'),
+      ('linux', '386'),
+      ('darwin', 'amd64'),
+      ('freebsd', 'amd64'),
+  )
+
+  for goos, goarch in cross_targets:
+    result = ctx.execute([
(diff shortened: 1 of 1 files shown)


bazel-io:

pmbethe09:
Jenkins, test this please.

pcj:
Looks like this adds about a minute for the travis tests to complete.  How is this going to degrade the user experience for rules_go users?  Although installing the go toolchain should be infrequent, when managing multiple projects and flux in the bazel ecosystem itself, it's not that rare.

In not necessarily against this, but Is there not another way to accomplish this by detecting when the `--cpu` flag is involved and do a `go install std` at that time?  Go install is smart about knowing if the stdlib is already compiled, so if it's already done, it returns quickly.

mikedanese:
I'm not entirely convinced that this is useful yet... still experimenting.

PR #185, opened 2016-11-03
support configuring xdefs from file

mikedanese:
-X link option is commonly used for build stamping. Unfortunately workspace status is not yet available from skylark. This allows users to receive build files containing x_defs and use these link options from go_binary and go_test rules.


bazel-io:

mikedanese:
I think a better implementation would be more similar to linkstamp.

PR #186, opened 2016-11-03
add support for injecting build stamps with link options

mikedanese:
for a workspace with BUILD_SCM_VERSION=10 defined as a volatile status and a go binary rule:
```
go_binary(
    ....
    stamp = True,
    linkstamp = "my/pkg/version",
)
```

The go binary will be linked with the option `-X my/pkg/version.BUILD_SCM_VERSION`

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -34,9 +34,21 @@ _DEFAULT_LIB = "go_default_library"
 
 _VENDOR_PREFIX = "/vendor/"
 
-go_filetype = FileType([".go", ".s", ".S"])
+go_filetype = FileType([
+    ".go",
+    ".s",
+    ".S",
+])
+
 # be consistent to cc_library.
-hdr_exts = ['.h', '.hh', '.hpp', '.hxx', '.inc']
+hdr_exts = [
+    ".h",
+    ".hh",
+    ".hpp",
+    ".hxx",
+    ".inc",
+]
+
 cc_hdr_filetype = FileType(hdr_exts)
 
 ################
@@ -119,11 +131,19 @@ def go_environment_vars(ctx):
                                     "GOARCH": "amd64"})
 
 def _emit_generate_params_action(cmds, ctx, fn):
-  cmds_all = ["set -e"]
+  cmds_all = [
+      # Use bash explicitly. /bin/sh is default, and it may be linked to a
+      # different shell, e.g., /bin/dash on Ubuntu.
+      "#!/bin/bash",
+      "set -e",
+  ]
   cmds_all += cmds
   cmds_all_str = "\n".join(cmds_all) + "\n"
   f = ctx.new_file(ctx.configuration.bin_dir, fn)
-  ctx.file_action( output = f, content = cmds_all_str, executable = True)
+  ctx.file_action(
+      output = f,
+      content = cmds_all_str,
+      executable = True)
   return f
 
 def emit_go_asm_action(ctx, source, out_obj):
@@ -391,6 +411,7 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
       ('../' * out_depth) + ctx.file.go_tool.path,
       "tool", "link", "-L", ".",
       "-o", _go_importpath(ctx),
+      '"${STAMP_XDEFS[@]}"',
   ]
 
   if x_defs:
@@ -414,8 +435,25 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
   # TODO(yugui) Remove this workaround once rules_go stops supporting XCode 7.2
   # or earlier.
   cmds += ["export PATH=$PATH:/usr/bin"]
+
   cmds += [
-    "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
+      "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
+      "STAMP_XDEFS=()",
+  ]
+
+  stamp_inputs = []
+  if ctx.attr.linkstamp:
+    # read workspace status files, converting "KEY value" lines
+    # to "-X $linkstamp.KEY=value" arguments to the go linker.
+    stamp_inputs = [ctx.info_file, ctx.version_file]
+    for f in stamp_inputs:
+      cmds += [
+          "while read -r stamp || [[ -n $stamp ]]; do",
+          "  STAMP_XDEFS+=(-X \"%s.$(echo $stamp | sed -e 's!^\([^ ][^ ]*\) *\(.*\)$!\\1=\\2!')\")" % ctx.attr.linkstamp,
+          "done < " + f.path,
+      ]
+
+  cmds += [
     "cd " + out_dir,
     ' '.join(link_cmd),
     "mv -f " + _go_importpath(ctx) + " " + ("../" * out_depth) + executable.path,
@@ -425,7 +463,7 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
 
   ctx.action(
       inputs = (list(transitive_libs) + [lib] + list(cgo_deps) +
-                ctx.files.toolchain + ctx.files._crosstool),
+                ctx.files.toolchain + ctx.files._crosstool) + stamp_inputs,
       outputs = [executable],
       executable = f,
       mnemonic = "GoLink",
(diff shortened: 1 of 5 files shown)


bazel-io:

pmbethe09:
How is this different from the 'x_defs' field already on the go_binary?

mikedanese:
@pmbethe09 currently workspace status is not available from skylark: https://github.com/bazelbuild/bazel/issues/1054 so it's not possible to use stamp info from the 'x_defs' field.

go/def.bzl
```diff
@@ -389,9 +390,24 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
+    "STAMP_XDEFS=()",
+  ]
+
+  stamp_input = []
+  if ctx.attr.stamp and ctx.attr.linkstamp:
+    stamp_input = [ctx.info_file, ctx.version_file]
+    for f in stamp_input:
+      cmds += [
+          "while read -r stamp || [[ -n $stamp ]]; do",
+          "if echo $stamp | grep -qe '^.*[[:space:]].*$'; then",
```
ixdy: it seems like it's not currently possible to access from skylark, per https://github.com/bazelbuild/bazel/issues/1054.

go/def.bzl
```diff
@@ -389,9 +390,24 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
-  cmds += ["export PATH=$PATH:/usr/bin"]
   cmds += [
+    "export PATH=$PATH:/usr/bin",
     "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
+    "STAMP_XDEFS=()",
+  ]
+
+  stamp_input = []
+  if ctx.attr.stamp and ctx.attr.linkstamp:
+    stamp_input = [ctx.info_file, ctx.version_file]
```
ixdy: what are `ctx.info_file` and `ctx.version_file`? they don't seem to be documented on https://www.bazel.io/versions/master/docs/skylark/lib/ctx.html.

go/def.bzl
ixdy: would adding a comment explaining what these lines of shell are doing help?

"reads workspace status file, converting "KEY value" lines to "-X KEY=$linkstamp.value" arguments to the go linker."

go/def.bzl
```diff
@@ -389,9 +390,24 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
+    for f in stamp_input:
+      cmds += [
+          "while read -r stamp || [[ -n $stamp ]]; do",
+          "if echo $stamp | grep -qe '^.*[[:space:]].*$'; then",
+          "STAMP_XDEFS+=(-X $(echo %s.${stamp} | tr ' ' '='));" % ctx.attr.linkstamp,
```
ixdy: if `linkstamp` is empty, this results in `-X .BUILD_TIMESTAMP=1234` - is what what we want?
also, is it legal for values to have spaces? do we need to add quotes around the values?

go/def.bzl
```diff
@@ -558,7 +574,8 @@ go_library = rule(
 go_binary = rule(
     go_binary_impl,
     attrs = go_library_attrs + _crosstool_attrs + {
-        "stamp": attr.bool(default = False),
+        "stamp": attr.bool(default = True),
```
ixdy: maybe update the root `README.md` explaining what `stamp` and `linkstamp` do?

pmbethe09:
Jenkins, test this please.

pmbethe09:
Jenkins, test this please.

mikedanese:
I haven't had a chance to get back to this yet.

go/def.bzl
mikedanese: They are not documented because they are alpha still.

mikedanese: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/SkylarkRuleContext.java#L816-L838

go/def.bzl
mikedanese: Can we do this if/when we decide we like the design? What's the policy on compatibility? I'd like someone from go team to weigh in on this before publicizing, but it unblocks us now.

go/def.bzl
```diff
@@ -389,9 +390,25 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
   # or earlier.
-  cmds += ["export PATH=$PATH:/usr/bin"]
   cmds += [
+    "export PATH=$PATH:/usr/bin",
     "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
+    "STAMP_XDEFS=()",
+  ]
+
+  stamp_input = []
+  if ctx.attr.stamp and ctx.attr.linkstamp:
```
ixdy: should we maybe `fail` if `stamp` is true but `linkstamp` isn't set? otherwise it might be confusing why `stamp` isn't working.

(alternately, do we need both? or just use nonempty `linkstamp` as the setting?)

mikedanese: I dropped stamp.

pmbethe09:
Jenkins, test this please.

pmbethe09:
seems to be failing on linux:
"
bazel-out/local-fastbuild/bin/examples/stamped_bin/bazel-out/local-fastbuild/bin/examples/stamped_bin/stamped_bin.a.GoLinkFile.params: 10: bazel-out/local-fastbuild/bin/examples/stamped_bin/bazel-out/local-fastbuild/bin/examples/stamped_bin/stamped_bin.a.GoLinkFile.params: Syntax error: "(" unexpected
____Building complete.
"

go/def.bzl
```diff
@@ -389,9 +402,25 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
+
+  stamp_input = []
+  if ctx.attr.linkstamp:
+    stamp_input = [ctx.info_file, ctx.version_file]
+    for f in stamp_input:
+      cmds += [
+          "while read -r stamp || [[ -n $stamp ]]; do",
+          # reads workspace status file, converting "KEY value" lines to "-X KEY=$linkstamp.value" arguments to the go linker.
+          "if echo $stamp | grep -qe '^.*[[:space:]].*$'; then",
+          "STAMP_XDEFS+=(-X \"$(echo \"%s.${stamp}\" | tr ' ' '=')\");" % ctx.attr.linkstamp,
```
ixdy: It seems like we can't use shell arrays here.

maybe just build up a string? there might be subtle quoting or escaping issues, but I'm not sure of a better way.

go/def.bzl
```diff
@@ -389,9 +402,25 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
+    "STAMP_XDEFS=()",
+  ]
+
+  stamp_input = []
+  if ctx.attr.linkstamp:
+    stamp_input = [ctx.info_file, ctx.version_file]
+    for f in stamp_input:
+      cmds += [
+          "while read -r stamp || [[ -n $stamp ]]; do",
+          # reads workspace status file, converting "KEY value" lines to "-X KEY=$linkstamp.value" arguments to the go linker.
```
ixdy: I think this comment is wrong - it's `-X $linkstamp.KEY=value`, right?

go/def.bzl
```diff
@@ -389,9 +402,25 @@ def emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
...
   # TODO(yugui) Remove this workaround once rules_go stops supporting XCode 7.2
   # or earlier.
-  cmds += ["export PATH=$PATH:/usr/bin"]
   cmds += [
+    "export PATH=$PATH:/usr/bin",
     "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
+    "STAMP_XDEFS=()",
+  ]
+
+  stamp_input = []
```
jayconrod: Nit: consider renaming this to `stamp_inputs` to make it clear it's a list.

go/def.bzl
jayconrod: Do we need the grep? It matches any line with whitespace, right? I'd imagine in some cases, the value for some key would be empty, and a line might not have any space. Maybe use sed instead?

    echo $stamp | sed -e 's%^\([^ ]\+\) \?\(.*\)$%-Xa/b.\1=\2%'

mikedanese:
@jayconrod thanks for finishing this off!

Issue #189, opened 2016-11-08
gazelle: handle case of go_proto_library and also checked in generated files


pmbethe09:
Some projects might try and stay "Go-compatible", and check-in the .pb.go files, but for bazel use a go_proto_library (whether from here or pubref).

When merging the new and old, gazelle should drop the new go_library and filegroup if it detects a go_proto_library(name="go_default_library")

jayconrod:
Closing old Gazelle issues. If Gazelle isn't doing something you want, please open a new issue at https://github.com/bazelbuild/bazel-gazelle.

We now use a `-proto` argument or directive to decide what rules to generate. The `fix` command can convert old rules to the new scheme.

PR #190, opened 2016-11-08
add special case to ignore gazelle if a go_proto_library is detected

pmbethe09:
Fixes https://github.com/bazelbuild/rules_go/issues/189


pmbethe09:
The travis failure appears transient:

"error: RPC failed; result=22, HTTP code = 502
fatal: The remote end hung up unexpectedly"

pmbethe09:
maybe a simpler thing would be a file-level
# gazelle: ignore

which tells gazelle to just leave the file alone.

PR #202, opened 2016-11-23
support cgo_libraries that should not create go_libraries

### LINKED ISSUE #201: cgo_library with dependency on types defined in .go files in the same package is broken
I can't figure out how to get this case to work. I have a `type Type struct` defined in types.go which is not a cgo file. I have a cgo_foo.go file which does import "C" and uses Type. The go tool handles this but I can't seem to figure  out how to get it to work with bazel rules. Here is a shot:

```
go_library(
    name = "go_default_library"
    srcs = ["types.go"],
    library = "foo_cgo",
)

cgo_library(
    name = "foo_cgo",
    srcs = ["foo_cgo.go"],
)
```

foo_cgo fails to compile `undefined: Type`. go build builds this package.

mikedanese:
Fixes #201

### DIFF
--- a/examples/cgo/skip_go_library/BUILD
+++ b/examples/cgo/skip_go_library/BUILD
@@ -1,19 +1,18 @@
 package(default_visibility = ["//visibility:public"])
 
-load("//go:def.bzl", "go_library", "cgo_library")
+load("//go:def.bzl", "go_library", "cgo_genrule")
 
 go_library(
     name = "go_default_library",
     srcs = [
         "types.go",
     ],
-    library = ":cgo_default_library",
+    library = ":cgo_codegen",
 )
 
-cgo_library(
-    name = "cgo_default_library",
+cgo_genrule(
+    name = "cgo_codegen",
     srcs = [
         "cgo_foo.go",
     ],
-    skip_go_library = True,
 )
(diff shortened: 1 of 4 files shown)


bazel-io:

pmbethe09:
Jenkins, test this please.

pmbethe09:
I am not entirely sure why this works, but happy to merge it if it unblocks you.

pmbethe09:
Looks like a real failure on darwin:
http://ci.bazel.io/job/PR-rules_go/201/BAZEL_VERSION=latest,PLATFORM_NAME=darwin-x86_64/console

mikedanese:
@pmbethe09 if you check out this branch and remove the `skip_go_library` flag from the build rule in `examples/cgo/skip_go_library/BUILD` then try to compile `//examples/cgo/skip_go_library:all`, you will see the problem. More details in #201.

I'll try to grab a hold of a mac. Without looking, I expect that the `#include <sys/eventfd.h>` is linux sepecific and is what is broken.

mikedanese:
This works because we get the cgo codegen output but skip the compilation of the go files. If cgo code gen files depend on types in non cgo files in the package, compilation of the go library will fail.

There might be a cleaner approach. I might give breaking out the cgo codegen to a public macro a shot.

pmbethe09:
Jenkins, retest this please.

PR #203, opened 2016-11-23
pass on deps to go_library generated by cgo_library

mikedanese:
otherwise deps is unused. I don't think this is intended.


bazel-io:

mikedanese:
Oops misread the code.

PR #204, opened 2016-11-23
go_proto_library now handles generated proto inputs as well as regular

pmbethe09:


examples/proto/gostyle/BUILD
```diff
@@ -1,8 +1,15 @@
 load("//proto:go_proto_library.bzl", "go_proto_library")
 
+genrule(
+    name = "copy",
+    srcs = ["gostyle.proto"],
+    outs = ["gostyle.gen.proto"],
+    cmd = "sed -e 's/ano.proto/any.proto/' $< > $@",
```
wlynch: I don't understand why this sed is necessary. ano.proto does not appear anywhere in gostyle.proto, and even if it did, why would we want this instead of just changing the proto?

proto/go_proto_library.bzl
```diff
@@ -53,95 +53,83 @@ def _go_prefix(ctx):
...
-  bazel_style = ctx.label.name != _DEFAULT_LIB + _PROTOS_SUFFIX
-  protos = list(ctx.files.srcs)
-  go_package_name = ""
-  if bazel_style:
-    go_package_name = "/" + ctx.label.name[:-len(_PROTOS_SUFFIX)]
-  m_import_path = ",".join(["M%s=%s%s%s" % (f.path, _go_prefix(ctx),
-                                            ctx.label.package, go_package_name)
-                            for f in ctx.files.srcs])
+def _collect_protos_import(ctx):
+  """Collect the list of transitive protos and m_import_path."""
```
wlynch: What does m_import_path correspond to?

proto/go_proto_library.bzl
```diff
@@ -53,95 +53,83 @@ def _go_prefix(ctx):
...
     protos += d._protos
     m_import_path += "," + d._m_import_path
-  use_grpc = ""
-  if ctx.attr.grpc:
-    use_grpc = "plugins=grpc,"
+  return list(protos), m_import_path
+
+def _drop_external(path):
+  """Drop leading '../' indicating an external dir of the form ../$some-repo."""
+  if path[:3] != "../":
```
wlynch: I would prefer to use startswith (https://www.bazel.build/versions/master/docs/skylark/lib/string.html#startswith) to make this more easily readable.

proto/go_proto_library.bzl
```diff
@@ -53,95 +53,83 @@ def _go_prefix(ctx):
...
+    return path
+  return "/".join(path.split("/")[2:])
```
wlynch: Why are we only interested in [2:]? [0] should be "..", but what is [1] and why don't we care about it?

proto/go_proto_library.bzl
```diff
@@ -53,95 +53,83 @@ def _go_prefix(ctx):
...
+  return proto_outs, "/" + ctx.label.name[:-len(_PROTOS_SUFFIX)]
+
+def _go_proto_library_gen_impl(ctx):
+  """Rule implementation that generates Go using protoc."""
+  proto_outs, go_package_name = _check_bazel_style(ctx)
+  m_import_path = ",".join(["M%s=%s%s%s" % (f.short_path, _go_prefix(ctx),
+                                            ctx.label.package, go_package_name)
+                            for f in ctx.files.srcs])
+  protos, mi = _collect_protos_import(ctx)
+  m_import_path += mi
```
wlynch: Why += instead of ",".join?

proto/go_proto_library.bzl
pmbethe09: Paths of the form Mpath/to.proto=foo.com/bar specify a mapping into the global Go namespace.

https://github.com/golang/protobuf#parameters

proto/go_proto_library.bzl

proto/go_proto_library.bzl
pmbethe09: From the comment, non-generated but external proto dependencies show up in .short_path as e.g.
'../com_github_golang_protobuf/ptypes/any/any.proto'
For both directory creation and destination, it is critical to drop the leading 2 path elements.

proto/go_proto_library.bzl
pmbethe09: that seems better.

PR #206, opened 2016-11-29
go_proto_library: Add clarifying doc to internal functions.

pmbethe09:


wlynch:
Thanks for the doc update! This LGTM, but can you update the PR message so that the squashed commit is more descriptive?

PR #208, opened 2016-11-29
update go_proto_repositories to use the latest external tags

pmbethe09:
also use faster http_archive for google/protobuf

### DIFF
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -291,27 +291,28 @@ def go_proto_repositories(shared=1, rules_go_repo_only_for_internal_use=None):
   new_go_repository(
       name = "com_github_golang_protobuf",
       importpath = "github.com/golang/protobuf",
-      commit = "1f49d83d9aa00e6ce4fc8258c71cc7786aec968a",
+      commit = "8ee79997227bf9b34611aee7946ae64735e6fd93",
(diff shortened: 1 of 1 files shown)


proto/go_proto_library.bzl
```diff
@@ -291,27 +291,28 @@ def go_proto_repositories(shared=1, rules_go_repo_only_for_internal_use=None):
...
       rules_go_repo_only_for_internal_use = rules_go_repo_only_for_internal_use,
   )
   if shared:
     # if using multiple *_proto_library, allows caller to skip this.
-    native.git_repository(
+    native.http_archive(
         name = "com_github_google_protobuf",
-        remote = "https://github.com/google/protobuf",
-        commit = "008b5a228b37c054f46ba478ccafa5e855cb16db",
+        url = "https://github.com/google/protobuf/archive/v3.1.0.tar.gz",
```
wlynch: There is a v3.1.0 tag https://github.com/google/protobuf/releases/tag/v3.1.0

wlynch: LGTM.

Explained offline- we are using this because git_repository does not support shallow clones at the moment. Using the http archive means that we are pulling less data down (only the source at the given commit instead of the whole history) so we have faster builds.

PR #209, opened 2016-11-30
gazelle: add ability to write # gazelle:ignore

pmbethe09:


go/tools/gazelle/merger/merger.go
```diff
@@ -26,7 +26,10 @@ import (
 	bzl "github.com/bazelbuild/buildifier/core"
 )
 
-const keep = "# keep" // marker in srcs or deps to tell gazelle to preserve.
+const (
+	gazelleIgnore = "gazelle:ignore" // marker in a BUILD file to ignore it.
```
wlynch: Lets document these features in the README.

go/tools/gazelle/merger/merger.go
```diff
@@ -51,6 +54,13 @@ func MergeWithExisting(newfile *bzl.File) (*bzl.File, error) {
 	if err != nil {
 		return nil, err
 	}
+	for _, s := range f.Stmt {
+		for _, c := range s.Comment().After {
+			if strings.Contains(c.Token, gazelleIgnore) {
```
wlynch: Do you want this to match if it maps to anything in the comment string, or should this be a prefix match instead?

pmbethe09: good idea

Issue #211, opened 2016-12-01
Compile actions should include the intended import path


creachadair:
The [Kythe](http://kythe.io/) tools need to keep track of the (putative) import path of each library being built. It would thus be very helpful if the `go tool compile` invocation included the `-p` flag. I believe the patch below is sufficient to add this.

Absent this, we have to recover that information by various disgusting string hackery, which is further complicated by `go_prefix`, and thus very brittle.

Better still would be if the `-o` output path already had the correct layout for the import mechanism to recognize it—I think that would permit most of the noise in the response file to be removed (the symlink forest is especially pernicious).

```diff --git a/go/def.bzl b/go/def.bzl
index baf535a..fdf6344 100644
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -200,7 +200,8 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
       ('../' * out_depth) + ctx.file.go_tool.path,
       "tool", "compile",
       "-o", ('../' * out_depth) + out_lib.path, "-pack",
-      "-I", "."
+      "-I", ".",
+      "-p", _go_importpath(ctx),
   ]
 
   # Set -p to the import path of the library, ie.
```

pmbethe09:
SGTM, do you want to send out a PR and assign to me?

creachadair:
Sure, will do, thanks. I'll start with the smaller part. I'd also like to see if we can eliminate the symlink forest, but I'll need to do more extensive testing for that. :)

PR #212, opened 2016-12-01
Set the import path in go tool compile invocations (#211).

creachadair:
This change permits tools to discover the nominal import path attributed to the
package being built, even if the actual paths don't match the standard layout.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -200,7 +200,8 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
       ('../' * out_depth) + ctx.file.go_tool.path,
       "tool", "compile",
       "-o", ('../' * out_depth) + out_lib.path, "-pack",
-      "-I", "."
+      "-I", ".",
+      "-p", _go_importpath(ctx),
   ]
 
   # Set -p to the import path of the library, ie.


bazel-io:

creachadair:
@pmbethe09 I'm not sure how to assign this to you per your request, but here is the PR we discussed.

pmbethe09:
Jenkins, test this please.

pmbethe09:
Looks like no good.
"github.com/bazelbuild/rules_go/bazel-out/local-fastbuild/bin/go/tools/filter_tags/filter_tags_test_main_test.go:8: import "github.com/bazelbuild/rules_go/go/tools/filter_tags/filter_tags_test" while compiling that package (import cycle)"

creachadair:
Hmm. Okay, I guess back to the drawing board for now.

Out of curiosity, is there any design doc for these rules floating around? A lot of this feels very contrary to the way the Go native build tools are intended to work, and I'm hoping to understand the rationale for this rule-architecture a little better.

pmbethe09:
IS https://github.com/bazelbuild/rules_go/pull/410 useful for what you were trying to do?

creachadair:
I can't really tell. I don't see any working examples (are there tests somewhere that I missed?)

I wound up reverse-transcribing the fiddly symlink forest that the current rules generate, from the contents of the response file it uses for the tool invocation. That worked well enough that I stopped looking for a better solution. I assume I'm going to have to throw it all out anyway once the new rules land.

creachadair:
Sorry I left this open; I don't intend to pursue it any further at this point.

PR #219, opened 2016-12-07
gazelle: Add ability to read/write to BUILD.bazel instead of BUILD

pmbethe09:
Start of fix for 
https://github.com/bazelbuild/rules_go/issues/144

After this, use it in new_go_repository.

PR #228, opened 2016-12-14
Fix support for BUILD.bazel files

twpayne:
This PR builds on the work in #219 and includes the ability to use `BUILD.bazel` files in `new_go_repository`, for example:

```
new_go_repository(
    name = "org_golang_x_text",
    build_file_name = "BUILD.bazel",
    commit = "a8b38433e35b65ba247bb267317037dee1b70cea",
    importpath = "golang.org/x/text",
)
```

### DIFF
--- a/go/tools/gazelle/gazelle/fix.go
+++ b/go/tools/gazelle/gazelle/fix.go
@@ -17,10 +17,18 @@ package main
 
 import (
 	"io/ioutil"
+	"os"
+	"path"
 
 	bzl "github.com/bazelbuild/buildifier/core"
 )
 
 func fixFile(file *bzl.File) error {
-	return ioutil.WriteFile(file.Path, bzl.Format(file), 0644)
+	if err := ioutil.WriteFile(file.Path, bzl.Format(file), 0644); err != nil {
+		return err
+	}
+	if path.Base(file.Path) != *buildFileName {
+		return os.Rename(file.Path, path.Join(path.Dir(file.Path), *buildFileName))
+	}
+	return nil
 }
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -180,7 +180,9 @@ func findBuildFile(repo string) (string, error) {
 			if fi.Mode().IsRegular() {
 				return p, nil
 			}
-		} else if !os.IsNotExist(err) {
+			continue
+		}
+		if !os.IsNotExist(err) {
 			return "", err
 		}
 	}
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -24,7 +24,9 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
 
 	bzl "github.com/bazelbuild/buildifier/core"
 	"github.com/bazelbuild/rules_go/go/tools/gazelle/generator"
@@ -33,15 +35,12 @@ import (
 )
 
 var (
-	buildName       = flag.String("build_name", "BUILD", "name of output build files to generate, defaults to 'BUILD'")
-	buildTags       = flag.String("build_tags", "", "comma-separated list of build tags. If not specified, GOOS and GOARCH are used.")
-	goPrefix        = flag.String("go_prefix", "", "go_prefix of the target workspace")
-	repoRoot        = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
-	mode            = flag.String("mode", "fix", "print: prints all of the updated BUILD files\n\tfix: rewrites all of the BUILD files in place\n\tdiff: computes the rewrite but then just does a diff")
-	validBuildNames = map[string]bool{
-		"BUILD":       true,
-		"BUILD.bazel": true,
-	}
+	buildFileName  = flag.String("build_file_name", "BUILD", "name of output build files to generate.")
+	buildTags      = flag.String("build_tags", "", "comma-separated list of build tags. If not specified, GOOS and GOARCH are used.")
+	goPrefix       = flag.String("go_prefix", "", "go_prefix of the target workspace")
+	repoRoot       = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
+	mode           = flag.String("mode", "fix", "print: prints all of the updated BUILD files\n\tfix: rewrites all of the BUILD files in place\n\tdiff: computes the rewrite but then just does a diff")
+	buildFileNames = []string{"BUILD.bazel", "BUILD"}
 )
 
 func init() {
@@ -56,8 +55,17 @@ var modeFromName = map[string]func(*bzl.File) error{
 	"diff":  diffFile,
 }
 
+func isValidBuildFileName(buildFileName string) bool {
+	for _, bfn := range buildFileNames {
+		if buildFileName == bfn {
+			return true
+		}
+	}
+	return false
+}
+
 func run(dirs []string, emit func(*bzl.File) error) error {
-	g, err := generator.New(*repoRoot, *goPrefix, *buildName, *buildTags)
+	g, err := generator.New(*repoRoot, *goPrefix, *buildFileName, *buildTags)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


go/tools/gazelle/gazelle/main.go
```diff
@@ -33,15 +33,12 @@ import (
...
+	buildFileName = flag.String("build_file_name", "BUILD", "name of output build files to generate, defaults to 'BUILD'")
+	goPrefix      = flag.String("go_prefix", "", "go_prefix of the target workspace")
+	repoRoot      = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
+	mode          = flag.String("mode", "fix", "print: prints all of the updated BUILD files\n\tfix: rewrites all of the BUILD files in place\n\tdiff: computes the rewrite but then just does a diff")
 
-	validBuildNames = map[string]bool{
-		"BUILD":       true,
-		"BUILD.bazel": true,
-	}
+	buildFileNames = []string{"BUILD.bazel", "BUILD"}
```
pmbethe09: Why this change?  Just so that iteration is in a consistent order?

go/tools/gazelle/generator/generator.go
```diff
@@ -101,7 +102,7 @@ func (g *Generator) Generate(dir string) ([]*bzl.File, error) {
 		if len(files) == 0 && rel != "" {
 			// "dir" was not a buildable Go package but still need a BUILD file
 			// for go_prefix.
-			files = append(files, emptyToplevel(g.goPrefix, g.buildName))
+			files = append(files, g.emptyToplevel(g.goPrefix))
```
pmbethe09: if you are going to make a receiver method, might as well not pass g.goPrefix either and just grab it there.

go/tools/gazelle/gazelle/main.go
twpayne: Exactly. The iteration order over Go maps is undefined (and is indeed randomized) so the old code would have undefined behaviour if both files were present. I believe that Bazel itself prioritises `BUILD.bazel` over `BUILD`.

bazel-io:

twpayne:
Thanks for the quick review! Comments addressed in 3057a7f9f3a41c72090aa3e6f8b2847212cf3fc7.

pmbethe09:
Jenkins, test this please.

go/tools/gazelle/gazelle/fix.go
```diff
@@ -17,10 +17,18 @@ package main
...
 	bzl "github.com/bazelbuild/buildifier/core"
 )
 
 func fixFile(file *bzl.File) error {
-	return ioutil.WriteFile(file.Path, bzl.Format(file), 0644)
+	if err := ioutil.WriteFile(file.Path, bzl.Format(file), 0644); err != nil {
+		return err
+	}
+	if path.Base(file.Path) != *buildFileName {
+		return os.Rename(file.Path, path.Join(path.Dir(file.Path), *buildFileName))
```
pmbethe09: nice.

go/tools/gazelle/gazelle/main.go
```diff
@@ -126,8 +132,8 @@ func main() {
 		}
 	}
 
-	if !validBuildNames[*buildName] {
-		log.Fatalf("-build_name %q must be in: %v", *buildName, validBuildNames)
+	if !isValidBuildFileName(*buildFileName) {
+		log.Fatalf("invalid build file name %q", *buildFileName)
```
pmbethe09: add back the list of valid names?

go/tools/gazelle/gazelle/main.go
```diff
@@ -145,23 +151,26 @@ func main() {
...
+func findBuildFile(repo string) (string, error) {
+	for _, base := range buildFileNames {
+		p := filepath.Join(repo, base)
+		fi, err := os.Stat(p)
 		if err != nil {
-			if os.IsNotExist(err) {
-				continue
-			}
-			return nil, "", err
+			continue
```
pmbethe09: If the stat fails for a reason other than os.IsNotExist, isn't that a sign of bad things?

pmbethe09:
I think there is still an issue which I had not resolved.
merger.go loads a file from the place that the new copy expect to write it -- but what if the current version is 'BUILD', but the -build_file_name is 'BUILD.bazel'.

I think the code will create a file for BUILD.bazel, not see anything at that location, so return the new version, and then write that.  Any customizations will be lost.

go/tools/gazelle/gazelle/main.go
twpayne: Done.

go/tools/gazelle/gazelle/main.go
twpayne: Agreed. Done.

twpayne:
> I think there is still an issue which I had not resolved.
> merger.go loads a file from the place that the new copy expect to write it -- but what if the current version is 'BUILD', but the -build_file_name is 'BUILD.bazel'.
> 
> I think the code will create a file for BUILD.bazel, not see anything at that location, so return the new version, and then write that. Any customizations will be lost.

My understanding in this case is that the `BUILD` file (including customizations) will be read and then will be re-written (including customizations) before being renamed to `BUILD.bazel`, i.e. that customizations will be preserved.

pmbethe09:
The path is something like this:
* First you generate a file for //some/dir/
** and it gets named with the passed in flag
https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/generator/generator.go#L142
* Then merger.MergeWithExisting gets called,
and it tries to load a file with the same name as 'newfile', and if that fails, then it just returns newfile and later writes that.

So if
//some/dir/BUILD
exists, but you run gazelle -build_file_name BUILD.bazel
I am quite sure that it will generate a new file
//some/dir/BUILD.bazel
leaving the BUILD file also in place.

twpayne:
Ah, you're right! (bug verified with a manual test). I've added a commit that fixes this.

go/tools/gazelle/gazelle/main.go
```diff
@@ -69,11 +77,26 @@ func run(dirs []string, emit func(*bzl.File) error) error {
...
-			}
-			bzl.Rewrite(f, nil) // have buildifier 'format' our rules.
-			if err := emit(f); err != nil {
+			existingFilePath, err := findBuildFile(path.Dir(f.Path))
+			if os.IsNotExist(err) { // No existing file, so write a new one
+				bzl.Rewrite(f, nil) // have buildifier 'format' our rules.
+				if err := emit(f); err != nil {
+					return err
+				}
+			} else if err == nil { // Existing file, so merge and maybe remove the old one
```
pmbethe09: Use continue and guard-clauses, so add a continue at the end of the above if loop, 
  continue
}
if err != nil {
  return err
}
this code w/ no indent.

go/tools/gazelle/merger/merger_test.go
```diff
@@ -93,7 +93,37 @@ func TestMergeWithExisting(t *testing.T) {
...
+	defer os.Remove(tmp.Name())
+	if _, err := io.WriteString(tmp, oldData); err != nil {
+		t.Fatal(err)
+	}
+	if err := tmp.Close(); err != nil {
+		t.Fatal(err)
+	}
+	newTmp, err := ioutil.TempFile(os.Getenv("TEST_TMPDIR"), "BUILD.bazel")
+	if err != nil {
+		t.Fatal(err)
```
pmbethe09: From here down, these should be t.Errorf, not Fatal (only use Fatal when continuing the test would likely lead to panic or similar)

pmbethe09: Looks like then only the last one should be t.Error -- if either this one or the next one fail, then the LHS might be nil, and continuing is suspect.

pmbethe09:
I am afraid I have merged other changes and you have to rebase, but I think as soon as that last Errorf is in this is LGTM and ready to merge.

twpayne:
Rebased and commits squashed (quicker and more accurate to resolve one merge conflict than one for each of ten commits). Ready for final review.

go/tools/gazelle/gazelle/main.go
```diff
@@ -145,23 +172,27 @@ func main() {
...
-				continue
+func findBuildFile(repo string) (string, error) {
+	for _, base := range buildFileNames {
+		p := filepath.Join(repo, base)
+		fi, err := os.Stat(p)
+		if err == nil {
+			if fi.Mode().IsRegular() {
+				return p, nil
 			}
-			return nil, "", err
```
pmbethe09: add a continue here, and then the 'else if' can become just 'if'

pmbethe09:
Jenkins, test this please.

go/tools/gazelle/gazelle/main.go
twpayne: Done.

twpayne:
Thanks for the merge!

PR #229, opened 2016-12-15
discrimante on test func signatures when generating tests

mikedanese:


bazel-io:

mikedanese:
Hi Clayton, while the discriminators I added are always true, the discriminator for the referenced type "testing" is not true if the "testing" package was aliased to some other name. Why anyone would do that, I'm not sure. I can definitely extract that information by running the typechecker but I can't decide if that is overkill for this.

pmbethe09:
LGTM, just need to fix merge conflicts.

smarterclayton:
Might be overkill, but I've seen aliases of "testing" in the wild (probably one in a thousand).

pmbethe09:
Jenkins, test this please.

PR #235, opened 2016-12-21
Gazelle: remove fixed shortcoming from README.md

twpayne:
This commit should have been part of #228.


bazel-io:

PR #241, opened 2017-01-06
Don't add tests and benchmarks twice.

mikedanese:
Tests and benchmarks are added to the end of this method.


mikedanese:
I messed up the rebase.

PR #245, opened 2017-01-17
change README in advance of 0.4.0 tag

pmbethe09:


wlynch:
LGTM, but do not submit until the 0.4.0 tag is live. I would rather users who are using rules_go for the first time pull an old release rather than get a 404.

PR #247, opened 2017-01-20
Get go rules working with execroot rearrangement

kchodorow:
I am getting ready to resubmit https://github.com/bazelbuild/bazel/issues/1262,
which significantly changes execroot paths (external/foo becomes ../foo, essentially).
Without this patch, the execroot change messes up the symlink tree the go rules
create.

This compensates for the path change in a backwards-compatible way:
bazel test //... passes for me with old & patched versions of bazel.

### DIFF
--- a/examples/proto/BUILD
+++ b/examples/proto/BUILD
@@ -1,4 +1,4 @@
-load("//go:def.bzl", "go_library")
+load("//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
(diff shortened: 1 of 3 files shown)


pmbethe09:
LGTM, but can you also look at 
https://github.com/bazelbuild/rules_go/blob/master/proto/go_proto_library.bzl
and make sure it is ready for this change as well?

kchodorow:
It seems to work!  It looks like you actually updated it a couple months ago (probably when I last tried to submit this change).  I added a test for it, lmk if that covers what you were thinking of.

pmbethe09:
LGTM

PR #249, opened 2017-01-25
Data is not included transitively in dependant rules

achew22:
This is one part PR, one part issue.

As it says in the description, data is not pulled in transitively to targets. Could you help me understand what I need to do to this sample to make it work?

### DIFF
--- a/examples/data/BUILD
+++ b/examples/data/BUILD
@@ -1,0 +1,16 @@
+load("//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["data.go"],
+    data = ["data.txt"],
+    visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = ["data_test.go"],
+    # Uncomment this line and the demo test will pass.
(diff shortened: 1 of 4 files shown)


pmbethe09:
Unlike Java, Go does not have a built-in way to embed resources in the compiled object.
AFAIK, people write custom code-generators to embed the resources in a .go file with a public variable/method for access.  So it makes no sense for data to be transitively inherited, as a go_binary would just ignore it.

We could do something similar in rules_go - a go_embed_data(...) rule, which takes one or more text files and makes a .go file with variables containing the text inside `` named by the file name (with some escaping of course).

amlinux:
Sample code doesn't ask for embedding data in the binary. It marks data.txt as a runtime dependency that should be available to the binary when it runs.

jayconrod:
@amlinux, that's correct, data.txt should be part of the runfiles, and it should be available to the test when it's run with "bazel test". I'm not sure yet why it isn't getting picked up. `runfiles` is a little underdocumented, so this may require digging into the Bazel source code.

jayconrod:
This should be fixed in #324, so I'll close this. Let me know if anything is not working as expected.

PR #251, opened 2017-01-27
point to latest buildifier

pmbethe09:
switch to new import path in go and BUILD files

PR #253, opened 2017-01-30
Fix dependencies on buildifier

adrianludwin:
Buildifier changed one of their packages from "core" to "build," thereby breaking Gazelle. This change makes Gazelle compile again, passes all tests and works correctly in my project.


adrianludwin:
Duplicate of https://github.com/bazelbuild/rules_go/pull/251

pmbethe09:
... it only breaks Gazelle at HEAD, and as you saw a companion change is in flight.

PR #263, opened 2017-02-06
Fix 'protoc' work directory deletion in go_proto_library.bzl

joseferminj:
go_proto_library.bzl creates a work dir and copy all the protos. It
removes it in each invocation but the 'rm' shell command does not
include the -r option.


bazel-io:

pmbethe09:
Jenkins, test this please.

PR #267, opened 2017-02-08
Integrates filter_tags into emit_go_compile_action.

robotoer:
This change adds basic golang build constraint support to rules_go by integrating the already existing `filter_tags` binary.

The meat of the change is in the `emit_go_compile_action` func where filter_tags is run against the list of go sources producing a filtered list of go sources that are passed to the go compile tool.

I also had to create a new set of go_library, go_binary, and go_test rules for building filter_tags without requiring itself (preventing a circular dependency). This means that tools built with these rules will not support build constraints, but generic builds will.

The overall effect of this change is making go_library and friends only compile sources that meet their build constraints.

Potential future work:
* Add support for defining custom tags in the BUILD file that should be considered "satisfied":
  
      go_library(
          tags=['release'],
          ...
      )

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -217,19 +217,39 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
   if _is_external(out_dir):
     out_depth -= 2
   cmds = symlink_tree_commands(out_dir, tree_layout)
+
+  extra_inputs = ctx.files.toolchain
+
+  # cd into the out_dir.
+  cmds.append("export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..")
+  cmds.append(" ".join(["cd", out_dir]))
+
+  # Apply build tags here.
+  use_build_tags = hasattr(ctx.attr, "filter_tags")
+  cleaned_source_paths = [prefix + _remove_external_prefix(i.path) for i in sources]
+  if use_build_tags:
+    filter_args = [
+        ('../' * out_depth) + ctx.executable.filter_tags.path,
+    ]
+    filter_args.extend(cleaned_source_paths)
+    cmds.append("export GOFILES=\"$({})\"".format(" ".join(filter_args)))
+    extra_inputs += [ctx.executable.filter_tags]
+  else:
+    cmds.append("export GOFILES='{}'".format(" ".join(cleaned_source_paths)))
+
   args = [
-      "cd ", out_dir, "&&",
       ('../' * out_depth) + ctx.file.go_tool.path,
       "tool", "compile",
       "-o", ('../' * out_depth) + out_lib.path, "-pack",
-      "-I", "."
+      "-I", ".",
+      "${GOFILES}",
   ]
 
   # Set -p to the import path of the library, ie.
   # (ctx.label.package + "/" ctx.label.name) for now.
-  cmds += [ "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
-    ' '.join(args + [prefix + _remove_external_prefix(i.path) for i in sources])]
-  extra_inputs = ctx.files.toolchain
+  cmds += [
+    ' '.join(args),
+  ]
 
   if extra_objects:
     extra_inputs += extra_objects
@@ -558,6 +578,15 @@ go_library_attrs = go_env_attrs + {
     ),
 }
 
+go_tool_attrs = {
+    "filter_tags": attr.label(
+        default = Label("//go/tools/filter_tags:filter_tags"),
+        cfg = "host",
+        executable = True,
+        allow_files = True,
+    ),
+}
+
 _crosstool_attrs = {
     "_crosstool": attr.label(
         default = Label("//tools/defaults:crosstool"),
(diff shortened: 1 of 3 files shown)


bazel-io:

pmbethe09:
Jenkins test this please

go/def.bzl
```diff
@@ -217,19 +217,39 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
...
+      "-I", ".",
+      "${GOFILES}",
   ]
 
   # Set -p to the import path of the library, ie.
   # (ctx.label.package + "/" ctx.label.name) for now.
-  cmds += [ "export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..",
-    ' '.join(args + [prefix + _remove_external_prefix(i.path) for i in sources])]
-  extra_inputs = ctx.files.toolchain
+  cmds += [
```
pmbethe09: fits on 1 line

go/def.bzl
```diff
@@ -217,19 +217,39 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
   if _is_external(out_dir):
     out_depth -= 2
   cmds = symlink_tree_commands(out_dir, tree_layout)
+
+  extra_inputs = ctx.files.toolchain
```
pmbethe09: should copy this

go/def.bzl
```diff
@@ -217,19 +217,39 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
...
+  # Apply build tags here.
+  use_build_tags = hasattr(ctx.attr, "filter_tags")
+  cleaned_source_paths = [prefix + _remove_external_prefix(i.path) for i in sources]
+  if use_build_tags:
+    filter_args = [
+        ('../' * out_depth) + ctx.executable.filter_tags.path,
+    ]
+    filter_args.extend(cleaned_source_paths)
+    cmds.append("export GOFILES=\"$({})\"".format(" ".join(filter_args)))
+    extra_inputs += [ctx.executable.filter_tags]
```
pmbethe09: can these 3 lines be more consistent?  .extend, then .append, then += [...]

robotoer:
Ah sorry, didn't see that someone had commented. There are 2 bugs I have found with the current implementation while testing:
- filter_tags removes cgo_generated go files incorrectly.
- The _cgo_codegen rule needs to only compile some of the generated cgo files as one is generated per platform supported by cgo.

I will update this PR in the next few days.

go/def.bzl
robotoer: Ah, yes will update.

go/def.bzl
```diff
@@ -592,6 +621,53 @@ go_binary = rule(
...
+    fragments = ["cpp"],
+    outputs = {
+        "lib": "%{name}.a",
+        "main_lib": "%{name}_main_test.a",
+        "main_go": "%{name}_main_test.go",
+    },
+    test = True,
+)
+
+# go_tool_* rules are for use building tools required by BUILD rules.
```
jayconrod: I don't think we need these rules. You should be able to add filter_tags to the set of tools we build with "go build" inside go_repositories. See _go_repositories_tools_impl in go/private/go_repositories.bzl. fetch_repo and gazelle are already built this way.

go/def.bzl
```diff
@@ -558,6 +578,15 @@ go_library_attrs = go_env_attrs + {
     ),
 }
 
+go_tool_attrs = {
+    "filter_tags": attr.label(
```
jayconrod: I think this should be added to go_env_attrs instead. It should be a private attribute ("_filter_tags"). I don't think anyone would want to override this.

go/def.bzl
```diff
@@ -217,19 +217,39 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
...
   cmds = symlink_tree_commands(out_dir, tree_layout)
+
+  extra_inputs = ctx.files.toolchain
+
+  # cd into the out_dir.
+  cmds.append("export GOROOT=$(pwd)/" + ctx.file.go_tool.dirname + "/..")
+  cmds.append(" ".join(["cd", out_dir]))
+
+  # Apply build tags here.
+  use_build_tags = hasattr(ctx.attr, "filter_tags")
```
jayconrod: If _filter_tags is in go_env_attrs (see comment below), then you won't need this test. Just assume it's there.

go/def.bzl
```diff
@@ -217,19 +217,39 @@ def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
...
+    cmds.append("export GOFILES='{}'".format(" ".join(cleaned_source_paths)))
+
   args = [
-      "cd ", out_dir, "&&",
       ('../' * out_depth) + ctx.file.go_tool.path,
       "tool", "compile",
       "-o", ('../' * out_depth) + out_lib.path, "-pack",
-      "-I", "."
+      "-I", ".",
+      "${GOFILES}",
```
jayconrod: I don't think this will correctly handle filenames with spaces. Could you add a TODO here or above? I'll fix this before the next release.

jayconrod:
@robotoer I discussed this PR with some other folks on the Go team. There is some code within Google that does tag filtering for Cgo in a way that go/build does not. filter_tags just relies on go/build, and it probably can't do what we need. We will probably open source this code (or something like it).

Not sure if you want to proceed with this PR. A lot of things will probably change around this area.

jayconrod:
@robotoer I'm closing this PR since we merged #311, which adds the same functionality. Thanks for bringing this up. I ended up taking a similar approach.

robotoer:
Awesome! Thanks for finishing this!

PR #277, opened 2017-02-24
Handle empty go_prefix in gcc invocation

amlinux:
Empty go_prefix is useful for projects not hosted in public repositories. rules_go handles empty prefix well, except for gcc linking where it produces "-L" flag without arguments. This PR removes "-L" if prefix is empty.


bazel-io:

amlinux:
I signed it!

jayconrod:
@pmbethe09 WDYT? Should the flag be "-L." instead of nothing if the prefix is empty? What could be linked here? Would anything else break if the prefix is empty?

pmbethe09:
I always use a dummy prefix so that gazelle can differentiate my code with Go-core libraries, so I don't really understand the need for an empty go-prefix.

With that said, removing -L seems fine to me.

PR #278, opened 2017-02-24
Enable test filtering with --test_filter

### LINKED ISSUE #272: go_test should filter tests using --test_filter
Bazel allows users to pass in a pattern for matching test cases within a test using [`--test_filter=filter-expression`](https://bazel.build/versions/master/docs/bazel-user-manual.html#flag--test_filter). This pattern is passed to the test runner through the `TESTBRIDGE_TEST_ONLY` environment variable.

The test runner generated by //go/tools:generate_test_main should treat this environment variable as a regular expression and filter tests accordingly. Currently, we pass a function to `testing.Main` which matches all tests.

jayconrod:
Modified generate_test_main.go to map the TESTBRIDGE_TEST_ONLY
environment variable (set by Bazel for --test_filter) onto the
-test.run flag, which is defined and used by the testing package.

Fixes #272

### DIFF
--- a/go/tools/generate_test_main.go
+++ b/go/tools/generate_test_main.go
@@ -128,6 +128,7 @@ func main() {
 	tpl := template.Must(template.New("source").Parse(`
 package main
 import (
+	"flag"
 	"os"
 	"testing"
 	"testing/internal/testdeps"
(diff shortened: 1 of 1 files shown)


pmbethe09:
LGTM

PR #279, opened 2017-02-27
cgo libraries are now built with -pthread

jayconrod:
This matches behavior of "go build", which compiles and links C code
with -pthread (or -mthreads on Windows).


pmbethe09:
LGTM

PR #282, opened 2017-02-28
gazelle: don't rewrite files with gazelle:ignore

### LINKED ISSUE #266: gazelle can modify a "# gazelle:ignore" BUILD file
Running gazelle to regenerate a BUILD file containing `# gazelle:ignore` can modify the existing file. It appears that the changes are only reordering/cleanup of the existing rules, with no structural changes. However, the docs say "# gazelle:ignore in a BUILD file will instruct gazelle to leave the file alone.", which isn't true.

One possible fix would be to update the documentation for accuracy.

Another possible fix would be to avoid writing out anything when the ignore directive appears. 

I can see arguments for both the cleanup mode (the existing behavior) and the strict non-intervention mode (what the documentation says). I think either of my proposals would address the problem.

The ideal solution might be to use the strict mode by default, then also provide a `--cleanup` mode which executes the existing cleanup behavior.

### To reproduce
```
mkdir -p gazelle_test/code
cd gazelle_test/
echo "package main\nfunc main() {}" > code/main.go
print "# gazelle:ignore\n\n\n# some comment" > BUILD.orig
print "# gazelle:ignore\n\n\n# some comment" > BUILD
gazelle --repo_root=. --go_prefix=code
diff code/BUILD.orig code/BUILD
```
produces:
```
3,4c3
< 
< # some comment
\ No newline at end of file
---
> # some comment
```
i.e. some whitespace was removed. Other changes are also possible, such as reordering of build rule attributes.

### Random implementation thought

One possible approach to implementing the strict mode might be to use something like this in gazelle's main.go where MergeWithExisting() is called.
https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/gazelle/main.go#L101
```
f, err := merger.MergeWithExisting(f, existingFilePath)
switch err.(type) {
    case nil:
        // write out the generated file as usual
    case GazelleIgnoreErr:  // "# gazelle:ignore" was seen
        // cancel further processing of this BUILD, but continue
        // the loop to process other BUILDs.
        continue  
    case error:
        return err // interrupt the loop and cancel program execution          
}
```

jayconrod:
Previously, Gazelle would avoid merging files with a "#
gazelle:ignore" comment, but it would still rewrite them. This could
result in small changes, e.g., whitespace.

Gazelle will no longer rewrite these files at all.

Fixes #266.

### DIFF
--- a/go/tools/gazelle/merger/merger_test.go
+++ b/go/tools/gazelle/merger/merger_test.go
@@ -92,6 +92,7 @@ go_library(
 
 type testCase struct {
 	previous, current, expected string
+	ignore                      bool
 }
 
 func TestMergeWithExisting(t *testing.T) {
@@ -104,8 +105,8 @@ func TestMergeWithExisting(t *testing.T) {
 	}
 	defer os.Remove(tmp.Name())
 	for _, tc := range []testCase{
-		{oldData, newData, expected},
-		{ignore, newData, ignore},
+		{oldData, newData, expected, false},
+		{ignore, newData, "", true},
 	} {
 		if err := ioutil.WriteFile(tmp.Name(), []byte(tc.previous), 0755); err != nil {
 			t.Fatal(err)
@@ -115,8 +116,16 @@ func TestMergeWithExisting(t *testing.T) {
 			t.Fatal(err)
 		}
 		afterF, err := MergeWithExisting(newF, tmp.Name())
-		if err != nil {
+		if _, ok := err.(GazelleIgnoreError); ok {
+			if !tc.ignore {
+				t.Fatalf("unexpected ignore: %v", err)
+			}
+			continue
+		} else if err != nil {
 			t.Fatal(err)
+		}
+		if tc.ignore {
+			t.Error("expected ignore")
 		}
 		if s := string(bzl.Format(afterF)); s != tc.expected {
 			t.Errorf("bzl.Format, want %s; got %s", tc.expected, s)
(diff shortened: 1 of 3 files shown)


jayconrod:
Travis builds are timing out due to an Amazon S3 outage. [Travis status](https://www.traviscistatus.com/incidents/hmwq9yy5dh9d)

I'll rerun the builds when it's back up.

pmbethe09:
Not sure about this one. 
Does '# gazelle:ignore' mean "leave this file completely alone", or just "only run buildifier".

IMHO, all of your BUILD files should be formatted by buildifier, so in the case of gazelle:ignore, rather than doing nothing, you should return the on-disk AST and then write the file back out.

jayconrod:
README.md says:

> `# gazelle:ignore` in a BUILD file will instruct gazelle to leave the file alone.

I agree that buildifier should be run in most cases, but I think users should have a way to opt out of it. If we want to preserve the old behavior, maybe we should rename the directive to "# gazelle:format-only"?

pmbethe09:
Hmm, good point.

go/tools/gazelle/merger/merger_test.go
```diff
@@ -115,9 +116,17 @@ func TestMergeWithExisting(t *testing.T) {
...
+		if _, ok := err.(GazelleIgnoreError); ok {
+			if !tc.ignore {
+				t.Fatalf("unexpected ignore: %v", err)
+			}
+			continue
+		} else if err != nil {
 			t.Fatal(err)
 		}
+		if tc.ignore {
+			t.Fatal("expected ignore")
```
pmbethe09: Error ?

jayconrod: Fixed. Will submit when Travis is back up.

PR #289, opened 2017-03-03
Add a framework for testing things that can't be checked through bazel

### LINKED ISSUE #287: Add non-Bazel integration tests
Portions of this repository's functionality can't be tested directly through Bazel. For example:

* There's no easy way to test that when a user supplies the --test_filter command line argument when running a go_test, the generated test runner filters the tests correctly.
* It would be good to check that Gazelle runs successfully on *this* repository (see #285).

I'm thinking of creating a Bash script, something like tests/run_non_bazel_tests.bash. This would run other scripts that check these kinds of things. The top-level script would be run by Travis.

jayconrod:
- Included an initial test case to check that --test_filter is
  observed by go_tests.
- The test framework is run through Travis CI.

Fixes #287

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ script:
       --genrule_strategy=standalone \
       --local_resources=400,1,1.0 \
       //...
+  - tests/run_non_bazel_tests.bash
 
 notifications:
   email: false
--- a/tests/run_non_bazel_tests.bash
+++ b/tests/run_non_bazel_tests.bash
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# This script executes tests that can't be executed through Bazel. For example,
+# we may need to test the behavior of the rules when certain flags are passed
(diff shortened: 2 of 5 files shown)


BUILD
```diff
@@ -19,3 +19,5 @@ lines_sorted_test(
     error_message = "Authors must be sorted by first name",
     file = "AUTHORS",
 )
+
+load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
```
pmbethe09: this appears to be a bug that it adds this not detecting it as the same as the first line in the file.

jayconrod: Oops, that shouldn't have made it into this commit.

I think this is left over from when I ran without setting -go_rules_bzl_only_for_internal_use. Maybe gazelle should detect the repository name (and whether it's running in this repository) automatically.

pmbethe09:
LGTM

PR #291, opened 2017-03-07
Support gc_goopts and gc_linkopts attributes

### LINKED ISSUE #217: Add support for gc_linkopts
With the same functionality as internal go rules.

jayconrod:
go_library, cgo_library, go_binary, and go_test now support the
gc_goopts attribute. This is a list of flags to pass to the Go
compiler.

go_binary and go_test also support the gc_linkopts attribute. This is
a list of flags to be passed to the Go linker.

Note that _emit_go_compile_action and _emit_go_link_action have had
small interface changes and are now private. I didn't see any uses of
these functions on GitHub, and I'd like to avoid having people depend
on them in the future.

Fixes #217

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -206,7 +206,8 @@ def _is_external(p):
   """Checks if the string starts with ../"""
   return p[0:3] == '../'
 
-def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
+def _emit_go_compile_action(ctx, sources, deps, out_lib,
+                            extra_objects, gc_goopts):
   """Construct the command line for compiling Go code.
   Constructs a symlink tree to accommodate for workspace name.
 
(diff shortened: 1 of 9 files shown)


jayconrod:
cc @alandonovan  @mikedanese

go/def.bzl
```diff
@@ -206,7 +206,8 @@ def _is_external(p):
   """Checks if the string starts with ../"""
   return p[0:3] == '../'
 
-def emit_go_compile_action(ctx, sources, deps, out_lib, extra_objects=[]):
+def _emit_go_compile_action(ctx, sources, deps, out_lib,
+                           extra_objects=[], gc_goopts=[]):
```
pmbethe09: I think skylark is immune to the python bug where a default argument of [] is bad, but I still see common skylark practice to follow the convention of 'None' as the default value and then a 
if not gc_goopts:
  gc_goopts = []

jayconrod: I'm not sure if Skylark is affected by this. I figured it's better to just make the arguments explicit though.

Will submit when CI is done again (Mac build is taking forever to run these days).

Issue #294, opened 2017-03-09
new_go_repository should export files other than .go


rosun82:
Amazon AWS unittest failure because of this.

bazel test -c opt @com_github_aws_aws_sdk_go//...

@com_github_aws_aws_sdk_go//aws/credentials:go_default_test              FAILED in 1 out of 2 in 0.1s
@com_github_aws_aws_sdk_go//aws/session:go_default_test                  FAILED in 1 out of 2 in 0.1s
@com_github_aws_aws_sdk_go//vendor/github.com/go-ini/ini:go_default_test FAILED in 1 out of 2 in 0.1s

The reason is aws/credentials/example.ini is not exported from BUILD but used in unittest.

pmbethe09:
How would we know to reference that file, or should we just always add such files to a 'data' attribute?

jayconrod:
I think Gazelle (which generates BUILD files inside new_go_repository) should add files in testdata directories to the data attribute of go_default_test rules, probably through a glob. The go tool has some special handling of this directory already, and we probably should, too.

Tests probably still won't pass because the initial directory of the test is logically the root of the repository, not the directory of the test sources.

I don't think there's any need for Gazelle to add other data files to the data attribute or export them. Custom BUILD files should be used if that's desired.

rosun82:
I like pmbethe09's solution better, maybe adding all these files to a target and put data=target in all unittest.

The point is, there could (should) be many open source go projects, where the author put data files in the code folder and "go test" happily passes (cannot blame them as they usually do not have 3 Googlers reviewing their code, and they do not start from using blaze). When you import those projects as bazel project, suddenly a ton of unittest fails, which creates a lot of trouble for everyone.

jayconrod:
How would Gazelle know which files to include though? Including a file makes it available for a test, but the test has to be re-run whenever that file changes. Data files also have to be copied when tests are run remotely. That might be annoying if a directory contains generated documentation or a definition index, or something else that changes frequently.

I could see making this behavior available behind an option, but I don't think it should be the default.

pmbethe09:
And separate point of @jayconrod, should the rules_go testrunner for go_test change to the code directory, which would be inline with how 'go test' works, but different than the bazel-style of running from the repository-root?

jayconrod:
The directory issue is a really good point. We should support both conventions. I've filed #299 to track that.

jayconrod:
[...]

PR #300, opened 2017-03-10
gazelle adds a data dependency for tests on testdata files

jayconrod:
If a directory named "testdata" is present in a Go package, now adds a
"data" attribute to newly generated rules that globs the contents of
the directory.

    go_test(
        name = "go_default_test",
        ...
        data = glob(["testdata/**"]),
    )

This attribute is included in newly generated internal and external
tests. It is not merged into existing test rules.

Related to #294

### DIFF
--- a/go/tools/gazelle/rules/construct.go
+++ b/go/tools/gazelle/rules/construct.go
@@ -27,6 +27,11 @@ type keyvalue struct {
 	value interface{}
 }
 
+type globvalue struct {
+	patterns []string
+	excludes []string
+}
+
 func newRule(kind string, args []interface{}, kwargs []keyvalue) (*bzl.Rule, error) {
 	var list []bzl.Expr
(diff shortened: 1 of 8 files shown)


go/tools/gazelle/rules/construct.go
```diff
@@ -77,7 +82,28 @@ func newValue(val interface{}) (bzl.Expr, error) {
 			list = append(list, elem)
 		}
 		return &bzl.ListExpr{List: list}, nil
-	default:
-		return nil, fmt.Errorf("not implemented %T", val)
+	case reflect.Struct:
+		if glob, ok := val.(globvalue); ok {
```
pmbethe09: following Go guard-clause style, this should be:
glob, ok := val.(globvalue)
if !ok {
  return nil, ... error ...
}

PR #307, opened 2017-03-13
Create library for interacting with Bazel from Go

achew22:
This commit creates a new package
`golang.com/bazelbuild/rules_go/tools/bazel` which calls out to the
Bazel CLI and allows you to run, build, test and query from Go.

Example uses of this library:

1.  Creating a tool to watch files required to build a target and
    retrigger `bazel test` when the files change.
2.  Creating a tool that prunes dependencies by inspecting the graph
    in combination with build artifacts to determine if a resource
    was actually used.
3.  Other things that I haven't imagined yet.

### DIFF
--- a/go/tools/bazel/BUILD
+++ b/go/tools/bazel/BUILD
@@ -1,0 +1,14 @@
+load("//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["bazel.go"],
+    visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "go_default_test",
+    size = "small",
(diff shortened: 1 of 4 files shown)


pmbethe09:
Seems like a useful library -- but does it belong in the core rules_go repository?

jayconrod:
Agree with Paul. This seems generally useful for all languages, but I think it's out of scope for rules_go.

Maybe chat with bazel-discuss or bazel-dev about whether this can be a top-level repo in the bazelbuild org? There are a few other tools in there like tulsi, so it's not without precedent.

jayconrod:
Closing PR, since a separate open source launch process is underway for this.

Looking forward to seeing this!

PR #308, opened 2017-03-14
Gazelle finds "gazelle:ignore" before and after top-level statements

### LINKED ISSUE #302: Extra newline required after `# gazelle:ignore`.
The gazelle [documentation](https://github.com/bazelbuild/rules_go/tree/master/go/tools/gazelle) does not indicate that a newline is required after the directive `# gazelle:ignore`. I assumed I didn't have the right version of gazelle installed.  It's only by chance that I added an extra newline while trying to figure out what was  going wrong.

jayconrod:
Previously, Gazelle only searched comments after top-level statements
for the "gazelle:ignore" comment. This caught comments at the top of
the file (since the buildifier parser adds an implicit top-level
statement to hold comments), but only if there was a blank line before
the first real statement.

With this change, comments before top-level statements are now
recognized.

Fixes #302

### DIFF
--- a/go/tools/gazelle/merger/merger.go
+++ b/go/tools/gazelle/merger/merger.go
@@ -57,12 +57,8 @@ func MergeWithExisting(newfile *bzl.File, existingFilePath string) (*bzl.File, e
 	if err != nil {
 		return nil, err
 	}
-	for _, s := range f.Stmt {
-		for _, c := range s.Comment().After {
-			if strings.HasPrefix(c.Token, gazelleIgnore) {
-				return nil, GazelleIgnoreError{c.Start}
-			}
-		}
+	if err := shouldIgnore(f); err != nil {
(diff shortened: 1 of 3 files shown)


go/tools/gazelle/merger/merger.go
```diff
@@ -57,12 +57,9 @@ func MergeWithExisting(newfile *bzl.File, existingFilePath string) (*bzl.File, e
 	if err != nil {
 		return nil, err
 	}
-	for _, s := range f.Stmt {
-		for _, c := range s.Comment().After {
-			if strings.HasPrefix(c.Token, gazelleIgnore) {
-				return nil, GazelleIgnoreError{c.Start}
-			}
-		}
+	err = shouldIgnore(f)
```
pmbethe09: push into if, use :=

PR #309, opened 2017-03-14
Combine extldflags lists when linking

### LINKED ISSUE #305: Using -extldflags in gc_linkopts gets ignored by go tool link
Related to #217 and #161 
When passing `-extldflags` with `gc_linkopts`:
```
go_binary(
    name = "foo",
    srcs = ["foo.go"],
    gc_linkopts = [
        "-extldflags \"-static\"",
    ],
)
```
the -extldflags gets added to `go tool link` twice due to `_emit_go_link_action` (https://github.com/bazelbuild/rules_go/blob/master/go/def.bzl#L442).
It looks like `go tool link` then only uses the second set of flags, meaning the flags passed in through `gc_linkopts` are ignored.
Ideally these flags should be merged so that only one `-extldflags` is emitted.

jayconrod:
_emit_go_link_action now scans gc_linkopts for -extldflags
arguments. The Go linker will only use the last list passed to it
(earlier lists are silently ignored), so we combine these into one
external flags list.

Fixes #305


go/def.bzl
```diff
@@ -431,6 +431,17 @@ def _emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
     if d.basename.endswith('.so'):
       dirname = _short_path(d)[:-len(d.basename)]
       ldflags += ["-Wl,-rpath,$ORIGIN/" + ("../" * pkg_depth) + dirname]
+  filtered_gc_linkopts = []
+  is_extldflags = False
```
pmbethe09: whenever I see this sort of loop-surviving boolean, it screams "factor out a method" to me.

jayconrod: Good idea. I've factored out _extract_extldflags, which returns an updated version of both lists.

Looks like something's going wrong with the Mac build; will fix on my laptop when I get home today.

tests/extldflags_static_link/extldflags_link_test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+result=0
+while [ $# -gt 0 ]; do
+  path=$1
+  pattern=$2
+  shift 2
+  if ! file --dereference "$path" | grep --quiet "$pattern"; then
```
pmbethe09: I don't think --dereference exists in the file command on Mac

jayconrod:
As it turns out, macOS [doesn't support static linking with gcc](https://developer.apple.com/library/content/qa/qa1118/_index.html), which I found pretty surprising.

I think I found a better way to test this though. Forcing Go to use the external linker and setting additional rpaths has an observable effect. It's also easy to do this in multiple -extldflags groups.

Not sure how we're going to manage static linking on macOS later on though.

PR #314, opened 2017-03-17
Remove support for tags from gazelle

### LINKED ISSUE #310: Support for build tags
(Work on this feature is already underway, but it's good to have an issue to track it).

Go supports [build tags](https://golang.org/pkg/go/build/#hdr-Build_Constraints) in a specially formatted comment at the top of Go files.

```go
// +build linux,386 darwin,!cgo
```

These tags provide conditional compilation with the go tool. The Bazel Go rules should interpret these tags and provide conditional compilation as well. There are three approaches I can think of for this.

1. Gazelle interprets these tags when scanning source files and generates BUILD files appropriate for the the platform Gazelle runs on. Gazelle does this already.
2. Gazelle could translate build tags into an elaborate set of `config_setting`s and `select`s.
3. The compile action could filter build files before providing them to the compiler during execution.

I believe (3) is the most reasonable solution. (1) means that BUILD files are platform-specific. Developers on macOS and Linux can't share BUILD files generated by Gazelle, and they can't be checked into VCS. (2) will create BUILD files that are difficult to maintain with tools. I'm not convinced it's even possible to reduce build tags into `config_setting`s and `select`s.

(3) doesn't have either of these problems. It simplifies BUILD files for users that don't want to generate them with Gazelle. BUILD files won't need to be updated (with Gazelle or otherwise) when tags are changed in source files.

#267 is a work-in-progress PR to support (3). I'm working on fixing this up and adding tests, and I expect to have something ready within the next week. After that's merged, I'll remove build tags support from Gazelle and have it match all sources.

jayconrod:
Since #311, the Bazel rules filter sources with build tags and
filename suffixes automatically.

Gazelle will no longer produce BUILD files which are pre-filtered.
This allows the same BUILD files to be used on multiple platforms.

Fixes #310

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -41,8 +41,7 @@ def _new_go_repository_impl(ctx):
   _go_repository_impl(ctx)
   gazelle = ctx.path(ctx.attr._gazelle)
 
-  cmds = [gazelle, '--go_prefix', ctx.attr.importpath, '--mode', 'fix',
-          "--build_tags", ",".join(ctx.attr.build_tags)]
+  cmds = [gazelle, '--go_prefix', ctx.attr.importpath, '--mode', 'fix']
   if ctx.attr.build_file_name:
       cmds += ["--build_file_name", ctx.attr.build_file_name]
   if ctx.attr.rules_go_repo_only_for_internal_use:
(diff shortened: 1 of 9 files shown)


pmbethe09:
Not quite sure why this is broken

jayconrod:
It looks like this is failing because golang.org/x/net/html/atom has sources for multiple packages, "atom" and "main". The source file for "main" had the tag `// +build ignore`, so we didn't pick it up before.

I think for now, Gazelle should match that one tag to avoid this error. Long term, we should tolerate multiple packages in the same directory. We've talked about doing that anyway for Blaze-style projects. We'll need to fork `go/build` for this, since the multiple package error comes from there.

jayconrod:
Bazel seems to be crashing at HEAD. I'm going to file a bug report.

PR #316, opened 2017-03-17
Enable cgo tag by default during compilation

jayconrod:
Previously, when calling filter_tags, _emit_go_compile_action set the
cgo tag conditionally, based on whether any cgo code was present. This
did not match the behavior of "go build", which has the cgo tag based
on the CGO_ENABLED environment variable.

The cgo tag is now enabled whether any cgo code is present or not.


achew22:
How does this affect compile times for rules_go based code?

jayconrod:
@achew22 It depends on the project. This means files with `// +build cgo` *will* be compiled and files with `// +build !cgo` *won't* be compiled, whether or not cgo code is present. This should match the behavior of "go build", so it's more of a correctness issue.

Issue #313, opened 2017-03-16
Allow gazelle to be aware of go_prefix mappings defined in the workspace.


wlynch:
# Summary

Gazelle currently does not handle remote dependencies with a go_prefix value that is not [URL-like](https://github.com/bazelbuild/rules_go/blob/9be311aedccedabf575d152a07daf17fe5d4fdf6/go/tools/gazelle/rules/generator.go#L398). For example, if a remote dep has a go_prefix of "foo", gazelle currently assumes that this package is a part of the standard library and should not try and resolve the dep.

# Proposal

Allow gazelle to be aware of go_prefix mappings defined in the workspace.

If gazelle encounters an import path matching this prefix, it should use this when it generates the deps. A prerequisite of this is to allow fetching of repos using non-URL like import paths.

### Phase 1: Modify fetch_repo to take in remote and import path. 

If given remote, assume that this is the exact URL and we shouldn't modify it.

This allows fetch_repo and go_repository to work on non-standard URL paths if the user specifies it explicitly. This has a few benefits: 

1) This brings the use of remote in line with [git_repository](https://bazel.build/versions/master/docs/be/workspace.html#git_repository.remote).
2) This makes go_repository rules easier to work with for private repositories that don't expose [go-import meta tags](https://golang.org/cmd/go/#hdr-Remote_import_paths). Currently teams get around this by using the new [git_repository](https://github.com/bazelbuild/bazel/blob/8d635fc57e37d53dceceebb3e0868005530f7d32/tools/build_defs/repo/git.bzl) rule instead.

### Phase 2: Make Gazelle aware of WORKSPACE files.

Gazelle already finds the the WORKSPACE file when running to determine the repo root. We should use the WORKSPACE to map go_prefixes to deps. If gazelle finds an import path that matches something in this map, then it uses that dep in its output. If gazelle cannot find a match, then it will default to the [existing resolve_external behavior](https://github.com/bazelbuild/rules_go/blob/9be311aedccedabf575d152a07daf17fe5d4fdf6/go/tools/gazelle/rules/resolve_external.go).

In the sandbox use case of new_go_repository where gazelle may not have access to the WORKSPACE file, the existing behavior should be sufficient. This is because you generally use new_go_repository because the remote dep has no bazel build files, therefore they won't override their default go_prefix, and their deps will also follow standard URL-like go import paths (following go get behavior). From an implementation standpoint, we shouldn't need to make any changes for this behavior since the dep override map described above should be empty.

[...]

PR #319, opened 2017-03-20
go_repository: add importpath and vcs flags (#313)

wlynch:
This change allows fetch_repo to fetch repositories without attempting
to determine the repo root if the user supplies the information
themselves. This allows users to be able to fetch repositories that the
vcs package would otherwise not be able to recognize.

This change also changes the behavior of the --remote flag to match the
git_repository rule.

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -27,12 +27,16 @@ def _go_repository_impl(ctx):
 
   # TODO(yugui): support submodule?
   # c.f. https://www.bazel.io/versions/master/docs/be/workspace.html#git_repository.init_submodules
-  remote = ctx.attr.remote if ctx.attr.remote else ctx.attr.importpath
+  remote = ctx.attr.remote
+  vcs = ctx.attr.vcs
+  importpath = ctx.attr.importpath
   result = ctx.execute([
       fetch_repo,
       '--dest', ctx.path(''),
       '--remote', remote,
-      '--rev', rev])
+      '--rev', rev,
+      '--vcs', vcs,
+      '--importpath', importpath])
   if result.return_code:
     fail("failed to fetch %s: %s" % (remote, result.stderr))
 
@@ -58,8 +62,9 @@ def _new_go_repository_impl(ctx):
 
 _go_repository_attrs = {
     "build_file_name": attr.string(),
-    "importpath": attr.string(mandatory = True),
+    "importpath": attr.string(),
     "remote": attr.string(),
+    "vcs": attr.string(default="git", values=["git", "hg", "svn", "bzr"]),
     "commit": attr.string(),
     "tag": attr.string(),
     "build_tags": attr.string_list(),
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -38,7 +38,7 @@ def _go_repository_impl(ctx):
       '--vcs', vcs,
       '--importpath', importpath])
   if result.return_code:
-    fail("failed to fetch %s: %s" % (remote, result.stderr))
+    fail("failed to fetch %s: %s" % (ctx.name, result.stderr))
 
 
 def _new_go_repository_impl(ctx):
@@ -64,7 +64,7 @@ _go_repository_attrs = {
     "build_file_name": attr.string(),
     "importpath": attr.string(),
     "remote": attr.string(),
-    "vcs": attr.string(default="git", values=["git", "hg", "svn", "bzr"]),
+    "vcs": attr.string(default="", values=["", "git", "hg", "svn", "bzr"]),
     "commit": attr.string(),
     "tag": attr.string(),
     "build_tags": attr.string_list(),
--- a/go/tools/fetch_repo/fetch_repo_test.go
+++ b/go/tools/fetch_repo/fetch_repo_test.go
@@ -72,7 +72,6 @@ func TestGetRepoRoot_error(t *testing.T) {
 		remote     string
 		cmd        string
 		importpath string
-		r          *vcs.RepoRoot
 	}{
 		{
 			label:  "importpath as remote",
--- a/go/tools/fetch_repo/fetch_repo_test.go
+++ b/go/tools/fetch_repo/fetch_repo_test.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+
+	"golang.org/x/tools/go/vcs"
+)
+
+var (
+	root = &vcs.RepoRoot{
+		VCS:  vcs.ByCmd("git"),
+		Repo: "https://github.com/bazeltest/rules_go",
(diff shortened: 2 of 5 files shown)


go/tools/fetch_repo/main.go
```diff
@@ -19,19 +19,45 @@ import (
...
 )
 
+func getRepoRoot(remote, cmd, importpath string) (*vcs.RepoRoot, error) {
+	r := &vcs.RepoRoot{
+		VCS:  vcs.ByCmd(cmd),
+		Repo: remote,
+		Root: importpath,
+	}
+
+	if remote != "" && importpath == "" && cmd == "" {
```
jayconrod: It's fine to make this an error. I don't think we need to maintain backward compatibility for this tool since it should only be used through the repository rules.

I'll document the change in the next release.

go/tools/fetch_repo/main.go
```diff
@@ -19,19 +19,45 @@ import (
 )
 
 var (
-	remote = flag.String("remote", "", "Go importpath to the repository fetch")
-	rev    = flag.String("rev", "", "target revision")
-	dest   = flag.String("dest", "", "destination directory")
+	remote     = flag.String("remote", "", "The URI of the remote repository. Must be used with the --vcs flag.")
+	cmd        = flag.String("vcs", "", "Version control system to use to fetch the repository. Should be one of: git,hg,svn,bzr. Must be used with the --remote flag.")
```
jayconrod: Check that `remote` and `cmd` are both set or neither set, probably in `main`.

go/tools/fetch_repo/main.go
```diff
@@ -19,19 +19,45 @@ import (
...
+	rev        = flag.String("rev", "", "target revision")
+	dest       = flag.String("dest", "", "destination directory")
+	importpath = flag.String("importpath", "", "Go importpath to the repository fetch")
 )
 
+func getRepoRoot(remote, cmd, importpath string) (*vcs.RepoRoot, error) {
+	r := &vcs.RepoRoot{
+		VCS:  vcs.ByCmd(cmd),
```
jayconrod: Check that `VCS` is not nil. Otherwise, we will get a panic later.

Also, I think this should part of the branch below. This assignment shouldn't dominate the call to `RepoRootForImportPath` where we overwrite it.

```go
if cmd != "" && remote != "" {
  r = &vcs.RepoRoot{...}
} else {
  r, err := vcs.RepoRootForImportPath(importpath, true)
  ...
}
```

go/tools/fetch_repo/fetch_repo_test.go
```diff
@@ -0,0 +1,75 @@
...
+var (
+	root = &vcs.RepoRoot{
+		VCS:  vcs.ByCmd("git"),
+		Repo: "https://github.com/bazeltest/rules_go",
+		Root: "github.com/bazeltest/rules_go",
+	}
+)
+
+func TestGetRepoRoot(t *testing.T) {
+	for _, tc := range []struct {
```
jayconrod: Please add a stub for `vcs.RepoRootForImportPath` since it accesses the network. Gazelle does this in [resolve_external_test.go](https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/rules/resolve_external_test.go).

This might make the tests a little tautological, but I guess the important thing is to verify that the function is called when `cmd` and `remote` aren't specified.

go/tools/fetch_repo/fetch_repo_test.go
```diff
@@ -0,0 +1,75 @@
+package main
```
jayconrod: Please add a rule for this test in the BUILD file.

go/tools/fetch_repo/main.go
```diff
@@ -19,19 +19,45 @@ import (
...
+		Repo: remote,
+		Root: importpath,
+	}
+
+	if remote != "" && importpath == "" && cmd == "" {
+		// User passed in old-style arguments. Assume old behavior for now, but give a warning.
+		log.Println("WARNING: --remote should be used with the --vcs flag. If this is an import path, use --importpath instead.")
+		importpath = remote
+	}
+	if cmd == "" || remote == "" {
```
pmbethe09: invert this

go/tools/fetch_repo/fetch_repo_test.go
wlynch: Done.

go/tools/fetch_repo/main.go
wlynch: Done. I did not do this in main so that I could test that this was an error being returned by getRepoRoot.

go/tools/fetch_repo/main.go
wlynch: Done (slightly different than what you suggested, but I believe it is still in line with what you wanted).

go/tools/fetch_repo/main.go
wlynch: Done.

go/tools/fetch_repo/main.go
wlynch: Done.

go/private/go_repository.bzl
```diff
@@ -27,12 +27,16 @@ def _go_repository_impl(ctx):
 
   # TODO(yugui): support submodule?
   # c.f. https://www.bazel.io/versions/master/docs/be/workspace.html#git_repository.init_submodules
-  remote = ctx.attr.remote if ctx.attr.remote else ctx.attr.importpath
+  remote = ctx.attr.remote
+  vcs = ctx.attr.vcs
+  importpath = ctx.attr.importpath
   result = ctx.execute([
```
jayconrod: It looks like you'll need to call fetch_repo differently depending on which attributes were passed in.

go/private/go_repository.bzl
```diff
@@ -27,12 +27,16 @@ def _go_repository_impl(ctx):
...
       fetch_repo,
       '--dest', ctx.path(''),
       '--remote', remote,
-      '--rev', rev])
+      '--rev', rev,
+      '--vcs', vcs,
+      '--importpath', importpath])
   if result.return_code:
     fail("failed to fetch %s: %s" % (remote, result.stderr))
```
jayconrod: Let's report ctx.name here instead of remote.

go/private/go_repository.bzl
wlynch: Done. The problem was the vcs value was always being set to "git" by default. Changed to default to "".

go/private/go_repository.bzl
wlynch: Done.

go/tools/fetch_repo/fetch_repo_test.go
wlynch: Done (previous commit).

PR #321, opened 2017-03-21
Allow setting custom Go version in go_repositories

### LINKED ISSUE #223: allow selecting go version in go_repositories()
It looks like `go_repositories()` roughly follows the latest released go version, without any option for projects to choose an explicit version.

Because there are frequently incompatible changes between go releases, we explicitly choose when to upgrade to a new version, and continue to build patches releases for old releases at the old go version.

As such, it'd be helpful if we could choose which version to use for builds, even if just the minor version (e.g. 1.6, 1.7, 1.8), which would track to the latest patch release (e.g. 1.7.4).

jayconrod:
go_repositories accepts three optional new arguments which allow
developers to use a custom version of Go.

* go_version: Go version string to use. Defaults to "1.8". Only "1.8"
  is supported now, but support will be added for new version in the
  future.
* go_linux, go_darwin: name of a repository containing the Go
  toolchain. This is useful for developers who want to fetch Go from
  an alternate source.

If go_linux or go_darwin is provided, go_version cannot also be
provided.

The omit_go argument is no longer supported.

TODO: manual (non-CI) tests.
TODO: update README.md.

Fixes #223

### DIFF
--- a/go/private/go_repositories.bzl
+++ b/go/private/go_repositories.bzl
@@ -148,11 +148,14 @@ def _go_repository_select_impl(ctx):
 
   # 1. Configure the goroot path
   if os_name == 'linux':
-    goroot = ctx.path(ctx.attr._linux).dirname
+    go_version = ctx.attr.go_linux_version
   elif os_name == 'mac os x':
-    goroot = ctx.path(ctx.attr._darwin).dirname
+    go_version = ctx.attr.go_darwin_version
   else:
     fail("Unsupported operating system: " + os_name)
+  if go_version == None:
+    fail("No Go toolchain provided for host operating system: " + os_name)
+  goroot = ctx.path(go_version).dirname
 
   # 2. Create the symlinks and write the BUILD file.
   gobin = goroot.get_child("bin")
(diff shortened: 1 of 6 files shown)


jayconrod:
@wlynch @pmbethe09 Does this interface make sense for specifying custom Go versions? Wanted to make sure before writing tests and updating documentation.

go/private/go_repositories.bzl
```diff
@@ -175,46 +178,68 @@ _go_repository_select = repository_rule(
...
-      build_file_content = "",
-      sha256 = "6fdc9f98b76a28655a8770a1fc8197acd8ef746dd4d8a60589ce19604ba2a120",
-      strip_prefix = "go",
+        name = "golang_darwin_amd64",
+        url = "https://storage.googleapis.com/golang/go%s.darwin-amd64.tar.gz" % go_version,
+        build_file_content = "",
+        sha256 = _GO_VERSIONS_SHA256[go_version]["darwin"],
+        strip_prefix = "go",
     )
+    go_linux = "@golang_linux_amd64"
```
wlynch: Looks like we're not respecting the user value of go_linux/go_darwin here.

jayconrod: This branch is only taken if go_linux, go_darwin are not set.

jayconrod:
This handles both the common case where a user wants to stay on a specific supported version of Go, and the customized case, where they pass in repo names.

If go_version is set, go_linux and go_darwin must not be set; internally, they will be set to the public repositories for that version.

If go_linux and go_darwin are set (either or both may be set),  go_version must not be set.

If none are set, we proceed as if go_version is "1.8" (will update to the latest as new versions are added).

jayconrod:
I've updated README.md and added a fairly elaborate test. PTAL.

jayconrod:
@wlynch @pmbethe09 Does this look okay in its current state?

pmbethe09:
LGTM

ixdy:
@jayconrod what is needed to support go1.7(.5) as well as go1.8?

jayconrod:
1.8 is the only supported version at the moment. It should be pretty easy to add back support for 1.7.5 though. There were some breaking changes in the `testing` package, and `//go/tools:generate_test_main` targets the 1.8 version of that package. We could detect Go 1.7.5 and generate the appropriate code though.

Issue #323, opened 2017-03-22
Any objection to add a ignore_go_package option in go_proto_library rule?


rosun82:
Background:

We have a bunch of protos put in a git repository, and for historical reasons they have go_package option specified. However, when switching to Bazel, the go_package option gets in the way, as the output .pb.go file is in the wrong place and wrong package hence cannot be used as "deps" for other go_proto_library rules.

Proposal, add an ignore_go_package option, when specified, simply remove "option go_package" line from the proto source code then compile.

I can handle the modification if there is no objection.

pmbethe09:
Long-term: we should fix go_proto_library.  But that might take some effort.

So short-term I would be ok with this hack if you want to write it.

jayconrod:
@rosun82 Would this work for you: write a genrule that removes `option go_package` with `grep -v` or something similar, then use the output as srcs for `go_proto_library` instead of the original.

rosun82:
I have tried that. The trouble is there are proto imports. A genrule has to produce a file with different name, say we rename file a.proto to a_fix.proto through genrule. If b.proto imports a.proto, it will not build

rosun82:
@pmbethe09 Sorry to disturb, but how to contribute to the rules_go repository exactly?

I have checked CONTRIBUTING.md, and found it extremely confusing. The link goes to contributing to bazel itself, but I could not find the repository for go rules at all.

PR #325, opened 2017-03-23
Adding a ignore_go_package_option option to the go_proto_library.

rosun82:
Once the option is on, we will first strip the option go_package = ... line from
the .proto file then compile, which avoids the problem where the compiled .pb.go
file is in the directory unexpected by bazel.

This partially resolves https://github.com/bazelbuild/rules_go/issues/323.

### DIFF
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -130,7 +130,7 @@ def _go_proto_library_gen_impl(ctx):
   srcs = list(ctx.files.srcs)
   dirs = set([s.short_path[:-1-len(s.basename)]
               for s in srcs + protos])
-  cmds += ["/bin/mkdir -p %s/%s" % (work_dir, _drop_external(d)) for d in dirs if d]
+  cmds += ["mkdir -p %s/%s" % (work_dir, _drop_external(d)) for d in dirs if d]
 
   if ctx.attr.ignore_go_package_option:
     # Strip the "option go_package" line from the proto file before compiling,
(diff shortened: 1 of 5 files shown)


bazel-io:

proto/go_proto_library.bzl
```diff
@@ -131,8 +131,21 @@ def _go_proto_library_gen_impl(ctx):
...
+  if ctx.attr.ignore_go_package_option:
+    # Strip the "option go_package" line from the proto file before compiling,
+    # c.f., https://github.com/bazelbuild/rules_go/issues/323
+    #
+    # NOTE: Using sed does not provide a perfect solution, build will break if
+    # the go_package option splits multiple lines. Use with caution.
+    cmds += ["/bin/sed '/^ *option  *go_package/d' %s > %s/%s" % (s.path, work_dir,
+                                                                  _drop_external(s.short_path))
+             for s in srcs]
+    cmds += ["/bin/cp %s %s/%s" % (s.path, work_dir, _drop_external(s.short_path))
```
rosun82: We only apply the stripping to src files, so that common things like "google/protobuf/..." will not be broken.

proto/go_proto_library.bzl
```diff
@@ -131,8 +131,21 @@ def _go_proto_library_gen_impl(ctx):
...
-  cmds += ["/bin/cp %s %s/%s" % (s.path, work_dir, _drop_external(s.short_path))
-           for s in srcs + protos]
+
+  if ctx.attr.ignore_go_package_option:
+    # Strip the "option go_package" line from the proto file before compiling,
+    # c.f., https://github.com/bazelbuild/rules_go/issues/323
+    #
+    # NOTE: Using sed does not provide a perfect solution, build will break if
+    # the go_package option splits multiple lines. Use with caution.
+    cmds += ["/bin/sed '/^ *option  *go_package/d' %s > %s/%s" % (s.path, work_dir,
```
jayconrod: Formatting nit: move the tuple after `%` to the next line

jayconrod:
Jenkins test this please

proto/go_proto_library.bzl
rosun82: Done.

rosun82:
Added a small test case demonstrating that the fix works.

jayconrod:
Looks good. I'll submit when tests are finished running.

Jenkins test this please.

rosun82:
Thanks a lot for approving my first open source CL : P

rosun82:
Seems test fails due to timeout (not related to this change) PTAL

jayconrod:
Glad to have your contribution. :) Hopefully not the last.

On macOS, I think the test is failing because sed is at `/usr/bin/sed`. Maybe just use `sed`? Should still be visible on PATH.

I have to run to a meeting, will take a closer look right after.

rosun82:
Fixed by adding both /bin and /usr/bin into environment variable before compiling.

jayconrod:
I think that's fixed it.

Jenkins test this please.

Issue #329, opened 2017-03-23
How can I use a custom protobuf version in third_party?


ivyxyz:
The protobuf and go's protobuf are hard coded as `@com_github_google_protobuf//:protoc` and `@com_github_golang_protobuf//protoc-gen-go`. How do I specify to use my local copy of protobuf at `//third_party/protobuf:protoc`?

Is it solvable by switching to `//external:protoc` and `//external:protoc-gen-go` in [here](https://github.com/bazelbuild/rules_go/blob/9069f96267a51e9a393ff04e20cee0e1fb26b275/proto/go_proto_library.bzl#L182) and use bind instead?

jayconrod:
We don't support using `protoc` other than the one that's hard-coded in.

If you change the labels you pointed to in a fork of `rules_go` though, it should work.

pmbethe09:
2 ways to solve this:
1) As you wrote, use "//external:" and add binds.  That either requires every user to add binds to all WORKSPACE files, or some logic in calling go_proto_repositories(custom_bind=1)
2) Fewer changes in the general case, is what google/protobuf does
https://github.com/google/protobuf/blob/master/protobuf.bzl#L173

have an input to the macro which defaults to the usual thing -- so almost all of us who use that do nothing, but you can override that field in your usage of go_proto_library

ivyxyz:
Thanks, I also prefer option 2 for adding a `protoc = "@com_github_google_protobuf//:protoc"` in go_proto_library and pass it to _go_proto_library_gen. Could you help to make the change because it will take me some effort to sign the cla?

PR #336, opened 2017-03-28
go_proto_library: Add ability to override protoc and protoc_gen_go binaries

pmbethe09:
Fixes https://github.com/bazelbuild/rules_go/issues/329


jayconrod:
LGTM. Doesn't seem like the flake is due to this change.

Jenkins test this please.

PR #337, opened 2017-03-29
Fix infinite recursion

ianthehat:
On windows, you get something that matches none of the terminating conditions, but filepath.Dir(dir) returns dir unchanged.


bazel-io:

PR #338, opened 2017-03-29
Using filepath rather than path

ianthehat:


bazel-io:

PR #346, opened 2017-03-30
Change from executable to command

ianthehat:
It's  a shell script, which is not executable on windows.


bazel-io:

jayconrod:
LGTM. I don't have access to a Windows machine and can't verify myself though.

I'm surprised this doesn't work in Windows. Since these files are created with `executable = True`, I expected Bazel would make this work.

Jenkins test this please.

ianthehat:
I think that means bazel can ask the host to run it directly, rather than
asking the shell to run it, which works on Linux because it understand the
shebang but Windows just thinks it's a text file...

PR #347, opened 2017-03-30
Filter go_test sources before generating test main file

### LINKED ISSUE #340: go_tests with build constraints fail to build
When generate_test_main.go generates test glue code, it parses test source files in order to discover which functions should be called. The Go rules pass in the list of sources to parse. Since #311 and #314, rules are expected to list all source files, including those build constraints (build tags and filename suffixes). Currently, the rules pass the unfiltered list of sources to generate_test_main, but they build with the filtered list, which means the generated main file may attempt to call functions that weren't compiled.

jayconrod:
generate_test_main now receives a list of source files filtered
according to build constraints (build tags and suffixes) using
filter_tags. This prevents generate_test_main from emitting calls to
functions that won't be compiled or linked into the test binary.

Fixes #340.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -581,19 +581,36 @@ def go_test_impl(ctx):
   lib_result = go_library_impl(ctx)
   main_go = ctx.outputs.main_go
   prefix = _go_prefix(ctx)
-
   go_import = _go_importpath(ctx)
 
-  args = (["--package", go_import, "--output", ctx.outputs.main_go.path] +
-          [i.path for i in lib_result.go_sources])
-
-  inputs = list(lib_result.go_sources) + list(ctx.files.toolchain)
+  cmds = [
+      'UNFILTERED_TEST_FILES=(%s)' %
+          ' '.join(["'%s'" % f.path for f in lib_result.go_sources]),
+      'FILTERED_TEST_FILES=()',
+      'while read -r line; do',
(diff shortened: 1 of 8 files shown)


jayconrod:
@ianthehat Would you mind reviewing this fix?

Jenkins is failing due to #349, not this PR.

go/def.bzl
```diff
@@ -581,19 +581,36 @@ def go_test_impl(ctx):
...
   main_go = ctx.outputs.main_go
   prefix = _go_prefix(ctx)
-
   go_import = _go_importpath(ctx)
 
-  args = (["--package", go_import, "--output", ctx.outputs.main_go.path] +
-          [i.path for i in lib_result.go_sources])
-
-  inputs = list(lib_result.go_sources) + list(ctx.files.toolchain)
+  cmds = [
```
ianthehat: I think this would be easier to read if it was a single multi line string using .format with named args, but it's not worth changing, just making a comment in passing....

jayconrod: I agree that would be mostly simpler, though it might be more opaque for quoted lists of items like `UNFILTERED_TEST_FILES`.

In the not-too-distant future, I'd like to rewrite all of this in Go, since the Bash code is difficult to work with (hard to read, hard to test, error-prone). For now, I'll leave it as is for consistency with other blocks of commands formatted the same way.

PR #352, opened 2017-04-03
Adds support for Go 1.7.5

xtrqt:
This PR fixes issues with not backward compatible unit test generator. This should unblock support for Go1.7.5. See #321

### DIFF
--- a/go/tools/generate_test_main.go
+++ b/go/tools/generate_test_main.go
@@ -20,11 +20,11 @@ package main
 import (
 	"flag"
 	"go/ast"
-	"go/build"
 	"go/parser"
 	"go/token"
 	"log"
 	"os"
+	"runtime"
 	"strings"
 	"text/template"
 )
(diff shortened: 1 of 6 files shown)


bazel-io:

xtrqt:
I signed it!

jayconrod:
Thanks for sending this! It's definitely something I wanted to restore support for, since this has been blocking some teams that haven't migrated to 1.8 yet.

I have a couple small comments, and I'd like to add a test to ensure this continues working in the future. It's probably easiest if I push those changes directly to your branch though.

jayconrod:
Jenkins test this please

go/tools/generate_test_main.go
```diff
@@ -82,33 +79,24 @@ func main() {
 		}
 	}
 
-{{ if .Version.go1_8 }}
+{{if eq .Version "go1.8"}}
```
twpayne: Won't this break when Go 1.9 is released? Should this be `if ge .Version "go1.8"` instead?

jayconrod: Yes, I could have done this better. A lot of things in the `testing` package are excluded from Go 1 compatibility guidelines, so I expect to have to fix this tool after every Go release. However, it shouldn't break automatically if the upstream API *doesn't* change.

PR #357, opened 2017-04-04
Gazelle: remove support for graphical multidiff

### LINKED ISSUE #342: Gazelle should use buildifier differ correctly or not at all
This issue is based on a performance issue mentioned in #335.

When Gazelle is invoked with `-mode diff`, it runs a command provided by github.com/bazelbuild/buildifier/differ. The differ package is supposed to provide different commands and different ways of invoking them, depending on whether a graphical diff tool is available and whether multi-diff is supported.

Gazelle incorrectly assumes multi-diff is not supported. It calls `Differ.Show` for each BUILD file without ever calling `Differ.Run`. This causes it to build a command line without actually running it.

I'm not actually sure a graphical diff is expected behavior for Gazelle. If it is not, we should just invoke `diff` directly without using the `differ` package. If it is, then we should call the `Run` method.

jayconrod:
buildifier/differ supports both graphical and non-graphical
diff. The graphical diff doesn't work though: it generates a
command line for tkdiff, but tkdiff doesn't actually accept commands
in that format.

With this change, we just print diffs in the simple unified format by
executing `diff -u`.

Fixes #342

### DIFF
--- a/go/tools/gazelle/gazelle/diff.go
+++ b/go/tools/gazelle/gazelle/diff.go
@@ -18,9 +18,9 @@ package main
 import (
 	"io/ioutil"
 	"os"
+	"os/exec"
 
 	bzl "github.com/bazelbuild/buildifier/build"
-	"github.com/bazelbuild/buildifier/differ"
 )
 
 func diffFile(file *bzl.File) error {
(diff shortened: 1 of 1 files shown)


pmbethe09:
LGTM

PR #360, opened 2017-04-06
Simplify symlink handling

jayconrod:
* Introduced _relative_path_from_dir which calculates the relative path
  from a directory to a file.
* Rewrote _symlink_tree_commands (now private) using this new function.
* Removed _remove_external_prefix and _is_external. They were no-ops
  since Bazel no longer expresses external paths as "../repo".
* Simplified path munging in compile and link actions and fixed
  quoting in a few places.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -59,10 +59,29 @@ def _go_prefix(ctx):
     prefix = prefix + "/"
   return prefix
 
-# TODO(bazel-team): it would be nice if Bazel had this built-in.
-def symlink_tree_commands(dest_dir, artifact_dict):
-  """Symlink_tree_commands returns a list of commands to create the
-  dest_dir, and populate it according to the given dict.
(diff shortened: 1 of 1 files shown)


jayconrod:
Tests are failing because of #361. I'll fix that first.

jayconrod:
Needs more work. `_is_external` and `_remove_external_prefix` are needed to support bazelbuild/bazel#1262.

pmbethe09:
has this been replaced by https://github.com/bazelbuild/rules_go/pull/375 ?

jayconrod:
This has been partially replaced by #375. That PR was originally based on this one, but I split out the relevant part so we could move forward.

There is still some refactoring I'd like to do, but it's probably better to wait for bazelbuild/bazel#1262 to be fixed so we can test against that.

I'll abandon this PR for now.

PR #362, opened 2017-04-06
Fetch buildifier from new location

jayconrod:
The github.com/bazelbuild/buildifier repository was renamed to
buildtools. With this change, we download from the new location but
keep the old import path.

Fixes #361

PR #367, opened 2017-04-07
Cherrypick 0.4.2

jayconrod:
Cherrypicked the following changes:

* **#319 go_repository: add importpath and vcs flags**. This allows us to have separate import paths and remote URLs for repositories.
* **#362 Fetch buildifier from new location**. This allows the buildifier zip to be extracted with the new name.

Note that this PR is onto a new branch, release/0.4.2, which currently points to 0.4.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,12 +148,15 @@ placed in the WORKSPACE.
 ## go\_repository
 
 ```bzl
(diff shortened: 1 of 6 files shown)


pmbethe09:
LGTM

PR #375, opened 2017-04-10
Support build constraints in C sources in cgo_library

### LINKED ISSUE #330: Build tag filtering for `cgo` rules
I think this is the issue I was actually seeing when I created #328. `cgo_library` rules appear to not properly filter based on build tags, and will try to run over `.go` files that shouldn't be built.

https://github.com/nelhage-stripe/bug_rules_go/tree/master/cgo_tags builds for me under `go build` and not `rules_go`, on `darwin_amd64`.

jayconrod:
C and C++ sources are now filtered using build constraints (tags and
suffixes) in cgo_library. This matches behavior in "go build".

Fixes #330

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -50,6 +50,22 @@ hdr_exts = [
 
 cc_hdr_filetype = FileType(hdr_exts)
 
+# Extensions of files we can build with the Go compiler or with cc_library.
+# This is a subset of the extensions recognized by go/build.
+cgo_filetype = FileType([
+    ".go",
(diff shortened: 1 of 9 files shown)


jayconrod:
@kchodorow Mind taking a look at `_relative_path_from_dir` in this PR? I think it will be compatible with the changes in https://github.com/bazelbuild/bazel/issues/1262, but I'm not sure how to test against that.

pmbethe09:
LGTM

PR #377, opened 2017-04-11
Update README.md in preparation for 0.4.3

jayconrod:
Also fix whitespace issue in table of contents.


pmbethe09:
LGTM.

side-bar: do we still consider these experimental?  or maybe Alpha now since we think the parts that work should work?

jayconrod:
Alpha sounds right. I've updated README.md. Also added cross compilation to the list of features not supported yet.

PR #378, opened 2017-04-12
Add wtool README.md and a link from Vendoring.md

pmbethe09:


jayconrod:
LGTM.

In the long term, I think gazelle should take over some of wtool's responsibilities (I'm drafting a design doc, will send soon). We definitely still need wtool now though, and it's good to have it better documented.

PR #381, opened 2017-04-15
Add bazel helper library

achew22:
After the discussion in bazelbuild/rules_webtesting#108 I figured I'd put together a PR to see what people think.

@DrMarcII, what do you think about this? Are you okay with me moving it? Are you okay with me making it //visibility:public?

@jayconrod what do you think about putting it here?

### DIFF
--- a/go/tools/bazel/bazel.go
+++ b/go/tools/bazel/bazel.go
@@ -22,6 +22,12 @@ import (
 	"path/filepath"
 )
 
+const TEST_SRCDIR = "TEST_SRCDIR"
+const TEST_TMPDIR = "TEST_TMPDIR"
+const TEST_WORKSPACE = "TEST_WORKSPACE"
+
+var defaultTestWorkspace = ""
+
 // Runfile returns an absolute path to the specified file in the runfiles directory of the running target.
 // It searches the current working directory, RunfilesPath() directory, and RunfilesPath()/TestWorkspace().
 // Returns an error if unable to locate RunfilesPath() or if the file does not exist.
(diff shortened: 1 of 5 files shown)


go/tools/bazel/bazel.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2017 Google Inc.
```
jayconrod: I think this needs to be "The Bazel Authors" instead of Google.

go/tools/bazel/bazel_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Errorf("Error finding runfiles path: %s", err)
+	}
+
+	if path == "" {
+		t.Errorf("Runfiles path is empty: %s", path)
+	}
+}
+
+func TestNewTmpDir(t *testing.T) {
+	//prefix := "new/temp/dir"
```
jayconrod: Can this line be removed?

go/tools/bazel/bazel_test.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2017 Google Inc.
```
jayconrod: "The Bazel Authors"

DrMarcII:
I would prefer you kept the ability to set the default workspace (maybe return an error from TestWorkspace function if default not set and TEST_WORKSPACE is not defined). When running tests directly from bazel-bin the TEST_WORKSPACE variable doesn't get set.

achew22:
@DrMarcII I have updated the API to support that functionality. PTAL

DrMarcII:
LGTM. Thanks.

achew22:
@DrMarcII any time. Thanks for doing the hard work of actually writing the code :wink: 

@jayconrod, I'm getting CI failures that I don't think are related to this. What would you like to do?

Also, if/when we merge this would you like me to squash it or do you use the GitHub UI to squash?

jayconrod:
Agree the failures aren't due to this change. Travis CI doesn't seem to be running the install script for Bazel on some configurations.

I'll go ahead and submit this now.

PR #383, opened 2017-04-17
In new_go_repository, generate BUILD.bazel files by default

### LINKED ISSUE #144: new_go_repository does not work with golang.org/x/text on macOS
Since golang.org/x/text/collate/build is a package, and macOS by default has a case-insensitive (but case-respecting) filesystem, I cannot use new_go_repository with that package.

Before new_go_repository existed, I was overlaying my own BUILD files using new_git_repository, which did not have this issue because I used on per repository, instead of one per package.

Apparently it is in some cases possible to convert your filesystem losslessly to case-sensitive using iPartition, but that doesn't work if you're encrypting your hard drive.

Can the new_go_repository rules be changed to use this same approach? Or is there a technical reason why that wouldn't work?

This came up because I'm actually trying to depend on an entirely different package, afero, using the following WORKSPACE

``` python
git_repository(
    name = "io_bazel_rules_go",
    remote = "https://github.com/bazelbuild/rules_go.git",
    tag = "0.2.0",
)
load("@io_bazel_rules_go//go:def.bzl", "go_repositories", "new_go_repository")
go_repositories()

new_go_repository(
  name = "org_golang_x_text",
  importpath = "github.com/golang/text",
  commit = "d69c40b4be55797923cec7457fac7a244d91a9b6",
)

new_go_repository(
  name = "com_github_pkg_sftp",
  importpath = "github.com/pkg/sftp",
  commit = "a71e8f580e3b622ebff585309160b1cc549ef4d2",
)

new_go_repository(
  name = "com_github_spf13_afero",
  importpath = "github.com/spf13/afero",
  commit = "52e4a6cfac46163658bd4f123c49b6ee7dc75f78",
)
```

The actual error is:

```
$ bazel build //...:all
ERROR: /private/var/tmp/_bazel_swsnider/d980a28a08d1d45d19e13e085f723430/external/com_github_spf13_afero/BUILD:5:1: no such package '@org_golang_x_text//unicode/norm': failed to generate BUILD files for github.com/golang/text: 2016/10/12 23:42:24 open /private/var/tmp/_bazel_swsnider/d980a28a08d1d45d19e13e085f723430/external/org_golang_x_text/collate/BUILD: is a directory
 and referenced by '@com_github_spf13_afero//:go_default_library'.
ERROR: Analysis of target '//compiler/cmd/pp:pp' failed; build aborted.
INFO: Elapsed time: 33.302s
```

jayconrod:
Generated BUILD files may conflict with other files or directories
named "build" on case insensitive file systems. "BUILD.bazel" is less
likely to conflict, and Bazel supports these files since 0.4.1.

The filename can still be overridden by specifying the build_file_name
attribute. This just changes the default.

Fixes #144

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -60,7 +60,7 @@ def _new_go_repository_impl(ctx):
 
 
 _go_repository_attrs = {
-    "build_file_name": attr.string(),
+    "build_file_name": attr.string(default="BUILD.bazel"),
     "importpath": attr.string(),
     "remote": attr.string(),
     "vcs": attr.string(default="", values=["", "git", "hg", "svn", "bzr"]),
--- a/tests/new_go_repository_build_name/local/BUILD
+++ b/tests/new_go_repository_build_name/local/BUILD
@@ -1,0 +1,10 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_prefix", "go_test")
+
+go_prefix("local")
+
+go_test(
+    name = "go_default_test",
+    srcs = ["local_test.go"],
+    tags = ["manual"],
+    deps = ["@remote//build:go_default_library"],
+)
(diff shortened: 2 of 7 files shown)


pmbethe09:
LGTM

jmhodges:
Thanks!

Issue #339, opened 2017-03-29
Gazelle generates partial BUILD files when multiple packages are present


jayconrod:
When generating a BUILD file for [golang.org/x/tools/imports](https://github.com/golang/tools/tree/master/imports), Gazelle generates a rule to build `go_default_library` but does not include any dependencies.

This happens because of the presence of mkindex.go. This source file is in the `main` package, but it is excluded with `// +build ignore`. Since #314, Gazelle reads all source files regardless of build tag. `go/build` reports a `MultiplePackageError` when processing this directory, which Gazelle silently ignores. When `go/build` returns this error, it does not include any imports in the returned package. Consequently, Gazelle behaves as if there are no dependencies.

PR #386, opened 2017-04-17
Ignore extra packages in Gazelle

jayconrod:
If Gazelle finds .go files in a directory with multiple distinct
package declarations, it will now only generate rules to build .go
files where the package declaration matches the directory name. If
there are multiple packages, and none match the directory name,
Gazelle will report a build.MultiplePackageError, as bove. If is a
single package that doesn't match the directory name, Gazelle will
still generate rules for it.

This change helps address #339. When we are no longer applying build
constraints by default, multiple packages are much more common; many
files are hidden with "// +build ignore".

### DIFF
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -93,7 +93,7 @@ func findPackage(bctx build.Context, dir string) (*build.Package, error) {
 func findPackageFiles(dir string) (packageGoFiles map[string][]os.FileInfo, otherFiles []os.FileInfo, err error) {
 	files, err := ioutil.ReadDir(dir)
 	if err != nil {
-		return
+		return nil, nil, err
 	}
 
 	packageGoFiles = make(map[string][]os.FileInfo)
@@ -128,7 +128,7 @@ func findPackageFiles(dir string) (packageGoFiles map[string][]os.FileInfo, othe
 		}
 		packageGoFiles[packageName] = append(packageGoFiles[packageName], file)
 	}
-	return
+	return packageGoFiles, otherFiles, nil
 }
 
 func defaultPackageName(dir string) string {
@@ -143,26 +143,29 @@ func defaultPackageName(dir string) string {
 func selectPackageName(packageGoFiles map[string][]os.FileInfo, dir string) (string, error) {
 	if len(packageGoFiles) == 0 {
 		return "", &build.NoGoError{Dir: dir}
-	} else if len(packageGoFiles) == 1 {
+	}
+
+	if len(packageGoFiles) == 1 {
 		var packageName string
 		for name, _ := range packageGoFiles {
 			packageName = name
 		}
 		return packageName, nil
-	} else {
-		defaultName := defaultPackageName(dir)
-		if _, ok := packageGoFiles[defaultName]; ok {
-			return defaultName, nil
-		}
-
-		err := &build.MultiplePackageError{Dir: dir}
-		for name, files := range packageGoFiles {
-			// Add the first file for each package for the error message.
-			// Error() method expects these lists to be the same length. File
-			// lists are guaranteed to be non-empty.
-			err.Packages = append(err.Packages, name)
-			err.Files = append(err.Files, files[0].Name())
-		}
-		return "", err
 	}
+
+	defaultName := defaultPackageName(dir)
+	if _, ok := packageGoFiles[defaultName]; ok {
+		return defaultName, nil
+	}
+
+	err := &build.MultiplePackageError{Dir: dir}
+	for name, files := range packageGoFiles {
+		// Add the first file for each package for the error message.
+		// Error() method expects these lists to be the same length. File
+		// lists must be non-empty. These lists are only created by
+		// findPackageFiles for packages with .go files present.
+		err.Packages = append(err.Packages, name)
+		err.Files = append(err.Files, files[0].Name())
+	}
+	return "", err
 }
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -17,19 +17,33 @@ package packages
 
 import (
 	"go/build"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"log"
 	"os"
+	"path"
 	"path/filepath"
+	"sort"
+	"strings"
 )
 
 // A WalkFunc is a callback called by Walk for each package.
 type WalkFunc func(pkg *build.Package) error
 
-// Walk walks through Go packages under the given dir.
+// Walk walks through directories under "root".
 // It calls back "f" for each package.
 //
 // It is similar to "golang.org/x/tools/go/buildutil".ForEachPackage, but
 // it does not assume the standard Go tree because Bazel rules_go uses
 // go_prefix instead of the standard tree.
+//
+// If a directory contains no buildable Go code, "f" is not called. If a
+// directory contains one package with any name, "f" will be called with that
+// package. If a directory contains multiple packages and one of the package
+// names matches the directory name, "f" will be called on that package and the
+// other packages will be silently ignored. If none of the package names match
+// the directory name, a *build.MultiplePackageError error is returned.
 func Walk(bctx build.Context, root string, f WalkFunc) error {
 	return filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,94 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+				log.Printf("%s: error parsing package clause: %v", filename, err)
+				continue
+			}
+
+			packageName = ast.Name.Name
+			if strings.HasSuffix(packageName, "_test") {
+				packageName = packageName[:len(packageName)-len("_test")]
+			}
+			packageGoFiles[packageName] = append(packageGoFiles[packageName], file)
+		} else {
```
pmbethe09: invert this and continue
if !isGo {
  otherFiles = append(otherFiles, file)
  continue
}

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,94 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+		if file.IsDir() {
+			continue
+		}
+
+		name := file.Name()
+		filename := filepath.Join(dir, name)
+		ext := path.Ext(name)
+		isGo := ext == ".go"
+
+		var packageName string
```
pmbethe09: remove and just use := later

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,94 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+	}
+	return
+}
+
+func defaultPackageName(dir string) string {
+	pname := filepath.Base(dir)
+	if pname == "." || pname == "/" {
+		// We'll only use this name at the root of the filesystem.
+		return "unnamed"
+	} else {
```
pmbethe09: drop else wrapper

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,94 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+		return pname
+	}
+}
+
+func selectPackageName(packageNames []string, dir string) (string, error) {
+	if len(packageNames) == 0 {
+		return "", &build.NoGoError{Dir: dir}
+	} else if len(packageNames) == 1 {
+		return packageNames[0], nil
+	} else {
```
pmbethe09: drop else wrapper

go/tools/gazelle/packages/walk_test.go
```diff
@@ -31,78 +31,162 @@ func tempDir() (string, error) {
...
-			t.Errorf("pkg.Name = %q; want %q", got, want)
+func createFiles(files []fileSpec) (dir string, err error) {
+	dir, err = tempDir()
+	if err != nil {
+		return
+	}
+	for _, f := range files {
+		path := filepath.Join(dir, f.path)
+		if strings.HasSuffix(f.path, "/") {
+			err = os.MkdirAll(path, 0700)
```
pmbethe09: push into if, use :=

go/tools/gazelle/packages/walk_test.go
```diff
@@ -31,78 +31,162 @@ func tempDir() (string, error) {
...
+			if err != nil {
+				return
+			}
```
pmbethe09: add continue, drop else wrapper

go/tools/gazelle/packages/walk_test.go
```diff
@@ -31,78 +31,162 @@ func tempDir() (string, error) {
...
+		} else {
+			err = os.MkdirAll(filepath.Dir(path), 0700)
```
pmbethe09: push into if, use :=

go/tools/gazelle/packages/walk_test.go
```diff
@@ -31,78 +31,162 @@ func tempDir() (string, error) {
...
+			if err != nil {
+				return
+			}
+			err = ioutil.WriteFile(path, []byte(f.content), 0600)
```
pmbethe09: push into if, use :=

go/tools/gazelle/packages/walk_test.go
```diff
@@ -31,78 +31,162 @@ func tempDir() (string, error) {
...
+		path := filepath.Join(dir, f.path)
+		if strings.HasSuffix(f.path, "/") {
+			err = os.MkdirAll(path, 0700)
+			if err != nil {
+				return
+			}
+		} else {
+			err = os.MkdirAll(filepath.Dir(path), 0700)
+			if err != nil {
+				return
```
pmbethe09: log this error? (and in other places)

go/tools/gazelle/packages/walk_test.go
```diff
@@ -31,78 +31,162 @@ func tempDir() (string, error) {
...
+			}
+			err = ioutil.WriteFile(path, []byte(f.content), 0600)
+			if err != nil {
+				return
+			}
 		}
-		n++
+	}
+	return
```
pmbethe09: not needed?

jayconrod:
@pmbethe09 Thanks for reviewing. I think I've addressed all your comments, but I can't see them inline in the code after force-pushing... sorry about that. I'll just add commits in the future.

The Travis tests were failing due to me incorrectly recognizing the "documentation" package, which go/build ignores as a special case. I've added a similar special case, and I've fixed the way I was constructing `build.MultiplePackageError`. This required changing the signature of `selectPackageName`.

PTAL.

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,103 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+func selectPackageName(packageGoFiles map[string][]os.FileInfo, dir string) (string, error) {
+	if len(packageGoFiles) == 0 {
+		return "", &build.NoGoError{Dir: dir}
+	} else if len(packageGoFiles) == 1 {
+		var packageName string
+		for name, _ := range packageGoFiles {
+			packageName = name
+		}
+		return packageName, nil
+	} else {
```
pmbethe09: remove 'else {' wrapper.

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,103 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+			return defaultName, nil
+		}
+
+		err := &build.MultiplePackageError{Dir: dir}
+		for name, files := range packageGoFiles {
+			// Add the first file for each package for the error message.
+			// Error() method expects these lists to be the same length. File
+			// lists are guaranteed to be non-empty.
+			err.Packages = append(err.Packages, name)
+			err.Files = append(err.Files, files[0].Name())
```
pmbethe09: are you sure files[0] exists?

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,103 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+	bctx.ReadDir = func(dir string) ([]os.FileInfo, error) {
+		return files, nil
+	}
+	return bctx.ImportDir(dir, build.ImportComment)
+}
+
+func findPackageFiles(dir string) (packageGoFiles map[string][]os.FileInfo, otherFiles []os.FileInfo, err error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return
```
pmbethe09: reads cleaner as
return nil, nil, err

go/tools/gazelle/packages/walk.go
```diff
@@ -52,3 +66,103 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
...
+		if packageName == "documentation" {
+			// go/build ignores this package.
+			continue
+		}
+		if strings.HasSuffix(packageName, "_test") {
+			packageName = packageName[:len(packageName)-len("_test")]
+		}
+		packageGoFiles[packageName] = append(packageGoFiles[packageName], file)
+	}
+	return
```
pmbethe09: easier to read if this is return packageGoFiles, otherFiles, nil

go/tools/gazelle/packages/walk.go
jayconrod: Done

go/tools/gazelle/packages/walk.go
jayconrod: Done

go/tools/gazelle/packages/walk.go
jayconrod: Done. Also dropped the else above.

go/tools/gazelle/packages/walk.go
jayconrod: I've updated the comment a bit to explain. `findPackageFiles` will only create an entry in `packageGoFiles` when it finds a .go file. So the list should never be empty.

PR #387, opened 2017-04-18
Add external attribute for new_go_repository

ains:
Add attribute "external" to new_go_repository, so that you can specify if a library that you depend on uses vendored dependencies.

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -47,7 +47,8 @@ def _new_go_repository_impl(ctx):
 
   cmds = [gazelle, '--go_prefix', ctx.attr.importpath, '--mode', 'fix',
           '--repo_root', ctx.path(''),
-          "--build_tags", ",".join(ctx.attr.build_tags)]
+          "--build_tags", ",".join(ctx.attr.build_tags),
+          "--external", ctx.attr.external]
   if ctx.attr.build_file_name:
       cmds += ["--build_file_name", ctx.attr.build_file_name]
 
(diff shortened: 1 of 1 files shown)


bazel-io:

jayconrod:
Hey @ains, thanks for submitting this. I'm hesitant to merge at this time though. My concern is that dependencies will be hidden outside of the local workspace. It would be too easy to end up with two different copies of the same library at different versions. Ideally, all dependencies should be either external (declared in WORKSPACE) or flattened in the vendor directory.

We're working on a number of improvements to gazelle to make it easier to deal with external dependencies. I'm planning to add some functionality to add missing dependencies recursively, which should make WORKSPACE more ergonomic.

One confounding factor is that Bazel has some [tentative plans](https://bazel.build/designs/2016/09/19/recursive-ws-parsing.html) to un-flatten the WORKSPACE file. If that happens in the future, we may want external vendoring to behave the same way (as in this PR).

ains:
Hey @jayconrod - thanks for the review.

So if I understand correctly there's no direct support for external dependencies which vendor their own libraries? That is they are forced to use the version that is declared in the WORKSPACE?

I've got a workaround for this at the moment, but with how vendoring is in current versions of Go it's totally possible to have a version of a library used in your code that is different from the one used in an external dependency which uses it's own vendor directory.

jayconrod:
Right, there's currently no direct support for this through `new_go_repository`. It is still possible with `new_git_repository` and a custom BUILD file, or by forking the git repository and adding your own BUILD files.

I know it's possible with Go to have multiple versions of the same package, but it's kind of an anti-pattern. I don't think the tools should encourage this.

jayconrod:
Closing for now. If Bazel implements hierarchical WORKSPACE files, we should revisit this though.

PR #391, opened 2017-04-18
Consolidate Bazel flags in a single place for tests

jayconrod:
* Test flags are now stored in .test-bazelrc. This is referenced by
  .travis.yml and by all non-Bazel test scripts.
* Common functionality for non-Bazel test scripts is now in
  non_bazel_tests_common.bash. This includes Bash functions that help
  run Bazel with appropriate options. All test scripts source this.
* Scripts that create new workspaces now run in batch mode. This
  prevents Bazel servers from hanging around in memory.


pmbethe09:
LGTM

PR #392, opened 2017-04-19
go_binary and go_test no longer provide a library to other rules

### LINKED ISSUE #379: go_binary and go_test should not expose .a files in outputs
`go_binary` and `go_test` are currently implemented as `go_library` with an additional output: the executable file. This means they also have the .a file output from `go_library`.

This confuses rules that depend on `go_binary` and `go_test` outputs. For example, a packaging rule or a size test would need to filter out the .a files, either in the rule implementation or using an intermediate `genrule`.

Binary rules for other languages don't seem to share this behavior. For simplicity, we should only output the executable. If anyone really needs the .a file, they can define a separate `go_library` rule with all the sources, then link that to `go_binary` or `go_test` through the `library` attribute.

jayconrod:
Previously, go_binary and go_test provided both an executable and the
library built from srcs as outputs to other rules. This was
inconsistent with other binary rules (cc_binary, etc.) and confusing
for rules that expected a single output. With this change, only the
executable file is exposed through the files provider.

Fixes #379

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -604,7 +604,7 @@ def go_binary_impl(ctx):
     lib=lib_out, executable=executable,
     gc_linkopts=_gc_linkopts(ctx))
 
-  return struct(files = set([executable]) + lib_result.files,
+  return struct(files = set([executable]),
                 runfiles = lib_result.runfiles,
                 cgo_object = lib_result.cgo_object)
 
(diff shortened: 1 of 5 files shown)


jayconrod:
We could. It doesn't seem like we need to though. It doesn't seem like `outputs` is visible to dependent rules. Bazel checks that we define actions for all outputs. It's used to populate the `files` provider if that isn't set explicitly. Other tests will see whatever's in `files` though.

It does seem a little messy. There's some benefit in being able to call `go_library_impl` from `go_binary_impl` and `go_test_impl`. The former creates an action to write `ctx.outputs.lib` which is used by all three rules.

pmbethe09:
LGTM

PR #396, opened 2017-04-20
Gazelle no longer applies build constraints by default

### LINKED ISSUE #339: Gazelle generates partial BUILD files when multiple packages are present

jayconrod:
The Go rules now apply build constraints to files in srcs of
go_library, cgo_library, go_test, and go_binary, so it's not usually
necessary to perform this filtering in Gazelle as well. Gazelle will
not apply build constraints (GOOS and GOARCH) by default anymore.
Generated BUILD files will be the same for all platforms.

Build tags can still be specified on the command flag, with -build_tags.
This can still be done with new_go_repository.

Fixes #339

### DIFF
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -36,7 +36,7 @@ import (
 
 var (
 	buildFileName  = flag.String("build_file_name", "BUILD", "name of output build files to generate.")
-	buildTags      = flag.String("build_tags", "", "comma-separated list of build tags. If not specified, GOOS and GOARCH are used.")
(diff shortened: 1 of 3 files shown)


pmbethe09:
LGTM.  Did tests not fire?

PR #407, opened 2017-04-24
Gazelle: use go_prefix to determine default package name in root directory

### LINKED ISSUE #406: gazelle package detection should use the import path basename, not the filesystem
#386 made gazelle disambiguate multiple packages in a directory by picking the one that matches the dirname on the filesystem.

This doesn't suffice for top-level packages in a repository, since they appear on disk using their workspace name, which will generally not match the package name. Using the last component of the import path would be more robust in this case.

You can see this problem with `golang.org/x/text`, which has a `// +build ignore` file at top-level. I've added a test repository here: https://github.com/nelhage-stripe/bug_rules_go 

```
$ bazel build @org_golang_x_text//...
ERROR: failed to generate BUILD files for golang.org/x/text: 2017/04/24 13:57:53 found packages text (doc.go) and main (gen.go) in /home/nelhage/.cache/bazel/_bazel_nelhage/cfff2155b8c2607d1b621640bfa2d05b/external/org_golang_x_text
.
INFO: Elapsed time: 8.005s
```

`gazelle` knows it's generating for `golang.org/x/text`, so disambiguating based on import path rather than fs path would correctly pick `text` .

jayconrod:
Gazelle determines a default package name for each directory. Usually
this is the directory base name, but semantically, it should be the
last component of the import path.

With this change, the default package name for the root directory of
the repository is now the last component of the import path in
go_prefix. This accounts for repositories checked out under names that
differ from their import paths, which is always the case inside
new_go_repository.

Fixes #406

### DIFF
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -56,8 +56,7 @@ func Walk(bctx build.Context, repoRoot, goPrefix, dir string, f WalkFunc) error 
 			return filepath.SkipDir
 		}
 
-		defaultName := defaultPackageName(path, repoRoot, goPrefix)
-		pkg, err := findPackage(bctx, defaultName, path)
+		pkg, err := findPackage(bctx, defaultPackageName(path, repoRoot, goPrefix), path)
 		if err != nil {
 			if _, ok := err.(*build.NoGoError); ok {
 				return nil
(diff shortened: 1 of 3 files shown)


go/tools/gazelle/packages/walk.go
```diff
@@ -56,7 +56,8 @@ func Walk(bctx build.Context, root string, f WalkFunc) error {
 			return filepath.SkipDir
 		}
 
-		pkg, err := findPackage(bctx, path)
+		defaultName := defaultPackageName(path, repoRoot, goPrefix)
```
pmbethe09: inline this?

jayconrod: Done

PR #411, opened 2017-04-26
Don’t stutter the directory

ianthehat:
_emit_generate_params_action uses ctx.new_file, which already bakes in the directory, so the path from the exec root is duplicated, this removes that

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -313,7 +313,7 @@ def _emit_go_compile_action(ctx, sources, deps, out_lib,
     objs = ' '.join([c.path for c in extra_objects])
     cmds += [ctx.file.go_tool.path + " tool pack r " + out_lib.path + " " + objs]
 
-  f = _emit_generate_params_action(cmds, ctx, out_lib.path + ".GoCompileFile.params")
+  f = _emit_generate_params_action(cmds, ctx, out_lib.basename + ".GoCompileFile.params")
 
   ctx.action(
       inputs = [f] + inputs + extra_inputs + [ctx.executable._filter_tags],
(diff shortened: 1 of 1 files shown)


robotoer:
Unfortunately, this breaks the builds when using `go_library` targets that have slashes (`/`) in their name.

It seems that when using something like (see: https://github.com/bazelbuild/rules_go/blob/master/go/def.bzl#L778-L780):
```
go_library_outputs = {
    "lib": "%{name}.a",
}
```
in combination with:
```
  # Notice the out_lib.basename call here
  f = _emit_generate_params_action(cmds, ctx, out_lib.basename + ".GoCompileFile.params")
```

causes slashes in the name of targets to be treated as a part of the filepath, not the filename thus causing the first portion of the target name to be stripped. If there are two go_libraries defined as such:
```
go_library(
    name="my/target/name/with/slashes/mygopkg",
    ...
)

go_library(
    name="my/target/mygopkg",
    ...
)
```

This will cause the above code to generate a duplicate file named: `mygopkg.a.GoCompileFile.params` producing an error that looks like:
```
ERROR: file 'services/playdoh/vendor/internal.a.GoCompileFile.params' is generated by these conflicting actions:
Label: //services/playdoh/vendor:google.golang.org/grpc/internal, //services/playdoh/vendor:github.com/grpc-ecosystem/grpc-gateway/runtime/internal
RuleClass: go_library rule
Configuration: a0f7cfb2b43155be453b0ec199580fee
Mnemonic: FileWrite
Action key: 98711db68060cd961cfd7ce3a24d4961, 0d8fecdea63c9606b1503267ab3c3381
Progress message: Writing script services/playdoh/vendor/internal.a.GoCompileFile.params
PrimaryInput: (null)
PrimaryOutput: Artifact:[[/home/robert/.cache/bazel/_bazel_robert/19f329a8d4726a5ab6808b6e18fda04b/execroot/uno]bazel-out/local-fastbuild/bin]services/playdoh/vendor/internal.a.GoCompileFile.params
Primary outputs are different objects: 266495146, 2072729152
Owner information: //services/playdoh/vendor:google.golang.org/grpc/internal a0f7cfb2b43155be453b0ec199580fee (1330916379 2070122513), //services/playdoh/vendor:github.com/grpc-ecosystem/grpc-gateway/runtime/internal a0f7cfb2b43155be453b0ec199580fee (1731755235 2070122513)
MandatoryInputs: are equal
Outputs: are equal
.
ERROR: Analysis of target '//services/playdoh/vendor:google.golang.org/grpc/internal' failed; build aborted.
INFO: Elapsed time: 1.689s
```

robotoer:
Hmm, maybe I should just open a new issue

PR #412, opened 2017-04-26
removing cd from the scripts

ianthehat:
Now we don’t use source links, the libraries can be found by search path rather than by cd
This means we no longer need depth magic or output copying, and c deps can be linked directly rather than copied as well.


ianthehat:
This has been superseded by the pull request that kills the symlink trees

PR #416, opened 2017-04-27
Don’t build symlink trees in bash

ianthehat:
Only the compile step is changed so far.
It now uses separate bazel actions per file, and uses copy rather than link.
This means that you no longer need to manage directories, or wipe out the tree each time, instead each entry is updated when it needs to be.
It is also more portable.
Incremental building should be faster, and a clean build should not be much slower if at all, except in the presence of huge libraries that take a long time to copy.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -90,11 +90,13 @@ def _relative_path_from_dir(from_dir, to_file):
     up = from_dir.count('/') + 1
   return up * "../" + to_file
 
-def build_file_tree(ctx, dest_dir, artifact_dict):
-  """build_file_tree creates the specified file structure.
+def _build_file_tree(ctx, dest_dir, artifact_dict):
+  """_build_file_tree creates the specified file structure.
 
   Args:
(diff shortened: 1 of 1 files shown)


go/def.bzl
```diff
@@ -90,6 +90,27 @@ def _relative_path_from_dir(from_dir, to_file):
     up = from_dir.count('/') + 1
   return up * "../" + to_file
 
+def build_file_tree(ctx, dest_dir, artifact_dict):
```
jayconrod: Rename to _build_file_tree to make this private.

go/def.bzl
```diff
@@ -90,6 +90,27 @@ def _relative_path_from_dir(from_dir, to_file):
...
+  """build_file_tree creates the specified file structure.
+
+  Args:
+    dest_dir: The destination directory, a string.
```
jayconrod: May want to comment that `dest_dir` should be a basename, not a path from the workspace root.

Issue #389, opened 2017-04-18
Gazelle: add missing dependencies to WORKSPACE


jayconrod:
Gazelle should provide a mechanism to add `new_go_repository rules` to the WORKSPACE file in order to resolve import paths for missing external dependencies. This should be behind a flag (say, `-update_workspace`). When an external dependency is added, its dependencies should be added recursively.

When a new dependency is added, Gazelle should try to use a known version of that dependency, if possible. For example, suppose we want to add a dependency on "foo" because our dependency "bar" has "foo" in its vendor directory, which is managed with Godep. We should read Godeps.json to determine the preferred version of "foo".

When a dependency that was automatically added by Gazelle is no longer needed, we should remove it from WORKSPACE (again, behind a flag). We should avoid removing any rules manually added by the developer. If this can't be done safely, we should at least emit a warning.

chrislovecnm:
Crazy question.  Would this be better or would integrating with dep be better?  Just throwing out ideas :)

chrislovecnm:
Pretty please?? Sugar on top.  Automating this would be HUGE

jayconrod:
This is definitely still planned, but a lot of work is needed.

The rough design right now is to add four commands:

* add-deps: add new dependencies, optionally including transitive dependencies. Possibly automatic, based on import paths that can't be resolved.
* update-deps: update existing dependencies to newer versions. Can optionally add or delete dependencies which are newly needed or no longer needed.
* rm-deps: remove existing dependencies. Possibly automatic, based on import paths.
* sync-deps: transcribe dependencies from dep's lock file or some other equivalent. This could probably be done earlier.

chrislovecnm:
Just starting simple where it adds a new block for the user with the expected name would be super. Just have the user out in the git URL and the sha.

If we can get someone in that helps with typing and then make it smart later.

Also I think bazel has feature flags? Fairly new to this project.

jayconrod:
> Just starting simple where it adds a new block for the user with the expected name would be super. Just have the user out in the git URL and the sha.

There's actually a tool that does this, but I intend to delete it as soon as Gazelle is capable of replacing this. You can find it in go/tools/wtool (`go get -u github.com/bazelbuild/rules_go/go/tools/wtool`). Normally, you'll want to use it with the `-asis` flag. For example:

```
wtool -asis golang.org/x/sys
```

This would add this block to WORKSPACE (it figures out the sha1 automatically):

```
go_repository(
[...]

Issue #388, opened 2017-04-18
Gazelle: warn when an import path can't be resolved


jayconrod:
When an import path references a package not provided by any known dependency, Gazelle should print a warning. This should happen whether the package is local or non-local and whether Gazelle is running in external or vendored mode.

This will require some awareness of WORKSPACE and the vendor directory that Gazelle doesn't have yet.

jayconrod:
Migrated to bazelbuild/bazel-gazelle#9

PR #420, opened 2017-04-27
Declare repository dependencies in child WORKSPACE files

kchodorow:
Bazel is moving to a model where child repository WORKSPACE files will be
parsed (see https://github.com/bazelbuild/bazel/issues/1943). This means
that top-level WORKSPACE files won't have to declare every dependency
(yay!) but, as a part of this, child repositories will have to start
declaring the repositories they use.

This should be a no-op, though, since the top-level WORKSPACE file
overrides all of these repository definitions.  So it doesn't really
matter what they are (hence the dummy paths).


pmbethe09:
So for every new_go_repository, you have to know what repos it depends on?

kchodorow:
Yes. Don't you have to now, anyway?  They repository definitions just go in the main WORKSPACE file.

pmbethe09:
maintaining this 2-way list manually sounds hard

pmbethe09:
<shrug>LGTM</shrug>

kchodorow:
How are people doing this now? Just building, seeing what deps are missing, adding a new_go_repo, and rebuilding?

pmbethe09:
pretty much.  there is a bare-bones tool in this repo 'wtool', which I use to add a new_go_repository, then build, then call wtool again for whatever was missing, etc.

jayconrod:
We're thinking of giving Gazelle more capabilities for managing the the WORKSPACE file. See #388 and #389. Even if it doesn't end up writing new dependencies to WORKSPACE, it at least needs to be able to read them. If they're hidden in other WORKSPACE files, I don't know how we'll do that. Bazel doesn't really provide any capabilities for querying external repositories, as far as I know.

Personally, I'd much rather have a flat WORKSPACE file.

go/private/go_repository.bzl
```diff
@@ -57,6 +57,13 @@ def _new_go_repository_impl(ctx):
   if result.return_code:
     fail("failed to generate BUILD files for %s: %s" % (
         ctx.attr.importpath, result.stderr))
+  ctx.file('WORKSPACE', """
+# Automatically generated by //go/private:go_repository.bzl
+workspace(name = "{workspace_name}")
+{repo_deps}
+""".format(
+    workspace_name = ctx.name,
+    repo_deps="\n".join(["local_repository(name='%s', path='/dummy')" % r for r in ctx.attr.repo_deps])))
```
jayconrod: Using `local_repository` here doesn't make sense to me. Shouldn't Bazel report a conflict if you declare a `local_repository` in one place and, for example, a `git_repository` in another place? Seems like that should count as different versions.

Maybe there should be a special `use_repository` rule?

jayconrod:
When are these declarations going to be mandatory in Bazel? I think we're going to have to improve our tooling to support this.

kchodorow:
If this is going to cause you guys a lot of issues, I might be able to get away with not having per-repo declarations as long as the main repo declares them.  Let me think about it and see if I can implement that tomorrow...

go/private/go_repository.bzl
kchodorow: I could add a built-in "dummy" repository rule, if that works better for you guys.

kchodorow:
Seems like that will be do-able.

PR #426, opened 2017-05-02
Fix Bazel install script in Travis CI

jayconrod:
Bazel added an additional build artifact to Jenkins builds:
bazel--without-jdk-installer.sh. This confused our installer script,
which was grepping the artifact name out of the build page. Now there
are two matching artifacts instead of one.

With this change we always fetch bazel--installer.sh for builds with
Bazel HEAD.


pmbethe09:
LGTM

Issue #350, opened 2017-03-31
cgo_library doesn't build if source files are in different directories


jayconrod:
For each .go file provided to `cgo_library`, `go tool cgo` generates a `.cgo1.go` file and a `.cgo2.c`file (other files are generated, too). When determining the names for the generated files, cgo [translates slashes, backslashes, and colons](https://github.com/golang/go/blob/a8b2e4a630a5991e91095d85c604dc1fa23c1e56/src/cmd/cgo/out.go#L519) in the input files into underscores. Because directory separators are removed, all generated files end up in the same directory.

In order to avoid this name mangling, `_cgo_codegen_impl` [changes into the directory of the first source file](https://github.com/bazelbuild/rules_go/blob/1566d5f41a66b927524d31673dac4de228e2ebcd/go/def.bzl#L840) before invoking cgo. If source files are in different directories though, later source files won't be found.

PR #429, opened 2017-05-03
Kill symlink_tree_commands

ianthehat:
This rips out all the cd’s in the scripts, kills all the softlinks and removes all the code that tries to work out depts and wether a file is “external”
This result in far simpler code, much shorter bash scripts, and far far less work at build time, at the cost of larger command lines due to the package search paths being added.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -173,15 +173,17 @@ def _emit_go_compile_action(ctx, step, sources, deps, libpaths, out_lib,
 
   Args:
     ctx: The skylark Context.
+    step: used to generate unique names for this step of a rule.
     sources: an iterable of source code artifacts (or CTs? or labels?)
     deps: an iterable of dependencies. Each dependency d should have an
       artifact in d.transitive_go_libraries representing all imported libraries.
+    libpaths: the set of paths to search for imported libraries.
     out_lib: the artifact (configured target?) that should be produced
     extra_objects: an iterable of extra object files to be added to the
       output archive file.
     gc_goopts: additional flags to pass to the compiler.
   """
-  inputs = depset() + sources
+  inputs = depset(sources)
   prefix = _go_prefix(ctx)
 
   cmds = [
@@ -232,16 +234,15 @@ def _emit_go_compile_action(ctx, step, sources, deps, libpaths, out_lib,
   # (ctx.label.package + "/" ctx.label.name) for now.
   cmds += [' '.join(args)]
 
-  inputs += ctx.files.toolchain
   if extra_objects:
     inputs += extra_objects
     objs = ' '.join([c.path for c in extra_objects])
     cmds += [ctx.file.go_tool.path + " tool pack r " + out_lib.path + " " + objs]
 
   f = _emit_generate_params_action(cmds, ctx, ctx.label.name + "." + step + ".GoCompileFile.params")
-  inputs += [f]
+  inputs += [f, ctx.executable._filter_tags] + ctx.files.toolchain
   ctx.action(
-      inputs = list(inputs) + [ctx.executable._filter_tags],
+      inputs = list(inputs),
       outputs = [out_lib],
       mnemonic = "GoCompile",
       command = f.path,
@@ -291,22 +292,22 @@ def go_library_impl(ctx):
   out_lib = ctx.new_file(lib_name)
   search_path = out_lib.path[:-len(lib_name)]
   gc_goopts = _gc_goopts(ctx)
-  transitive_libs = depset([out_lib])
-  transitive_lib_paths = [search_path]
+  transitive_go_libraries = depset([out_lib])
+  transitive_go_library_paths = [search_path]
   for dep in deps:
-     transitive_libs += dep.transitive_go_libraries
-     transitive_cgo_deps += dep.transitive_cgo_deps
-     transitive_lib_paths += dep.transitive_go_library_paths
+    transitive_go_libraries += dep.transitive_go_libraries
+    transitive_cgo_deps += dep.transitive_cgo_deps
+    transitive_go_library_paths += dep.transitive_go_library_paths
 
   _emit_go_compile_action(ctx,
-    step="lib",
-    sources=go_srcs,
-    deps= deps,
-    libpaths=transitive_lib_paths, 
-    out_lib=out_lib, 
-    extra_objects=extra_objects, 
-    gc_goopts=gc_goopts,
-    )
+      step = "lib",
+      sources = go_srcs,
+      deps = deps,
+      libpaths = transitive_go_library_paths, 
+      out_lib = out_lib, 
+      extra_objects = extra_objects, 
+      gc_goopts = gc_goopts,
+  )
 
   dylibs = []
   if cgo_object:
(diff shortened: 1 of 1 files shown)


go/def.bzl
```diff
@@ -382,16 +287,26 @@ def go_library_impl(ctx):
...
+  transitive_libs = depset([out_lib])
+  transitive_lib_paths = [search_path]
   for dep in deps:
-     transitive_libs += dep.transitive_go_library_object
+     transitive_libs += dep.transitive_go_libraries
      transitive_cgo_deps += dep.transitive_cgo_deps
-     transitive_importmap += dep.transitive_go_importmap
+     transitive_lib_paths += dep.transitive_go_library_paths
+
+  _emit_go_compile_action(ctx,
```
jayconrod: More style nits: 4 space indentation for function call arguments. `ctx` should be on its own line. Use space before and after = for named arguments. Closing ) should either be on the same line as the last argument or in the same column as the first line.

This file doesn't follow these rules consistently, but I think they are followed most of the time. Maybe we can add a lint check or some automatic reformatting in the future.

go/def.bzl
```diff
@@ -575,9 +458,7 @@ def _emit_go_link_action(ctx, importmap, transitive_libs, cgo_deps, lib,
       ]
 
   cmds += [
```
jayconrod: This can be a single line now.

go/def.bzl
```diff
@@ -891,57 +761,56 @@ def _cgo_codegen_impl(ctx):
...
-      # to prevent cgo from prefixing mangled directory names to the output
-      # files.
-      # TODO(#350): this causes errors if sources are in multiple directories.
-      'cd "%s/src/$(dirname %s)"' % (out_dir, _short_path(ctx.files.srcs[0])),
-      'unfiltered_go_files=(%s)' % ' '.join(["'%s'" % f.basename for f in go_srcs]),
+      'unfiltered_go_files=(%s)' % ' '.join(["'%s'" % f.path for f in go_srcs]),
       'filtered_go_files=()',
       'for file in "${unfiltered_go_files[@]}"; do',
-      '  if "$filter_tags" -cgo -quiet "$file"; then',
+      '  name=$(basename "$file" .go)',
```
jayconrod: Nit: `stem` would be a better name for this variable.

go/def.bzl
```diff
@@ -891,57 +761,56 @@ def _cgo_codegen_impl(ctx):
...
       '  fi',
       'done',
       '"$GOROOT/bin/go" tool cgo -objdir "$objdir" -- %s "${filtered_go_files[@]}"' %
           ' '.join(['"%s"' % copt for copt in copts]),
-      'rm -f $objdir/_cgo_.o $objdir/_cgo_flags']
+      # Rename the outputs using glob so we don't have to understand cgo's mangling
+      # TODO(#350): might be fixed by this?.
+      'for file in "${filtered_go_files[@]}"; do',
+      '  name=$(basename "$file" .go)',
+      '  mv $objdir/*$name.cgo1.go $objdir/$name.cgo1.go',
```
jayconrod: This will break if `objdir` or `name` have spaces in them.

I don't really have a solution to propose, other than to eventually rewrite this in Go. It's good to fix #350, but I think this problem is just as bad.

go/def.bzl
```diff
@@ -382,16 +287,26 @@ def go_library_impl(ctx):
...
-  out_lib = ctx.outputs.lib
+  lib_name = _go_importpath(ctx) + ".a"
+  out_lib = ctx.new_file(lib_name)
+  search_path = out_lib.path[:-len(lib_name)]
   gc_goopts = _gc_goopts(ctx)
-  _emit_go_compile_action(ctx, go_srcs, deps, out_lib, extra_objects, gc_goopts)
-
-  transitive_libs = set([out_lib])
-  transitive_importmap = {out_lib.path: _go_importpath(ctx)}
+  transitive_libs = depset([out_lib])
```
jayconrod: Might be good to name these `transitive_go_libraries` and `transitive_go_library_paths` to be consistent with the provider.

go/def.bzl
```diff
@@ -891,57 +761,56 @@ def _cgo_codegen_impl(ctx):
...
       '  fi',
       'done',
       '"$GOROOT/bin/go" tool cgo -objdir "$objdir" -- %s "${filtered_go_files[@]}"' %
           ' '.join(['"%s"' % copt for copt in copts]),
-      'rm -f $objdir/_cgo_.o $objdir/_cgo_flags']
+      # Rename the outputs using glob so we don't have to understand cgo's mangling
+      # TODO(#350): might be fixed by this?.
+      'for file in "${filtered_go_files[@]}"; do',
+      '  name=$(basename "$file" .go)',
+      '  mv $objdir/*$name.cgo1.go $objdir/$name.go',
```
jayconrod: (Sorry if this is a duplicate; I wrote a comment earlier, but it disappeared).

This will break if `objdir` or `name` contains a space. I don't think there's a way to both glob and quote here. You could improve it with `"$objdir/"*"$name.cgo.go"`; cgo name mangling might still emit spaces that the glob would catch though.

go/def.bzl
```diff
@@ -264,25 +175,15 @@ def _emit_go_compile_action(ctx, sources, deps, out_lib,
     ctx: The skylark Context.
     sources: an iterable of source code artifacts (or CTs? or labels?)
     deps: an iterable of dependencies. Each dependency d should have an
```

go/def.bzl
```diff
@@ -264,25 +175,15 @@ def _emit_go_compile_action(ctx, sources, deps, out_lib,
...
-  # Build a symlink tree so that that deps are in directories that match
-  # their import paths.
-  tree_layout = {}
-  inputs = []
-  for d in deps:
-    importpath = d.transitive_go_importmap[d.go_library_object.path]
-    tree_layout[d.go_library_object] = importpath + ".a"
-
-  inputs += list(sources)
+  inputs = depset() + sources
```

go/def.bzl
```diff
@@ -317,26 +218,30 @@ def _emit_go_compile_action(ctx, sources, deps, out_lib,
...
   if extra_objects:
-    extra_inputs += extra_objects
+    inputs += extra_objects
     objs = ' '.join([c.path for c in extra_objects])
     cmds += [ctx.file.go_tool.path + " tool pack r " + out_lib.path + " " + objs]
 
-  f = _emit_generate_params_action(cmds, ctx, out_lib.basename + ".GoCompileFile.params")
-
+  f = _emit_generate_params_action(cmds, ctx, ctx.label.name + "." + step + ".GoCompileFile.params")
+  inputs += [f]
```

go/def.bzl

go/def.bzl
```diff
@@ -382,16 +287,26 @@ def go_library_impl(ctx):
...
   gc_goopts = _gc_goopts(ctx)
-  _emit_go_compile_action(ctx, go_srcs, deps, out_lib, extra_objects, gc_goopts)
-
-  transitive_libs = set([out_lib])
-  transitive_importmap = {out_lib.path: _go_importpath(ctx)}
+  transitive_libs = depset([out_lib])
+  transitive_lib_paths = [search_path]
   for dep in deps:
-     transitive_libs += dep.transitive_go_library_object
+     transitive_libs += dep.transitive_go_libraries
```

go/def.bzl
ianthehat: we need it to be automatic...
I am basically incapable of either noticing or remembering white space rules

go/def.bzl

go/def.bzl

go/def.bzl
ianthehat: I have made the change.
I think it could be properly fixed with $(find "$objdir" -maxdepth 1 -name "*$name.cgo.go") but that adds a dependency on a working version of find, so I decided not to.
The script is already not safe if there are spaces in the name (for instance, the rm at the bottom) so in the short term it's not a problem, and we plan to kill this for a go program in the next couple of weeks anyway.

jayconrod: Sounds good. The go program will fix all of this.

Issue #430, opened 2017-05-03
go_library (and other rules) don't work with target names containing slashes


robotoer:
Unfortunately, this: https://github.com/bazelbuild/rules_go/pull/411 breaks when using `go_library` targets that have slashes (`/`) in their name.

It seems that when using something like (see: https://github.com/bazelbuild/rules_go/blob/master/go/def.bzl#L778-L780):
```
go_library_outputs = {
    "lib": "%{name}.a",
}
```
in combination with:
```
  # Notice the out_lib.basename call here
  f = _emit_generate_params_action(cmds, ctx, out_lib.basename + ".GoCompileFile.params")
```

causes slashes in the name of targets to be treated as a part of the filepath, not the filename thus causing the first portion of the target name to be stripped. If there are two go_libraries defined as such:
```
go_library(
    name="my/target/name/with/slashes/mygopkg",
    ...
)

go_library(
    name="my/target/mygopkg",
    ...
)
```

This will cause the above code to generate a duplicate file named: `mygopkg.a.GoCompileFile.params` producing an error that looks like:
```
ERROR: file 'services/playdoh/vendor/internal.a.GoCompileFile.params' is generated by these conflicting actions:
Label: //services/playdoh/vendor:google.golang.org/grpc/internal, //services/playdoh/vendor:github.com/grpc-ecosystem/grpc-gateway/runtime/internal
RuleClass: go_library rule
Configuration: a0f7cfb2b43155be453b0ec199580fee
Mnemonic: FileWrite
Action key: 98711db68060cd961cfd7ce3a24d4961, 0d8fecdea63c9606b1503267ab3c3381
Progress message: Writing script services/playdoh/vendor/internal.a.GoCompileFile.params
PrimaryInput: (null)
PrimaryOutput: Artifact:[[/home/robert/.cache/bazel/_bazel_robert/19f329a8d4726a5ab6808b6e18fda04b/execroot/uno]bazel-out/local-fastbuild/bin]services/playdoh/vendor/internal.a.GoCompileFile.params
Primary outputs are different objects: 266495146, 2072729152
Owner information: //services/playdoh/vendor:google.golang.org/grpc/internal a0f7cfb2b43155be453b0ec199580fee (1330916379 2070122513), //services/playdoh/vendor:github.com/grpc-ecosystem/grpc-gateway/runtime/internal a0f7cfb2b43155be453b0ec199580fee (1731755235 2070122513)
MandatoryInputs: are equal
Outputs: are equal
[...]

PR #431, opened 2017-05-03
Fixes go rules so that target names with slash characters can be used.

robotoer:
Addresses https://github.com/bazelbuild/rules_go/issues/430

Replaces `/` characters with `_` in golang target names.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -388,7 +388,7 @@ def go_library_impl(ctx):
     _emit_go_asm_action(ctx, src, asm_hdrs, obj)
     extra_objects += [obj]
 
-  out_lib = ctx.outputs.lib
+  out_lib = _output_lib(ctx)
   gc_goopts = _gc_goopts(ctx)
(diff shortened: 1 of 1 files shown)


bazel-io:

robotoer:
This fix does open up a new issue where if you were to have two go_libraries defined as such:
```
go_library(
    name="my/target/name",
    ...
)

go_library(
    name="my_target_name",
    ...
)
```

The build would fail as those two targets would create files with the same names.

jayconrod:
Thanks for sending this PR. #430 is definitely a problem we should fix, but I'm concerned this solution will break the correspondence between Bazel labels, output files, and Go import paths. Since this changes the output file name, the compiler and linker would not be able to find the correct dependencies. I don't think we have any tests that cover that, but we'll add some.

I'll add some more comments on #430.

robotoer:
No longer necessary as this has been fixed in #429

PR #438, opened 2017-05-04
Split the pack and compile steps

ianthehat:
This changes the compile step to output a .o in all cases, and then use a subsequent pack step to build the .a
This removes the last read modify write action from the bash scripts, and reduces the compile bash script to something we can retool

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -243,20 +243,20 @@ def _emit_go_compile_action(ctx, step, sources, deps, libpaths, out_object, gc_g
       env = go_environment_vars(ctx),
   )
 
-def _emit_go_pack_action(ctx, out_lib, archives):
+def _emit_go_pack_action(ctx, out_lib, objects):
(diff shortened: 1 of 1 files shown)


ianthehat:
This is failing tests because it requires the GOROOT changes in #434 to be submitted first

go/def.bzl
```diff
@@ -240,19 +237,32 @@ def _emit_go_compile_action(ctx, step, sources, deps, libpaths, out_lib,
...
+      env = go_environment_vars(ctx),
+  )
+
+def _emit_go_pack_action(ctx, out_lib, archives):
+  """Construct the command line for packing objects together.
+
+  Args:
+    ctx: The skylark Context.
+    out_lib: the archive that should be produced
+    objects: an iterable of object or archive files to be added to the output archive file.
```
jayconrod: s/objects/archives/

ianthehat: fixed

PR #443, opened 2017-05-08
Kill the compile script

ianthehat:
Add the ability to filter a command line and then run it to filter_tags
Add the ability to abs arguments to filter_tags
Use the new filter tags to kill the remnants of the compile bash script

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -170,13 +170,11 @@ def _go_importpath(ctx):
     path = path[1:]
   return path
 
-def _emit_go_compile_action(ctx, step, sources, deps, libpaths, out_object, gc_goopts):
+def _emit_go_compile_action(ctx, sources, deps, libpaths, out_object, gc_goopts):
   """Construct the command line for compiling Go code.
-  Constructs a symlink tree to accommodate for workspace name.
 
   Args:
     ctx: The skylark Context.
-    step: used to generate unique names for this step of a rule.
     sources: an iterable of source code artifacts (or CTs? or labels?)
     deps: an iterable of dependencies. Each dependency d should have an
       artifact in d.transitive_go_libraries representing all imported libraries.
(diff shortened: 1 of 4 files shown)


go/tools/filter_tags/filter_tags.go
```diff
@@ -55,16 +60,76 @@ func filterFilenames(bctx build.Context, inputs []string) ([]string, error) {
...
+}
+
 func main() {
 	flag.Parse()
 
 	bctx := build.Default
 	bctx.BuildTags = strings.Split(*tags, ",")
 	bctx.CgoEnabled = *cgo
 
+	if *run {
```
jayconrod: It would be better to extract this functionality into a new binary, maybe `filter_cmd`.

go/tools/filter_tags/filter_tags.go
```diff
@@ -29,6 +30,10 @@ var (
 	cgo   = flag.Bool("cgo", false, "Sets whether cgo-using files are allowed to pass the filter.")
 	quiet = flag.Bool("quiet", false, "Don't print filenames. Return code will be 0 if any files pass the filter.")
 	tags  = flag.String("tags", "", "Only pass through files that match these tags.")
+	run   = flag.Bool("run", false, "Run mode, filter a command line and then run it.")
+
+	absMarker    = "≡"
+	filterMarker = ""
```
jayconrod: Private use symbols are an elegant way to add a side channel into the command line. I intended to use JSON for this earlier, but as we discussed, it makes the Bazel command line opaque for anyone running with `--subcommands`, and it causes problems if we need to preserve order among arguments in different lists (e.g., filtered and non-filtered).

A more conventional solution would be to have flags before arguments that need special processing, e.g., `-abs`, `-filter`. That may require escaping or rejecting arguments that start with '-'. Although it's less elegant, I think it's a less surprising solution, and it would make the command lines more readable. Up to you.

If you do keep these symbols, it would be good to add comments, both here and in def.bzl, indicating these are in the private use area. Definitely good to have these declared as global constants.

go/def.bzl
```diff
@@ -170,76 +170,41 @@ def _go_importpath(ctx):
...
-  args += gc_goopts + ['"${FILTERED_GO_FILES[@]}"']
-
-  # Pack extra objects into an archive, if provided.
-  # Set -p to the import path of the library, ie.
-  # (ctx.label.package + "/" ctx.label.name) for now.
-  cmds += [' '.join(args)]
-
-  f = _emit_generate_params_action(cmds, ctx, ctx.label.name + "." + step + ".GoCompileFile.params")
-  inputs += [f, ctx.executable._filter_tags] + ctx.files.toolchain
+  args += gc_goopts + go_sources + cgo_sources
```
jayconrod: Consider adding a `--` in there somewhere.

go/def.bzl
```diff
@@ -170,76 +170,41 @@ def _go_importpath(ctx):
...
-      '    FILTERED_GO_FILES+=("$line")',
-      '  fi',
-      'done < <(\'%s\' -cgo "${UNFILTERED_GO_FILES[@]}")' % ctx.executable._filter_tags.path,
-      'if [ ${#FILTERED_GO_FILES[@]} -eq 0 ]; then',
-      '  echo no buildable Go source files in %s >&1' % str(ctx.label),
-      '  exit 1',
-      'fi',
-  ]
-
+  go_sources = ["" + i.path for i in sources if not i.basename.startswith("_cgo")]
```
jayconrod: It probably doesn't matter, but we may want to preserve order here.

go/tools/filter_tags/filter_tags.go
ianthehat: As we discussed, moving to a model that uses a longer string rather than a magic character, and a different "escaping" prefix, but keeping the rest of the model the same

go/tools/filter_tags/filter_tags.go
```diff
@@ -55,16 +60,76 @@ func filterFilenames(bctx build.Context, inputs []string) ([]string, error) {
...
+		if strings.HasPrefix(in, absMarker) {
+			in, _ = filepath.Abs(in[len(absMarker):])
+		}
+		if !strings.HasPrefix(in, filterMarker) {
+			// not a filter candidate
+			args = append(args, in)
+			continue
+		}
+		filename := in[len(filterMarker):]
+		fullPath, err := filepath.Abs(filename)
```
ianthehat: done

go/tools/filter_tags/filter_tags.go
```diff
@@ -55,16 +60,76 @@ func filterFilenames(bctx build.Context, inputs []string) ([]string, error) {
...
+		if _, err := os.Stat(filename); err != nil {
+			// not a valid file, don't filter
+			args = append(args, filename)
+			continue
+		}
+		dir, base := filepath.Split(fullPath)
+		matches, err := bctx.MatchFile(dir, base)
```
ianthehat: Not sure what you meant, but I have changed the flow so much I am not sure the comment would still apply anyway

go/tools/filter_tags/filter_tags.go
```diff
@@ -55,16 +60,76 @@ func filterFilenames(bctx build.Context, inputs []string) ([]string, error) {
...
+		if err != nil {
+			//match test failure, return it
+			return err
+		}
+		if !matches {
+			// file should be filtered...
+			filtered++
+			continue
+		}
+		// not a match, add it
```
ianthehat: Re worded, I find they use of "match" for filters very ambiguous, if it matches, does that mean it is filtered or passed...

go/tools/filter_tags/filter_tags.go
```diff
@@ -55,16 +60,76 @@ func filterFilenames(bctx build.Context, inputs []string) ([]string, error) {
...
+		args = append(args, filename)
+		unfiltered++
+	}
+	// args should now be filtered
+	// if all possible filter candidates were removed, then don't run the command
+	if filtered > 0 && unfiltered == 0 {
+		os.Exit(1)
```
ianthehat: done

go/tools/filter_tags/filter_tags.go
ianthehat: done

go/def.bzl

go/def.bzl
ianthehat: not needed, the presence of the command to exec always terminates the argument processing

go/tools/filter_exec/filter_exec.go
```diff
@@ -0,0 +1,111 @@
...
+// runCommand goes through it's arguments filtering out source code files that do not match
+// the supplied build context, and expanding the current working directory where needed.
+// It then invokes the executable with the remaining result.
+func runCommand(bctx build.Context, executable string, input []string) error {
+	var err error
+	args := []string{}
+	unfiltered := 0
+	filtered := 0
+	for _, in := range input {
+		if strings.HasPrefix(in, escapeMarker) {
```
jayconrod: I don't think this escaping is quite correct.

If only one verb is allowed, then the conditions checking `absMarker` and `filterMarker` below should be exclusive: only one should be matched, and we shouldn't examine any text after the match. I think it would be cleaner to have all three conditions together, combined with the conditions below. Something like this:

```go
if strings.HasPrefix(in, escapeMarker) {
  in = in[len(escapeMarker):]
} else if strings.HasPrefix(in, absMarker) {
  in = in[len(absMarker):]
  in, err := filepath.Abs(in)
  ...
} else if strings.HasPrefix(in, filterMarker) {
  in = in[len(filterMarker):]
  dir, base := filepath.Split(in)
  ...
}
args = append(args, in)
```

If more than one verb can be matched, we'd need a loop. `escapeMarker` should terminate that loop when matched.

ianthehat: It's deliberate. A loop allowing any ordering would be overkill, and then you would have to decide if it should complain if you repeat, and all that extra, I allow you if you are specifying filter to also specify that it should be abs'd first, which is a useful combination.
This is not a general purpose tool, it's invoked entirely by our rules, so I am fine with feature restriction that increases code simplicity, so long as it has the feature set we need right now.

PR #444, opened 2017-05-08
Fix stamped_bin

ianthehat:
It was not actually doing any stamping, or testing wether the stamping worked.
It has been converted from a binary to a test, that validates that the timestamp variable in the included library is correctly replaced.


ianthehat:
I think we need to put some thought into what's a test, what's an example, and how to make sure in the future that examples remain correct. For now, I think we just maintain the examples folder to make sure they remain at least partially useful, and then restructure them properly when the dust settles on major changes.

PR #446, opened 2017-05-08
merge go_repository rules

ianthehat:
This is a proof of concept change to merge new_go_repository into
go_repository.
Instead it autmatically runs gazelle on repositories that do not
supply a build file.
It also adds the ability to fetch a go repository by source blob
instead, with the same automatic running of gazelle if needed.
This is fully backwards compatible, but going forward would allow
gazelle to only generate go_repository rules, and to automatically
switch to source blobs when it knows it can resolve one (rather
than using the vcs)

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -14,8 +14,8 @@
 
 
 def _go_repository_impl(ctx):
-  fetch_repo = ctx.path(ctx.attr._fetch_repo)
-
+  # TODO(iancottrell): Decide if commit and tag apply to non vcs forms
+  # TODO(iancottrell): Decide if commit and tag should remain mutually exclusive
   if ctx.attr.commit and ctx.attr.tag:
     fail("cannot specify both of commit and tag", "commit")
   if ctx.attr.commit:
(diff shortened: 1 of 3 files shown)


go/private/go_repository.bzl
```diff
@@ -25,67 +25,84 @@ def _go_repository_impl(ctx):
...
-      '--remote', remote,
-      '--rev', rev,
-      '--vcs', vcs,
-      '--importpath', importpath])
-  if result.return_code:
-    fail("failed to fetch %s: %s" % (ctx.name, result.stderr))
+  if ctx.attr.url:
+    # explicit source url
+    if ctx.attr.vcs:
+        fail("cannot specify both of vcs and url")
```
jayconrod: Nit: 2 space indent for blocks.

go/private/go_repository.bzl
```diff
@@ -25,67 +25,84 @@ def _go_repository_impl(ctx):
...
-    "build_tags": attr.string_list(),
-    "_fetch_repo": attr.label(
-        default = Label("@io_bazel_rules_go_repository_tools//:bin/fetch_repo"),
-        allow_files = True,
-        single_file = True,
-        executable = True,
-        cfg = "host",
-    ),
-}
+      fail("failed to generate BUILD files for %s: %s" % (
```
jayconrod: Nit: 2 space indent

go/private/go_repository.bzl
```diff
@@ -25,67 +25,84 @@ def _go_repository_impl(ctx):
...
+        # Attributes for a repository that cannot be inferred from the import path
+        "vcs": attr.string(default="", values=["", "git", "hg", "svn", "bzr"]),
+        "remote": attr.string(),
 
-new_go_repository = repository_rule(
-    implementation = _new_go_repository_impl,
-    attrs = _go_repository_attrs + {
+        # Attributes for a repository that comes from a source blob not a vcs
+        "url": attr.string(),
+        "strip_prefix": attr.string(),
```
jayconrod: I'm really hesitant to encourage the use of `strip_prefix`, since renaming a GitHub repository can retroactively break everything that depends on it. I wonder if we should provide an alternative, maybe `strip_count` that removes the first n directory components? If we keep `strip_prefix`, we should definitely document the risk of using it.

go/private/go_repository.bzl
```diff
@@ -95,3 +112,5 @@ new_go_repository = repository_rule(
         ),
     },
 )
+
+new_go_repository = go_repository
```
jayconrod: Might be good to have a comment explaining what `new_go_repository` was for future generations.

go/private/go_repository.bzl
```diff
@@ -25,67 +25,84 @@ def _go_repository_impl(ctx):
...
+        '--remote', ctx.attr.remote,
+        '--rev', rev,
+        '--vcs', ctx.attr.vcs,
+        '--importpath', ctx.attr.importpath,
+    ])
+    if result.return_code:
+      fail("failed to fetch %s: %s" % (ctx.name, result.stderr))
 
+  if (ctx.attr.disable_build_file_generation or 
+      ctx.path('BUILD').exists or
```
jwpowell055: A couple of pathological cases might be worth considering:

  1. What if there's file named BUILD but it comes from the veeeerrry similar build system used by that other company? E.g. https://buckbuild.com/rule/go_binary.html or http://www.pantsbuild.org/go_readme.html?  This could be solved using a flag to _force_ the generation of a Bazel BUILD file, or BUILD.bazel as is typical of new_git_repository.

  2. What if Bazel BUILD files already exist but only in subdirectories of the base repo?  This may happen if my repo contains only packages that should be imported as, say, "github.com/jwpowell055/foo_repo/bar_package".

go/private/go_repository.bzl
ianthehat: me to, but that's the tool that download_and_extract gives us right now, the same thing exposed by http_archive, and I am expecting these rules to be written and updated by gazelle. Also, github is not the only place we might get source archives from, and that's very much a github concert, so I decided to stick with the norm for bazel rather than go our own way.
When we are doing it correctly, either you will get an identical archive, and things will be fine, or you won't and the sha won't match so we will complain early and clearly, and you can run gazelle to make it all better again.

go/private/go_repository.bzl
ianthehat: done

go/private/go_repository.bzl
ianthehat: done

go/private/go_repository.bzl
ianthehat: done

go/private/go_repository.bzl
ianthehat: For case 1, I think there is no way to cope with it correctly, as you will need special processing to strip out BUILD files that bazel must not see as well as generate new ones, which may involve a custom tool to do a conversion, I think when that situation occurs, you are going to need a whole new rule to cope with it, or something more complicated than just running gazelle. I am not opposed to adding an extra flag when we find we need it, but I would really want an actual case to work on before I do.
I guess I could change ctx.attr.disable_build_file_generation to a tri state (off, auto, on) and default to auto?

ianthehat: We don't support repo's without a root BUILD that sets the go_prefix...
But assuming we start to, either gazelle will run, add a root build file, and not damage the sub directory ones (which hopefully have all been generated by people using gazelle anyway) or you disable it with disable_build_file_generation.

Issue #409, opened 2017-04-26
Proposal: Gazelle build tag handling


jayconrod:
### Background

In #396, we stopped applying build constraints by default in Gazelle. The goal of this change was to  produce platform-independent BUILD files. Unfortunately, this has caused several problems.

1. We now generate rules for files with `// +build ignore` or other tags indicating they will never be built. These rules frequently end up empty when build constraints are applied at compile-time. If there are files declaring multiple packages in a directory, Gazelle may report a `MultiplePackageError` even if `go/build` would not have reported that error because of ignored files. This affects golang.org/x/sys/unix/linux. See #399.
2. We are generating rules for files with `// +build appengine`. We don't support AppEngine yet. This pulls in AppEngine dependencies from external repositories, which probably aren't defined. This affects several godoc-related packages in golang.org/x/tools. In general, we should avoid adding dependencies on packages imported by files that are never built.
3. In #402, generated rules include dependencies and flags that aren't available on the target system. Specifically, Windows libraries are included in a Linux build. This proposal won't fix that issue, but I mention it for completeness.

There must be a compromise between not applying build constraints at all and applying build constraints that only match the platform Gazelle is run on.

### Proposal

Gazelle should determine whether a file *may be* built, using tags it knows are *maybe true*. When Gazelle is invoked with `-build_tags`, it will interpret those tags as *maybe true*. Other tags will be interpreted as *never true*. By default, Gazelle will treat all OS, architecture, and Go version tags as *maybe true*.

Gazelle will parse and evaluate `// +build` comments in source files to determine whether they may be built. The logic for evaluating conditions with Maybe and Never is different than True and False:

* NOT Maybe = Maybe. NOT Never = Maybe.
* Maybe AND Maybe = Maybe. Never AND X = Never.
* Maybe OR X = Maybe. Never OR Never = Never.

This will let us evaluate conditions like `// +build windows,!linux`. If `windows` and `linux` were both True, this would evaluate to False, and we would not build a file with this condition. This approach is necessarily conservative since [SAT is NP-complete](https://en.wikipedia.org/wiki/Boolean_satisfiability_problem): we may still include more files than necessary. For example, a contradictory condition like `// +build ignore,!ignore` would still be built.

With this proposal, rules generated by Bazel will include files that may be built from any platform. The normal build constraint logic will still be applied by the rules at compile-time to determine which files should actually be built.

[...]

PR #447, opened 2017-05-08
Gazelle select (1/x): consolidate parameters in packageReader

jayconrod:
This is the first of several changes which will make Gazelle put
platform-specific sources, dependencies, and options into select
expressions.

This groups several related values into a new struct,
packageReader. This avoids passing the same parameters to every
function. I will add several more methods on this struct in an
upcoming change.

Related: #409

### DIFF
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -29,14 +29,6 @@ import (
 	"strings"
 )
 
-// packageReader holds the state used to read package metadata from a directory.
-// This mostly exists to avoid passing the same of parameters to every function.
-type packageReader struct {
-	bctx               build.Context
-	repoRoot, goPrefix string
-	dir                string
-}
-
 // A WalkFunc is a callback called by Walk for each package.
 type WalkFunc func(pkg *build.Package) error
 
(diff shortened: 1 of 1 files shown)


go/tools/gazelle/packages/walk.go
```diff
@@ -28,6 +29,14 @@ import (
 	"strings"
 )
 
+// packageReader holds the state used to read package metadata from a directory.
+// This mostly exists to avoid passing the same of parameters to every function.
```
ianthehat: These comments imply a passive structure, something without methods, whereas it seems to be an active one.

go/tools/gazelle/packages/walk.go
```diff
@@ -28,6 +29,14 @@ import (
...
+type packageReader struct {
+	bctx               build.Context
+	repoRoot, goPrefix string
```
ianthehat: why do repoRoot and goPrefix belong together, when dir does not?

go/tools/gazelle/packages/walk.go
jayconrod: I've changed the comments to indicate this. Also moved the definition below `WalkFunc` and `Walk`, closer to the methods.

go/tools/gazelle/packages/walk.go
jayconrod: Done. Originally, I just had one `packageReader`, and I set the `dir` each time the callback passed to `filepath.Walk` was called.

PR #448, opened 2017-05-08
Gazelle select (2/x): add Package, Target, PlatformStrings

jayconrod:
Added three structs that contain metadata for a Go package in a
directory. These will be used to generate rules in place of
go/build.Context.

The main difference between these and go/build.Context is that these
structs divide files by target instead of type, and they are capable
of list platform-specific files, imports, and flags for multiple
platforms.

Related: #409

### DIFF
--- a/go/tools/gazelle/packages/package.go
+++ b/go/tools/gazelle/packages/package.go
@@ -57,21 +57,20 @@ func (p *Package) HasGo() bool {
 	return p.Library.HasGo() || p.CgoLibrary.HasGo() || p.Binary.HasGo() || p.Test.HasGo() || p.XTest.HasGo()
 }
 
-// firstGoFile returns the name of a .go file and true if the package contains
-// at least one .go file, or "" and false otherwise. Used by HasGo and for
-// error reporting.
-func (p *Package) firstGoFile() (string, bool) {
-	if f, ok := p.Library.firstGoFile(); ok {
-		return f, true
+// firstGoFile returns the name of a .go file if the package contains at least
+// one .go file, or "" otherwise. Used by HasGo and for error reporting.
+func (p *Package) firstGoFile() string {
+	if f := p.Library.firstGoFile(); f != "" {
+		return f
 	}
-	if f, ok := p.CgoLibrary.firstGoFile(); ok {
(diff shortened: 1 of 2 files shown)


go/tools/gazelle/packages/package.go
```diff
@@ -0,0 +1,118 @@
...
+
+// HasGo returns true if at least one target in the package contains a
+// .go source file. If a package does not contain Go code, Gazelle will
+// not generate rules for it.
+func (p *Package) HasGo() bool {
+	return p.Library.HasGo() || p.CgoLibrary.HasGo() || p.Binary.HasGo() || p.Test.HasGo() || p.XTest.HasGo()
+}
+
+// firstGoFile returns the name of a .go file and true if the package contains
+// at least one .go file, or "" and false otherwise. Used by HasGo and for
```
ianthehat: Given that a valid go file cannot be the empty string, what is the benefit of the extra bool?

go/tools/gazelle/packages/package.go
```diff
@@ -0,0 +1,118 @@
...
+
+package packages
+
+import "strings"
+
+// Package contains metadata about a Go package extracted from a directory.
+// It fills a similar role to go/build.Package, but it separates files by
+// target instead of by type, and it supports multiple platforms.
+type Package struct {
+	Dir  string
```
ianthehat: To support the longer term goals of directories with multiple packages generating multiple go_library rules, should there be an extra level in here?

type Directory {
  Path string
  Packages []Package
}

And remove the Dir member from Package?
Or would you just return a slice of Package directly and have a dir in each one?

jayconrod: The second one: planning to return `[]Package` eventually.

I could see taking `Dir` out at that point and passing it in as a separate parameter to `rules.Generator.Generate`, but I'm not sure I want to do that refactoring yet.

go/tools/gazelle/packages/package.go
jayconrod: Trying to avoid overloading a single value, but I guess it is unnecessarily cluttered. Fixed.

PR #449, opened 2017-05-10
Gazelle select (3/x): add fileInfo and related methods

jayconrod:
fileInfo holds information used to decide how to build a file.

packageReader.goFileInfo and .otherFileInfo create fileInfo. These
methods examine file name, build tags, package declaration, imports,
and cgo directives.

The functions in this change imitate logic from go/build. Unexported
functions from go/build are copied where possible.

Related: #409

### DIFF
--- a/go/tools/gazelle/packages/fileinfo.go
+++ b/go/tools/gazelle/packages/fileinfo.go
@@ -80,7 +80,7 @@ type taggedOpts struct {
 	opts []string
 }
 
-// extCategory indicates how a file should be treated, based on extention.
+// extCategory indicates how a file should be treated, based on extension.
 type extCategory int
 
 const (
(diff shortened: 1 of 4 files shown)


go/tools/gazelle/packages/fileinfo.go
```diff
@@ -0,0 +1,538 @@
...
+}
+
+// taggedOpts a list of compile or link options which should only be applied
+// if the given set of build tags are satisfied.
+type taggedOpts struct {
+	tags string
+	opts []string
+}
+
+// extCategory indicates how a file should be treated, based on extention.
```
jayconrod: Done

go/tools/gazelle/packages/fileinfo.go
```diff
@@ -0,0 +1,538 @@
...
+// fileNameInfo returns information that can be inferred from the name of
+// a file. It does not read data from the file.
+func fileNameInfo(dir, name string) fileInfo {
+	ext := path.Ext(name)
+
+	// Determine test, goos, and goarch. This is intended to match the logic
+	// in goodOSArchFile in go/build.
+	var isTest bool
+	var goos, goarch string
+	i := strings.Index(name, "_")
```
jayconrod: Done. This was copied from go/build originally, but I agree this is more readable.

jayconrod:
Agree about exposing some stuff in go/build. Not sure what the best path is there. Maybe some new exported methods in that package in the Go 1.10 timeframe?

Issue #427, opened 2017-05-02
gazelle needs to run faster


mikedanese:
gazelle is unreasonably slow in the kubernetes repository. It is 2-3 orders of magnitude slower than gazel. For example:
```console
$ pwd
/usr/local/google/home/mikedanese/go/src/k8s.io/kubernetes/pkg
$ time gazel pkg
gazel pkg  3.09s user 0.57s system 141% cpu 2.583 total
$  time (cd pkg/; gazelle -mode fix > gazelle.log 2>&1)
( cd pkg/; gazelle -mode fix > gazelle.log 2>&1; )  5.34s user 4.12s system 1% cpu 10:10.35 total
```
That's 2 seconds vs 10 minutes. I don't know if this is the only issue, but gazel got a huge speedup when we started caching calls to the build.Context, a la [this](https://github.com/mikedanese/gazel/blob/v17/gazel/gazel.go#L127-L146).

jayconrod:
Agreed that gazelle is unacceptably slow for Kubernetes. I think #353 is the cause of most of this slowness.

I'll leave this issue open to remind us to re-evaluate performance once #353, since there are probably other significant causes.

jayconrod:
@mikedanese Could you check performance with the current Gazelle is acceptable? #353 is fixed now.

jayconrod:
@mikedanese Gentle ping. Is performance acceptable now?

We'll be moving away from `build.Context` soon, but I don't expect the performance to be significantly difference, since the new solution does much of the same work.

mikedanese:
Yes perf is now accpetable. ~2 seconds for the whole repo using:

gazelle -external vendored

PR #450, opened 2017-05-10
Gazelle: cache lookups using vcs.RepoRootForImportPath

### LINKED ISSUE #353: Gazelle in external mode is slow for repositories with lots of dependencies
In [resolve_external.go](https://github.com/bazelbuild/rules_go/blob/cfdcbdc1d17e6dc3c48bbda4fce760949e58e381/go/tools/gazelle/rules/resolve_external.go#L54), Gazelle calls [`vcs.RepoRootForImportPath`](https://godoc.org/golang.org/x/tools/go/vcs#RepoRootForImportPath), which can be very slow (it may fetch information about repositories over the network). This gets called for every external import in every .go file.

One solution would be to call [`vcs.RepoRootForImportPathStatic`](https://godoc.org/golang.org/x/tools/go/vcs#RepoRootForImportPathStatic) instead, which does not do a network fetch. This will work fine for well-known sites like GitHub, but it won't be able to identify the repository root for other sites, unless it's qualified in the import path.

Another solution would be to read the repositories in the WORKSPACE file, which have full URLs, and attempt to resolve using those *before* calling `RepoRootForImportPath`.

jayconrod:
This change fixes a performance problem in the external
resolver.

For each import path in a repository, Gazelle determines a label for
the corresponding dependency. For external dependencies, the external
name in the label is based on the repository root URL, so Gazelle
looks this up using the vcs package. The lookup process is expensive,
since it involves HTTP requests and redirects for unqualified paths.

This change caches lookup results. It also unifies the special case
logic into this cache.

Fixes #353
Related #427

### DIFF
--- a/go/tools/gazelle/rules/resolve_external.go
+++ b/go/tools/gazelle/rules/resolve_external.go
@@ -16,6 +16,7 @@ limitations under the License.
 package rules
 
 import (
+	"path"
 	"strings"
 
 	"golang.org/x/tools/go/vcs"
@@ -84,17 +85,17 @@ func findCachedRepoRoot(importpath string) string {
 				// The import path is shorter than expected. Treat as a miss.
 				return ""
 			}
-			// Cache hit.
+			// Cache hit. Restore n components of the import path to get the
+			// repository root.
 			return subpaths[len(subpaths)-n-1]
 		}
 
 		// Prefix not found. Remove the last component and try again.
-		i := strings.LastIndexByte(importpath, '/')
-		if i < 0 {
+		importpath = path.Dir(importpath)
+		if importpath == "." || importpath == "/" {
 			// Cache miss.
 			return ""
 		}
-		importpath = importpath[:i]
 		subpaths = append(subpaths, importpath)
 	}
 }
--- a/go/tools/gazelle/rules/resolve_external.go
+++ b/go/tools/gazelle/rules/resolve_external.go
@@ -28,19 +28,18 @@ var (
 )
 
 // ImportPathToBazelRepoName converts a Go import path into a bazel repo name
- // following the guidelines in http://bazel.io/docs/be/functions.html#workspace
- func ImportPathToBazelRepoName(importpath string) string {
-      components := strings.Split(importpath, "/")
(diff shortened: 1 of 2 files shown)


go/tools/gazelle/rules/resolve_external.go
```diff
@@ -70,22 +70,55 @@ func (e externalResolver) resolve(importpath, dir string) (label, error) {
...
+				return ""
+			}
+			// Cache hit.
+			return subpaths[len(subpaths)-n-1]
 		}
-		l := len(known)
-		if idx := strings.Index(importpath[l:], "/"); idx != -1 {
-			return importpath[:l+idx]
+
+		// Prefix not found. Remove the last component and try again.
```
ianthehat: Maybe use 
  importpath = path.Dir(importpath)
  if importpath == "." {
    return ""
  }
instead?

go/tools/gazelle/rules/resolve_external.go
```diff
@@ -70,22 +70,55 @@ func (e externalResolver) resolve(importpath, dir string) (label, error) {
...
-		if !strings.HasPrefix(importpath, known) {
-			continue
+	for {
+		if n, ok := repoRootCache[importpath]; ok {
+			if n >= len(subpaths) {
+				// The import path is shorter than expected. Treat as a miss.
+				return ""
+			}
+			// Cache hit.
+			return subpaths[len(subpaths)-n-1]
```
ianthehat: It took me a fair while to work out why subpaths[len(subpaths)-n-1] was the right answer, not sure what comment you could add to make it more clear.
Maybe something like
"We found a cached vcs root, restore n elements to get the repository root"

jayconrod: Sorry about that, I know it's confusing. I thought about doing something with splitting and joining strings, but I thought it might be just as confusing and less efficient.

I've updated the comment.

go/tools/gazelle/rules/resolve_external.go
jayconrod: Thanks, that's much better.

PR #451, opened 2017-05-10
Gazelle select (4/x): convert and maps to select expressions

jayconrod:
newValue now converts maps to to a call to the select function with a
dictionary argument. The keys of the dict are the keys of the map
converted to strings. The values are converted recursively with
newValue.

newValue also supports packages.PlatformStrings. This is converted
into a list if it only has generic strings, a select call if it only
has platform-specific strings, or a concatenation of the two if it has
both.

These cases are not used yet, since rules doesn't use packages.Package
yet. This prepares for later PRs.

Related: #409

### DIFF
--- a/go/tools/gazelle/rules/construct.go
+++ b/go/tools/gazelle/rules/construct.go
@@ -18,8 +18,10 @@ package rules
 import (
 	"fmt"
 	"reflect"
+	"sort"
 
 	bzl "github.com/bazelbuild/buildifier/build"
+	"github.com/bazelbuild/rules_go/go/tools/gazelle/packages"
 )
 
 type keyvalue struct {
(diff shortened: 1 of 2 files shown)


go/tools/gazelle/rules/construct.go
```diff
@@ -82,30 +87,73 @@ func newValue(val interface{}) (bzl.Expr, error) {
...
-		glob, ok := val.(globvalue)
-		if !ok {
-			return nil, fmt.Errorf("not implemented %T", val)
+
+	case reflect.Map:
+		rkeys := rv.MapKeys()
+		sort.Slice(rkeys, func(i, j int) bool {
+			return rkeys[i].String() < rkeys[j].String()
+		})
+		args := make([]bzl.Expr, 0, len(rkeys))
```
ianthehat: The only path out of here that does not set every element of args abandons it totally, so you might as well just make it the right size and avoid append.
	args := make([]bzl.Expr, len(rkeys))
	for i, rk := range rkeys {
		k := &bzl.StringExpr{Value: rk.String()}
		v, err := newValue(rv.MapIndex(rk))
		if err != nil {
			return nil, err
		}
		args[i] = &bzl.KeyValueExpr{Key: k, Value: v})
	}

jayconrod: Done

PR #454, opened 2017-05-11
Bulk reformat

ianthehat:
first run of
bazel run //:bzl_format

### DIFF
--- a/go/private/bzl_format.bzl
+++ b/go/private/bzl_format.bzl
@@ -38,8 +38,7 @@ py_binary(
 
 
 _bzl_format_yapf_repository = repository_rule(
-    implementation = _bzl_format_yapf_repository_impl,
-    attrs = {})
+    implementation = _bzl_format_yapf_repository_impl, attrs = {})
 
 
 def bzl_format_repositories():
@@ -61,8 +60,7 @@ def _run_yapf_impl(ctx):
 
 
 _run_yapf = rule(
-    _run_yapf_impl,
-    attrs = {
+    _run_yapf_impl, attrs = {
         "yapf": attr.label(),
     })
 
--- a/go/private/files_equal_test.bzl
+++ b/go/private/files_equal_test.bzl
@@ -12,15 +12,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """Tests that two files contain the same data."""
+
 
 def _impl(ctx):
   if ctx.file.golden == ctx.file.actual:
     fail("GOLDEN and ACTUAL should be different files")
   ctx.file_action(
-      output=ctx.outputs.executable,
-      content="\n".join([
+      output = ctx.outputs.executable,
+      content = "\n".join([
           "#!/bin/bash",
           "function checksum() {",
           "  if command -v openssl >/dev/null; then",
@@ -43,9 +43,9 @@ def _impl(ctx):
           "  exit 1",
           "fi",
       ]),
-      executable=True)
-  return struct(runfiles=ctx.runfiles([ctx.file.golden,
-                                       ctx.file.actual]))
+      executable = True)
+  return struct(runfiles = ctx.runfiles([ctx.file.golden, ctx.file.actual]))
+
 
 def _runpath(f):
   """Figures out the proper runfiles path for a file, using voodoo"""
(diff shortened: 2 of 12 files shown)


go/private/bzl_format.bzl
```diff
@@ -38,8 +38,7 @@ py_binary(
 
 
 _bzl_format_yapf_repository = repository_rule(
-    implementation = _bzl_format_yapf_repository_impl,
-    attrs = {})
+    implementation = _bzl_format_yapf_repository_impl, attrs = {})
```
pmbethe09: skylark-style is one key-value pair to a line, so not a fan of this re-format.  Can yapf be configured?

go/private/go_repositories.bzl
```diff
@@ -16,53 +16,59 @@ load("//go/private:go_repository.bzl", "go_repository", "new_go_repository")
...
-        repo = 'https://github.com/bazelbuild/buildtools',
-        commit = '81c36a8418cb803d381335c95f8bb419ad1efa27',
-    ),
-    'tools': struct(
-        importpath = 'golang.org/x/tools',
-        repo = 'https://github.com/golang/tools',
-        commit = '3d92dd60033c312e3ae7cac319c792271cf67e37',
-    )
+    'buildtools':
+        struct(
```
pmbethe09: this does not seem to be an improvement

go/private/files_equal_test.bzl
```diff
@@ -54,18 +54,17 @@ def _runpath(f):
...
-            allow_files = True,
-            single_file = True),
-        "actual": attr.label(
-            mandatory = True,
-            allow_files = True,
-            single_file = True),
-        "error_message": attr.string(
-            default="FILES DO NOT HAVE EQUAL CONTENTS"),
+        "golden":
+            attr.label(
```
pmbethe09: not better?

proto/go_proto_library.bzl
```diff
@@ -300,45 +320,45 @@ _well_known_imports = [
...
   if shared:
     # if using multiple *_proto_library, allows caller to skip this.
     native.http_archive(
         name = "com_github_google_protobuf",
         url = "https://github.com/google/protobuf/archive/v3.2.0.tar.gz",
         strip_prefix = "protobuf-3.2.0",
-        sha256 = "2a25c2b71c707c5552ec9afdfb22532a93a339e1ca5d38f163fe4107af08c54c",
-    )
+        sha256 =
+        "2a25c2b71c707c5552ec9afdfb22532a93a339e1ca5d38f163fe4107af08c54c")
```
pmbethe09: not better

proto/go_proto_library.bzl
```diff
@@ -300,45 +320,45 @@ _well_known_imports = [
...
   new_go_repository(
       name = "org_golang_x_net",
       commit = "4971afdc2f162e82d185353533d3cf16188a9f4e",
-      importpath = "golang.org/x/net",
-  )
+      importpath = "golang.org/x/net")
   new_go_repository(
       name = "org_golang_google_grpc",
       tag = "v1.0.4",
-      importpath = "google.golang.org/grpc",
```
pmbethe09: should stay like before to match BUILD syntax of trailing ',' and ')' on next line

jayconrod:
I turned a few knobs, but I couldn't get it to work the way I wanted it. Most of the issues Paul pointed out are still present.

Some options I tried:

* `split_before_first_argument`: if a call is split, the split always happens before the first argument. Pretty sure we want this.
* `split_arguments_when_comma_terminated`: this gives us one argument per line if the last argument ends with a comma. Seems reasonable, but we'll need to manually add commas to a lot of calls.
* `indent_dictionary_value`: if a dictionary key and value don't fit on the same line, the value will be indented.
* `coalesce_brackets`: combines (, [, { when it makes sense.

Some remaining problems:

* `coalesce_brackets` does not play nice with `split_arguments_when_comma_terminated`. The closing paren gets placed on the same line as the last argument after the comma, which looks awkward. Example:

```bzl
_bzl_format_yapf_repository = repository_rule(
    implementation = _bzl_format_yapf_repository_impl,
    attrs = {},)
```

* If a dictionary key and value don't fit on one line and the value is a call (e.g., struct), the split should happen before the first argument instead of before the call. I don't think yapf can do this. I'm willing to live with it though.
* If a keyword argument is split, the value is on the same column as the key. I didn't see a way to indent it.

ianthehat:
We have decided that we can't quite get yapf to produce a format we are happy with, so we are going to wait for a proper skylark formatter.

Issue #140, opened 2016-10-13
Coverage of go test


GinFungYJF:
Does the go rule still have no support for calculating the coverage of go test now? If so, is there any way to calculate it?

yugui:
It does not support coverage right now.
Probably we need to wait for https://github.com/bazelbuild/bazel/issues/1118.

GinFungYJF:
Any update on this? Or is there any usable workaround like 
https://github.com/bazelbuild/bazel/issues/1118  ?

pmbethe09:
AFAIK no one is working on this.  If it is something you need, you may have
to be the one to implement it.

ghasemloo:
I think we can fix this directly, go tool itself support coverage independent of bazel.
I think what needs to be done is adding something similar to the block regarding coverage from go test main (https://golang.org/src/cmd/go/test.go?h=cover+flag#L1522)
to generate test main template: (https://github.com/bazelbuild/rules_go/blob/master/go/tools/generate_test_main.go#L129).

I tried it but I don't fully understand how flags are parsed in the go test, they are defined in 
https://golang.org/src/cmd/go/testflag.go

GinFungYJF:
@ghasemloo When I run go test -cover -x  pkg, I see that it will do some preproccessing for the files that are not *_test.go files in the pkg before they are compiled.
```
go/pkg/tool/linux_amd64/cover -mode set -var GoCover_0 -o .A.go pkg/A.go
go/pkg/tool/linux_amd64/cover -mode set -var GoCover_0 -o .B.go pkg/B.go
...
```
Is it helpful to you?

linuxerwang:
It would be great if the future bazel-go-coverage also supports external-test besides standard unit tests. Here are a few references of this technique:

https://blog.cloudflare.com/go-coverage-with-external-tests/
https://husobee.github.io/golang/test/coverage/2015/11/17/external-test-coverage.html
https://www.elastic.co/blog/code-coverage-for-your-golang-system-tests

I hope I could find time to work on it.

pmbethe09:
@linuxerwang -- sounds great.  Feel free to send some ideas here and then
of course send a PR .

GinFungYJF:
Has anybody already have any idea about implementing it ? I don't know where to start :(

jayconrod:
@GinFungYJF We haven't given this a lot of thought. It seems clear this will be a lot of work though.

[...]

PR #455, opened 2017-05-14
Instrument test coverage, and modify the generated main test program

linuxerwang:
Add test coverage for rules_go.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -182,6 +182,9 @@ def _emit_go_compile_action(ctx, sources, deps, libpaths, out_object, gc_goopts)
     out_object: the object file that should be produced
     gc_goopts: additional flags to pass to the compiler.
   """
+  if ctx.coverage_instrumented():
+    sources = _emit_go_cover_action(ctx, sources)
+
   # Compile filtered files.
   args = [
       "-cgo",
@@ -206,6 +209,8 @@ def _emit_go_compile_action(ctx, sources, deps, libpaths, out_object, gc_goopts)
       env = go_environment_vars(ctx),
   )
 
+  return sources
+
 def _emit_go_pack_action(ctx, out_lib, objects):
   """Construct the command line for packing objects together.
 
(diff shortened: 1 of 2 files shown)


bazel-io:

linuxerwang:
> In general it looks good, is it ready to use? If so could you give an example of how to use it?

It's not fully ready to use, since I need a way to make the report easy to reach for CI tools. That part needs more of my spare time.

To use it, simply run the standard bazel coverage:

```
$ bazel coverage examples/lib:lib_test

$ cat bazel-testlogs/examples/lib/lib_test/test.log 
2 + 3 = 5
PASS
coverage: 40.0% of statements
```

> Do you have any ideas as to how we could add a test to make sure we don't break it?

The breakage of the normal "bazel test" runs are prevented by current unit tests. The breakage of the future "bazel coverage" runs might need some addition in the CI scripts. Is the rules_go current using jenkins for CI?

> From my understanding of Bazel coverage I was expecting to see instrumented_files being returned in the rules providers, is it not needed?

Good point. I'll put some spare time on making it work as expected. But the current implementation is functioning fine, or at least it's not breaking the existing "bazel test" runs.

Thanks for the input!

go/tools/generate_test_main.go
```diff
@@ -39,6 +59,29 @@ type Cases struct {
 	HasTestMain      bool
 	Version17        bool
 	Version18OrNewer bool
+	Cover            []coverInfo
```
jayconrod: Could this be replaced with `CoverVars []CoverVar`? Seems like a flat list would work well here, since we only range over this. This would let you drop the `coverInfo` struct.

jayconrod:
For testing, I'll add a test script once the other changes are addressed, so don't worry about that. I think it will be enough to run "bazel coverage" on a test with known coverage and verify the output (as in your example). Later, we'll need to verify integration with Bazel.

For reference, this is related to #140.

Jenkins test this please

go/def.bzl
```diff
@@ -223,6 +228,41 @@ def _emit_go_pack_action(ctx, out_lib, objects):
       env = go_environment_vars(ctx),
   )
 
+def _emit_go_cover_action(ctx, sources):
+  """Construct the command line for test coverage instrument.
+
+  Args:
+    ctx: The skylark Context.
+    sources: an iterable of Go source code.
```
linuxerwang: Done.

go/def.bzl
```diff
@@ -223,6 +228,41 @@ def _emit_go_pack_action(ctx, out_lib, objects):
...
+
+  Returns:
+    A list of Go source code files which might be coverage instrumented.
+  """
+  outputs = []
+  # TODO(linuxerwang): make the mode configurable.
+  args = ("tool", "cover", "--mode=set")
```
linuxerwang: Done.

go/def.bzl
```diff
@@ -223,6 +228,41 @@ def _emit_go_pack_action(ctx, out_lib, objects):
...
+  count = -1
```
linuxerwang: Done.

go/tools/generate_test_main.go
linuxerwang: It is actually copied from $GOROOT/src/cmd/go/test.go. I suggest we keep it so that it will be easier to compare and update with the reference impl. Also, the map is needed because we need to track cover vars for each file.
Thanks!

go/def.bzl
```diff
@@ -223,6 +228,41 @@ def _emit_go_pack_action(ctx, out_lib, objects):
...
+
+  for src in sources:
+    if not src.path.endswith(".go") or src.path.endswith("_test.go"):
+      outputs += [src]
+      continue
+
+    count += 1
+    coverVar = "GoCover_%d" % count
```
linuxerwang: Done.

go/tools/generate_test_main.go
jayconrod: Fair enough, sounds good.

go/private/go_repositories.bzl
```diff
@@ -219,14 +219,14 @@ def go_repositories(
            (go_version, " ".join(_GO_VERSIONS_SHA256.keys())))
     native.new_http_archive(
         name = "golang_linux_amd64",
-        url = "https://storage.googleapis.com/golang/go%s.linux-amd64.tar.gz" % go_version,
+        url = "http://localhost:3737/static/tools/bazel/go%s.linux-amd64.tar.gz" % go_version,
```
jayconrod: Please revert.

linuxerwang: Sorry, my bad. Reverted.

jayconrod:
Thanks!

Jenkins test this please

PR #458, opened 2017-05-16
Enable mono repos

ianthehat:
The previous work on cleaning up the library system now allows arbitrary remapping of library import paths.
This exposes that functionality directly on the library rule, and adds an example of using it.
The enables “mono” repos, where different sub directories represents radically different import paths.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -69,13 +69,6 @@ cgo_filetype = FileType([
 
 ################
 
-def _go_prefix(ctx):
-  """slash terminated go-prefix"""
-  prefix = ctx.attr.go_prefix.go_prefix
-  if prefix != "" and not prefix.endswith("/"):
-    prefix = prefix + "/"
-  return prefix
-
 def go_environment_vars(ctx):
   """Return a map of environment variables for use with actions, based on
   the arguments. Uses the ctx.fragments.cpp.cpu attribute, if present,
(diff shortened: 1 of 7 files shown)


go/def.bzl
```diff
@@ -374,14 +372,11 @@ def _emit_go_link_action(ctx, transitive_go_library_paths, transitive_go_librari
   """Sets up a symlink tree to libraries to link together."""
   config_strip = len(ctx.configuration.bin_dir.path) + 1
   pkg_depth = executable.dirname[config_strip:].count('/') + 1
-  prefix = _go_prefix(ctx)
 
   ld = "%s" % ctx.fragments.cpp.compiler_executable
   extldflags = _c_linker_options(ctx) + [
       "-Wl,-rpath,$ORIGIN/" + ("../" * pkg_depth),
   ]
-  if prefix:
```
jayconrod: So this isn't needed anymore after de-symlink-ification?

go/def.bzl
```diff
@@ -607,6 +601,7 @@ go_library_attrs = go_env_attrs + {
             "transitive_cgo_deps",
         ],
     ),
+    "import_path": attr.string(),
```
jayconrod: For consistency with `go_repository`, let's call this `importpath`.

go/def.bzl
ianthehat: yes, this was just leftover cruft I spotted on the way through

go/def.bzl

ianthehat:
I want to leave it as a mostly hidden feature for now, I think we need to revisit the documentation very thoroughly sometime soon, but things are so in flux now that I think we ought to just fix it where it is wrong and leave documenting new features until things are stable.

Issue #456, opened 2017-05-15
Gazelle fails on case-insensitive file systems when repository contains file named "build"


ains:
When a repository contains a file named "build" for example https://github.com/coreos/fleet

Gazelle will fail to generate on a case-insensitive FS even when given the "build_file_name" flag with "BUILD.bazel". I'm presuming this is because findBuildFile uses the hardcoded list of build file names, and will return the "build" file, which Gazelle will attempt to merge with.

Workaround atm is to delete the file in a fork of the repo but potential fix proposals are

1) Allow buildFileNames to be passed in as a flag
2) Add a --[no]merge flag to Gazelle, as I'm not sure it should be attempting to merge w/ build files in external repositories anyway

ains:
Currently opted for 1) and defaulting to build_file_name == valid_build_file_names for new_go_repository which unblocks us for now but open to discussing better issues.

https://github.com/improbable-io/rules_go/commit/4f0fb98aa0e6158c4e1d27ab6927e39558e5f4df

ianthehat:
The approach I took when fixing this issue for the go_repository rule detecting the root build file was to check whether the "build" file is a file, rather than just that it exists, and always use BUILD.bazel when a directory is found. I think the same thing would work inside gazelle, and cover this case easily, I will try to add both a test and a fix.

ains:
The issue with coreos/pkg coreos/fleet and coreos/etcd is that the "build" file is a file, it's just a bash script as opposed to a bazel build file, but the fact that it exists / is a file is good enough for Gazelle

ianthehat:
Sorry, I misunderstood problem.
Honestly I think the right answer is for all tooling to work only with BUILD.bazel by default, and for BUILD files to be a thing for the history books, the more explicit name is better for almost every possible use case. However, given that is not the case for where we are right now, I think your solution is mostly the best choice.
I think I would probably change it so that we don't take a new flag, but just split build buildFileName on commas and if there are multiple entries, the first one is used when generating new files, otherwise we start getting into the crazy world of "what if the value of buildFileName is not in validBuildFileNames".

ains:
@ianthehat agreed, will clean up commit and send a PR

PR #461, opened 2017-05-18
Determine valid build file names from buildFileName flag

ains:
Fixes https://github.com/bazelbuild/rules_go/issues/456

### DIFF
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -35,13 +35,12 @@ import (
 )
 
 var (
-	buildFileName  = flag.String("build_file_name", "BUILD", "name of output build files to generate.")
-	buildTags      = flag.String("build_tags", "", "comma-separated list of build tags. If not specified, Gazelle will not\n\tfilter sources with build constraints.")
-	external       = flag.String("external", "external", "external: resolve external packages with new_go_repository\n\tvendored: resolve external packages as packages in vendor/")
-	goPrefix       = flag.String("go_prefix", "", "go_prefix of the target workspace")
-	repoRoot       = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
-	mode           = flag.String("mode", "fix", "print: prints all of the updated BUILD files\n\tfix: rewrites all of the BUILD files in place\n\tdiff: computes the rewrite but then just does a diff")
-	buildFileNames = []string{"BUILD.bazel", "BUILD"}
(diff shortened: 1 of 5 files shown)


bazel-io:

go/tools/gazelle/gazelle/main.go
```diff
@@ -35,13 +35,12 @@ import (
...
 
 var (
-	buildFileName  = flag.String("build_file_name", "BUILD", "name of output build files to generate.")
-	buildTags      = flag.String("build_tags", "", "comma-separated list of build tags. If not specified, Gazelle will not\n\tfilter sources with build constraints.")
-	external       = flag.String("external", "external", "external: resolve external packages with new_go_repository\n\tvendored: resolve external packages as packages in vendor/")
-	goPrefix       = flag.String("go_prefix", "", "go_prefix of the target workspace")
-	repoRoot       = flag.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
-	mode           = flag.String("mode", "fix", "print: prints all of the updated BUILD files\n\tfix: rewrites all of the BUILD files in place\n\tdiff: computes the rewrite but then just does a diff")
-	buildFileNames = []string{"BUILD.bazel", "BUILD"}
+	buildFileName = flag.String("build_file_name", "BUILD,BUILD.bazel", "comma-separated list of valid build file names.\nThe first element of the list is the name of output build files to generate.")
```
ianthehat: Let's reverse this while we are in here and make BUILD.bazel the default?

ains:
Updated @ianthehat although wondering if we want to remove the default
`"build_file_name": attr.string(default="BUILD.bazel,BUILD"),`
As we don't pass the --build_file_name flag to Gazelle if it's empty, and then we don't have the default duplicated

ianthehat:
I almost suggested that, and then I wondered if it was nicer for users of the rules to not have to run the binary to find out what value that attribute would have. But now I think about it even more, I am not sure they should have to care, if we are doing it right nobody is ever going to want to set, or even know about that attribute. Basically I am ambivalent, do whichever you like, I will approve either one as soon as the tests are passing :)

ains:
Ah, we'll leave it for now - test should be fixed :)

achew22:
This change has the affect that when gazelle is run in a directory with `gazelle .` it removes any previous BUILD files and replaces them with BUILD.bazel. Is the objective to get everyone to migrate to BUILD.bazel files? If so could we document that? If not could we switch the default back to 'BUILD' and update the bzl to pass in the flag?

jayconrod:
I don't think the intent of this change was to rename existing files. I'll open an issue and fix it soon, hopefully today.

For new files though, the default will still be BUILD.bazel, which we'd like to encourage. BUILD has too many collisions with existing files and directories.

Issue #460, opened 2017-05-17
Should we remove the go_repository rule?


ianthehat:
*TLDR:* go_repository is overly complex, and causes problems for efficient distributed build systems, and we shouldn't really need it.

# Why does it exist?
The go repository rule does two main things that could _not_ be encompassed by a non go specific rule.
* Understands go package import redirection
* Generates BUILD files by running gazelle after download

# Why is it a problem?
Running arbitrary programs at workspace fetch time is not very friendly to systems using `bazel fetch` to prefill an image before running on isolated servers.

Running programs that use proprietary protocols to fetch content from the internet directly is not good for companies that want cached and verified copies of external dependencies, and run their entire build systems with only access to that cache.

Attempting to build those programs from source during that process is both very awkward (you cannot use bazel to build them for instance) and also not friendly to systems that don't expect `blaze fetch` to not do build work. There is also no way to not build the tools if they are not used, as you are not part of a build graph. There are lots of things it would be nice for gazelle to do in the future that will increase its dependencies considerably, and managing these while still building in this fashion is going to get harder and harder.

# How would it take to fix it

The goal would be to just use the normal repository rules, new_http_repository, new_git_repository. For this to work there are a few considerations. 

Most of the work of fetch_repo needs to be done at workspace editing time, rather than repository fetch. In other words, given an import path you need to work out the final fully resolved repository url and vcs type **before** you write the workspace rule, so that you can pick the right one.

The normal repository rules do not support adding BUILD files throughout the tree, but only adding a root BUILD file, so we need to generate these full repository BUILD files instead.

The BUILD files must be generated ahead of time, rather than just in time, so they would also be generated as the repository was added to the workspace. This means they have to be checked in somehwere.

# What are any other advantages

The BUILD files for external dependencies are no longer magic and invisible. This means if gazelle cannot fully cope with a dependency, you can hand edit the resulting BUILD file to fix it easily. It also makes it easier for users to examine the rules for their dependencies, and understand the full dependency tree.

It makes it easy to have a repository of BUILD files for common dependencies. While it would always be better to upstream BUILD files into the original repository anywhere that such changes would be accepted, this does make it easier to have a library of BUILD files for repositories where that is not accepted/feasible.

# How would we do it

Until recently doing this was not feasible, but with the current go rules we can now have packages independant of directory layout, and single build files for entire trees of packages.

[...]

PR #464, opened 2017-05-23
Improve go_repository when downloading from url

DrMarcII:
Replace url with urls attribute.
Require commit or tag only if urls is not set.
Disallow commit and tag if urls are set.

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 def _go_repository_impl(ctx):
-  if ctx.attr.url or ctx.attr.urls:
+  if ctx.attr.urls:
     disallowed = []
     # explicit source url
     if ctx.attr.vcs:
@@ -94,7 +94,6 @@ go_repository = repository_rule(
         "remote": attr.string(),
 
         # Attributes for a repository that comes from a source blob not a vcs
-        "url": attr.string(),
         "urls": attr.string_list(),
         "strip_prefix": attr.string(),
         "type": attr.string(),
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -14,19 +14,15 @@
 
 def _go_repository_impl(ctx):
   if ctx.attr.urls:
-    disallowed = []
     # explicit source url
     if ctx.attr.vcs:
-      disallowed.append("vcs")
+      fail("cannot specify both of urls and vcs", "vcs")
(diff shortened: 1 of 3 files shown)


ianthehat:
Honestly, I think we should just kill the url attribute and only allow urls, it would make more sense, and I don't think there will be so many users of this feature already that we can't just make a breaking change.

I wanted to leave the ability to set commit and or tag on rules, even in the presence of urls, and I don't even think they should be exclusive. This would allow us to write tools that automatically upgrade the commit id, repository url, sha256 and strip prefix for you, for instance to the current value of the tag/branch for instance.
If you want to make them optional if urls is set, I am fine with that.

Of course, if I have my way with #460 we will kill all of this, but that's a slightly longer term thing.

DrMarcII:
As the rules are written, commit and tag have no effect if url or urls are set, so allowing them is potentially misleading.

DrMarcII:
I did go ahead and delete the url attribute.

go/private/go_repository.bzl
```diff
@@ -13,26 +13,34 @@
...
-  else:
-    fail("neither commit or tag is specified", "commit")
-
-  if ctx.attr.url:
+  if ctx.attr.urls:
+    disallowed = []
     # explicit source url
     if ctx.attr.vcs:
-      fail("cannot specify both of vcs and url")
+      disallowed.append("vcs")
```
jayconrod: I think it would be simpler to just have an explicit `fail` in each branch.

go/private/go_repository.bzl
```diff
@@ -13,26 +13,34 @@
...
+    if ctx.attr.commit:
+      disallowed.append("commit")
+    if ctx.attr.tag:
+      disallowed.append("tag")
+    if disallowed:
+      fail("cannot specify and of %s if url or urls is specified", "url")
     ctx.download_and_extract(
         url = ctx.attr.url,
```
jayconrod: Can remove this argument, since ctx.attr.url shouldn't exist anymore.

`urls` is undocumented in `download_and_extract`. Any idea when it was added? Will this break our compatibility with Bazel 0.4.4?

DrMarcII: It is actually still url, it just now accepts either a string or a list of strings. Not sure when it changed.

DrMarcII:
Update README.md and tests/popular_repos/WORKSPACE.in.

Also deleted references to new_go_repository in those locations.

PR #472, opened 2017-05-25
Write rules in terms of a toolchain object

ianthehat:
This modifies all the rules to access the tools and files of a go toolchain
through a single struct, in preparation for that struct being an automatically
selected toolchain.
Accessing this struct is hidden behind a get_toolchain function for now, to
allow it's behaviour to be easily modified.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -94,7 +94,8 @@ def go_environment_vars(ctx):
   }
   env = {}
   if hasattr(ctx.file, "go_tool"):
-    env["GOROOT"] = ctx.file.go_tool.dirname + "/.."
+    go_toolchain = get_go_toolchain(ctx)
+    env["GOROOT"] = go_toolchain.root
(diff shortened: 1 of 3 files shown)


go/def.bzl
```diff
@@ -625,8 +630,33 @@ go_env_attrs = {
...
+    ),
+    "_extract_package": attr.label(
+        default = Label("//go/tools/extract_package"),
+        executable = True,
+        cfg = "host",
+    ),
 }
 
+
+def get_go_toolchain(ctx):
```
jayconrod: Can this be private? Looks like it isn't used outside of def.bzl right now, but maybe later?

ianthehat: I started trying to break up def.bzl and move some of these functions into the private area, which ironically means they need to be public functions, and the name of this is a left over from that, but it did not go well (cgo is a spiderweb), I will try again sometime soon, for now making it private.

PR #478, opened 2017-05-26
When running gazelle with --mode=diff, show missing BUILD file path instead of /dev/null when it does not exist.

stefan-improbable:


bazel-io:

stefan-improbable:
Closing this PR, I'm going to send the same PR this via a different user.

PR #484, opened 2017-05-26
Gazelle select (5/x): rewrite merge logic

### LINKED ISSUE #480: Gazelle removes keep comments from existing sources
Suppose we have the following BUILD.bazel file:

```bzl
load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_prefix")

go_prefix("github.com/jayconrod/scratch")

go_library(
    name = "go_default_library",
    srcs = [
        "foo.go",  # keep
    ],
)
```

When we run gazelle on this, if `foo.go` exists, gazelle will remove the `# keep` comment.

```bzl
load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_prefix")

go_prefix("github.com/jayconrod/scratch")

go_library(
    name = "go_default_library",
    srcs = ["foo.go"],
)
```

When gazelle merges lists of strings, it appends strings with a `# keep` comment to the generated list of strings, then sorts and uniqs the list (using buildifier). `foo.go` will be a duplicate, and the uniq can remove the string with the `# keep` comment.

jayconrod:
* Merging logic now handles calls to select with dictionaries for
  conditional srcs and deps.
* Most of the merging logic is rewritten. Instead of having "src and
  "dest" and modifying dest in place, most functions now have "old"
  and "gen" and return "merged".

Fixes #480
Related to #409

### DIFF
--- a/go/tools/gazelle/merger/merger.go
+++ b/go/tools/gazelle/merger/merger.go
@@ -104,7 +104,7 @@ func mergeRule(gen, old *bzl.CallExpr) *bzl.CallExpr {
 		} else {
 			genAttr := genRule.Attr(k)
 			var err error
-			mergedAttr, err = mergeAttr(genAttr, oldAttr)
+			mergedAttr, err = mergeExpr(genAttr, oldAttr)
 			if err != nil {
 				// TODO: add a verbose mode and log errors like this.
 				mergedAttr = genAttr
@@ -125,11 +125,20 @@ func mergeRule(gen, old *bzl.CallExpr) *bzl.CallExpr {
 	return &merged
 }
 
-// mergeAttr combines information from gen and old and returns an updated attr.
-// The attrs may be strings, lists, a select call with a dict, or a list plus
-// a select call. An error is returned if the attrs can't be merged, for example
-// because they are in a format Gazelle doesn't understand.
-func mergeAttr(gen, old bzl.Expr) (bzl.Expr, error) {
+// mergeExpr combines information from gen and old and returns an updated
+// expression. The following kinds of expressions are recognized:
+//
+//   * nil
+//   * strings (can only be merged with strings)
+//   * lists of strings
+//   * a call to select with a dict argument. The dict keys must be strings,
+//     and the values must be lists of strings.
+//   * a list of strings combined with a select call using +. The list must
+//     be the left operand.
+//
+// An error is returned if the expressions can't be merged, for example
+// because they are not in one of the above formats.
+func mergeExpr(gen, old bzl.Expr) (bzl.Expr, error) {
 	if _, ok := gen.(*bzl.StringExpr); ok {
 		if shouldKeep(old) {
 			return old, nil
--- a/go/tools/gazelle/merger/merger.go
+++ b/go/tools/gazelle/merger/merger.go
@@ -46,8 +46,8 @@ var (
 	}
 )
 
-// MergeWithExisting merges genfile with an existing build file at
-// existingFilePath and returns genfile.
+// MergeWithExisting merges genFile with an existing build file at
+// existingFilePath and returns the merged file.
 func MergeWithExisting(genFile *bzl.File, existingFilePath string) (*bzl.File, error) {
 	oldData, err := ioutil.ReadFile(existingFilePath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


go/tools/gazelle/merger/merger.go
```diff
@@ -46,112 +46,300 @@ var (
...
+	}
+
+	return &merged
+}
+
+// mergeAttr combines information from gen and old and returns an updated attr.
+// The attrs may be strings, lists, a select call with a dict, or a list plus
+// a select call. An error is returned if the attrs can't be merged, for example
+// because they are in a format Gazelle doesn't understand.
+func mergeAttr(gen, old bzl.Expr) (bzl.Expr, error) {
```
ianthehat: I think the fact that this is currently used to merge attributes is not a good reason to name it that.
In theory you could call this on anything that was an expression with a type that was understood (merging globals springs to mind)
I would just call it mergeExpression and doucment the expression types it works with.
The same is true of the other methods below.

jayconrod: Sounds good, renamed to `mergeExpr` and improved documentation.

go/tools/gazelle/merger/merger.go
```diff
@@ -46,112 +46,309 @@ var (
...
+	oldRule := bzl.Rule{Call: old}
+	merged := *old
+	merged.List = nil
+	mergedRule := bzl.Rule{Call: &merged}
+
+	for _, k := range oldRule.AttrKeys() {
+		oldAttr := oldRule.Attr(k)
+		var mergedAttr bzl.Expr
+		if !mergeableFields[k] {
+			mergedAttr = oldAttr
```
pmbethe09: long else blocks are sad...  
oldAttr is known to exists, so here just
mergedRule.SetAttr(k, oldAttr)
continue

and then avoid the long else as the following code requires no indent.

go/tools/gazelle/merger/merger.go
```diff
@@ -46,112 +46,309 @@ var (
...
+				return nil, d, nil
+			}
+		}
+	case *bzl.BinaryExpr:
+		if expr.Op == "+" {
+			if l, ok := expr.X.(*bzl.ListExpr); ok {
+				if call, ok := expr.Y.(*bzl.CallExpr); ok && len(call.List) == 1 {
+					if x, ok := call.X.(*bzl.LiteralExpr); ok && x.Token == "select" {
+						if d, ok := call.List[0].(*bzl.DictExpr); ok {
+							return l, d, nil
```
pmbethe09: this is incredibly hard to read.  Prefer guard-clauses with error return to avoid indent.
if expr.Op != "+" {
  return nil, fmt.Errorf(...
}
l, ok := expr.X.(*bzl.ListExpr)
if ok {
  return nil, fmt.Errorf(...
}
... and so forth.

go/tools/gazelle/merger/merger.go
jayconrod: Done

go/tools/gazelle/merger/merger.go
jayconrod: Done. I was trying to avoid some redundancy with the error messages, but it made the code very ugly. I wish Go had pattern matching; it's perfect for this.

jayconrod:
I've made some additional changes.

* In `mergeRule`, copy unnamed arguments from `oldRule`. This is needed to preserve the argument to `go_prefix`. We might want to consider merging these in the future, but I'm trying to match the previous behavior in this PR.
* In `mergeRule`, copy attribute-level comments during merging. Currently, we lose some expression level comments (we did before, too). I'll file a bug to track that.
* Renamed `attrListAndDict` to `exprListAndDict`.

PTAL

go/tools/gazelle/merger/merger.go
ianthehat: The problem with flattening (and the reason I did not suggest it) is now if you want to handle a non plus binary expression, you have to un-flatten again, and the more cases you add the worse it gets.
I have hit this a few times doing ast manipulation code in the past, not found a really good answer yet...
The cleanest I have found is to have "matching" functions that do all the tests flattened like  you do now, and return all valid parts, so it's one if per pattern. That generally that leads to inefficiencies where you test the same things in many conditions though, and because they are often nested they never get inlined well, but at least it means everywhere you want to test for the same pattern you are using the same code to do it.

jayconrod: Let's keep this as it is for now. I don't expect we'll be supporting other patterns, but we can extract this into other functions if we need to.

go/tools/gazelle/merger/merger.go
```diff
@@ -46,112 +46,333 @@ var (
...
+		return expr, nil, nil
+	case *bzl.CallExpr:
+		if x, ok := expr.X.(*bzl.LiteralExpr); ok && x.Token == "select" && len(expr.List) == 1 {
+			if d, ok := expr.List[0].(*bzl.DictExpr); ok {
+				return nil, d, nil
 			}
 		}
+	case *bzl.BinaryExpr:
+		if expr.Op != "+" {
+			return nil, nil, fmt.Errorf("expression could not be matched")
```
pmbethe09: Might have changed each error message to be more informative...

jayconrod: Sorry, that was lazy of me. #493 improves the error messages.

PR #487, opened 2017-05-27
Allow specification of alternate gRPC deps

### LINKED ISSUE #442: Allow vendored dependencies for `go_proto_library`
As it stands, whenever `go_proto_repositories()` is invoked it automatically adds `new_go_repository` entries for `org_golang_x_net` and `org_golang_google_grpc`, regardless of if they exist or not yet.  The `go_proto_library` rule also adds dependencies on `@org_golang_x_net//context:go_default_library` and `@org_golang_google_grpc//:go_default_library` (if the proto has servies) and `@com_github_golang_protobuf//proto:go_default_library` even if it doesn't.

If the project using these rules has vendored `golang.org/x/net` or `google.golang.org/grpc` this can create version clashing which will cause compilation to fail due to conflicting object files or conflicting definitions during build.

I think allowing for overriding the dependency paths just like the rules allows overriding protoc and protoc_gen_go would allow for more flexibility for users.

tomwilkie:
Fixes #442 

Have tested this and it seem to work, but this is my first skylark, so be kind!

### DIFF
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -223,6 +223,7 @@ def go_proto_library(name, srcs = None, deps = None,
                      protoc = "@com_github_google_protobuf//:protoc",
                      protoc_gen_go = "@com_github_golang_protobuf//protoc-gen-go",
                      rules_go_repo_only_for_internal_use = "@io_bazel_rules_go",
+                     grpc_deps = None,
                      **kwargs):
   """Macro which generates and compiles protobufs for Go.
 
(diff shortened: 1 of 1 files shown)


bazel-io:

pmbethe09:
I saw https://github.com/bazelbuild/rules_go/pull/488 first -- should we go with that one?

tomwilkie:
Yeah no problem!

jayconrod:
Thanks for sending this! I'll close it though since #488 landed first.

PR #494, opened 2017-05-31
Gazelle select (7/x): parse platform constraints

jayconrod:
Add a command line flag (-platforms), which specifies the platforms
Gazelle should generate BUILD files for. Each platform has a label
(corresponding to a config_setting) and a comma-separated list of tags
that are true on that platform.

config_settings are provided for Darwin, Linux, and Windows on amd64
by default in //go/platform. Gazelle references these platforms by
default. Users can define their own config_settings and reference them
on the command line.

The command line flag is parsed and added to generator.Generator, but
it's not used yet.

Related #409


go/tools/gazelle/packages/package.go
```diff
@@ -15,7 +15,46 @@ limitations under the License.
...
+)
+
+// PlatformConstraints is a map from config_setting labels (for example,
+// "@io_bazel_rules_go//go/platform:linux_amd64") to a sets of build tags
+// that are true on each platform (for example, "linux,amd64").
+type PlatformConstraints map[string]map[string]bool
+
+// ParsePlatformConstraints parses a string that contains all platforms and
+// tags on each platform. The string must contain an even number of fields,
+// separated by whitespace. Even fields (counting from 0) are platform labels.
```
ianthehat: Flags that are required to contain whitespace are the spawn of the devil....
Not making the label tag separator distinct from the pair separator makes the parsing harder, makes reading it harder, and requires things like "must be an even number".
Also I think they are inverted in logic from the way most people will want to thing about them, they will want to say "these tags should be this label"
```linux,amd64=@io_bazel_rules_go//go/platform:linux_amd64+darwin,amd64=@io_bazel_rules_go//go/platform:darwin_amd64+windows,amd64=@io_bazel_rules_go//go/platform:windows_amd64```
where the + operator is open to anything that is not shell or label specific I think would be better

Given the regularity of the rules, and the fact that we hope this is a temporary solution, it might be even better to support a pattern instead, or just specify the package where the constraints live and always generate label names using strings.Join(tags, "_")

Comment (jayconrod) [inline go/tools/gazelle/packages/package.go]: This won't necessarily be a temporary solution, so it would be good to get this right early on. Agree that whitespace is a bad idea. No opinion about whether tags or labels should be first.

It should be rare that users need to set this flag manually. It should only be needed if they want to support fewer platforms (for example, they don't want any mention of `linux` in their BUILD files) or if they want to add custom `config_setting` rules, implied by a custom set of tags.

Instead of +, how do you feel about semicolons?

```
linux,amd64=@io_bazel_rules_go//go/platform:linux_amd64;darwin,amd64=@io_bazel_rules_go//go/platform:darwin_amd64
```

As we add more options to Gazelle, especially complicated options like this, we might want to consider supporting per-project configuration files.

ianthehat: That's why I suggested just specifying the package, that way the per project specific customization is the BUILD file that lists the config settings, which has almost the same list of things in it that otherwise you have to replicate on the gazelle command line, and still have in the build file anyway. The only difference is that then you could split your config settings amongs multiple packages, which I don't think is enough of an advantage to be worth the complexity.
I feel like we ought to be trying to move towards gazelle having fewer to no command line options, not growing more.

jayconrod: Gazelle needs a table containing `config_setting` labels for each platform and lists of tags that are true for each platform. The tags will usually be GOOS and GOARCH, but there may also be tags for specific environments (for example, `appengine`, `js`) or specific features that would pull in dependencies that shouldn't always be compiled.

We need a simple way to express this table on the command line. There is some redundancy here that could be eliminated by requiring some conventions on `config_setting` naming. Tag lists must be specified though. Two suggestions you mentioned above (as I understand them):

* We could require all `config_setting`s to be in the same package so we only have to write the package name once instead of repeating it for each entry. If users define their own `config_setting`s, they would need to alias or copy the predefined rules if they still wanted to use them.
* We could derive `config_setting` names from the tags. For example, `linux,amd64` becomes `linux_amd64`. This prevents customization of the platform name, but I agree, that's of limited value.

Since the package name and the tag lists still need to be specified, maybe we would end up with two smaller options?

`-platform_package @io_bazel_rules_go//go/platforms -platform_tags darwin,amd64;linux,amd64;windows,amd64`

WDYT?

ianthehat: Part of my worry is this...
	android	arm
	darwin	386
	darwin	amd64
	darwin	arm
	darwin	arm64
	dragonfly	amd64
	freebsd	386
	freebsd	amd64
	freebsd	arm
	linux	386
	linux	amd64
	linux	arm
	linux	arm64
	linux	ppc64
	linux	ppc64le
	linux	mips
	linux	mipsle
	linux	mips64
	linux	mips64le
	netbsd	386
	netbsd	amd64
	netbsd	arm
	openbsd	386
	openbsd	amd64
	openbsd	arm
	plan9	386
	plan9	amd64
	solaris	amd64
	windows	386
	windows	amd64
Is the full list of goos/goarch pairs that make up a platform, anything that asks a user to repeat that list on a command line is probably a broken user interface, especially as it's hard coded in the go tools.
I think we should presume that full set exists, and just list the additional tags that might apply, and assume they apply to all those combinations.
You can even pick the GOOS and GOARCH lists out of go/build/syslist.go and just assume all combinations are valid.

jayconrod: Per offline discussion, at this time, the platforms will not be configurable. Each platform will just be GOOS,GOARCH. We'll support platforms that both Go and Bazel work on.

We'll figure out configurability when we find a compelling use case.

PR #501, opened 2017-06-01
Gazelle select (8/x): methods for adding files to Package

jayconrod:
Related #409

### DIFF
--- a/go/tools/gazelle/packages/package.go
+++ b/go/tools/gazelle/packages/package.go
@@ -274,7 +274,7 @@ func uniq(ss []string) []string {
 	}
 	result := ss[:1]
 	prev := ss[0]
-	for _, s := range ss {
+	for _, s := range ss[1:] {
 		if s != prev {
 			result = append(result, s)
 			prev = s
--- a/go/tools/gazelle/packages/package.go
+++ b/go/tools/gazelle/packages/package.go
@@ -17,6 +17,7 @@ package packages
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+// "buildTags" is a set of tags that are true on all platforms.
+// "platforms" is a map from platform names (labels referencing config_settings)
+// to sets of tags that are true on those platforms. The tag sets may include
+// "buildTags".
+//
+// An error is returned if a file is buildable but invalid (for example, a
+// test .go file containing cgo code). Files that are not buildable will not
+// be added to any target (for example, .txt files).
+func (p *Package) addFile(info fileInfo, cgo bool, buildTags map[string]bool, platforms PlatformConstraints) error {
+	var t *Target
```
ianthehat: I feel like this function would be easier to read as
```
	switch {
	case info.isXTest:
		if info.isCgo {
			return fmt.Errorf("%s: use of cgo in test not supported", info.path)
		}
		p.XTest.addFile(info, buildTags, platforms)
	case info.isTest:
		if info.isCgo {
			return fmt.Errorf("%s: use of cgo in test not supported", info.path)
		}
		p.Test.addFile(info, buildTags, platforms)
	case info.isCgo || cgo && (info.category == cExt || info.category == hExt || info.category == csExt):
		p.CgoLibrary.addFile(info, buildTags, platforms)
	case info.category == goExt || info.category == sExt || !cgo && info.category == hExt:
		p.Library.addFile(info, buildTags, platforms)
	}
	return nil
```

go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+		}
+		t = &p.XTest
+	} else if info.isTest {
+		if info.isCgo {
+			return fmt.Errorf("%s: use of cgo in test not supported", info.path)
+		}
+		t = &p.Test
+	} else if info.isCgo || cgo && (info.category == cExt || info.category == hExt || info.category == csExt) {
+		t = &p.CgoLibrary
+	} else if info.category == goExt || info.category == sExt || !cgo && info.category == hExt {
```
ianthehat: isn't the !cgo redundant, it will have been matched in the condition above if true?
(I did not fix this in my suggested switch version above btw)

go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+func (ps *PlatformStrings) Clean() {
+	sort.Strings(ps.Generic)
+	ps.Generic = uniq(ps.Generic)
+
+	genSet := make(map[string]bool)
+	for _, s := range ps.Generic {
+		genSet[s] = true
+	}
+
+	if ps.Platform != nil {
```
ianthehat: Invert the condition to a return and you can undent the rest of the function

go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+		for n, ss := range ps.Platform {
+			ss = remove(ss, genSet)
+			if len(ss) == 0 {
+				delete(ps.Platform, n)
+				continue
+			}
+			sort.Strings(ss)
+			ps.Platform[n] = uniq(ss)
+		}
+		if len(ps.Platform) == 0 {
```
ianthehat: Why is this necessary? Does something really care about Platform == nil rather than len(Platform) == 0? if so, it probably should not.

go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+	ps.Generic = uniq(ps.Generic)
+
+	genSet := make(map[string]bool)
+	for _, s := range ps.Generic {
+		genSet[s] = true
+	}
+
+	if ps.Platform != nil {
+		for n, ss := range ps.Platform {
+			ss = remove(ss, genSet)
```
ianthehat: I feel like this whole bit would be much easier to understand (and probably faster) if you just converted it to a map, then twiddled the bools (possibly converting back to a list at the end if you really really need a list rather than a map), rather than sorting and uniquing

go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+		if !remove[ss[r]] {
+			ss[w] = ss[r]
+			w++
+		}
+	}
+	return ss[:w]
+}
+
+func uniq(ss []string) []string {
+	if len(ss) <= 1 {
```
ianthehat: pointless precondition? first test of the loop does this anyway

go/tools/gazelle/packages/package.go
```diff
@@ -136,3 +137,171 @@ func (ts *PlatformStrings) firstGoFile() string {
...
+		return ss
+	}
+
+	var r, w int
```
ianthehat: only w is used outside the loop, and r is the true loop control variable, so you might as well declare and init (with :=) w outside and r inside
also you can compare ss[r] to ss[w] becuase r-1 will have been copied down there anyway
also, r is used as a straight up iterator...
also, you might as well just cheat and use append rather than count and reslice, it's not going to be much worse in performance and much easier to understand
```
func uniq(ss []string) []string {
	result := ss[:0]
	previous := ""
	for _, s := range ss {
		if s != previous {
			result = append(result, s)
			previous = s
		}
	}
	return result
}
```

go/tools/gazelle/packages/package.go
jayconrod: Done

go/tools/gazelle/packages/package.go
jayconrod: Yes, it is redundant. Thought it would make the condition more readable, but maybe it's just confusing.

go/tools/gazelle/packages/package.go
jayconrod: Done

go/tools/gazelle/packages/package.go
jayconrod: It's important that this stays a list that normally allows duplicates. The sorting and uniquing isn't important for the final result (buildifier does this on the AST), but it's useful to have predictable results for unit testing.

`PlatformStrings` is used for files, flags, and imports. `Clean` doesn't need to be called on files or flags. Files are only added once in a predictable order. Flags are not sorted, and duplicates are allowed.

`Clean` does need to be called on imports. Imports can be added repeatedly if they appear in different files, and they can be both generic and platform-specific. It should be possible to prevent duplication and to move platform-specific imports to the generic list if they are added to the generic list later, but it would require some extra internal data structures in `PlatformStrings` that would only be used for imports. It seemed simpler to `Clean` after everything is added.

go/tools/gazelle/packages/package.go
jayconrod: Nothing cares about this, but it made the tests a little easier to write (don't need to create an empty map), and it saves a little memory.

go/tools/gazelle/packages/package.go
jayconrod: I've mostly taken your proposed code. I think I need that precondition again now though.

PR #503, opened 2017-06-02
Absoulte paths for load statements

ianthehat:
Change all internal load statements to absolute paths otherwise we end up with
duplicate instances of the scripts, with thing like providers of the same name
that don't match.


jayconrod:
When you submit this, please mention in the squashed commit message this is a workaround for  bazelbuild/bazel#3115.

PR #505, opened 2017-06-03
Support stamping in x_def attributes

### LINKED ISSUE #432: support stamping multiple packages in one binary
We seemingly need to stamp multiple packages now (https://github.com/kubernetes/kubernetes/issues/45298), but the `go_binary` `linkstamp` option only supports a single package.

Any ideas besides adding a new attribute to allow specifying multiple packages? (`linkstamp_pkgs`?)

cc @mikedanese

ixdy:
This PR fixes #432 by augmenting the `x_defs` attribute of `go_binary` and `go_test` rules to refer to variables in the workspace status files.

This uses the unofficial `{VAR}` syntax I [implemented](https://github.com/bazelbuild/rules_docker/pull/17) in the `rules_docker` repo, as suggested by @mikedanese in https://github.com/bazelbuild/rules_go/issues/432#issuecomment-300245203; I'm not sure if there's a consensus style for this sort of thing, yet.

As of now this implicitly turns on stamping if it finds any `x_defs` in the `{VAR}` style, but we could also potentially make it explicit with a `stamp=1` attribute.

### DIFF
--- a/examples/stamped_bin/stamped_bin_test.go
+++ b/examples/stamped_bin/stamped_bin_test.go
@@ -22,7 +22,16 @@ import (
 )
 
 func TestStampedBin(t *testing.T) {
-	if stamp.BUILD_TIMESTAMP == stamp.NOT_A_TIMESTAMP {
+	// If we use an x_def when linking to override BUILD_TIMESTAMP but fail to
+	// pass through the workspace status value, it'll be set to empty string -
+	// overridden but still wrong. Check for that case too.
+	if stamp.BUILD_TIMESTAMP == stamp.NOT_A_TIMESTAMP || stamp.BUILD_TIMESTAMP == "" {
 		t.Errorf("Expected timestamp to have been modified, got %s.", stamp.BUILD_TIMESTAMP)
+	}
+	if stamp.XdefBuildTimestamp == "" {
+		t.Errorf("Expected XdefBuildTimestamp to have been modified, got %s.", stamp.XdefBuildTimestamp)
+	}
+	if stamp.PassIfEmpty != "" {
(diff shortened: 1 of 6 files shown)


ixdy:
also, I was clever here and extended the existing stamping functionality with yet more shell dependencies (`grep` and `cut`), but I could probably rework this into a short external go utility program instead.

examples/stamped_bin/stamped_bin_test.go
```diff
@@ -22,7 +22,13 @@ import (
 )
 
 func TestStampedBin(t *testing.T) {
-	if stamp.BUILD_TIMESTAMP == stamp.NOT_A_TIMESTAMP {
+	if stamp.BUILD_TIMESTAMP == stamp.NOT_A_TIMESTAMP || stamp.BUILD_TIMESTAMP == "" {
```
ixdy: When overridden with a stamped `x_def` (e.g. [here](https://github.com/ixdy/bazel-rules_go/blob/30d98e0cb710a10e04655330d73c1b150fa3fda5/examples/stamped_bin/BUILD#L18)) if the variable doesn't exist or the implementation doesn't work (which is what happened to me first. :))

This is mostly covered by other cases now, so I can revert this if you'd like.

ixdy: (that said, `BUILD_TIMESTAMP` should always exist, I'm guessing, so it might be nice to continue to test that both stamping codepaths are actually passing it through and not just passing an empty `x_def`)

ixdy: added a comment to explain this case.

PR #509, opened 2017-06-05
Gazelles select (10/10): wire in new package logic

### LINKED ISSUE #409: Proposal: Gazelle build tag handling

jayconrod:
* Rewrite packages/walk.go. Instead of producing build.Package using
  go/build, we now produce packages.Package using the logic that has
  been added in the last several changes in this series.
* Rewrite generator functions in rules/generator.go. We now produce
  rules using packages.Package. There is a new generic rule-creating
  method, generateRule, which the other methods are implemented with.

Fixes #409

### DIFF
--- a/go/tools/gazelle/generator/generator.go
+++ b/go/tools/gazelle/generator/generator.go
@@ -19,7 +19,6 @@ package generator
 
 import (
 	"fmt"
-	"go/build"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -39,7 +38,7 @@ type Generator struct {
 	repoRoot      string
 	goPrefix      string
 	buildFileName string
-	bctx          build.Context
+	buildTags     map[string]bool
 	platforms     packages.PlatformConstraints
 	g             rules.Generator
 }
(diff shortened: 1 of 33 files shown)


jayconrod:
Note: this change is blocked on #510 being fixed.

PR #515, opened 2017-06-07
Support select in cgo_library srcs and deps

### LINKED ISSUE #510: cgo_library assumes srcs is a list
This prevents the use of `select` with `cgo_library` which is important, since cgo sources are frequently platform-specific.

jayconrod:
`cgo_library` should allow `srcs` and `cdeps` to be arbitrary expressions
that produce lists of appropriate sources. In particular, they should
allow calls to select, which allows sources and dependencies to be
included conditionally.

This change introduces a new rule, `_cgo_select_files`, which copies
files that match a given set of suffixes (possibly inverted).
`_setup_cgo_library` and related functions are rewritten to
use this. Some other small fixes are included, such as demangling cgo
outputs.

In the future, we'll use output groups with filegroups instead of
`_cgo_select_files`, but they aren't supported in Bazel 0.4.4 (our
current minimum).

Fixes #510

### DIFF
--- a/go/private/cgo.bzl
+++ b/go/private/cgo.bzl
@@ -63,11 +63,8 @@ def cgo_genrule(name, srcs,
   )
   _cgo_genrule(
       name = name,
-      srcs = cgogen.go_thunks + [
-          cgogen.gotypes,
-          cgogen.outdir + "/_cgo_import.go",
-      ],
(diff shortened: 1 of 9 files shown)


jayconrod:
Sorry for the extra work. Let me know if I can help with the merge.

PR #524, opened 2017-06-08
Add the compile build tool

ianthehat:
This uses the same style as the asm tool, and removes the need for filter_exec


go/tools/builders/compile.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	gotool := args[0]
+	args = args[1:]
+	sources := []string{}
+	goargs := []string{}
+	bctx := build.Default
+	bctx.CgoEnabled = true
+	for i, s := range args {
+		if s == "--" {
+			if i < len(args) {
```
jayconrod: I think you can get rid of this condition. It should always be true.

go/tools/builders/compile.go
```diff
@@ -0,0 +1,90 @@
...
+				goargs = args[i+1:]
+			}
+			break
+		}
+		sources = append(sources, s)
+	}
+	// filter the soure list to the ones that apply
```
jayconrod: Typo in "soure". But consider rewording to "Apply build constraints to the source list".

go/tools/builders/compile.go
```diff
@@ -0,0 +1,90 @@
...
+
+func run(args []string) error {
+	// process the args
+	if len(args) < 2 {
+		return fmt.Errorf("Usage: compile gotool [sources] -- <extra options>")
+	}
+	gotool := args[0]
+	args = args[1:]
+	sources := []string{}
+	goargs := []string{}
```
jayconrod: Consider renaming to `goopts`. Keep the `goargs` as a separate variable below when you build the full argument list with `tool`, `compile` and sources.

go/tools/builders/compile.go
ianthehat: done

go/tools/builders/compile.go
ianthehat: It happens if the -- is the last entry in the list

go/tools/builders/compile.go
ianthehat: done

PR #527, opened 2017-06-08
Update popular_repos.bash

jayconrod:
* Add and build golang.org/x/sys
* Add github.com/mattn/go-sqlite3, but don't build it yet.
* Reorganize excludes and remove excludes that are now working.


jayconrod:
Tests working now. PTAL.

PR #533, opened 2017-06-12
Add go link wrapper tool

ianthehat:
Move all link stamping behavior into the tool.
This kills another use of bash.
There is now no more bash outside of the cgo rules.


examples/stamped_bin/BUILD
```diff
@@ -7,6 +7,7 @@ go_test(
     x_defs = {
         "github.com/bazelbuild/rules_go/examples/stamped_bin/stamp.PassIfEmpty": "",
         "github.com/bazelbuild/rules_go/examples/stamped_bin/stamp.XdefBuildTimestamp": "pass",
+        "github.com/bazelbuild/rules_go/examples/stamped_bin/stamp.XdefInvalid": "{Undefined_Var}",  # undefined should set the var to empty
```
jayconrod: Update this comment and the others below. Should be left alone now.

examples/stamped_bin/stamp/stamp.go
```diff
@@ -23,3 +23,6 @@ var PassIfEmpty = "fail"
 
 // an xdef should set this to nonempty
 var XdefBuildTimestamp = ""
+
+// an xdef with an missing key should leave this alone
```
jayconrod: Typo: "a missing key"

go/tools/builders/link.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	gotool := args[0]
+	args = args[1:]
+	linkargs := []string{}
+	goopts := []string{}
+	bctx := build.Default
+	bctx.CgoEnabled = true
+	for i, s := range args {
+		if s == "--" {
+			if i < len(args) {
```
jayconrod: I still don't think this condition is necessary. `i` should be strictly less than `len(args)` in the loop. `args[i+1:]` should be safe if `i` is the last index.

ianthehat: yes, you are right, I was not reading my own code correctly (originally there was a thing where the first value was not stripped, and the iteration index did not match the slice index, and I forgot I stopped doing that because it was messy)

examples/stamped_bin/stamp/stamp.go
ianthehat: done

examples/stamped_bin/BUILD
ianthehat: done

PR #534, opened 2017-06-13
Update README.md in preparation for 0.5.0 release

jayconrod:
[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,10 @@ They currently do not support (in order of importance):
 * coverage
 * test sharding
 
-Note: The latest version of these rules (0.5.0) require Bazel ≥ 0.4.5 to work.
-If you use these rules at master, you may need a newer version of Bazel.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,11 +59,8 @@ They currently do not support (in order of importance):
 * coverage
 * test sharding
 
-Note: since 0.4.x this repo requires bazel ≥ 0.4.4 to function (due to the use of
-BUILD.bazel files in bazelbuild/buildtools).
-
-Note: on master this repo requires bazel ≥ 0.4.5 to function (due to the use of
```
ianthehat: Might be worth keeping this comment and moving it to bazel 0.5 while you are there, as we will be checking in some of the producer switch straight after.
Or should we just generally have a note that says master is only guaranteed to work with the latest shipping bazel for future proofing?

PR #535, opened 2017-06-13
Gazelle: simplify error handling

jayconrod:
* Most errors in Gazelle are non-fatal. Most entry points in each
  package no longer return errors. Errors will generally be logged
  using log.Print, and a nil value or a partial success value will be
  returned.
* packageReader.warn is replaced by log.Print. There is no longer a
  meaningful difference between warnings and errors.
* In cases where errors could only occur due to bugs in Gazelle (i.e.,
  most of rules), log.Panic is used instead of log.Print.

### DIFF
--- a/go/tools/gazelle/merger/merger.go
+++ b/go/tools/gazelle/merger/merger.go
@@ -19,6 +19,7 @@ package merger
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"sort"
 	"strings"
 
@@ -30,14 +31,6 @@ const (
 	keep          = "# keep"           // marker in srcs or deps to tell gazelle to preserve.
 )
 
-type GazelleIgnoreError struct {
-	posn bzl.Position
-}
-
-func (e GazelleIgnoreError) Error() string {
-	return fmt.Sprintf("%s found on line %d", gazelleIgnore, e.posn.Line)
-}
-
 var (
 	mergeableFields = map[string]bool{
 		"srcs":    true,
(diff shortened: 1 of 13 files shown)


jayconrod:
Open question: should Gazelle return a non-zero error code if any errors were reported? If so, we'll need a new error logger that records whether any errors were reported.

go/tools/gazelle/merger/merger.go
```diff
@@ -47,26 +40,30 @@ var (
...
+		return nil
 	}
 
 	oldStmt := oldFile.Stmt
 	var newStmt []bzl.Expr
 	for _, s := range genFile.Stmt {
 		genRule, ok := s.(*bzl.CallExpr)
 		if !ok {
-			return nil, fmt.Errorf("got %v expected only CallExpr in %q", s, genFile.Path)
+			log.Panicf("got %v expected only CallExpr in %q", s, genFile.Path)
```
ianthehat: Why panic instead of fatal?

go/tools/gazelle/rules/construct.go
```diff
@@ -134,37 +116,28 @@ func newValue(val interface{}) (bzl.Expr, error) {
...
-		default:
-			return nil, fmt.Errorf("not implemented %T", val)
+			return &bzl.BinaryExpr{X: gen, Op: "+", Y: sel}
 		}
-
-	default:
-		return nil, fmt.Errorf("not implemented %T", val)
 	}
+
+	log.Panicf("type not supported: %T", val)
```
ianthehat: Another panic

go/tools/gazelle/merger/merger.go
jayconrod: This error should only happen due to a bug in `rules` or `packages`. It should not be triggered by user input. I'd like to use `Fatal` for unrecoverable user errors (e.g., invalid flags on the command line) and `panic` for programmer errors so we get a stack.

PR #538, opened 2017-06-14
In repository rules, wrap execute commands with "env -i".

### LINKED ISSUE #536: 0.5.0 doesn't work with $GOBIN env
When $GOBIN env is set, the binaries for rule 0.5.0 like fetch_repo and gazelle go into $GOBIN, but the go rule tries to find from `cache/.../external/io_bazel_rules_go_repository_tools/bin/fetch_repo` and fails:

```
ERROR: /home/jhahn/work/migrate2/io/datastore/BUILD:3:1: no such package '@com_google_cloud_go//datastore': Traceback (most recent call last):
	File "/home/jhahn/.cache/bazel/_bazel_jhahn/97b14c2b3573f5bc7394d4aa4b0b4fcf/external/io_bazel_rules_go/go/private/go_repository.bzl", line 45
		ctx.execute([ctx.path(ctx.attr._fetch_repo),...", <9 more arguments>])
	File "/home/jhahn/.cache/bazel/_bazel_jhahn/97b14c2b3573f5bc7394d4aa4b0b4fcf/external/io_bazel_rules_go/go/private/go_repository.bzl", line 46, in ctx.execute
		ctx.path(ctx.attr._fetch_repo)
Not a file: /home/jhahn/.cache/bazel/_bazel_jhahn/97b14c2b3573f5bc7394d4aa4b0b4fcf/external/io_bazel_rules_go_repository_tools/bin/fetch_repo and referenced by '//io/datastore:go_default_library'.
ERROR: Analysis of target '//io/datastore:go_default_library' failed; build aborted.
```

I think the rule should ignore $GOBIN env when compiling binaries.

Thanks,

jayconrod:
Fixes #536

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -42,14 +42,18 @@ def _go_repository_impl(ctx):
       fail("if vcs is specified, remote must also be")
     # TODO(yugui): support submodule?
     # c.f. https://www.bazel.io/versions/master/docs/be/workspace.html#git_repository.init_submodules
-    result = ctx.execute([
-        ctx.path(ctx.attr._fetch_repo),
-        '--dest', ctx.path(''),
-        '--remote', ctx.attr.remote,
-        '--rev', rev,
-        '--vcs', ctx.attr.vcs,
-        '--importpath', ctx.attr.importpath,
-    ])
+    result = env_execute(
+        ctx,
+        [
+            ctx.path(ctx.attr._fetch_repo),
+            '--dest', ctx.path(''),
+            '--remote', ctx.attr.remote,
+            '--rev', rev,
+            '--vcs', ctx.attr.vcs,
+            '--importpath', ctx.attr.importpath,
+        ],
+        environment = {"PATH": ctx.os.environ["PATH"]},  # to find git
+    )
     if result.return_code:
       fail("failed to fetch %s: %s" % (ctx.name, result.stderr))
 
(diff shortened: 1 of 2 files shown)


ecnovash:
@jayconrod: 

This patch caused VCS failures, as VCS that relies on special configs can no longer find their dot configs from `$HOME` as `env -i` stripped it off.

I think given the problem presented in #536, `env -u` with a short GO related env variable blacklist is more preferable, alternatively please add `$HOME` to environment.

Please let me know if I should raise another bug.

jayconrod:
@ecnovash Could you open a new issue with more details about what information is missing and what's failing? Maybe there's another way to do this.

We probably won't add `$HOME` to the environment. Bazel doesn't force repository rules to be hermetic, but they should be to the extent possible. If repository rules access configuration files in developers' home directories, that breaks hermeticity. Different people may have different settings and will get different build results. This could also cause problems for caching and remote execution.

ecnovash:
@jayconrod Will do. 

The thing is that we set proxy in the gitconfig to fetch certain repositories that are behind firewall.

We used to set `HTTP_PROXY` in environment as suggested by bazel [doc](https://docs.bazel.build/versions/master/external.html#using-proxies), and then moved to the config file under the builder account's home directory.  

Unfortunately, none of them works now after this commit.

jayconrod:
@ecnovash We can whitelist `HTTP_PROXY`. Would that work for you? I'm not sure what git recognizes exactly. A brief search shows it might be `http_proxy` and `https_proxy` (lower case), so we can add those, too.

PR #541, opened 2017-06-14
Change cgo rules to not copy so much

ianthehat:
Many of the copies were not needed, all the filtering ones.
This also removes 3 of the 4 directories that only existed to hold the copies.
_cgo_codegen was removed, it was not being used any more.
This also kills a bunch more uses of bash

### DIFF
--- a/go/private/cgo.bzl
+++ b/go/private/cgo.bzl
@@ -125,14 +125,13 @@ def cgo_library(name, srcs,
 
 def _cgo_filter_srcs_impl(ctx):
   go_toolchain = get_go_toolchain(ctx)
-  out_dir = ctx.attr.out_dir
   srcs = ctx.files.srcs
   outputs = []
   cmds = []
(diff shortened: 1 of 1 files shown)


ianthehat:
Jenkins test this please

go/private/cgo.bzl
```diff
@@ -161,47 +160,25 @@ _cgo_filter_srcs = rule(
...
-    ctx.action(
-        inputs = [f] + ctx.files.srcs,
-        outputs = outputs,
-        mnemonic = "CgoSelectFiles",
-        command = f.path,
-    )
-
+      outputs += [src]
+  if ctx.attr.at_least_one_file and not outputs:
+    empty_file = ctx.new_file("empty_file" + ctx.attr.exts[0])
```
ianthehat: This is all about to change again as I merge it down and replace it with go code...

PR #542, opened 2017-06-14
Don't include non-.go files as go_test srcs

pmcalpine:
Non .go files (example: foo_test.py) were being included as go_test
sources despite fileInfo.isTest's comment indicating that it should
never be true for non-go files.

This change moves the detection of isTest after the extCategory is
calculated and only sets isTest to true if category is not .

Tested:
* go test github.com/bazelbuild/rules_go/go/tools/gazelle/packages
* bazel test //...

### DIFF
--- a/go/tools/gazelle/packages/fileinfo.go
+++ b/go/tools/gazelle/packages/fileinfo.go
@@ -144,7 +144,7 @@ func fileNameInfo(dir, name string) fileInfo {
 	var goos, goarch string
 	l := strings.Split(name[:len(name)-len(ext)], "_")
 	if len(l) >= 2 && l[len(l)-1] == "test" {
-		isTest = true && category != unsupportedExt && category != ignoredExt
+		isTest = category == goExt
 		l = l[:len(l)-1]
 	}
 	switch {
--- a/go/tools/gazelle/packages/fileinfo.go
+++ b/go/tools/gazelle/packages/fileinfo.go
@@ -117,25 +117,6 @@ const (
 func fileNameInfo(dir, name string) fileInfo {
 	ext := path.Ext(name)
 
-	// Determine test, goos, and goarch. This is intended to match the logic
-	// in goodOSArchFile in go/build.
-	var isTest bool
-	var goos, goarch string
-	l := strings.Split(name[:len(name)-len(ext)], "_")
-	if len(l) >= 2 && l[len(l)-1] == "test" {
-		isTest = true
-		l = l[:len(l)-1]
-	}
-	switch {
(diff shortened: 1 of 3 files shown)


bazel-io:

go/tools/gazelle/packages/fileinfo.go
```diff
@@ -157,6 +138,25 @@ func fileNameInfo(dir, name string) fileInfo {
 		category = ignoredExt
 	}
 
+	// Determine test, goos, and goarch. This is intended to match the logic
+	// in goodOSArchFile in go/build.
+	var isTest bool
+	var goos, goarch string
+	l := strings.Split(name[:len(name)-len(ext)], "_")
+	if len(l) >= 2 && l[len(l)-1] == "test" {
+		isTest = true && category != unsupportedExt && category != ignoredExt
```
ianthehat: Not sure this is the right test, this would still allow any other category through, I think it should check that the category is exactly the go one.
I also think it would be easier to make the much smaller change of 
```isTest:   isTest && category == goExt```
down below
I'm not even convinced that the comment is how things should be, I think we should allow non go files to be marked as test and the go_test rule should be adding only files with the go category, but I will leave that decision to Jay

Comment (jayconrod) [inline go/tools/gazelle/packages/fileinfo.go]: I think this would be simpler as:

```
if category == goExt && len(l) >= 2 && l[len(l)-1] == "test" {
  isTest = true
}
```

+1 to the last thing Ian said though. Would be good for `Package.addFile` to have an early check that ignores files with `ignoredExt` and `unsupportedExt`.

File: go/tools/gazelle/packages/fileinfo.go [hunk 2]
```diff
@@ -157,6 +138,25 @@ func fileNameInfo(dir, name string) fileInfo {
 		category = ignoredExt
 	}
 
+	// Determine test, goos, and goarch. This is intended to match the logic
+	// in goodOSArchFile in go/build.
+	var isTest bool
+	var goos, goarch string
+	l := strings.Split(name[:len(name)-len(ext)], "_")
+	if len(l) >= 2 && l[len(l)-1] == "test" {
+		isTest = category == goExt
```
jayconrod: I updated the test to check goExt, which is the only extension that can be part of a `go_test` (per logic in `go/build`). In `goodOSArchFile`, the `test` suffix is recognized and removed from all files before checking for GOOS and GOARCH suffixes, which still apply

jayconrod:
Jenkins test this please

pmcalpine:
Thanks you beat me to it! :)

PR #549, opened 2017-06-15
Gazelle: generate BUILD files for "testdata" directories containing code

### LINKED ISSUE #415: Gazelle should recognize testdata directories that contain code
`golang.org/x/crypto/ssh/testdata` is a Go package that contains Go code. Gazelle treats this as a directory containing data files, which breaks tests that depend on it as a normal package.

Gazelle should check whether testdata directories contain Go code. If they do, it should treat them as normal directories. If they don't, it should continue treating them as data.

jayconrod:
Gazelle will now recurse into "testdata" directories and generate
BUILD files if there are buildable .go files present.

Generated go_test rules will only have 'data = glob(["testdata/**"])'
if a "testdata" directory exists and does not contain .go files or
BUILD or BUILD.bazel files in any subdirectory.

Fixes #415

### DIFF
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -19,7 +19,6 @@ import (
 	"go/build"
 	"io/ioutil"
 	"log"
-	"os"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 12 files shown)


go/tools/gazelle/packages/walk.go
```diff
@@ -43,51 +42,57 @@ type WalkFunc func(pkg *Package)
...
+		hasTestdata := false
+		for _, f := range files {
+			base := f.Name()
+			if f.IsDir() && base != "" && base[0] != '.' {
+				hasPackage := visit(filepath.Join(path, base))
+				if base == "testdata" {
+					hasTestdata = !hasPackage
+				}
+				subdirHasPackage = subdirHasPackage || hasPackage
+			} else if !f.IsDir() && (base == "BUILD" || base == "BUILD.bazel") {
```
ianthehat: Should this use validBuildFileNameSlice?

jayconrod: Yes. I'll push a separate change that makes this available to packages.

PR #550, opened 2017-06-15
A better version of cgo_genrule

ianthehat:


mikedanese:
This doesn't break kubernetes. LGTM.

PR #553, opened 2017-06-16
Gazelle: refactor configuration plumbing

jayconrod:
* Added a new "config" package with a "Config" struct that contains
  most of the configuration information that Gazelle gets from flags
  (go prefix, repo root, etc.). Defaults and utility methods are also
  moved to this package.
* Added newConfiguration, a function which builds this
  configuration. Tags are preprocessed, and paths are canonicalized so
  we don't have to worry about that later.
* Functions which previously accepted arguments that are found in
  Config now just accept *config.Config.
* Deleted packageReader, since it had a very similar purpose.

### DIFF
--- a/go/tools/gazelle/config/config.go
+++ b/go/tools/gazelle/config/config.go
@@ -1,0 +1,128 @@
+/* Copyright 2017 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"go/build"
+)
+
+// Config holds information about how Gazelle should run. This is mostly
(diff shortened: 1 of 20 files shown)


go/tools/gazelle/config/config.go
```diff
@@ -0,0 +1,128 @@
...
+	// ExternalMode indicates imports should be resolved to external dependencies
+	// (declared in WORKSPACE).
+	ExternalMode DependencyMode = iota
+
+	// VendorMode indicates imports should be resolved to libraries in the
+	// vendor directory.
+	VendorMode
+)
+
+func DependencyModeFromString(s string) (DependencyMode, error) {
```
ianthehat: Public function with no docstring

go/tools/gazelle/gazelle/main.go
```diff
@@ -144,23 +115,7 @@ func main() {
...
-		if *goPrefix, err = loadGoPrefix(*repoRoot); err != nil {
-			if !os.IsNotExist(err) {
-				log.Fatal(err)
-			}
-			log.Fatalf("-go_prefix not set and no root BUILD file found")
-		}
-	}
-
-	genericTags, err := parseBuildTags(*buildTags)
+	c, err := newConfiguration()
```
ianthehat: I would change newConfiguration to take *flag.FlagSet, and pass in flag.CommandLine here so that access to the global is isolated to this point only.
Opens the door to change to a locally declared flagset in the future rather than global flags.

jayconrod:
Thanks! I'll address these comments in a follow-up PR. I need to merge #549 with this one in order to address feedback there, and it would be good if Travis can run in parallel.

PR #554, opened 2017-06-16
Fix go_test chdir to package directory

### LINKED ISSUE #540: go_test rules in remote repositories fail to change directory
`go_test` rules that are declared in a subdirectory of an external repository do not successfully change directory at the beginning of the test. This prevents them from accessing data files using paths relative to the test package directory. This is why many of the tests in popular_repos.bash are failing.

jayconrod:
The main go file generated by go_test includes a call to os.Chdir to
change to the package directory so that data files may be accessed by
package-relative paths. This is the same behavior as "go test".

The chdir now includes the workspace directory, which is necessary
for tests in remote workspaces. We also detect errors when changing
directories.

Fixes #540

### DIFF
--- a/go/private/cgo.bzl
+++ b/go/private/cgo.bzl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@io_bazel_rules_go//go/private:common.bzl", "get_go_toolchain", "emit_generate_params_action", "go_filetype", "cgo_filetype", "cc_hdr_filetype", "hdr_exts")
+load("@io_bazel_rules_go//go/private:common.bzl", "get_go_toolchain", "emit_generate_params_action", "go_filetype", "cgo_filetype", "cc_hdr_filetype", "hdr_exts", "pkg_dir")
 load("@io_bazel_rules_go//go/private:library.bzl", "go_library")
 load("@io_bazel_rules_go//go/private:binary.bzl", "c_linker_options")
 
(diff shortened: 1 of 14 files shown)


ixdy:
This change seems to have broken the kubernetes e2e tests, and I'm stumped on a proper fix.
We do things pretty weirdly, however:
* we use the [Ginkgo](https://github.com/onsi/ginkgo) framework and test runner
* we actually run tests outside bazel (largely for legacy reasons), and we precompile the tests into a binary which we pass to `ginkgo`.
* outside bazel, we build the test binary using `go test -c`
* inside bazel, we have a [genrule](https://github.com/kubernetes/kubernetes/blob/60007128032869eadb44a831e53834384716db80/test/e2e/BUILD#L132-L141) which copies the `go_default_test` binary to a binary named `e2e.test`.

And then we run roughly `ginkgo bazel-bin/test/e2e/e2e.test`, and things used to work.

With this change, we now get this error message:
```
Jul 28 16:07:47.682: INFO: Overriding default scale value of zero to 1
Jul 28 16:07:47.682: INFO: Overriding default milliseconds value of zero to 5000
2017-07-28 16:07:47.733763 I | could not change to test directory: chdir test/e2e: no such file or directory

Ginkgo ran 1 suite in 233.960091ms
Test Suite Failed
```

I know we're doing like 10 things wrong here, but any suggestions on the easiest way to get around this?
I'm not even sure where it's running that it can't chdir into a `test/e2e` directory.

ixdy:
Hm, so I'm still not entirely sure how `ginkgo` plays into this, but this seems to break the kubernetes workflow, in which we use a compiled test binary separate from the kubernetes source to run e2e tests.

This now fails because the test package directory doesn't exist.

Maybe there needs to be a way to get a prebuilt test binary out of the go rules?

ixdy:
FYI I took discussion of this to #670

PR #559, opened 2017-06-20
Full bazel tests from inside bazel

ianthehat:
This makes a new system using bazel rules and sh_test to run bazel tests from
inside bazel.
This moves only cgo_library_root_dir as a prototype, but it should hold for all
the others.

To run the same tests as before do any of:
    bazel test --test_tag_filters=-bazel //...
To only run the "non_bazel_tests"
    bazel test --test_tag_filters=bazel //...
To run all the tests you ought to be running
    bazel test //...

Better method of running tests when trying to work fast
    bazel test --test_size_filters=-large //...

Even better methods of running tests when trying to work fast that are wrong at
the moment because we are not being disciplined about our test tagging
    bazel test --test_size_filters=small //...
    bazel test --test_timeout_filters=short //...


tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
+_script_template = """
+ANCHOR=$(readlink -f {ANCHOR})
+BASE=$(dirname $ANCHOR)
+CACHE_FILE=$(realpath {CACHE})
+BAZEL_CACHE=$(dirname $CACHE_FILE)
+cd {PACKAGE}
+bazel {ARGS} {TARGET}
```
jayconrod: Most of these template variable expansions should be quoted, but be careful with `ARGS`. Might run into trouble if there's a space above wherever rules_go is checked out.

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+      anchor = anchor,
+      cache = cache,
+      go_version = go_version,
+  )
+  native.sh_test(
+      name = name,
+      size = "large",
+      timeout="short",
+      srcs = [script_name],
+      tags = ["local", "bazel"] + tags,
```
jayconrod: What will use the "bazel" tag?

tests/bazel_tests.bzl
ianthehat: All changed, but also done.

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+  ))
+  # Build the bazel startup args
+  args = []
+  if ctx.attr.batch:
+    args += ["--batch"]
+  # Add the command and any command specific args
+  args += [ctx.attr.command]
+  if ctx.attr.command == "build":
+    args += _build_args
+  elif ctx.attr.command == "test":
```
ianthehat: done

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+  return struct(
+    files = depset([script]),
+    runfiles = ctx.runfiles([workspace])
+  )
+
+_bazel_test_script = rule(
+    _bazel_test_script_impl,
+    attrs = {
+        "batch": attr.bool(default=True),
+        "command": attr.string(mandatory=True),
```
ianthehat: done

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+        "args": attr.string_list(default=[]),
+        "target": attr.string(),
```
ianthehat: done

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+        "anchor": attr.label(allow_files=True, single_file=True),
+        "cache": attr.label(allow_files=True, single_file=True),
```
ianthehat: they no longer exist.
They (and the new equivalent) are not implicit because I want to make sure I use the same value in two places, and thus are explicitly passed.
The tests have no control over the value anyway, they are supplied by the wrapper function to the hidden rule.

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+        "go_version": attr.string(),
+    }
+)
+
+def bazel_test(name, batch = None, command = None, args=None, target = None, go_version = None, tags=[]):
+  script_name = name+"_script"
+  anchor = "//:README.md"
```
ianthehat: Done

tests/bazel_tests.bzl
```diff
@@ -0,0 +1,132 @@
...
+      args = args,
+      target = target,
+      anchor = anchor,
+      cache = cache,
+      go_version = go_version,
+  )
+  native.sh_test(
+      name = name,
+      size = "large",
+      timeout="short",
```
ianthehat: done

tests/bazel_tests.bzl
ianthehat: It's so you can filter test that invoke bazel easily

go/BUILD
```diff
@@ -0,0 +1,5 @@
+filegroup(
+  name = "all_rules",
```
ianthehat: done

tests/BUILD
```diff
@@ -0,0 +1,13 @@
+load(":bazel_tests.bzl", "md5_sum")
+
+md5_sum(
+  name = "all_rules",
```
ianthehat: done

tests/BUILD
```diff
@@ -0,0 +1,13 @@
...
+  srcs = ["//go:all_rules"],
+  visibility = ["//visibility:private"],
+)
+
+filegroup(
```
ianthehat: nope, and for no good reason....
For some reason using the md5 rule directly in a runtime data list does not work, but wrapping it in a filegroup first does.

tests/bazel_tests.bzl
```diff
@@ -46,57 +31,66 @@ def _bazel_test_script_impl(ctx):
...
 
-  ctx.file_action(output=script, executable=True, content=_script_template.format(
-    ANCHOR = ctx.file.anchor.path,
-    CACHE = ctx.file.cache.path,
-    PACKAGE = ctx.label.package,
-    ARGS = " ".join(args),
-    TARGET = ctx.attr.target,
-  ))
+  for ext in ctx.files.externals:
+    root = ext.owner.workspace_root
```
jayconrod: Might be simpler to iterate over `ctx.attr.externals` and access `ext.label`.

tests/bazel_tests.bzl
```diff
@@ -114,19 +108,36 @@ md5_sum = rule(
     attrs = { "srcs": attr.label_list(allow_files=True) },
 )
 
-def _test_cache_impl(ctx):
+def _test_environment_impl(ctx):
+  execroot, _, ws = str(ctx.path(".")).rpartition("/external/")
+  if not ctx.name == ws:
```
jayconrod: `ctx.name != ws`

PR #565, opened 2017-06-21
Gazelle: kill generator package

jayconrod:
* Removed the generator package. Its remaining functionality is
  distributed to rules and main.
* rules.Generator.Generate now returns *bzl.File instead of
  []bzl.Expr. The rules package will add load and go_prefix statements
  as appropriate.
* main.run now calls packages.Walk and rules.Generator.Generate
  directly.
* Gazelle will no longer update BUILD files in the repository root
  directory unless it was one of the directories Gazelle traversed.


go/tools/gazelle/gazelle/main.go
```diff
@@ -43,35 +44,63 @@ var modeFromName = map[string]emitFunc{
...
+	if err != nil {
+		log.Print(err)
+		return
+	}
+	oldFile, err := bzl.Parse(oldPath, oldData)
+	if err != nil {
+		log.Print(err)
+		return
+	}
+	mergedFile := merger.MergeWithExisting(genFile, oldFile)
```
swsnider: This change now causes Gazelle to segfault whenever there's a BUILD file it processes with `# gazelle:ignore` at the top level, since this causes you to pass a nil to buildifier, which is expecting a real file object.

jayconrod: Thanks for pointing this out, and sorry for the break! PR #594 will fix this.

PR #568, opened 2017-06-22
Strict dependency checking

### LINKED ISSUE #566: go_library should fail when a direct dependency is missing
Suppose we have this BUILD file:

```

go_library(
    name = "go_default_library",
    srcs = ["lib.go"],
    deps = [":a"],
)

go_library(
    name = "a",
    srcs = ["a.go"],
    deps = [":b"],
)

go_library(
    name = "b",
    srcs = ["b.go"],
)
```

If lib.go imports both "a" and "b", Bazel should report an error because the dependency on "b" wasn't explicitly specified in `deps`. Currently, no error is reported. Transitive dependencies must be available to the compiler, since we need type information, so the compiler won't be able to check this.

jayconrod:
go_library, go_binary, and go_test now check that their sources don't
import any packages (other than standard ones) that are not provided
by direct dependencies (including dependencies from the "library"
attribute).

Fixes #566


go/tools/builders/compile.go
```diff
@@ -19,30 +19,45 @@ package main
...
 	gotool := args[0]
 	args = args[1:]
 	sources := []string{}
 	goopts := []string{}
+	deps := []string{}
 	bctx := build.Default
 	bctx.CgoEnabled = true
+	depNext := false
+argsLoop:
 	for i, s := range args {
```
ianthehat: This has become too much like custom flag parsing for my liking
How about just use the flags package, the -- will force it to stop processing args anyway, so we can consume flags.Args() into the goops easily.

You will want something like the below to allow you to consume mutiple -dep args into a single list, and maybe do the same for srcs, probably put it in a separate flags.go so we can can use the same approach in all the builders.

```
// multiFlag allows repeated string flags to be collected into a slice
type multiFlag struct {
	values *[]string
}

func (m multiFlag) String() string {
	if len(*m.values) == 0 {
		return ""
	}
	return fmt.Sprint(*m.values)
}

func (m multiFlag) Set(v string) error {
	(*m.values) = append(*m.values, v)
	return nil
}
```

jayconrod: Good idea. I reused multiFlag in flags.go, which is already used for the linker.

ianthehat: That's so much nicer, thanks for doing that.
It's nicer even without the -deps than it was before, I will convert all the other builders to the same method I think.

jayconrod:
@ianthehat I've migrated the integration test to `bazel_test`. It required modifying the test script template to record the return code. Not sure if it's the best approach, but I think it will work for gc_opts_unsafe.bash. PTAL.

Issue #557, opened 2017-06-19
Gazelle: incorporate generated .go sources into rules


jayconrod:
A key strength of Bazel is its ability to handle generated code. Gazelle should make it easy to take advantage of this. If Gazelle can determine that an existing rule generates .go files, it should incorporate those files into `go_library` and `go_test` rules in the same package.

```bzl
genrule(
    name = "gen_parser",
    srcs = ["gopher.y"],
    outs = ["gopher.go"],
    cmd = ["go tool yacc -o gopher.go -p parser gopher.y"],
)

# gazelle could generate this rule
go_library(
    name = "go_default_library",
    srcs = ["gopher.go"],
)
```

### Detecting generated code

Gazelle generates rules by scanning source files in a directory. Gazelle could use existing BUILD files as another source of information. It should look for files with .go suffixes in `out` and `outs` attributes of top-level rules. This would cover `genrule` and `go_embed_data` when it's supported (#304). It would also cover `_go_proto_library_gen` if we ever made that a standalone public rule.

There are some limitations to this approach. We won't be able to detect .go files created implicitly by rules or inside macros unless we run `bazel query`, which would be very slow, especially if it needs to download additional dependencies. We won't be able to detect sources created with `go generate`, since generate comments don't list any inputs, outputs, or dependencies.

### Incorporating generated code

Gazelle uses `packages.Package` to represent the sources and dependencies for a package. Gazelle should add detected files to `Package` before rules are generated from it. Generated files should be treated as if they have no imports or build tags, since we cannot infer those without actually generating them. We should respect filename suffixes though: `foo_linux_test.go` should be treated as a Linux-specific test file.

### Excluding existing files

Because `go generate` is not part of `go build`, it's common for people to check in generated files so that `go get` Just Works. This is not necessary with Bazel, but we need to avoid any problems caused by this.

If a rule generates a file with the same name as an existing file, Bazel will use the generated file at build time as if the existing file didn't exist. Gazelle should probably incorporate imports and build tags from the existing file. This is unclear, since the file won't be seen at build time, but the programmer's intent is most likely for the generated and existing files to be equivalent.

[...]

PR #577, opened 2017-06-26
Gazelle: parse BUILD files in packages.Walk

jayconrod:
This is in preparation for increased awareness of generated
code. packages.Walk now discovers and parses existing BUILD files. A
*bzl.File is passed to the callback if one was found.

Gazelle will need to read BUILD files in order to build
packages.Package structures for two reasons:

1. Gazelle will discover generated files by examining attributes of
rules in existing BUILD files.
2. Some files can be excluded using comments in BUILD files.

Related #557

### DIFF
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -23,16 +23,18 @@ import (
 	"path/filepath"
 	"strings"
 
+	bzl "github.com/bazelbuild/buildtools/build"
 	"github.com/bazelbuild/rules_go/go/tools/gazelle/config"
 )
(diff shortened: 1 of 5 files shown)


go/tools/gazelle/packages/walk.go
```diff
@@ -23,16 +23,18 @@ import (
 	"path/filepath"
 	"strings"
 
+	bzl "github.com/bazelbuild/buildtools/build"
```
ianthehat: I think this name is probably going to end up in confusion, bzl is a different dialect of skylark to BUILD files.

PR #573, opened 2017-06-23
Switching from new_go_repository to go_repository

ianthehat:
Also adds a print to the compatibility shim to warn people to switch, and changes the documentation to not refer to new_go_repository any more.

PR #580, opened 2017-06-26
wtool: add go_repository rules instead of new_go_repository

jayconrod:
Also, automatically load go_repository if it isn't loaded already.


jayconrod:
Oops, I forgot #573 updated wtool, too. Why don't you land that one first, and I'll fix any merge conflicts that come up here (I added a word to a comment, so there will be a conflict there).

PR #581, opened 2017-06-26
Gazelle: be aware of generated and excluded files

### LINKED ISSUE #557: Gazelle: incorporate generated .go sources into rules

jayconrod:
Gazelle now detects files generated inside rules of existing BUILD
files and incorporates them into the rules it generates. For example,
if a genrule lists "foo.go" in its "outs" attribute, Gazelle will add
"foo.go" to the go_library it generates.

Files may be excluded by adding a comment of the form:
"# gazelle:exclude foo.go"

The exclusion applies both to existing files and generated files.

Fixes #557

### DIFF
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -60,10 +60,11 @@ func Walk(c *config.Config, dir string, f WalkFunc) {
 		haveError := false
 		for _, base := range c.ValidBuildFileNames {
 			oldPath := filepath.Join(path, base)
-			oldData, err := ioutil.ReadFile(oldPath)
-			if os.IsNotExist(err) {
+			st, err := os.Stat(oldPath)
+			if os.IsNotExist(err) || err == nil && st.IsDir() {
 				continue
 			}
+			oldData, err := ioutil.ReadFile(oldPath)
 			if err != nil {
 				log.Print(err)
 				haveError = true
--- a/go/tools/gazelle/packages/walk.go
+++ b/go/tools/gazelle/packages/walk.go
@@ -19,6 +19,7 @@ import (
 	"go/build"
 	"io/ioutil"
 	"log"
+	"os"
 	"path"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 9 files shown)


go/tools/gazelle/packages/walk.go
```diff
@@ -126,19 +127,29 @@ func findPackage(c *config.Config, dir string, oldFile *bzl.File, hasTestdata bo
...
 	// List the files in the directory and split into .go files and other files.
 	// We need to process the Go files first to determine which package we'll
 	// generate rules for if there are multiple packages.
+	var goFiles, otherFiles []string
 	files, err := ioutil.ReadDir(dir)
 	if err != nil {
 		log.Print(err)
 		return nil
 	}
 	for _, file := range files {
```
ianthehat: I feel like this loop ought to be merged with packages.Walk for a couple of reasons.
First, speed, stating files is expensive.
Second, it would be nice if 
```# gazelle:exclude subdir```
could work to prevent gazelle from descending into a directory, without replicating the parsing an a second place.

jayconrod: Done, sounds good. I like the idea of being able to exclude subdirectories. May be useful for x/tools in the future.

PR #591, opened 2017-06-28
Only pass direct targets to compile

ianthehat:
They have all the transitive data they need inside them.
This should make complex builds considerably faster, although we have no useful
benchmark to prove it.
Also cleaned up the compile argument handling while I was in there.

### DIFF
--- a/go/private/library.bzl
+++ b/go/private/library.bzl
@@ -53,22 +53,26 @@ def emit_library_actions(ctx, sources, deps, cgo_object, library):
   lib_name = importpath + ".a"
   out_lib = ctx.new_file(lib_name)
   out_object = ctx.new_file(ctx.label.name + ".o")
-  search_path = out_lib.path[:-len(lib_name)]
+  searchpath = out_lib.path[:-len(lib_name)]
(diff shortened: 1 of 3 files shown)


go/private/library.bzl
```diff
@@ -218,7 +226,7 @@ def emit_go_compile_action(ctx, sources, libs, lib_paths, direct_paths, out_obje
     args += ["-src", src]
   for dep in direct_paths:
     args += ["-dep", dep]
-  args += ["--", "-o", out_object.path, "-trimpath", ".", "-I", "."]
+  args += ["-o", out_object.path, "-trimpath", ".", "-I", "."]
```
jayconrod: I think the Travis failure is due to removing "--" here. That needs to come before `gc_goopts` to prevent those options from being parsed.

ianthehat: Yeah. I had already spotted that.
I was trying to avoid the -- in favour of compile actually understanding the command line so it can really deal with any args it sees. For instance if someone puts a -I in the gc_goopts it also needs to be abs'd, but for now....

PR #594, opened 2017-06-29
Gazelle: check for ignored files before emitting

jayconrod:
merger.MergeWithExisting returns nil when it finds a
"# gazelle:ignore" directive. run must check for this before
formatting and writing the file.


jayconrod:
Thanks to @swsnider for pointing this out. This is a case where the unit tests pass, but we didn't write enough integration tests.

swsnider:
Thanks!

PR #596, opened 2017-06-30
Get directories from parsed FlagSet

ains:
The directories here should come from the flagSet we have parsed from the passed in args and not from the global Flag.args()


bazel-io:

jayconrod:
Thanks for fixing this!

Jenkins test this please.

PR #599, opened 2017-06-30
Start adding integration tests for Gazelle

jayconrod:
* Added config.Logger which should be called instead of log.Print and
  log.Fatal. Tests can override methods in this to trap errors.
* Added integration_test.go with a couple small tests. More to come.


go/tools/gazelle/config/config.go
```diff
@@ -47,6 +47,10 @@ type Config struct {
 
 	// DepMode determines how imports outside of GoPrefix are resolved.
 	DepMode DependencyMode
+
+	// Log is used to print errors and warnings to the user. It should be used
```
ianthehat: I am not sure what this gets you over just using
logs := &bytes.Buffer{}
log.SetOutput(logs)
in the tests?

jayconrod: Reverted

Issue #486, opened 2017-05-27
Support vendor dirs not in root of the repo


tomwilkie:
Gazelle expects the vendor directory to be in the root of the repo.  This isn't always the case.  Would it be possible to pass the vendor directory location?

jayconrod:
Seems like something we should support.

I'm currently wrapping up some improvements for handling build constraints in Gazelle. Once that work is done, I'll be focusing on improving Gazelle's management of dependencies. This will fit into that project.

pwittrock:
Any update on this?  I need to run carefully crafted `sed` scripts after running gazelle.

the equivalent of `go list -f '{{ join .Deps  "\n"}}' | grep` might be able to provide this

jayconrod:
#859 will solve this.

At the moment, Gazelle converts import paths to Bazel deps just by transforming the string without checking whether something is there or not. That prevents us from using vendor directories that aren't at the root of the repository.

#859 is a proposal to index all of the packages in a repository by import path, then choose something appropriate. I'm planning to work on that as soon as we ship support for proto rules (#868). It should be in the major release after the current one we're about to tag, hopefully in a month or so.

jayconrod:
This should be working now.

Issue #194, opened 2016-11-14
new_go_repository: vendored code interfering with non-vendored


benley:
I'm trying to build https://github.com/golang/gddo using new_go_repository, and I used [wtool](https://github.com/bazelbuild/rules_go/tree/master/go/tools/wtool) to generate the various new_go_repository [WORKSPACE entries](https://gist.github.com/benley/f2aeb9de8eedffda0e9cbc2a219ac0dd) for it.

It appears that gddo has all its dependencies vendored into the gddo repo, which isn't too uncommon.  Unfortunately it looks like gazelle is building the vendored-in copy of various things, but translating those libraries' internal dependencies to point at the non-vendored copy of the same library, which may not even be at the same version.  The result is some strange `visibility` errors:

```
$ bazel build @com_github_golang_gddo//...
ERROR: /private/var/tmp/_bazel_bstaffin/26abb02a80899f40a981f66c2d82ca06/external/com_github_golang_gddo/vendor/golang.org/x/net/trace/BUILD:3:1: Target '@org_golang_x_net//internal/timeseries:go_default_library' is not visible from target '@com_github_golang_gddo//vendor/golang.org/x/net/trace:go_default_library'. Check the visibility declaration of the former target if you think the dependency is legitimate.
ERROR: Analysis of target '@com_github_golang_gddo//vendor/golang.org/x/net/trace:go_default_library' failed; build aborted.
INFO: Elapsed time: 0.089s
```

Is there a known workaround for this?

pmbethe09:
I think you have hit a feature of gazelle, directories called "internal"
get a more restrictive visibility:

https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/rules/generator.go#L258

Probably this feature should be turned off for gazelle running under
new_go_repository

GinFungYJF:
Has this feature been turned off for gazelle running under new_go_repository now?

jayconrod:
This should be working now. Gazelle now resolves import paths using all libraries in a repository, including in vendor directories (subject to vendoring visibility). external / vendored mode only applies to libraries that can't be resolved this way.

Issue #114, opened 2016-09-22
gazelle vendor support


hochhaus:
Is vendoring automatically supported when using gazelle/`new_go_repository()`?

As a specific example, when using `new_go_repository()` with a repository that uses vendoring (eg: https://github.com/rsc/letsencrypt) I see that gazelle creates BUILD files for the vendored repositories.

For example, adding the following to my WORKSPACE:

```
new_go_repository(
    name = "io_rsc_letsencrypt",
    commit = "a18c646c3d0772313b7b53c78847bb3eb6463f0b",
    importpath = "rsc.io/letsencrypt",
)
```

generates the following BUILD files:

```
ahochhaus@ahochhaus-pc:~/ll$ find -L bazel-* -name BUILD | grep letsen
bazel-ll/external/io_rsc_letsencrypt/BUILD
bazel-ll/external/io_rsc_letsencrypt/vendor/github.com/xenolf/lego/acme/BUILD
```

However when I build, the vendored repository (`com_github_xenolf_lego`) can't be found.

pcj:
(chiming in)

There would have to be a `WORKSPACE` file written at `$(bazel info output_base)/external/com_github_xenolf_lego` for this to work.  Does that file exist?  Otherwise I'd think it one would have to refer to it as `@io_rsc_letsencrypt//vendor/github.com/xenolf/lego/acme:go_default_library`.

pmbethe09:
So running gazelle creates deps by using a consistent conversion string
from Go import-path to a workspace-name.
That is true both for manual local gazelle, and gazelle underneath a
new_go_repository.

However in both cases that assumes you have all of the right things in your
WORKSPACE file.

Presumably what we need is either a stand-alone "workzelle", or something
running as part of local invocation of 'gazelle -workspace=fix ...'
This would collect all of the remote-repo deps as referenced by gazelle,
and add them to your WORKSPACE, as well as look at the transitive
dependencies.
Note: this is because bazel require a flattening of all transitive
dependencies into your local WORKSPACE file -- it doesn't read WORKSPACE
files read from any remote repos.

I have started working on such a thing, but want to touch base with @yugui
to plan

yugui:
> I have started working on such a thing, but want to touch base with @yugui to plan

Both of `workzelle` and `-workspace=fix` sound good for me. But probably `workzelle` is simpler.

[...]

Issue #62, opened 2016-08-04
Using vendored libraries that depend on other libraries


lingz:
So as with your approach to vendoring, I manually created a BUILD file at my //goroot/src/github.com/etc location and include all the *.go files in the dependency. However, as the dependency imports other packages (also in the form of github.com), it will throw an "can't find import:" during compilation.

Is there any way to use vendored packages like this, which might have a long chain of dependencies?

pmbethe09:
Due to the diamond-dependency problem, bazel has chosen not to load
dependencies of dependencies.  Instead, in your WORKSPACE file, you will
need to flatten and provide all of the transitive dependencies.

lingz:
Are there any tools that can partially or fully automate this? I read about
glaze in some of these issues but it seems like there's no timeline on
releasing that.

jayconrod:
Closing as (mostly) obsolete. [gazelle](https://github.com/bazelbuild/bazel-gazelle) can resolve dependencies, both vendoring an external. It can also import external repositories from dep.

PR #601, opened 2017-07-03
Adds go_repository options for using vendored code

jmillikin-stripe:
There are two commits here:

First, I've added a `--vendor_prefix` flag to Gazelle. Setting this flag lets the caller adjust where Gazelle looks for vendored dependencies, for cases where the repository owner didn't use the default `vendor/` location.

Second, I added attributes to the `go_repository()` rule to let Gazelle use vendored dependencies in external repositories. I understand this is not quite the preferred approach of having everything in `WORKSPACE`, but it makes life significantly easier when our build wants to depend on a single `go_binary()` that itself requires ~dozens of vendored dependencies.

The motivating use case for us is bundling a compiled-from-Git etcd (https://github.com/coreos/etcd) into some Docker images built with Bazel's docker rules. With these changes, it's possible to depend on the etcd binary from our build without having to maintain a permanent "etcd plus BUILD rules" fork.

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -73,6 +73,10 @@ def _go_repository_impl(ctx):
             "--build_tags", ",".join(ctx.attr.build_tags)]
     if ctx.attr.build_file_name:
         cmds += ["--build_file_name", ctx.attr.build_file_name]
+    if ctx.attr.build_vendor:
+        cmds += ["--external", "vendored"]
+        if ctx.attr.build_vendor_prefix:
+            cmds += ["--vendor_prefix", ctx.attr.build_vendor_prefix]
     cmds += [ctx.path('')]
     result = env_execute(ctx, cmds)
     if result.return_code:
@@ -102,6 +106,11 @@ go_repository = repository_rule(
         "build_file_name": attr.string(default="BUILD.bazel,BUILD"),
         "build_file_generation": attr.string(default="auto", values=["on", "auto", "off"]),
         "build_tags": attr.string_list(),
+
+        # Attributes for a repository that uses build file generation with
+        # vendored dependencies.
+        "build_vendor": attr.bool(default=False),
+        "build_vendor_prefix": attr.string(default="vendor/"),
 
         # Hidden attributes for tool dependancies
         "_fetch_repo": attr.label(
--- a/go/tools/gazelle/config/config.go
+++ b/go/tools/gazelle/config/config.go
@@ -47,6 +47,10 @@ type Config struct {
 
 	// DepMode determines how imports outside of GoPrefix are resolved.
 	DepMode DependencyMode
+
+	// VendorPrefix is the import prefix for vendored dependencies. It will
+	// usually be "vendor/".
+	VendorPrefix string
 }
 
 var DefaultValidBuildFileNames = []string{"BUILD.bazel", "BUILD"}
--- a/go/tools/gazelle/gazelle/main.go
+++ b/go/tools/gazelle/gazelle/main.go
@@ -162,6 +162,7 @@ func newConfiguration(args []string) (*config.Config, emitFunc, error) {
 	external := fs.String("external", "external", "external: resolve external packages with go_repository\n\tvendored: resolve external packages as packages in vendor/")
 	goPrefix := fs.String("go_prefix", "", "go_prefix of the target workspace")
 	repoRoot := fs.String("repo_root", "", "path to a directory which corresponds to go_prefix, otherwise gazelle searches for it.")
+	vendorPrefix := fs.String("vendor_prefix", "vendor/", "import prefix for vendored dependencies.")
 	mode := fs.String("mode", "fix", "print: prints all of the updated BUILD files\n\tfix: rewrites all of the BUILD files in place\n\tdiff: computes the rewrite but then just does a diff")
 	if err := fs.Parse(args); err != nil {
 		if err == flag.ErrHelp {
@@ -232,6 +233,8 @@ func newConfiguration(args []string) (*config.Config, emitFunc, error) {
 			return nil, nil, fmt.Errorf("-go_prefix not set and not root BUILD file found")
 		}
 	}
+
+	c.VendorPrefix = *vendorPrefix
 
 	c.DepMode, err = config.DependencyModeFromString(*external)
 	if err != nil {
--- a/go/tools/gazelle/rules/generator.go
+++ b/go/tools/gazelle/rules/generator.go
@@ -68,7 +68,7 @@ func NewGenerator(c *config.Config) Generator {
 	case config.ExternalMode:
 		e = newExternalResolver()
 	case config.VendorMode:
-		e = vendoredResolver{}
+		e = newVendoredResolver(c)
 	default:
 		return nil
 	}
--- a/go/tools/gazelle/rules/resolve_vendored.go
+++ b/go/tools/gazelle/rules/resolve_vendored.go
@@ -1,11 +1,23 @@
 package rules
 
+import (
+	"github.com/bazelbuild/rules_go/go/tools/gazelle/config"
+)
+
 // vendoredResolver resolves external packages as packages in vendor/.
-type vendoredResolver struct{}
+type vendoredResolver struct {
+	prefix string
+}
+
+var _ labelResolver = (*vendoredResolver)(nil)
+
+func newVendoredResolver(c *config.Config) *vendoredResolver {
+	return &vendoredResolver{prefix: c.VendorPrefix}
+}
 
 func (v vendoredResolver) resolve(importpath, dir string) (label, error) {
 	return label{
-		pkg:  "vendor/" + importpath,
+		pkg:  v.prefix + importpath,
 		name: defaultLibName,
 	}, nil
 }


bazel-io:

jmillikin-stripe:
Related issues:
* https://github.com/bazelbuild/rules_go/issues/62 ("Using vendored libraries that depend on other libraries")
* https://github.com/bazelbuild/rules_go/issues/194 ("new_go_repository: vendored code interfering with non-vendored")
* https://github.com/bazelbuild/rules_go/issues/114 ("gazelle vendor support")
* https://github.com/bazelbuild/rules_go/issues/486 ("Support vendor dirs not in root of the repo")

jayconrod:
Could you explain more about how this would be used for etcd? I see there is a cmd/vendor directory. If I understand correctly, you'd like to be able to add etcd as an external repository in some other workspace, and you'd like the generated BUILD files to respect vendor directories that aren't at the top level.

Rather than adding a new command line argument to Gazelle, I think it would be better for Gazelle to detect and use vendor directories at any place in the tree automatically (#486). I don't think we should support vendoring in directories named something other than `vendor` though. We should also be careful not to let vendored packages be visible outside of the subtree that contains `vendor`. Basically, we want Gazelle to generate BUILD files that match "go build" conventions.

The main deficiency of Gazelle is that it currently does a fairly simple transformation from import paths to labels without actually looking at where depended-upon packages are located or whether they exist. I'm about to start a larger project that should fix that. Improved vendoring will be part of it.

jmillikin-stripe:
> If I understand correctly, you'd like to be able to add etcd as an external repository in some other workspace, and you'd like the generated BUILD files to respect vendor directories that aren't at the top level.

That's correct. The only thing our builds depend on from the etcd external repository is the etcd binary.

> Rather than adding a new command line argument to Gazelle, I think it would be better for Gazelle to detect and use vendor directories at any place in the tree automatically (#486).

I'm somewhat uncomfortable at the idea of Gazelle trying to auto-detect where the vendor directory is, considering how many ways there are for users to fiddle with `$GOPATH`. It's preferable to me if Gazelle could be told where to expect things.

> I don't think we should support vendoring in directories named something other than `vendor` though.

Could you expand on that? I don't understand why the Bazel rules should be more restrictive about directory layout than the underlying Go build system. Ideally, Bazel should be able to build any reasonably-shaped Go codebase.

jayconrod:
> I'm somewhat uncomfortable at the idea of Gazelle trying to auto-detect where the vendor directory is, considering how many ways there are for users to fiddle with $GOPATH. It's preferable to me if Gazelle could be told where to expect things.

We'd like Gazelle to be simple enough to be run automatically by editors when a file is saved. Within Google, there's an analogous tool called Glaze which does this very well. In that context, there's no opportunity to pass arguments on the command line. The behavior of Gazelle should be determined by convention whenever possible.

> Could you expand on that? I don't understand why the Bazel rules should be more restrictive about directory layout than the underlying Go build system. Ideally, Bazel should be able to build any reasonably-shaped Go codebase.

I just meant that Gazelle shouldn't allow directories with other names (for example, `third_party`) to be treated as `vendor`, since that breaks the "go build" convention. Gazelle should support `vendor` in other locations though. `cmd/vendor` works fine in "go build" right now, but not in Gazelle, and we should fix that. It requires something more sophisticated than simple import path translation though.

jmillikin-stripe:
> We'd like Gazelle to be simple enough to be run automatically by editors when a file is saved. Within Google, there's an analogous tool called Glaze which does this very well. In that context, there's no opportunity to pass arguments on the command line. The behavior of Gazelle should be determined by convention whenever possible.

Do you think it would make sense to have two separate modes for Gazelle? It seems like the job of "generate BUILD files in a non-Bazel source tree" is very different from "refresh existing BUILD files", and it would be a shame if the usefulness of Bazel's Go rules were limited by google3 usability requirements.

Remember that `third_party` rules assume BUILD files will be submitted alongside the external source, which is not something practical to do when pulling external dependencies from GitHub.

> `cmd/vendor` works fine in "go build" right now, but not in Gazelle, and we should fix that. It requires something more sophisticated than simple import path translation though.

The etcd build scripts run something like `ln -s "${PWD}/cmd/vendor "${GOPATH}/src"` before invoking `go build`, I'm not sure it's possible for Gazelle to detect when this sort of thing is happening.

If the concern is about the final directory being named "vendor", would you be OK with a flag like `--vendor_prefix=cmd` and then always joining `vendor` to the end of it?

jayconrod:
> Do you think it would make sense to have two separate modes for Gazelle? It seems like the job of "generate BUILD files in a non-Bazel source tree" is very different from "refresh existing BUILD files", and it would be a shame if the usefulness of Bazel's Go rules were limited by google3 usability requirements.

Perhaps, but this would add complexity. This would mean some options are available in one mode but not another. I'd prefer to have very little reliance on the command line.

> Remember that third_party rules assume BUILD files will be submitted alongside the external source, which is not something practical to do when pulling external dependencies from GitHub.

This is roughly what we want to do though. In #460, we're planning to remove `go_repository`. Instead of generating BUILD files when running `bazel build`, you'd generate them ahead of time with Gazelle, put them in a directory, then use them with `new_git_repository`. This makes it a lot easier to customize builds for dependencies that don't quite follow conventions.

There is also [some discussion](https://groups.google.com/forum/#!msg/bazel-discuss/yisAi4YUQjc/x8rFP4FnAQAJ) on the bazel-discuss mailing list about having a central repository of BUILD files for various libraries. See [meeting notes](https://docs.google.com/document/d/1VuoIsaqG1PwlncvMRY-fcW7e8lsY0SkEpaBpXZnZRIE/edit).

> If the concern is about the final directory being named "vendor", would you be OK with a flag like --vendor_prefix=cmd and then always joining vendor to the end of it?

I think we should have a more general solution. If Gazelle sees a directory named `cmd/vendor/` as it traverses the tree, it should be able to resolve imports from sources in `cmd/` (and subdirectories) to packages in `cmd/vendor/`. It should support multiple vendor directories in different parts of the tree with appropriate visibility. We should be able to detect these automatically without a command line option.

jmillikin-stripe:
> I think we should have a more general solution. If Gazelle sees a directory named `cmd/vendor/` as it traverses the tree, it should be able to resolve imports from sources in `cmd/` (and subdirectories) to packages in `cmd/vendor/`.

That wouldn't help in this case. The etcd code layout assumes there's a single vendor directory, it just happens they've named it `cmd/vendor/` instead of `vendor/`. For example, `//etcdmain/:go_default_library` depends on `//cmd/vendor/github.com/cockroachdb/cmux:go_default_library`.

> Perhaps, but this would add complexity. This would mean some options are available in one mode but not another. I'd prefer to have very little reliance on the command line.

Will the `--external` flag still exist, or is that going to be removed?

> This is roughly what we want to do though. In #460, we're planning to remove `go_repository`. Instead of generating BUILD files when running `bazel build`, you'd generate them ahead of time with Gazelle, put them in a directory, then use them with `new_git_repository`. This makes it a lot easier to customize builds for dependencies that don't quite follow conventions.

What's the timeline for this work? From the discussion on #460 it sounds like removing `go_repository` is still in the early design phase and has no concrete plans to land in master any time soon. Would it be OK to merge this PR so `go_repository` does what we need? We can revisit when it comes time to remove `go_repository` entirely.

jayconrod:
I took another look at etcd, and it's stranger than I thought. `cmd/tools` is a symlink to `../tools`. So that would mean that packages would build when built with the prefix `github.com/coreos/etcd/cmd/tools` but not with `github.com/coreos/etcd/tools` even though it's the same directory. This will confuse a lot of tools. `go build` works but `go list` does not for example. Bazel may be inconsistent as well. You could end up building the same package multiple times under different import paths.

I think supporting this in Gazelle is too much of a special case, and I'd rather not accept this PR. This is a situation where hand-written build files are needed.

When Gazelle supports flat mode (#608), it will help generate a file that can be used with `new_git_repository` after some manual adjustment. Flat mode will be a prerequisite for a lot of the other work, so it should happen soon rather than later, hopefully in the next week or two.

> Will the `--external` flag still exist, or is that going to be removed?

Not sure yet. Ideally, it would be removed.

> What's the timeline for this work? From the discussion on #460 it sounds like removing go_repository is still in the early design phase and has no concrete plans to land in master any time soon.

As a rough estimate, this should be done by the end of August. I'll be working on this primarily after Gophercon.

jmillikin-stripe:
It's now been ~3 months since this PR, and there doesn't seem to be any movement toward removing `go_repository`. Could you please reconsider your decision?

jayconrod:
I'm still planning to remove `go_repository`. #859 is a proposal to index the import paths of libraries in a repository and then resolve paths based on that. That should make make vendored mode largely obsolete, since vendored libraries will be found automatically. I'm not sure whether that would solve the etcd problem adequately though, since vendored libraries should only be visible below the vendor's parent directory.

In the mean time, I'm okay with landing the portion of this PR that allows vendoring to be turned on in `go_repository`. I'd suggest that attribute be named `build_external` and be an `attr.string` with a set of legal values instead of `attr.bool`. I don't want to support a configurable location for vendor though.

jmillikin-stripe:
Done. I moved it to a new PR since nearly all the discussion here is about the vendor prefix.

PR #610, opened 2017-07-06
Gazelle: don't emit visibility attrs if default_visibility is declared

### LINKED ISSUE #583: Gazelle and visibility
Gazelle needs to attempt to match user intent when merging build files with respect to visibility attributes.
Running it on kubernetes currently adds public visibility to almost all the rules.

jayconrod:
* If an existing BUILD file contains 'package(default_visibility = ...)',
  Gazelle will no longer set visibility attributes in generated
  rules, since this would be overriding user intent.
* A new Generator instance is now created per package, since generator
  now contains a visibility setting.
* LabelResolver is still created once in order to preserve the
  external cache. This interface is now public and will be moved to a
  new packge in the future.
* TestGenerator now searches testdata for directories with BUILD.want
  files instead of having an explicit list of directories.

Fixes #583


go/tools/gazelle/rules/resolve.go
```diff
@@ -56,3 +55,36 @@ func (l label) String() string {
...
+func NewLabelResolver(c *config.Config) LabelResolver {
+	var e LabelResolver
+	switch c.DepMode {
+	case config.ExternalMode:
+		e = newExternalResolver()
+	case config.VendorMode:
+		e = vendoredResolver{}
+	}
+
+	return &unifiedResolver{
```
ianthehat: Was trying to work out how this fits in to this cl, is this the start of an auto mode?

jayconrod: Kind of. This change means `rules.Generator` is instantiated for each package, since it now has package-specific state. That may be expanded if we support other directives or inferences in existing BUILD files. I wanted to keep a single instance of `rules.LabelResolver` though, so I needed to make it public so `main.run` could pass it to `rules.NewGenerator`.

I took the opportunity to do a little refactoring when I extracted this constructor from `rules.NewGenerator`. I'd like to pull the resolver into a new package later though before I add new functionality.

PR #613, opened 2017-07-07
Gazelle: sort srcs and deps string lists in buildifier order

### LINKED ISSUE #584: Gazelle attribute sorting
At the moment gazelle sorts
a/b/dir:target
a/b:target
where the more natural order (and the one kubernetes uses) would be
a/b:target
a/b/dir:target

jayconrod:
This is a workaround for bazelbuild/buildtools#122. When buildifier
sorts lists, it only processes list expressions as arguments for
certain attributes. It doesn't process lists inside select or lists
concatenated with select. When that bug is fixed, this code can be
removed.

Fixes #584

### DIFF
--- a/go/tools/gazelle/rules/sort_labels.go
+++ b/go/tools/gazelle/rules/sort_labels.go
@@ -1,0 +1,128 @@
+package rules
+
+import (
+	"sort"
+	"strings"
+
+	bf "github.com/bazelbuild/buildtools/build"
+)
+
+var (
+	goRuleKinds = map[string]bool{
+		"cgo_library": true,
+		"go_binary":   true,
+		"go_library":  true,
+		"go_test":     true,
+	}
+	sortedAttrs = []string{"srcs", "deps"}
+)
+
+// SortLabels sorts lists of strings in "srcs" and "deps" attributes of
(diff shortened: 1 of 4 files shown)


jayconrod:
I ran this version of Gazelle on Kubernetes, and it's starting to look pretty good. There are a lot of changes, but nearly all of them are reasonable now (e.g., moving platform specific files to `select`).

go/tools/gazelle/rules/sort_labels.go
```diff
@@ -0,0 +1,128 @@
...
+		if !goRuleKinds[r.Kind()] {
+			continue
+		}
+		for _, key := range []string{"srcs", "deps"} {
+			attr := r.AttrDefn(key)
+			if attr == nil {
+				continue
+			}
+			bf.Walk(attr.Y, func(e bf.Expr, _ []bf.Expr) {
+				list, ok := e.(*bf.ListExpr)
```
ianthehat: I would probably have extracted this to a named function.
I don't think it's closing over any variables, so it's reasonably discrete, and I generally try to avoid long lambdas with return statements, it's really hard to see what the return will do.

jayconrod: Done

Issue #161, opened 2016-10-22
support fully static binaries


mikedanese:
Users should be able to produce fully static binaries from go source, especially exclusively go source.

pmbethe09:
The current binaries presumably have a shared dependency on libpthread and libc?

I get the same result on linux if I say 'go install github.com/bazelbuild/rules_go/go/tools/gazelle/gazelle as when I 
bazel build //go/tools/gazelle/gazelle:gazelle

What is the magic flag to 'go build' to make it fully static?

mikedanese:
```
CGO_ENABLED=0 go install github.com/bazelbuild/rules_go/go/tools/gazelle/gazelle
```
Should produce a static binary.

powelljo:
I was just looking at this issue myself and see it's been discussed here already.  Has there been any progress made on passing along values of environment variables that Go builds can depend on?  I see there is limited support, specifically for `GOOS` and `GOARCH`, already [https://github.com/bazelbuild/rules_go/blob/c7e02c851a8faf803f260800e7387248ccd47683/go/def.bzl#L83].

GinFungYJF:
Any update about this? If no, are there any workarounds?

pmbethe09:
@jayconrod may look into this

jayconrod:
I'm definitely planning to improve this. Cross compilation is a slightly higher priority in the short term though.

mikedanese:
Also related: https://github.com/bazelbuild/rules_go/issues/217

Specifically, this would partially solve (although I believe this uses libc rather than the go libc implementation):

```python
go_binary(
    name = "foo",
    srcs = ["foo.go"],
    gc_linkopts = [
        "-linkmode external",
        "-extldflags -static",
    ],
)
```

GinFungYJF:
@mikedanese The gc_linkopts attribute seems to has not been implemented.

mikedanese:
@GinFungYJF It is not implemented in the opensource rules but it does exist in the internal rules (which are completely different). #217 proposes we implement similar feature in opensource.

jayconrod:
`gc_linkopts` has been merged for a little while now. So static builds should work even with cgo present.

```bzl
go_binary(
    name = "main",
    srcs = ["main.go"],
    go_linkopts = [
      "-extldflags",
      "-static",
    ],
    library = ":cgo_lib",
)

cgo_library(
    name = "cgo_lib",
    srcs = ["cgo_lib.go"],
[...]

PR #617, opened 2017-07-11
Add optional static outputs to `go_binary`.

jmillikin-stripe:
This PR changes`go_binary` to define implicit outputs for static binaries, using the `gc_linkopts` advice from <https://github.com/bazelbuild/rules_go/issues/161>. For example, a binary target `//foo/bar:main` would now have an additional implicit `//foo/bar:main.static` target alongside. The `.static` prefix was chosen by analogy with `.stripped` for stripped C binaries.


bazel-io:

achew22:
As a 3rd option, `go_binary` could be changed so that it creates a static binary by default

jayconrod:
I think it would be better to solve this in the rule instead of Gazelle. We might want to build binaries in a lot of configurations, and I don't think we should have a separate target for each.

For `go_binary`, an [output group](https://docs.bazel.build/versions/master/skylark/rules.html#output-groups) for static binaries would probably make the most sense. The rule would create files and actions for multiple configurations, but you only get the regular one when you build without flags. You could pass in `--output_groups=static` or something like that on the command line to get static builds.

We've also talked about using toolchains for different configurations. That makes the most sense for, opt, race, asan, etc., when something needs to be done differently for the entire build, but for static builds, we just need a few extra flags at the end.

jmillikin-stripe:
> I think it would be better to solve this in the rule instead of Gazelle. We might want to build binaries in a lot of configurations, and I don't think we should have a separate target for each.

Done.

I used an implicit output rather than an output group because Bazel seems to build all output groups by default. If I ran `bazel build //examples/bin:bin` on a Mac, the command would fail because MacOS doesn't support static linking. An implicit output lets the static target be built if requested, but doesn't build by default.

ianthehat:
Jay is right that we should fix this in the rule.
I don't have a clear enough picture of when people will want this to know what the right choice is though. The difficulty is that rules are only allows a single executable output, for good reasons, which means one of the build modes is going to be a lesser citizen if we do it with multiple outputs.
Basically, the question is will people ever want both versions in a single build, will they want to be able to switch between them for any given binary, will they want different binaries in different modes, and will they ever want to use a static form as a tool in the build.

jayconrod:
@ianthehat I thought we could have multiple executable outputs if they are declared using `ctx.new_file`. Does that not work? They might need to be requested explicitly, but you could build more than one file at a time from the command line.

I think this is a good approach, but I'd like to take some time to tinker with it a bit. I'm at GopherCon this week, so I'll be a bit slow.

ianthehat:
For a rule to be "bazel run"able it must be marked executable = True, if you do so, it pre-declares exactly one output file with a name based on the rule name, which will be used when you either run that label, or use it as a tool in another rule.
You can produce more files that have the executable bit set on them, but that's a different concept.
In other words, we are both right, we are just talking about different things :)

The output groups would definitely work, I have another change that uses them and it does not build  output groups that were not requested, but they are a bit confusing to get right, partly because there are not many examples of using them right now.

achew22:
The nearest symmetry for this I can see is in java_binary's implicit [`<name>_deploy.jar`](https://docs.bazel.build/versions/master/be/java.html#java_binary) which is, I believe, basically a statically linked jar. When you `bazel run` those targets I believe it runs the non-`_deploy.jar` variety.

jmillikin-stripe:
Maybe I'm using the output groups wrong? Here's what I tried to return from `_go_binary_impl`:

```python
  return [
      struct(
          files = depset([ctx.outputs.executable]),
          runfiles = lib_result.runfiles,
          cgo_object = lib_result.cgo_object,
      ),  
      OutputGroupInfo(
          static = depset([ctx.outputs.static_executable]),
      ),  
  ]
```

Without that change, only the "default" binary is built:

```console
$ bazel-0.5.2 build //examples/bin
INFO: Found 1 target...
Target //examples/bin:bin up-to-date:
  bazel-bin/examples/bin/bin
INFO: Elapsed time: 0.197s, Critical Path: 0.00
```

With the output group, the static binary is also built:

```console
$ bazel-0.5.2 build //examples/bin
INFO: Found 1 target...
Target //examples/bin:bin up-to-date:
  bazel-bin/examples/bin/bin.static
  bazel-bin/examples/bin/bin
```

> Basically, the question is will people ever want both versions in a single build, will they want to be able to switch between them for any given binary, will they want different binaries in different modes

I don't know of a use case for different versions of the same binary, but I can confirm that we want the ability to have a single build both static and dynamic targets.

> will they ever want to use a static form as a tool in the build.

This seems unlikely. If Bazel just built it, then the libraries should still be available.

ianthehat:
Yeah, you are using them wrong :)
The hidden knowledge that is not in the documentation is that you don't want to declare it using outputs at all, instead you want to use `ctx.new_file`, otherwise it is implicitly in the default output group. I am also not entirely sure if it works if you don't declare the `DefaultInfo` provider, I have never tried it with the implicit one cause by returning a `struct`.
In general I have come to the conclusion that you should *always* use `ctx.new_file` and never the other output declaration forms, with the exception of the `executable = True` form that you have to use for tools.

> I don't know of a use case for different versions of the same binary, but I can confirm that we want the ability to have a single build both static and dynamic targets.

That rules out switching it with either a config setting or on the primary toolchain (and toolchain selection on a dependancy is not here now or any time soon)
It does still mean that a `static = True` on the binary rule itself is still an option though.

jmillikin-stripe:
Using `ctx.new_file` allows output groups to work as documented, but seems to prevent other rules from directly referencing the static binary:

```python
static_executable = ctx.new_file(ctx.attr.name + ".static")
  return [
      struct( ... ),  
      OutputGroupInfo(
          static = depset([static_executable]),
      ),  
  ]
```

```console
$ bazel build //examples/bin --output_groups=static
INFO: Found 1 target...
Target //examples/bin:bin up-to-date:
  bazel-bin/examples/bin/bin.static
```

```console
bazel build //examples/bin:bin.static
```

ianthehat:
You never want to try to use file names when you are dealing with outputs, instead you want to pick information out of the providers. Apart from anything else the filename would be wrong on windows anyway.
So assuming you have a rule that wants to use the static form, you would have it in your attr's as a label, and thus ctx.attr.mydep would give you a target.
OutputGroupInfo is just a provider, you need to pick it out of the target and then access it's members.
`ctx.attr.mydep[OutputGroupInfo].static`

jmillikin-stripe:
If I have a BUILD file like this:

```python
pkg_tar(
    name = "image_files",
    files = ["@com_github_coreos_etcd//:com_github_coreos_etcd.static"],
    symlinks = { 
        "/bin/etcd": "/com_github_coreos_etcd.static",
    },  
)
```

How would I adapt it to use OutputGroupInfo? Do I have to write a custom build rule for each dependency?

ianthehat:
You can use a special mode of filegroup to pick up specific output groups.
Essentially you use it to switch from a non default group of it's inputs, to the default group of it's outputs...
```
filegroup(
    name = "etcd.static",
    srcs = ["@com_github_coreos_etcd//:com_github_coreos_etcd"],
    output_group = "static",
)
pkg_tar(
    name = "image_files",
    files = [":etcd.static"],
    symlinks = { 
        "/bin/etcd": "/com_github_coreos_etcd.static",
    },  
)
```

jmillikin-stripe:
Thanks, that looks like it's doing the right thing. I've updated the PR to use output groups.

go/private/binary.bzl
```diff
@@ -33,11 +35,32 @@ def _go_binary_impl(ctx):
...
+
+  return [
+      DefaultInfo(
+          files = depset([ctx.outputs.executable]),
+          runfiles = lib_result.runfiles,
+          cgo_object = lib_result.cgo_object,
+      ),
+      OutputGroupInfo(
+          static = depset([static_executable]),
+      ),
```
ianthehat: Minor points for discussion, could always be done in a different change later...

Should we also have a named group for the dynamic form so you can be explicit about which one you want?
What should we do on platforms where static is not possible? 
* Not declare the static group
* have the static depset empty
* leave it failing when trying to build

Should we have a way of allowing rules to use the static if available, and the dynamic otherwise, or are the consuming rules always going to have to modify their behaviour anyway?

go/private/binary.bzl
```diff
@@ -33,11 +35,32 @@ def _go_binary_impl(ctx):
     gc_linkopts=gc_linkopts(ctx),
     x_defs=ctx.attr.x_defs)
 
-  return struct(
-      files = depset([ctx.outputs.executable]),
-      runfiles = lib_result.runfiles,
-      cgo_object = lib_result.cgo_object,
-  )
+  # Static linking (in the 'static' output group)
+  static_linkopts = [
```
ianthehat: If we are going to add lots of variations on this theme, we might have to come up with a nice way to do this repeatedly, otherwise this function is going to get very repetitive and fragile.

jayconrod:
Also looks good to me. However, we need to bump our minimum version of Bazel, since 0.4.5 doesn't support output groups. I'll put together a PR to update to 0.5.2, and that will land first.

(We were planning to update anyway when platform / toolchain support lands).

jayconrod:
Jenkins test this please

jayconrod:
Thanks again for submitting this! And sorry for the delay -- GopherCon has kept me really busy this week.

PR #623, opened 2017-07-14
Don't mix declared and old-style providers

vladmos:
Now declared providers are used for both go_bootstrap_toolchain and go_toolchain.

In the future it will be possible to return a single declare provider (not in a list), but now due to a bug such providers are treated like old-style structs, therefore both rule implementation functions return singleton lists now for compatibility.

### DIFF
--- a/go/private/common.bzl
+++ b/go/private/common.bzl
@@ -53,10 +53,7 @@ cgo_filetype = FileType([
 ])
 
 def get_go_toolchain(ctx):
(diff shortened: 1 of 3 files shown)


vladmos:
Jenkins, test this please

damienmg:
retest this please (ci.bazel.io failure where due on test on ci)

vladmos:
Looks like Travis tests are broken or flaky, I don't know how to restart them but they fail only in one configuration.

PR #625, opened 2017-07-16
Switching to true toolchains

ianthehat:
Now fully working with the latest version of toolchains.
Passes all tests if you set BAZEL enviroment variable to tell it which version to run in integration tests.
mac->unix is the only working cross compilation, the go side should work for anything go supports, but that's the only c++ cross I can get to work out of the box.


jayconrod:
I won't be able to try this out until Wednesday, but it looked good on the branch earlier. Feel free to TBR it.

This works with github.com/katre/bazel/tree/toolchains, correct?

ianthehat:
Updated to the current tip of master (with all the provider and cgo changes)

ianthehat:
I am going to abandon this pull request in favor of an updated one directly to master that we will submit only once bazel 0.5.4 is out.

PR #627, opened 2017-07-18
Add go tool cgo wrapper

ianthehat:
Does all the work of filtering and choosing cgo sources, replaces all the ugly shell script.
This is a prerequisite for the cgo_library merge.


go/tools/builders/cgo.go
```diff
@@ -0,0 +1,175 @@
...
+	gotool := args[0]
+	if err := flags.Parse(args[1:]); err != nil {
+		return err
+	}
+
+	// apply build constraints to the source list
+	// also pick out the cgo sources
+	bctx := build.Default
+	bctx.CgoEnabled = true
+	cgo_srcs := []string{}
```
jayconrod: Nit: camelCase

go/tools/builders/cgo.go
```diff
@@ -0,0 +1,175 @@
...
+		if !strings.HasSuffix(in, ".go") {
+			// Not a go file, just filter, assume C or C-like
+			if match {
+				// not filtered, copy over
+				if err := ioutil.WriteFile(out, data, 0644); err != nil {
+					return err
+				}
+			} else {
+				// filtered, make empty file
+				if err := ioutil.WriteFile(out, []byte("//Empty file"), 0644); err != nil {
```
jayconrod: I would actually leave these files empty. // may not be a valid comment prefix for assembly dialects compiled by the external compiler.

go/tools/builders/cgo.go
```diff
@@ -0,0 +1,175 @@
...
+					return err
+				}
+			}
+			continue
+		}
+
+		// Go source, must produce both c and go outputs
+		cOut := strings.TrimSuffix(out, ".cgo1.go") + ".cgo2.c"
+		if err != nil {
```
jayconrod: Error check seems out of place. Did you mean to check that the suffix was trimmed?

go/tools/builders/cgo.go
ianthehat: Done, bouncing between skylark and go I think I am going to make a lot of these mistakes :)

go/tools/builders/cgo.go
ianthehat: Done. The go toolchain presumes all assembler files support c style comments for build tags, but it's probably safer anyway.

go/tools/builders/cgo.go
ianthehat: Nope, just left over from when there was another bit of code in there, removed.

PR #640, opened 2017-07-21
An example using go-bindata

ianthehat:
The basic rule is sufficient for normal use as well.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,7 +1,6 @@
 workspace(name = "io_bazel_rules_go")
 
 load("@io_bazel_rules_go//go:def.bzl", "go_repositories", "go_repository")
-
 go_repositories()
 
 # Needed for examples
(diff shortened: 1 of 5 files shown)


ianthehat:
This is just an example of how to use go-bindata, or any other go file generation tool like it. Lots of people want to do these kinds of things, having a best practice example is probably worth it.
 It does not add an actual required dependancy on the go rules themselves, even though it is referred to from the workspace.
I have been thinking that it might be nicer to keep our tests and examples in a separate place so they are not even in the same repo as the main rules.

There are lots of different solutions to packing data files problem out there, with all sorts of different features, supporting all sorts of compression, or zero copy, or http serving, or fake file systems etc. None of them belong directly in rules_go in my opinion, they do a lot more than just put a data file in a go file, and I don't think we can or should try to create a new way of doing this it really is a library concern. 
go-bindata the most popular, so it makes a good example though.

Issue #641, opened 2017-07-21
cover before cgo


jayconrod:
Currently, we create a `go_library` with cgo by splitting .go files into pure Go and C, then compiling the pure .go files as a normal `go_library`. This means coverage instrumentation is added after cgo runs.

Unfortunately, this is the only way coverage will work at the moment. The cover tool in the SDK deletes cgo comments (as of Go 1.9). "go build" also runs cgo, then coverage. I hope to get this switched upstream at some point (issue to be filed), but we'll need to be aware of the Go version when ordering these.

ianthehat:
+jayconrod I believe this is now how we want it to be?

jayconrod:
Yes, this is fixed now.

PR #642, opened 2017-07-21
Don't apply coverage to go_binary or go_test sources

### LINKED ISSUE #629: Coverage fails with "GoCover_0 redeclared in this block"
```
$ bazel coverage ...:*
ERROR: /.../BUILD.bazel:10:1: null failed: Process exited with status 1 [sandboxed].
bazel-out/local-fastbuild/.../main_GoCover_0.cover_GoCover_0.cover.go:47: GoCover_0 redeclared in this block
	previous declaration at bazel-out/local-fastbuild/.../main_GoCover_0.cover_GoCover_0.cover.go:32
```

The generated `main_GoCover_0.cover_GoCover_0.cover.go` has:

```
var GoCover_0 = struct {
	Count	[2]uint32
	Pos	[3 * 2]uint32
	NumStmt	[2]uint16
}{
	Pos: [3 * 2]uint32{
		9, 10, 0x2e000d,
		10, 12, 0x3002e,
	},
	NumStmt: [2]uint16{
		1,
		1,
	},
}

var GoCover_0 = struct {
	Count     [2]uint32
	Pos       [3 * 2]uint32
	NumStmt   [2]uint16
} {
	Pos: [3 * 2]uint32{
		9, 11, 0x2e000d, // [0]
		11, 14, 0x3002e, // [1]
	},
	NumStmt: [2]uint16{
		2, // 0
		2, // 1
	},
}
```

jayconrod:
Coverage is now applied in emit_library_actions instead of
emit_go_compile_action. emit_library_actions accepts a want_coverage
parameter, which is only true for go_library. go_test and go_binary
should not apply coverage to their own sources, but will continue to
apply coverage to sources from the library attribute.

In the future, we may want to apply coverage before running cgo
(#641), so this provides an easy way to prevent coverage being applied
a second time in go_library.

Fixes #629

### DIFF
--- a/go/private/library.bzl
+++ b/go/private/library.bzl
@@ -16,7 +16,7 @@ load("@io_bazel_rules_go//go/private:common.bzl", "get_go_toolchain", "DEFAULT_L
 load("@io_bazel_rules_go//go/private:asm.bzl", "emit_go_asm_action")
 load("@io_bazel_rules_go//go/private:providers.bzl", "GoLibrary", "GoSource")
 
-def emit_library_actions(ctx, sources, deps, cgo_object, library):
+def emit_library_actions(ctx, sources, deps, cgo_object, library, want_coverage):
   go_toolchain = get_go_toolchain(ctx)
 
   go_srcs = depset([s for s in sources if s.basename.endswith('.go')])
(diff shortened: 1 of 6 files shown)


go/private/library.bzl
```diff
@@ -118,6 +121,7 @@ def _go_library_impl(ctx):
       deps = ctx.attr.deps,
       cgo_object = cgo_object,
       library = ctx.attr.library,
+      want_coverage = ctx.coverage_instrumented(),
```
ianthehat: Is this not meant to use the want_coverage that was passed in?

ianthehat:
I was looking at this some more, and I think we ought to not return the covered sources from the emit_library action, instead we should use ctx.coverage_instrumented(ctx.attr.library) in the test rule to detect that we ought to ask for the library in covered mode.

go/private/library.bzl
jayconrod: `want_coverage` is not passed into `go_library` explicitly.

PR #648, opened 2017-07-24
README: show how to run equivalent of ./... and stream output

kevinburke:
I was confused about the syntax for doing these two things with Bazel
- basically how to match the output of "go test", so it might be
helpful to explicitly show users what command they should be running
with "bazel test".


bazel-io:

PR #657, opened 2017-07-27
README: clarify sample bazel test command

prattmic:
bazel test //... runs all tests in the entire workspace, not just under
the current working directory, as "go test ./..." implies. Note that and
clarify that "go test ./..." is only equivalent if it were run from the
go_prefix directory inside GOPATH.

bazel test ...:all could be used to run all tests under the CWD, but in
my experience it is more common to remain at workspace root anyways, and
specify packages you care about (//pkg/...) rather than changing to that
directory.

@kevinburke @jayconrod


bazel-io:

PR #667, opened 2017-07-28
Add -known_import flag

pmcalpine:
Private repositories can take a long time to resolve, and if the repo
host does not properly support go-import meta tags then gazelle may not
be able to resolve them at all.

This change allows the author of WORKSPACE files to add specially
crafted comments which instruct gazelle to update the external
resolver's repo cache. As a result, gazelle can be instructed to skip
resolution of certain imports.

### DIFF
--- a/go/tools/gazelle/config/config.go
+++ b/go/tools/gazelle/config/config.go
@@ -47,6 +47,9 @@ type Config struct {
 
 	// DepMode determines how imports outside of GoPrefix are resolved.
 	DepMode DependencyMode
+
+	// KnownImports is a list of imports to add to the external resolver cache
+	KnownImports []string
 }
 
 var DefaultValidBuildFileNames = []string{"BUILD.bazel", "BUILD"}
--- a/go/tools/gazelle/resolve/resolve.go
+++ b/go/tools/gazelle/resolve/resolve.go
@@ -76,7 +76,7 @@ func NewLabelResolver(c *config.Config) LabelResolver {
 	var e LabelResolver
 	switch c.DepMode {
 	case config.ExternalMode:
-		e = newExternalResolver()
+		e = newExternalResolver(c.KnownImports)
 	case config.VendorMode:
 		e = vendoredResolver{}
 	}
--- a/go/tools/gazelle/gazelle/BUILD
+++ b/go/tools/gazelle/gazelle/BUILD
@@ -5,6 +5,7 @@ go_library(
     srcs = [
         "diff.go",
         "fix.go",
+        "flags.go",
         "main.go",
         "print.go",
     ],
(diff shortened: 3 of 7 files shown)


bazel-io:

pmcalpine:
Hi. I have been keeping this functionality on my own fork for a while and have rewritten it enough times while rebasing that I'm wondering if you'd consider for the mainline.

This is a quick-and-dirty solution that I've been using for the past few months. My overarching objective is to be able to update the externalResolver's cache. If this implementation doesn't suit do you think there's another one that would?

pmcalpine:
Thinking over lunch: this is actually two changes
1. ability to add to the external resolver cache. This could be done with a new command line flag.
2. ability to configure gazelle via magic comments in the WORKSPACE file.

Please consider these independently. I'm happy to take on #1 asap.

go/tools/gazelle/resolve/resolve.go
```diff
@@ -77,6 +77,12 @@ func NewLabelResolver(c *config.Config) LabelResolver {
 	switch c.DepMode {
 	case config.ExternalMode:
 		e = newExternalResolver()
+	case config.KnownMode:
+		er := newExternalResolver()
```
jayconrod: It would be better to pass additional imports to `newExternalResolver` instead of modifying the cache directly here.

go/tools/gazelle/config/config.go
```diff
@@ -107,6 +144,12 @@ const (
 	// (declared in WORKSPACE).
 	ExternalMode DependencyMode = iota
 
+	// KnownMode indicates imports should be resolved via ExternalMode unless
```
jayconrod: There doesn't need to be an extra mode of this. `KnownImports` can just be empty if there aren't any.

pmcalpine: SGTM.

go/tools/gazelle/resolve/resolve.go
pmcalpine: If I'm adding KnownImports to config.Config{...} should I move the default known imports from resolve_external.go to config.go?

pmcalpine:
Ok. I'll turn this into a repeatable command line option.

Thanks!

go/tools/gazelle/resolve/resolve.go
jayconrod: That's a good question. Let's keep them in resolve_external.go for now. Config is created in a lot of places for tests, so that would make this change more complicated. I may refactor it later.

pmcalpine:
PTAL.

jayconrod:
Thanks, looks good!

Jenkins test this please.

PR #668, opened 2017-07-28
Gazelle: fix load statements before merge

jayconrod:
This change is the first in a series that will remove cgo_library from
Gazelle, now that it's no longer needed. The new functionality here is
not wired in yet so that the large change can be reviewed in smaller
pieces.

This adds a new function, fixLoads, which modifies load statements in
existing BUILD files. Symbols will only be loaded for rules that will
actually used. Multiple load statements may be consolidated into a
single statement. Load statements and symbols not generated by Gazelle
will not be modified.

This function will be applied after cgo_library rules are replaced or
merged with go_library rules. The cgo_library symbol should not be
loaded after this.


go/tools/gazelle/merger/fix.go
```diff
@@ -0,0 +1,189 @@
...
+		x, ok := c.X.(*bf.LiteralExpr)
+		if !ok {
+			continue
+		}
+
+		if x.Token == "load" {
+			if len(c.List) == 0 {
+				continue
+			}
+			if label, ok := c.List[0].(*bf.StringExpr); ok && label.Value == "@io_bazel_rules_go//go:def.bzl" {
```
ianthehat: We have talked about the fact that io_bazel_rules_go is the wrong name, but it's too hard to change.
Should we extract it to a constant, and think about wether the new fixing ability could update this name for us?

jayconrod:
I'll need to think some more about whether this should be behind a new verb. The harm in leaving cgo_library in place without squashing it is we'll lose user-added `# keep` strings. If we fix it automatically though, anything that built `cgo_default_library` explicitly will break.

PR #674, opened 2017-07-31
Only download SDK once

### LINKED ISSUE #671: Go binary is downloaded twice
On my Mac, I observe the download count get to about 89MB, and then reset to 0 and start downloading again. Project is: https://github.com/kevinburke/rest, running the "make test" target.

I'm not sure the best way to reproduce and not sure I'll have time to do it.

jayconrod:
Bazel repository rules may be restarted if they depend on a file that
isn't available yet using a Label. If a download has already occurred,
this means the download will happen again (unless
--experimental_repository_cache is in use).

We should download archives after Labels have been resolved.

Fixes #671

### DIFF
--- a/go/private/repository_tools.bzl
+++ b/go/private/repository_tools.bzl
@@ -29,15 +29,6 @@ filegroup(
 """
 
 def _go_repository_tools_impl(ctx):
-  # We have to download this directly because the normal version is based on go_repository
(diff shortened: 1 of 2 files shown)


kevinburke:
Thanks! Looking forward to pulling this in to my projects

PR #676, opened 2017-07-31
go_test: add rundir attribute

### LINKED ISSUE #670: support compiling tests as binary artifacts
We currently build a few of our e2e tests outside of Bazel using `go test -c`, as these tests are run later after the system-under-test has been started (using a separate framework).

Under Bazel, we [use a genrule that depends on the `go_test` target](https://github.com/kubernetes/kubernetes/blob/d8205661b700c6f99c33ea0ac1e5ed3e49c202b2/test/e2e/BUILD#L132-L141), but after #554 this stopped working, as the test first attempts to chdir to the test package, failing if it can't. And the way we run the tests, this always fails.

Is there a more obvious way we could do this that it'd actually work?

jayconrod:
This lets developers specify the directory where a test will run.

Fixes #670

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -26,6 +26,7 @@ import (
 	"go/token"
 	"log"
 	"os"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strconv"
@@ -56,6 +57,7 @@ type coverInfo struct {
 type Cases struct {
 	Package          string
 	RunDir           string
+	AltRunDir        string
 	TestNames        []string
 	BenchmarkNames   []string
 	HasTestMain      bool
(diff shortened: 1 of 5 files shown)


ixdy:
Confirmed that this fixes #670 for us if we use `rundir = "."`. Thanks for the quick fix!

kevinburke:
This is a little confusing, since `go test` for a given package always sets the CWD to the directory of the package

jayconrod:
@kevinburke That will be the default behavior. `rundir` will only be needed if you need another directory.

Or did you mean it's confusing that `rundir` is relative to the repo root and not the package directory?

kevinburke:
I guess my initial reaction on reading the documentation was, "oh, does Bazel not run tests in the package cwd?" Like the fact the setting exists made me more curious that the default was something not sensible. I guess each package has a BUILD file but it took me a second to realize that.

It's probably fine.

jayconrod:
Ah, yes, the Bazel default is to run everything (tests, binaries, builders) from the workspace root directory (which can even be different than the repo root directory for external targets). This is a nice convention with Google because we have a huge monorepo, but it's incompatible with nearly all Go tests outside of Google, which is why the code generated by `go_test` changes to the package directory before calling `TestMain` or running tests.

go/tools/builders/generate_test_main.go
```diff
@@ -166,9 +168,17 @@ func coverRegisterFile(fileName string, counter []uint32, pos []uint32, numStmts
 {{end}}
 
 func main() {
+{{if .AltRunDir}}
```
ianthehat: I wonder if instead of baking it into the binary through the template, we should just add a standard flag to the test main, and set it by default in the go_test.
That way if you build the test binary and then invoke it by hand the default behavior is not surprising (few tests actually look at the file system, and yet you would require the runfiles directory to be able to run them by hand)

jayconrod:
Per offline discussion, tests will now only change directory when run by Bazel. The generated test code will check whether the `TEST_SRCDIR` environment variable is defined (Bazel's test runner defines it), and then change if it is.

@ixdy, this means you won't have to set `rundir` at all for tests that get executed in another environment.

ixdy:
Thanks for the update. I'll test this new version and verify whether it still works.

ixdy:
Confirmed, this works as expected for Kubernetes. Thanks!

PR #680, opened 2017-08-01
Gazelle squash cgo library

jayconrod:
squashCgoLibrary removes cgo_library rules with the default name and
merges their attributes with go_library rules that reference them via
the library attribute. If no go_library rule exists, a new one will be
created.

This function is called from FixFile, which is not wired in yet, since
Gazelle doesn't generate combined go_library rules yet.


jayconrod:
(This is based on top of #679)

go/tools/gazelle/merger/fix.go
```diff
@@ -25,7 +26,211 @@ import (
 // FixFile updates rules in oldFile that were generated by an older version of
 // Gazelle to a newer form that can be merged with freshly generated rules.
 func FixFile(oldFile *bf.File) *bf.File {
-	return fixLoads(oldFile)
+	fixedFile := squashCgoLibrary(oldFile)
+	return fixLoads(fixedFile)
```
ianthehat: As we add more fixers in here, order is going to be very important (and fragile), might be worth adding comments about which things must happen before others and why (like fixLoads is probably last, because other rules might change what needs loading)

PR #682, opened 2017-08-02
Update README with details about `# gazelle:exclude`

ixdy:
`# gazelle:exclude` was added in #581 but the README wasn't updated.

I think this explanation is correct, but please correct me if I'm misunderstanding it.


ixdy:
squashed, thanks!

PR #683, opened 2017-08-02
Switching to true toolchains

ianthehat:
Working go toolchain, with cross compilation support.
You can cross compile mac->linux (including cgo) using
```bazel build --cpu k8```
You can cross compile linux->windows (pure go only) using
```bazel build --experimental_platforms=@io_bazel_rules_go//go/toolchain:windows_amd64```
but the file name will be wrong (no .exe extension) and build tag select statements may be broken.

### CODE REVIEW

File: go/private/go_tool_binary.bzl [hunk 1]
```diff
@@ -11,14 +11,12 @@
...
-
-go_bootstrap_toolchain_type = toolchain_type()
+load('//go/private:go_toolchain.bzl', 'go_toolchain_core_attrs')
 
 def _go_bootstrap_toolchain_impl(ctx):
-  return [go_bootstrap_toolchain_type(
-      exec_compatible_with = ctx.attr.exec_compatible_with,
-      target_compatible_with = ctx.attr.target_compatible_with,
+  return [platform_common.ToolchainInfo(
+      type = Label("@io_bazel_rules_go//go:bootstrap_toolchain"),
```
katre: Commit https://github.com/bazelbuild/bazel/commit/bb87c12dda25ebad124907625a9cdb973cf9e1f2 isn't in Bazel 0.5.4, but in the next release this line can be removed. Is it worth trying to get that cherry-picked to 0.5.4 to simplify removing this?

ianthehat: It's not that offensive.
If it's easy, sure, but it's not worth it if there is hassle or stress involved.

ianthehat:
Added an extra commit that supports the "host" toolchain.

building with `--experimental_platforms=@io_bazel_rules_go//go/toolchain:host`
will now cause it to detect and use the hosts go sdk.
It checks for `GOROOT` first, if not set it tries to find `go` and use that to
work out the `GOROOT`

AmandaCameron:
Toying with this PR in our codebase, I'm curious if it's supposed to be ignoring _foo.go files? Specifically, github.com/boltdb/boltdb is failing to compile because it's expecting a function defined in various platform-specific files to provide the actual impl.

jayconrod:
@AmandaCameron You're just getting errors when cross-compiling, not when building for the host platform, right?

@ianthehat Will go/builders/compile.go see the correct GOOS / GOARCH? It looks like they are set correctly, but it would be good to have a test for this.

ianthehat:
Yes, the GOOS and GOARCH come from the toolchain and will be correct.
The error is almost certainly what I meant when I said "build tag select statements may be broken".
We cannot write select statements in terms of the platform constraints yet, so they are still based on the cpu tag, which means they are wrong if you just use --experimental_platforms.
This should be fixed by the next version of Bazel, it's being worked on, but until then the only way round it is to remove the select statements that gazelle generates, and rely on the later tag processing that happens inside the compile action.

AmandaCameron:
Might be me mis-undrstanding something about the bazel cross-compile stuff, but yes this was from a cross-compile of darwin-> linux -- bazel seems to barf at `--cpu k8` on the darwin machine.

katre:
You need to use --experimental_platforms (and create a platform description
of the linux system) to use toolchain-based cross-compilation.

A valid platform might look like:
platform(
  name = 'linux',
  constraint_values = ['@bazel_tools//platforms:x86_64',
'@bazel_tools//platforms:linux'])

And then instead of --cpu k8 you need '--experimental_platforms=//:linux'
(or whatever the label of your new platform target is).

PR #687, opened 2017-08-03
Improve error message when no buildable cgo sources are found

jayconrod:


go/tools/builders/cgo.go
```diff
@@ -132,7 +135,15 @@ func run(args []string) error {
 			if err := ioutil.WriteFile(cOut, []byte(""), 0644); err != nil {
 				return err
 			}
-		} else if isCgo {
+			continue
+		}
+
+		if pkgName != "" && pkg != pkgName {
```
ianthehat: Is there any reason this test could not go above the if !match part?

jayconrod: I'm not sure there is a definite right answer here, but I don't think files that were filtered out should cause this error.

PR #688, opened 2017-08-03
Gazelle: add fix and update commands

jayconrod:
Gazelle will now accept a command verb as a first argument, which will
determine what should be done. Currently, two commands are supported:

  update - Gazelle will create new BUILD files or update existing
    BUILD files. This is the default; if Gazelle is run without a
    command, it will do this.
  fix - in addition to the changes made in update, Gazelle will make
    potentially breaking changes. Currently, it will squash
    cgo_library and consolidate load statements.

More commands will be added in the future, for example for managing
dependencies.

### DIFF
--- a/go/private/gazelle.bzl
+++ b/go/private/gazelle.bzl
@@ -20,7 +20,7 @@ $BASE/{gazelle} {args} $@
 """
 
 def _gazelle_script_impl(ctx):
-  args = ctx.attr.args
+  args = [ctx.attr.command] + ctx.attr.args
   args += [
       "-repo_root", "$WORKSPACE",
       "-go_prefix", ctx.attr._go_prefix.go_prefix,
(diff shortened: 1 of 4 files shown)


jayconrod:
Jenkins test this please

ianthehat:
Maybe also update gazelle.bzl?

jayconrod:
Done. I changed some things from being default in the macro to being default in the rule. PTAL.

ianthehat:
Looks good, thanks.
My only remaining question is whether we think that there could be any confusion between the two meanings of fix.
```
gazelle fix -mode=fix
```
The trouble is I think it's the mode that's badly named, but changing that is not backwards compatible.

jayconrod:
I'm concerned about that, too. I think we should rename `-mode` to `-output_mode` and `-mode=fix` to `-output_mode=write`. The old arguments and names can be preserved for now as aliases, and we can print a warning when they are used. We can remove the old names in a couple months. WDYT?

ianthehat:
Sounds good. output_mode is a bit verbose, but people won't use it much I don't think.

I was thinking of adding a new mode (verify) that has no output but causes gazelle to return non zero if there were any differences (for presubmit hook types of usage), does --output_mode=verify make sense?

jayconrod:
I like that idea, but I think the output would still be useful if verification fails. Maybe this should be the behavior of diff mode: return non zero if any of the diff commands were non empty.

ianthehat:
I was thinking it would print the filenames that were different, and return non zero.
Most people will never see or look at the output, they will just go, "oh, I forgot to run gazelle, best do that" and move on, if you really need to see the output, you will probably run it again by hand with a different mode.
the reason I wanted a different mode is because things like ide's might want diff mode and need to distinguish gazelle succeeding with diffs and aborting with partial diffs.

ixdy:
I would like a `verify` mode that exists nonzero if there are any changes `gazelle` would write. We have a similar option in `kazel` which we use as one of our [CI checks](https://github.com/kubernetes/kubernetes/blob/66bfab89151ce613fadedad15a848613685788ae/hack/verify-bazel.sh):
```shell
if ! kazel -validate -print-diff -root="$(kube::realpath ${KUBE_ROOT})" ; then
  echo
  echo "Run ./hack/update-bazel.sh"
  exit 1
fi
```

(We also have an older `-dry-run` option, which just prints out what files it would change, without showing the diff, but it always exits 0. `-validate` implies `-dry-run` but exits nonzero if there are changes.)

PR #695, opened 2017-08-04
Cleaning up the GoSource provider usage

ianthehat:
We use it to manage all source lists, and propagate both original and modified
sources sets from rules.
This is a preparatory step for cleaning up both the library attribute and the
cover rules.

### DIFF
--- a/go/private/common.bzl
+++ b/go/private/common.bzl
@@ -68,3 +68,29 @@ def pkg_dir(workspace_root, package_name):
   if package_name:
     return package_name
   return "."
+
+def dict_of(st):
+  """Converts struct objects into dictionaries."""
+  data = dict()
+  for key in dir(st):
+    value = getattr(st, key, None)
+    if value != None: # skip methods
+      data[key] = value
+  return data
+
+def merge(a, b):
+  """Merges two structs by assuming that all fields present support +="""
+  data = dict_of(a)
+  for key, value in dict_of(b).items():
+    if key in data:
+      data[key] += value
+    else:
+      data[key] = value
+  return data
+
+def replace(a, b):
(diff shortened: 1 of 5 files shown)


go/private/common.bzl
```diff
@@ -68,3 +68,29 @@ def pkg_dir(workspace_root, package_name):
   if package_name:
     return package_name
   return "."
+
+def dict_of(st):
```
jayconrod: Can this be private?

go/private/common.bzl
```diff
@@ -68,3 +68,29 @@ def pkg_dir(workspace_root, package_name):
...
+  """Converts struct objects into dictionaries."""
+  data = dict()
+  for key in dir(st):
+    value = getattr(st, key, None)
+    if value != None: # skip methods
+      data[key] = value
+  return data
+
+def merge(a, b):
```
jayconrod: It seems strange for `merge` and `replace` to accept structs and return dicts. What do you think about taking a constructor as another argument and returning a struct?

go/private/common.bzl
```diff
@@ -68,3 +68,29 @@ def pkg_dir(workspace_root, package_name):
...
+    if key in data:
+      data[key] += value
+    else:
+      data[key] = value
+  return data
+
+def replace(a, b):
+  """Returns a dict where values from b are used when present in both a and b"""
+  data = dict_of(a)
+  for key, value in dict_of(b).items():
```
jayconrod: I think this is equivalent to `data.update(dict_of(b))`

ianthehat:
I am dropping this for #699.
Some of it is the same, but I removed the GoSource provider instead of improving it, pass around flat lists instead of exploded forms, and dropped replace and merge completely.

PR #701, opened 2017-08-08
Gazelle: improve documentation

jayconrod:


go/tools/gazelle/README.md
```diff
@@ -1,35 +1,274 @@
...
+### Command line
+
+```
+gazelle <command> [flags...] [package-dirs...]
+```
+
+The first argument to Gazelle may be one of the commands below. If no command
+is specified, `update` is assumed.
+
+<table class="table table-condensed table-bordered table-params">
```
ianthehat: Did you try to make the tables with [Markdown](https://help.github.com/articles/organizing-information-with-tables/)?
It' not quite as powerful, but it's so much cleaner.

jayconrod: I looked into this, but I don't think I can use it. Markdown tables can't include multi-line content like paragraphs (used in the flags table) or lists. Each row of the table has to fit on a single (long) line, so it will be even less readable than the HTML.

PR #708, opened 2017-08-10
Fix up creation of GoLibrary providers

ianthehat:
We create them directly inside emit_library_actions now, so we dont have to do a
struct to provider field by field copy.
the cgo_object is now passed in it's own provider, this is partly to enable a
future clean up of the cgo chain, and partly because it contains mutable fields.
This leaves GoLibrary as something that can now be put in a depset.

### DIFF
--- a/go/private/library.bzl
+++ b/go/private/library.bzl
@@ -98,7 +98,7 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
       lib_paths = direct_search_paths_race,
       direct_paths = direct_import_paths,
       out_object = race_object,
-      gc_goopts = gc_goopts + ["-race"],
+      gc_goopts = gc_goopts + ("-race",),
   )
   emit_go_pack_action(ctx, race_lib, [race_object] + extra_objects)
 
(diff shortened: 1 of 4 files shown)


go/private/library.bzl
```diff
@@ -113,63 +114,52 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
...
-    direct_deps = deps,
-    transitive_cgo_deps = transitive_cgo_deps,
-    transitive_go_libraries = transitive_go_library_deps + [out_lib],
-    transitive_go_libraries_race = transitive_go_library_deps_race + [race_lib],
-    transitive_go_library_paths = transitive_go_library_paths,
-    transitive_go_library_paths_race = transitive_go_library_paths_race,
-    gc_goopts = gc_goopts,
-  )
+  return [
+    GoLibrary(
```
ianthehat: Done

go/private/library.bzl
```diff
@@ -113,63 +114,52 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
...
-  lib_result = emit_library_actions(ctx,
+  golib, cgolib = emit_library_actions(ctx,
       srcs = ctx.files.srcs,
       deps = ctx.attr.deps,
       cgo_object = cgo_object,
       library = ctx.attr.library,
       want_coverage = ctx.coverage_instrumented(),
   )
 
+  depset([golib])
```
ianthehat: Nope, was left over from me trying to work out which fields were mutable as early as possible.
Thanks!

PR #709, opened 2017-08-10
Rewrite go_test in terms of emit_library_actions

ianthehat:
This prevents duplicating the compile and pack logic, and cleans up replicating
the handling of special cases (that was actaully wrong in the test code)

### DIFF
--- a/go/private/library.bzl
+++ b/go/private/library.bzl
@@ -16,7 +16,7 @@ load("@io_bazel_rules_go//go/private:common.bzl", "get_go_toolchain", "DEFAULT_L
 load("@io_bazel_rules_go//go/private:asm.bzl", "emit_go_asm_action")
 load("@io_bazel_rules_go//go/private:providers.bzl", "GoLibrary", "CgoLibrary")
 
-def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
+def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage, importpath, golibs=[]):
   go_toolchain = get_go_toolchain(ctx)
   dep_runfiles = [d.data_runfiles for d in deps]
   if library:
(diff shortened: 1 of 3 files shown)


go/private/test.bzl
```diff
@@ -59,45 +59,32 @@ def _go_test_impl(ctx):
...
+      srcs = [main_go],
+      deps = [],
+      cgo_object = None,
+      library = None,
+      want_coverage = False,
+      importpath = ctx.label.name + "~testmain~",
+      golibs = [golib],
+  )
+
+  go_library_paths=golib.transitive_go_library_paths
```
jayconrod: Nit: space around = in this block

Also, consider making this an if-else, rather than just an if. I'm a little uncomfortable with `go_libraries` and `go_library_paths` being assigned more than once along the same path.

go/private/library.bzl
```diff
@@ -16,7 +16,7 @@ load("@io_bazel_rules_go//go/private:common.bzl", "get_go_toolchain", "DEFAULT_L
 load("@io_bazel_rules_go//go/private:asm.bzl", "emit_go_asm_action")
 load("@io_bazel_rules_go//go/private:providers.bzl", "GoLibrary", "CgoLibrary")
 
-def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
+def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage, importpath, golibs=[]):
```
jayconrod: I think `lib_name` or `lib_stem` might be a better name than `importpath`. There are a couple cases in this PR where the argument is not actually an import path.

go/private/test.bzl
ianthehat: Did the spaces.
I  would still need to introduce the variables, and then reassign them inside the else, so I don't think it really adds anything (except more lines), but it would mean linkopts was different from the other two because it appends rather than replaces, I think the result would be more confusing, especially as I expect more modes to arrive that may not change all the values in the future.

go/private/library.bzl
ianthehat: it has to be importpath, it's stored on the returned provider as that.
It really must be the import path for the generated library, it's just that we generate synthetic non importable packages for main/test packages, and I can't just call them by their name because they can collide. I could pass a name separately from an import path, and then verify that they were the same if importpath is set, but that seems fairly pointless?

jayconrod: Yeah, introducing another very similar parameter would be confusing. Better keep it as-is.

PR #710, opened 2017-08-10
Report coverage metrics using original source filenames.

jmillikin-stripe:
Currently, the metrics are being emitted for the generated
coverage-instrumented file, which makes `go tool cover` unable to render
the coverage HTML correctly. Reporting coverage with the original source
filename lets `go tool cover` work.

### DIFF
--- a/go/private/library.bzl
+++ b/go/private/library.bzl
@@ -80,8 +80,9 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
     transitive_go_library_paths_race += golib.transitive_go_library_paths_race
 
   go_srcs = source.go
+  coverage_map = {}
   if want_coverage:
-    go_srcs = _emit_go_cover_action(ctx, out_object, go_srcs)
+    go_srcs, coverage_map = _emit_go_cover_action(ctx, out_object, go_srcs)
 
   emit_go_compile_action(ctx,
       sources = go_srcs,
@@ -121,6 +122,7 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage):
           searchpath = searchpath,
           searchpath_race = searchpath_race,
           srcs = join_srcs(struct(**transformed)),
+          coverage_filename_map = coverage_map,
           importpath = importpath,
           direct_deps = deps,
           transitive_cgo_deps = transitive_cgo_deps,
@@ -278,10 +280,13 @@ def _emit_go_cover_action(ctx, out_object, sources):
     sources: an iterable of Go source files.
 
   Returns:
-    A list of Go source code files which might be coverage instrumented.
+    A tuple of:
+      * A list of Go source code files which might be coverage instrumented.
+      * A dict from generated coverage file to original source file name.
   """
   go_toolchain = get_go_toolchain(ctx)
   outputs = []
+  outputs_map = {}
   # TODO(linuxerwang): make the mode configurable.
   count = 0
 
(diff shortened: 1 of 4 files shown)


bazel-io:

ianthehat:
I have just rewritten most of the cover system to work a very different way...
You can see it [here](https://github.com/ianthehat/rules_go/tree/cover)

I think that version will already fix this issue along with a host of other things as well, and enable much more.
Please feel free to try that branch and tell me if it does what you need, I plan to get it submitted probably tomorrow or Monday.
Do you have any suggestions for how to neatly integrate the cover tools use? How are you doing it at the moment, just using `bazel cover` to build the binaries and then running them by hand?
I am trying to decide how much to use the go tool way, and how much to go the bazel coverage route (convert to lcov format) and what the user experience should be.
Also if you need/want any of the other modes (count, atomic) let me know, and how you might want to trigger their use.

jmillikin-stripe:
Thanks, I'll try that new branch out tomorrow. At a glance it seems like it'll work, since `src.short_path` is getting propagated to `coverRegisterFile`.

My ideal behavior would have the generated test main automatically set `--test.coverprofile` to something that ends up in `bazel-testlogs`, and have Bazel run `go tool cover -html` as a Skylark action when a test is run in coverage mode (again, writing to testlogs).

Both of those are blocked on upstream Bazel work. Undeclared outputs don't seem to be working properly, and there's no support at all for coverage reports in Skylark. For the latter something could be hacked up with an `exec.Command` at the end of generated main, but that's useless if the files get thrown away by the sandbox.

I would very much prefer that Bazel coverage be implemented in a backend-agnostic way. If Bazel can't get Go-oriented coverage working, they'll be no hope for weirder stuff like Ruby's SimpleCov or Haskell's HPC. Intuitively it doesn't seem like this should be hard, as long as the coverage data files have predictable file paths.

jmillikin-stripe:
That branch seems to have broken coverage support entirely:

```
$ bazel coverage --test_output=errors --test_arg=-test.coverprofile=/tmp/go-rules-coverage.dat //tests/coverage:go_default_test
INFO: Using default value for --instrumentation_filter: "//tests/coverage".
INFO: Override the above default with --instrumentation_filter
INFO: Found 1 test target...
FAIL: //tests/coverage:go_default_test (see /private/var/tmp/_bazel_jmillikin/135605aba55136b03bd3dd1e66a9f2ac/execroot/io_bazel_rules_go/bazel-out/local-fastbuild/testlogs/tests/coverage/go_default_test/test.log).
INFO: From Testing //tests/coverage:go_default_test:
==================== Test output for //tests/coverage:go_default_test:
testing: cannot use -test.coverprofile because test binary was not built with coverage enabled
Use --strategy=TestRunner=standalone to disable sandboxing for the failing actions.
Target //tests/coverage:go_default_test up-to-date:
  bazel-bin/tests/coverage/go_default_test
INFO: Elapsed time: 0.301s, Critical Path: 0.06s
//tests/coverage:go_default_test                                         FAILED in 0.1s
  /private/var/tmp/_bazel_jmillikin/135605aba55136b03bd3dd1e66a9f2ac/execroot/io_bazel_rules_go/bazel-out/local-fastbuild/testlogs/tests/coverage/go_default_test/test.log

Executed 1 out of 1 test: 1 fails locally.
```

ianthehat:
Yeah, sorry that branch had a temporarily broken version in it because it was extracted from a stack and put up just so I could show you what I was doing, I did not test that I extracted it correctly.
The pull request #713 is the full form

jmillikin-stripe:
Thanks! I've verified that current HEAD sets coverage filenames correctly.

```
$ bazel coverage --test_output=errors --test_arg=-test.coverprofile=/tmp/go-rules-coverage.dat //tests/coverage:go_default_test
[...]
$ cat /tmp/go-rules-coverage.dat
mode: set
tests/coverage/lib.go:3.20,5.2 1 1
tests/coverage/lib.go:7.20,9.2 1 0
```

PR #711, opened 2017-08-10
Clean up transitive information

ianthehat:
Instead of having the GoLibrary provider have lots of transitive fields, we now
have one transitive field full of GoLibrary providers.
This cleans up a lot of code, and also allows much more powerful tools to be
written on top of the new collated GoLibrary set.

### DIFF
--- a/go/private/binary.bzl
+++ b/go/private/binary.bzl
@@ -30,10 +30,8 @@ def _go_binary_impl(ctx):
   # Default (dynamic) linking
   emit_go_link_action(
       ctx,
-      transitive_go_libraries=golib.transitive_go_libraries,
-      transitive_go_library_paths=golib.transitive_go_library_paths,
-      cgo_deps=golib.transitive_cgo_deps,
-      libs=depset([golib.library]),
+      library=golib,
+      mode="",
       executable=ctx.outputs.executable,
       gc_linkopts=gc_linkopts(ctx),
       x_defs=ctx.attr.x_defs,
@@ -47,10 +45,8 @@ def _go_binary_impl(ctx):
   static_executable = ctx.new_file(ctx.attr.name + ".static")
   emit_go_link_action(
       ctx,
-      transitive_go_libraries=golib.transitive_go_libraries,
-      transitive_go_library_paths=golib.transitive_go_library_paths,
-      cgo_deps=golib.transitive_cgo_deps,
-      libs=depset([golib.library]),
+      library=golib,
+      mode="",
       executable=static_executable,
       gc_linkopts=gc_linkopts(ctx) + static_linkopts,
       x_defs=ctx.attr.x_defs,
@@ -60,10 +56,8 @@ def _go_binary_impl(ctx):
   race_executable = ctx.new_file(ctx.attr.name + ".race")
   emit_go_link_action(
     ctx,
-    transitive_go_libraries=golib.transitive_go_libraries_race,
-    transitive_go_library_paths=golib.transitive_go_library_paths_race,
-    cgo_deps=golib.transitive_cgo_deps,
-    libs=depset([golib.race]),
+    library=golib,
+    mode="_race",
     executable=race_executable,
     gc_linkopts=gc_linkopts(ctx) + ["-race"],
     x_defs=ctx.attr.x_defs,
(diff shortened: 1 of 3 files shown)


go/private/binary.bzl
```diff
@@ -164,8 +158,7 @@ def _extract_extldflags(gc_linkopts, extldflags):
       filtered_gc_linkopts += [opt]
   return filtered_gc_linkopts, extldflags
 
-def emit_go_link_action(ctx, transitive_go_library_paths, transitive_go_libraries, cgo_deps, libs,
-                         executable, gc_linkopts, x_defs):
+def emit_go_link_action(ctx, library, mode, executable, gc_linkopts, x_defs):
```
jayconrod: Could you add a doc string to this function? I don't think the purpose of `mode` is obvious from the implementation of this function, or from a lot of the call sites, so it would be good to have some explanation.

go/private/binary.bzl
```diff
@@ -175,18 +168,24 @@ def emit_go_link_action(ctx, transitive_go_library_paths, transitive_go_librarie
...
   link_opts = [
       "-L", "."
   ]
-  for path in transitive_go_library_paths:
-    link_opts += ["-L", path]
+  libs = depset()
+  cgo_deps = depset()
+  for golib in depset([library]) + library.transitive:
+    libs += [getattr(golib, "library" + mode)]
+    link_opts += ["-L", getattr(golib, "searchpath" + mode)]
```
jayconrod: This feels under-abstracted to me. Perhaps `mode` could be a symbol enum and we could either turn it into a property name suffix inside this function or just do the branch. That would also let us add `"-race"` to `link_opts` implicitly, rather than relying on the caller to pass it in.

go/private/library.bzl
```diff
@@ -110,26 +97,23 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage, im
...
-          importpath = importpath,
-          direct_deps = deps,
-          transitive_cgo_deps = transitive_cgo_deps,
-          transitive_go_libraries = transitive_go_library_deps + [out_lib],
-          transitive_go_libraries_race = transitive_go_library_deps_race + [race_lib],
-          transitive_go_library_paths = transitive_go_library_paths,
-          transitive_go_library_paths_race = transitive_go_library_paths_race,
-          gc_goopts = gc_goopts,
-          runfiles = runfiles,
+          importpath = importpath, # The import path for this library
```
jayconrod: I think these comments should be in provider.bzl where `GoLibrary` is defined instead of here.

go/private/library.bzl
```diff
@@ -209,38 +193,31 @@ def go_importpath(ctx):
...
   return path
 
-def get_gc_goopts(ctx):
-  gc_goopts = tuple(ctx.attr.gc_goopts)
-  if ctx.attr.library:
-    gc_goopts += ctx.attr.library[GoLibrary].gc_goopts
-  return gc_goopts
-
-def emit_go_compile_action(ctx, sources, libs, lib_paths, direct_paths, out_object, gc_goopts):
+def emit_go_compile_action(ctx, sources, golibs, mode, out_object, gc_goopts):
```
jayconrod: Same comments about `mode`, doc string, and abstraction as `emit_go_link_action`.

go/private/library.bzl
ianthehat: I plan to fully document the providers when I am finished changing them, as they are part of the stable api, but the documentation needs to go in def.bzl where they are aliased for public consumption, and I am not sure how to document them cleanly, I think I will have to just make up a syntax, as there is no prior art to work from.

go/private/binary.bzl
ianthehat: I started doing this, but I need to change a few places and it's hard to do cleanly before the cover cl lands as that one rearranges the source tranformation step to not need the output object pre declared, I would like to fix this in a follow on cl.

go/private/binary.bzl
ianthehat: Done, mostly, more to come when I change mode handling in a follow up

go/private/library.bzl
ianthehat: Same replies as well :)

PR #712, opened 2017-08-11
Gazelle: Split LabelResolver into Labeler and Resolver

jayconrod:
This is a prerequisite for flat mode.

* Labeler is responsible for generating names for library, binary, and
  test rules in a directory described by a relative path from the
  repository root.
* Labeler has two implementations: hierarchicalLabeler is the current,
  default behavior. flatLabeler will be used in flat mode.
* Resolver resolves absolute import paths to labels. It uses Labeler
  in its implementation. The caller is now responsible for cleaning
  relative import paths.

### DIFF
--- a/go/tools/gazelle/config/config.go
+++ b/go/tools/gazelle/config/config.go
@@ -48,8 +48,11 @@ type Config struct {
 	// DepMode determines how imports outside of GoPrefix are resolved.
 	DepMode DependencyMode
 
-	// KnownImports is a list of imports to add to the external resolver cache
+	// KnownImports is a list of imports to add to the external resolver cache.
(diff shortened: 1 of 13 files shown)


jayconrod:
I was also thinking about this in the context of multiple Go packages in the same directory. I was planning to add an extra parameter to the methods in this interface for the package name.

That doesn't solve the conflict between binary and library rules though. My solution here was to add "_cmd" as a suffix for all binary rules. I don't think binary and library rules should have the same name if only one is present. If you start out with just a binary, but you add a _test.go file later, Gazelle would need to split it into binary, library, and test, changing the name of the binary. Seems like that could break things.

PR #713, opened 2017-08-11
Switching coverage to a better mechanism

ianthehat:
This works by tracking all coverage variables from the full transitive dependacy
set, which allows multiple packages to be covered at once.
This also normalises the names of cover variables so that two separate covers of
the same file will produce actions that merge.
It also tracks the cover variables directly, rather than inferring them from the
file names, which also means the test generator does not need to process the
generated files.

@linuxerwang you might want to check these changes, as it's based heavily on the work you did.

### DIFF
--- a/go/private/library.bzl
+++ b/go/private/library.bzl
@@ -22,10 +22,12 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage, im
   direct = depset(golibs)
   gc_goopts = tuple(ctx.attr.gc_goopts)
   cgo_deps = depset()
+  cover_vars = ()
   if library:
     golib = library[GoLibrary]
     cgolib = library[CgoLibrary]
     srcs = golib.transformed + srcs
+    cover_vars += golib.cover_vars
     direct += golib.direct
     dep_runfiles += [library.data_runfiles]
     gc_goopts += golib.gc_goopts
@@ -69,10 +71,10 @@ def emit_library_actions(ctx, srcs, deps, cgo_object, library, want_coverage, im
     transitive += golib.transitive
 
   go_srcs = source.go
-  cover_vars = ()
   if want_coverage:
-    go_srcs, cover_vars = _emit_go_cover_action(ctx, go_toolchain, go_srcs)
-    
+    go_srcs, cvars = _emit_go_cover_action(ctx, go_toolchain, go_srcs)
+    cover_vars += cvars
+
   emit_go_compile_action(ctx,
       sources = go_srcs,
       golibs = direct,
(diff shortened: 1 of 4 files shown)


go/private/test.bzl
```diff
@@ -41,21 +41,30 @@ def _go_test_impl(ctx):
...
+      '--rundir',
+      run_dir,
+      '--output',
+      main_go.path,
+  ]
+  cover_vars = []
+  covered_libs = []
+  for golib in depset([golib]) + golib.transitive:
+    if golib.cover_vars:
+        covered_libs += [golib]
```
jayconrod: Nit: 2-space indent

go/private/test.bzl
```diff
@@ -66,7 +75,7 @@ def _go_test_impl(ctx):
       library = None,
       want_coverage = False,
       importpath = ctx.label.name + "~testmain~",
-      golibs = [golib],
+      golibs = [golib] + covered_libs,
```
jayconrod: We'll end up passing `golib` in multiple times here. This could go in another CL, but I think `emit_library_actions` should accept a depset instead of a list for this parameter.

go/private/test.bzl
ianthehat: Done

go/private/test.bzl
ianthehat: It actually does not matter because emit_library_actions immediately collects the libs into depsets internally (direct and transitive)

go/private/library.bzl
```diff
@@ -272,8 +273,9 @@ def _emit_go_cover_action(ctx, out_object, sources):
       outputs += [src]
       continue
 
-    cover_var = "GoCover_%d" % count
-    out = ctx.new_file(out_object, out_object.basename + '_' + src.basename[:-3] + '_' + cover_var + '.cover.go')
+    cover_var = "Cover_" + src.basename[:-3].replace("-", "_").replace(".", "_")
+    cover_vars += ["{}={}".format(cover_var,src.short_path)]
```
linuxerwang: nit: one space after ",".

go/private/test.bzl
```diff
@@ -41,21 +41,30 @@ def _go_test_impl(ctx):
...
   main_go = ctx.new_file(ctx.label.name + "_main_test.go")
+  arguments = [
+      '--package',
+      golib.importpath,
+      '--rundir',
+      run_dir,
+      '--output',
+      main_go.path,
+  ]
+  cover_vars = []
```
linuxerwang: Seem not used.

go/tools/builders/generate_test_main.go
```diff
@@ -93,6 +62,7 @@ import (
 	"flag"
 	"log"
 	"os"
+	"fmt"
```
linuxerwang: Move up to line 63.

PR #714, opened 2017-08-11
Gazelle: prepare rules.Generator for flat mode

jayconrod:
* GeneratePrefix, GenerateRules, and GenerateLoad are now separate
  methods. Generate is a convenience method that calls them together.
* generator has a new property, buildRel, the path from the repository
  root to the directory containing the build file. In flat mode, this
  will be different from pkg.Rel.

This is a prerequisite for flat mode.

### DIFF
--- a/go/tools/gazelle/rules/generator.go
+++ b/go/tools/gazelle/rules/generator.go
@@ -34,19 +34,36 @@ type Generator interface {
 	// contains rules for each non-empty target in "pkg". It also contains
 	// "load" statements necessary for the rule constructors. If this is the
 	// top-level package in the repository, the file will contain a
-	// "go_prefix" rule.
+	// "go_prefix" rule. This is a convenience method for the other methods
+	// in this interface.
 	Generate(pkg *packages.Package) *bf.File
+
+	// GeneratePrefix generates a go_prefix rule. This should be in the
+	// top-level build file for the repository.
+	GeneratePrefix() bf.Expr
+
+	// GenerateRules generates a list of rules for targets in "pkg".
+	GenerateRules(pkg *packages.Package) []bf.Expr
+
+	// GenerateLoad generates a load statement for the symbols referenced
+	// in "stmts". Returns nil if rules is empty.
+	GenerateLoad(stmts []bf.Expr) bf.Expr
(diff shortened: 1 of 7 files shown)


jayconrod:
This isn't quite ready yet.

* `go_test` rules need an explicit `rundir` attribute, since by default they'll run in the Bazel package directory (in this case, the repo root), not the Go package directory.
* `go_library` rules for Go packages at the repo root need an explicit `importpath` to work with `go_importpath` in library.bzl. Probably better to set `importpath` for all of them though.

jayconrod:
Added some improvements to this. I still haven't fixed up cgo option paths, but that will come in another CL. PTAL.

go/tools/gazelle/rules/generator.go
```diff
@@ -227,10 +237,14 @@ func (g *generator) generateXTest(pkg *packages.Package, library string) bf.Expr
...
-	return g.generateRule(pkg.Rel, "go_test", name, "", "", pkg.HasTestdata, pkg.XTest)
+	var rundir string
+	if g.c.StructureMode == config.FlatMode {
+		rundir = pkg.Rel
+	}
+	return g.generateRule(pkg.Rel, "go_test", name, "", "", "", rundir, pkg.HasTestdata, pkg.XTest)
 }
 
-func (g *generator) generateRule(pkgRel, kind, name, visibility, library string, hasTestdata bool, target packages.Target) bf.Expr {
+func (g *generator) generateRule(pkgRel, kind, name, visibility, library, importpath, rundir string, hasTestdata bool, target packages.Target) bf.Expr {
```
ianthehat: This is starting to get very unwieldy, 7 strings and a bool, hard to read what the arguments mean at the call sites.
Maybe think about a nicer pattern (but not for this pr)

jayconrod: Completely agree. I drafted another commit this morning to fix this, but I'll need to merge it with this before sending it.

The plan is replace `generateRule` with a `commonAttrs` methods that returns `[]keyvalue` and only deals with attributes used by all rules. Other methods will add their own attributes, then call `newRule` directly.

PR #722, opened 2017-08-14
Gazelle: simplify rule generation

jayconrod:
* Replace generator.generateRule with commonAttrs, which just returns
  []keyvalue containing attributes relevant for all rules. Other
  methods add kind-specific attributes to this, then call newRule
  directly.
* newRule sorts attributes using the same order as buildifier.

### DIFF
--- a/go/tools/gazelle/rules/construct.go
+++ b/go/tools/gazelle/rules/construct.go
@@ -36,6 +36,8 @@ type globvalue struct {
 }
 
 func newRule(kind string, args []interface{}, kwargs []keyvalue) *bf.CallExpr {
+	sort.Sort(byAttrName(kwargs))
+
 	var list []bf.Expr
 	for _, arg := range args {
 		list = append(list, newValue(arg))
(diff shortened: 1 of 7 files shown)


go/tools/gazelle/rules/construct.go
```diff
@@ -145,6 +147,55 @@ func newValue(val interface{}) bf.Expr {
...
+
+var _ sort.Interface = byAttrName{}
+
+func (s byAttrName) Len() int {
+	return len(s)
+}
+
+// namePriority is the sort order for rule attributes.
+// Copied from github.com/bazelbuild/buildifier/build/rewrite.go
+var namePriority = map[string]int{
```
ianthehat: We already depend on buildifier, and NamePriority is a public symbol, why not use it directly rather than copy it?

jayconrod: Done. I had an old version where this wasn't exported yet.

PR #726, opened 2017-08-16
Make .bzl files compatible with future versions of Bazel

vladmos:
In Bazel 0.6 the `+=` operator on lists will have mutating semantics instead of copying. Some .bzl files have to be updated to be compatible with both semantics.

Protobuf is updated for the same reason.

### DIFF
--- a/go/private/cgo.bzl
+++ b/go/private/cgo.bzl
@@ -34,7 +34,7 @@ def _mangle(ctx, src):
 
 def _cgo_codegen_impl(ctx):
   go_toolchain = get_go_toolchain(ctx)
-  linkopts = ctx.attr.linkopts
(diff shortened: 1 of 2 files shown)


vladmos:
Please release a new version with this update so that it can be imported from other projects that use rules_go, such as rules_webtesting. Thanks.

ianthehat:
There are some major changes pending, that may be affected. How do you find which cases need fixing and which don't?
We can't easily tag a new version right now, there are lots of breaking changes we don't want to inflict on people in a trickle, so we are not going to tag a new version until after we can get toolchain support submitted and we are sure it is stable, which is waiting for Bazel 0.5.4 to ship. If it's really important then we need to branch from the current release tag and cherry pick.

PR #732, opened 2017-08-16
Add go_path rule

ianthehat:
This is used to produce a GOPATH like export of the transitive set of sources
used to build specific targets.
You can run it in either copy or link mode.
Link mode allows tools that mutate the original source to run as well

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_prefix")
+load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_prefix", "go_path")
 load("@io_bazel_rules_go//go/private:lines_sorted_test.bzl", "lines_sorted_test")
 load("@io_bazel_rules_go//proto:go_proto_library.bzl", "go_google_protobuf")
 
@@ -28,3 +28,21 @@ gazelle(
 
 # This could be any file, used as an anchor point for the directory in tests
 exports_files(["README.md"])
+
+go_path(
+    name = "all_srcs",
+    deps = [
+        "//go/tools/builders:asm",
+        "//go/tools/builders:compile",
+        "//go/tools/builders:embed",
+        "//go/tools/builders:filter_tags",
+        "//go/tools/builders:generate_test_main",
+        "//go/tools/builders:link",
+        "//go/tools/builders:cgo",
+        "//go/tools/builders:md5sum",
+        "//go/tools/gazelle/gazelle:gazelle",
+        "//go/tools/fetch_repo:fetch_repo",
+        "//go/tools/wtool:wtool",
+    ],
+    tags = ["manual"],
+)
--- a/go/private/path.bzl
+++ b/go/private/path.bzl
@@ -1,0 +1,101 @@
+# Copyright 2014 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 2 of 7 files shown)


BUILD
```diff
@@ -28,3 +28,21 @@ gazelle(
 
 # This could be any file, used as an anchor point for the directory in tests
 exports_files(["README.md"])
+
+go_path(
+    name = "all_srcs",
+    deps = [
+        "//go/tools/builders:asm",
```
jayconrod: How is this list generated? Should it include tests and libraries?

Might be good to add a script or command in the comments to generate this. Maybe `bazel query` will work after toolchains lands?

go/private/path.bzl
```diff
@@ -0,0 +1,101 @@
...
+EXPERIMENTAL: the go_path rule is still very experimental
+Please do not rely on it for production use, but feel free to use it and file issues
+""")
+  # First gather all the library rules
+  golibs = depset()
+  for dep in ctx.attr.deps:
+    golib = dep[GoLibrary]
+    golibs += [golib]
+    golibs += golib.transitive
+  seen_libs = {}
```
jayconrod: Nit: move these two lines below the comment, since they are part of the next logical step.

go/private/path.bzl
```diff
@@ -0,0 +1,101 @@
...
+        fail("Duplicate path {}".format(outpath))
+      seen_paths[outpath] = True
+      out = ctx.new_file(outpath)
+      package_files += [out]
+      outputs += [out]
+      if ctx.attr.mode == "copy":
+        ctx.template_action(template=src, output=out, substitutions={})
+      elif ctx.attr.mode == "link":
+        ctx.action(
+            command="ln -s $(realpath $1) $2",
```
jayconrod: Quote `$1` and `$2`

go/private/path.bzl
```diff
@@ -0,0 +1,101 @@
...
+        )
+      else:
+        fail("Invalid go path mode '{}'".format(ctx.attr.mode))
+    packages += [struct(
+      golib = golib,
+      dir = outdir,
+      files = package_files,
+    )]
+  envscript = ctx.new_file("{}/setenv.sh".format(ctx.label.name))
+  gopath = envscript.short_path[:-len(envscript.basename)]
```
jayconrod: I think this will leave a trailing /. Probably won't harm anything, but it would be good to trim it off.

go/private/path.bzl
```diff
@@ -0,0 +1,101 @@
...
+  #TODO: also set the GOROOT, need new toolchains first
+  ctx.file_action(envscript, content="""
+export GOPATH={gopath}
```
jayconrod: Quote `{gopath}`

BUILD
ianthehat: Right now there is no good automated way to generate that list....
In theory adding all test and binary targets to the list ought to be sufficient for most cases, but there is nothing to stop you having multiple of these rules with differing subsets of the dependency list, and if you have libraries with no tests you expose for other repositories to consume they would not be covered by that either.
For now, lets leave it fully manual until we know more about how people want to use it.

go/private/path.bzl
ianthehat: switched to rpartition instead

go/private/path.bzl
ianthehat: Done

go/private/path.bzl
ianthehat: Done

go/private/path.bzl
ianthehat: Done

Issue #721, opened 2017-08-14
Deprecate and remove go_prefix


jayconrod:
`go_prefix` is used to translate Bazel labels (for example, `//foo/bar:go_default_library`) to Go import paths (`example.com/repo/foo/bar`) and vice versa. There are a number of problems with `go_prefix`, and we'd like to stop using it and eventually stop supporting it.

* Go rules implicitly depend on `//:go_prefix`. This means there must be a `go_prefix` defined in the repository root package, even if there are no other Go rules in that package.
* This does not work well for Go subtrees that are part of larger repositories. The actual import path of libraries within the repository may not correspond to `<go_prefix>/<subtree>/<library>`.
* This does not work for vendoring. We have special cases for vendoring in several places.
* It's confusing and unexpected that `go_prefix` is required for `go_binary`. For binary-only repositories, it shouldn't be required at all.

We would like to move away from using `go_prefix`. `go_library` already has an `importpath` attribute, which allows the import path to be set explicitly. This is currently optional, but it may be required in the future.

We will deprecate `go_prefix` slowly in several steps to avoid breaking people.

* Gazelle will add `importpath` attributes to new and existing `go_library` rules. These import paths will match the import paths that already would be computed, based on the current `go_prefix` and the relative path within the repository.
* `go_prefix` will no longer be mandatory for `go_library` rules that specify `importpath`. We should be able to accomplish this with computed dependencies.
* Gazelle will no longer generate `go_prefix` rules. When Gazelle is run with the "fix" command, it will remove existing `go_prefix` rules.
* At some point in the distant future, we may remove `go_prefix` altogether (this might not actually happen, but it's the plan of record).

nictuku:
Not sure if it's related, but do we need importpath on `go_binary` and `go_image`? One currently has to set either //:go_prefix or importpath on the specific rule.

Does this Issue imply that this requirement will also eventually be removed? That would be nice.

jayconrod:
At the moment, `importpath` is needed for `go_binary` if `//:go_prefix` does not exist. I can't comment on `go_image`.

Internally, a `go_binary` rule describes a package that needs to be compiled, just like `go_library`. The `importpath` determines the filename for this package. The difference is that `go_binary` gets linked and doesn't get imported by other rules, so the filename doesn't matter much for compilation.

[...]

PR #733, opened 2017-08-16
Make go_prefix optional

### LINKED ISSUE #720: Hard to provider rules_go examples / tests
## Challenge
I provide Bazel rules.  I want to add Go examples / tests.  I do **NOT** want to require all users of my rules to import `rules_go`.

Today to use `rules_go`, three main elements are required.
1. Import `rules_go` in `WORKSPACE`
1. `go_prefix` in `/BUILD`.
1. `go_binary` in `.../BUILD`.

It is this second requirement that is problematic because it is on the `BUILD` path that Bazel loads for *every* `.bzl` file in the repo.

## Strawman
I think the simplest solution would be to enable folks in my situation to override the baked-in `//:prefix` target, so that we can move it into `.../BUILD` and enable rule consumers to dodge the `rules_go` requirement.

jayconrod:
go_prefix is used to generate Go import paths from Bazel labels. Until
now, it has been mandatory, since there's an implicit dependency from
all go_library, go_binary, and go_test rules on //:go_prefix. With
this change, that dependency is not present when the importpath
attribute is specified. This means that if all of the Go rules in a
repository specify importpath, there's no need to define a go_prefix
rule in the repository root.

Fixes #720
Related #721

### DIFF
--- a/go/private/binary.bzl
+++ b/go/private/binary.bzl
@@ -13,13 +13,19 @@
 # limitations under the License.
 
 load("@io_bazel_rules_go//go/private:common.bzl",
-  "get_go_toolchain",
-  "go_filetype",
-  "compile_modes",
-  "NORMAL_MODE",
-  "RACE_MODE",
+    "NORMAL_MODE",
+    "RACE_MODE",
+    "compile_modes",
+    "get_go_toolchain",
+    "go_filetype",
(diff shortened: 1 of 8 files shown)


jayconrod:
cc @mattmoor

PR #735, opened 2017-08-16
Go vet rule

ianthehat:
This adds go_vet_test
It uses the gopath structure built by one or more go_path rules to run go tool
vet, the test fails if the vet tool reports any issues.


go/private/tools.bzl
```diff
@@ -0,0 +1,69 @@
+# Copyright 2014 The Bazel Authors. All rights reserved.
```
jayconrod: s/2014/2017/

Will `go_fmt_test` and `go_fix_test` also be in this file in the future? tools.bzl is an unfortunately generic name, even though it is accurate. Maybe go_tools.bzl or tool_tests.bzl would be better.

go/private/tools.bzl
```diff
@@ -0,0 +1,69 @@
...
+
+def _go_vet_generate_impl(ctx):
+  print("""
+EXPERIMENTAL: the go_vet_test rule is still very experimental
+Please do not rely on it for production use, but feel free to use it and file issues
+""")
+  go_toolchain = get_go_toolchain(ctx)
+  script_file = ctx.new_file(ctx.label.name+".bash")
+  gopath = []
+  files = ctx.files.data
```
jayconrod: In #726, they copied lists from `ctx.attr` (as `ctx.attr.linkopts[:]`). That seems to imply `ctx.files.data` will be immutable in the future.

go/private/tools.bzl
ianthehat: Not all things would be tests.
I was going to stick all the tools in one file, but how about this structure, a tools directory with a bzl file per tool, so go/private/tools/vet.bzl

go/private/tools.bzl
ianthehat: Changed

spxtr:
This looks really great and solves one of our annoying issues! One question: It looks like the suggested way to use this is to say
```python
go_vet_test(
    name = "vet",
    data = [":all_srcs"]
) 
```

For Kubernetes, running `go vet` on every file takes about 5 minutes, but running it on a given package might just take a moment. Unless I'm misunderstanding this, even making a trivial docs change will lead to a 5-minute-long `//:vet` when we say `bazel test //...`. That's not optimal, but not the end of the world. It's what we are currently doing, anyway.

ixdy:
@spxtr maybe `gazelle` could auto-generate go vet tests in each package? possibly with a tag so it could be selected/deselected easily?

lint checks in bazel are weird.

jayconrod:
I'm speculating here, but for v2 of these tests, I wonder if we could produce some artifact after a successful test of a package, then only re-test when the inputs change.

The question is would it be possible to save a lot of extra configuration (avoiding `go_vet_test` rules in each package) or would you have to re-specify the dependency graph. I imagine it could be pretty simple, since we have all the dependency information in analysis in `go_path`.

Would also be good to test a subset of packages. Maybe vendored packages have warnings you don't want to hear about, for example.

ianthehat:
This is not intended as the long term interface to vet, it was written more as a proof that the path code was good enough to write these tools on top of. It is easy to make it slightly more flexible if you want to start using it now (for instance, making it non recursive, or support a list of packages to actually check), and I am happy to do so, but it will definitely change again in the future.

Per package vet tests with correct dependancies are clearly the way to go, but there are some costs.... Every separate vet test has to build it's own gopath copy that includes that package and all dependencies. If we add it as a separate rule, what about all the other checks (gosimple, gofmt, errcheck, staticcheck, unused etc)
One rule for each of those for each go_library rule is going to get very messy, and also it would be weird to have gazelle write rules for every static code checker in existence.

I am currently considering the concept of adding a new toolchain type (go_validator) and having a go_validate_test() rule that uses that toolchain to do it's work. Then you could control the set of static checks by switching the toolchain registry. This would allow gazelle to add exactly one extra rule per library, while still giving users full control. It's still very much a nascent idea though, I need to try it out and see if it holds up in the real world first.

PR #737, opened 2017-08-17
Gazelle: transform paths in cgo options

jayconrod:
generator.options now transforms package-relative paths in cgo options
into repo-root relative paths.

### DIFF
--- a/go/tools/gazelle/rules/generator.go
+++ b/go/tools/gazelle/rules/generator.go
@@ -322,10 +322,47 @@ func (g *generator) dependencies(imports packages.PlatformStrings, pkgRel string
 	return deps
 }
 
-func (g *generator) options(opts packages.PlatformStrings, rel string) packages.PlatformStrings {
-	// TODO(jayconrod): paths in options (for example, include directories) should
-	// be interpreted relative to the Go package. If the Go package is different
-	// than the Bazel package (as it may be in flat mode), these paths will not
-	// be correct. We should adjust them here, but they are difficult to identify.
+// options transforms package-relative paths in cgo options into repository-
+// root-relative paths that Bazel can understand. For example, if a cgo file
+// in //foo declares an include flag in its copts: "-Ibar", this method
+// will transform that flag into "-Ifoo/bar".
+func (g *generator) options(opts packages.PlatformStrings, pkgRel string) packages.PlatformStrings {
+	fixPath := func(opt string) string {
+		if strings.HasPrefix(opt, "/") {
+			return opt
+		}
+		return path.Clean(path.Join(pkgRel, opt))
+	}
+
+	fixOpts := func(optStr string) (string, error) {
+		opts := strings.Split(optStr, " ")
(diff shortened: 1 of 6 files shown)


go/tools/gazelle/rules/generator.go
```diff
@@ -322,10 +322,47 @@ func (g *generator) dependencies(imports packages.PlatformStrings, pkgRel string
...
+			if isPath {
+				opts[i] = fixPath(opt)
+				isPath = false
+				continue
+			}
+
+			for _, prefix := range prefixes {
+				if strings.HasPrefix(opt, prefix) {
+					if len(opt) > len(prefix) {
+						opts[i] = prefix + fixPath(opt[len(prefix):])
```
ianthehat: I think this would also handle incorrect things, -framework_root would try to fixupPath("_root")
It's only valid for single character forms I think  (which would be length two as you include the -), -quotemypath is not valid, you must have the space (not sure if it accepts -iquote=mypath, never tried, but if it does then you need to handle that as a third case)

go/tools/gazelle/rules/generator.go
```diff
@@ -322,10 +322,47 @@ func (g *generator) dependencies(imports packages.PlatformStrings, pkgRel string
...
+func (g *generator) options(opts packages.PlatformStrings, pkgRel string) packages.PlatformStrings {
+	fixPath := func(opt string) string {
+		if strings.HasPrefix(opt, "/") {
+			return opt
+		}
+		return path.Clean(path.Join(pkgRel, opt))
+	}
+
+	fixOpts := func(optStr string) (string, error) {
+		opts := strings.Split(optStr, " ")
```
ianthehat: This really ought to use something like https://github.com/mattn/go-shellwords so it's safe for quoted args with spaces and things...

jayconrod: Yeah, we shouldn't split and join so much. I've changed `packages.taggedOpts` back to containing a `[]string`. In order to preserve the grouping and to avoid introducing a new data structure that contains `[][]string`, I'm adding a sentinel string at the end of each group. `generator.options` fixes paths in individual options, then calls `packages.JoinOptions`, which knows what to do with that sentinel string.

go/tools/gazelle/rules/generator.go
```diff
@@ -322,10 +322,47 @@ func (g *generator) dependencies(imports packages.PlatformStrings, pkgRel string
...
+		prefixes := []string{"-I", "-L", "-F", "-iquote", "-isystem", "-framework"}
```
jayconrod: Done. I don't expect this to be configurable, but it could be in config/constants.go if it were.

go/tools/gazelle/rules/generator.go
jayconrod: Fair point about `-framework`. I've split the prefixes into short and long lists. The short options like -I are in both.

`-iquotemypath` actually is recognized by gcc, but I've stopped handling it here because it is very strange to see. Would be easy to add it back in though.

`-iquote=mypath` is not supported. I don't know of any options with paths that take that form.

PR #750, opened 2017-08-28
Optionally omit external repositories

drigz:
This follows the style used by rules_closure:
https://github.com/bazelbuild/rules_closure/blob/master/closure/repositories.bzl

My use case: I'm extending buildtools to work with BUILD files for external repositories. Until this is merged upstream, I want to use my own fork in my workspace. However, `go_repositories` insists on using its own version, and Bazel won't let me override this:

`ERROR: Cannot redefine repository after any load statement in the WORKSPACE file`

I could include my fork with a different name, but Bazel doesn't like that either (although just a warning right now):

`Workspace name [in repo WORKSPACE] (@com_github_bazelbuild_buildtools) does not match the name given in the repository's definition (@com_github_drigz_buildtools); this will cause a build error in future versions`

Passing `omit_com_github_bazelbuild_buildtools = True` makes it easy to use my fork.


ianthehat:
I think this is a bad pattern that we should be trying to kill rather than to spread. If we end up depending on something with lots of dependencies (which is likely as gazelle gets more complicated) we end up with a massive list of arguments to that function.
This is something that is much more general and that Bazel needs to have good features for. I believe in the longer term, nested workspaces are supposed to fix all these issues, but in the shorter term it should either be possible to have a first-declaration wins model so you can load it in your workspace and have Bazel ignore the declaration by rules_go, or have a way to ask the repository context if a given external name is already taken so you don't need an explicit parameter. The latter is probably fairly easy and a decent general purpose tool anyway.

jayconrod:
+1 to what @ianthehat said. This will be unsustainable as more dependencies are added. Bazel itself should be improved so that this is more ergonomic. For reference there is a [recursive WORKSPACE design document](https://www.bazel.build/designs/2016/09/19/recursive-ws-parsing.html), which is supposed to solve these problems. I don't believe this is actively being worked on, but I'll follow up with the Bazel core team to see what's happening.

Some possible workarounds in the short term:

* Use `--override_repository` on the Bazel command line. You can add this to the `tools/bazel.rc` file in your project so you don't have to type it out every time.
* Fork rules_go and apply this change.
* Inline `go_repositories()` in your WORKSPACE file, changing `com_github_bazelbuild_buildtools` as needed. This means you will depend on some private repository rules that may change.

jayconrod:
[Question I sent to bazel-discuss](https://groups.google.com/forum/#!topic/bazel-discuss/EBeFucaAg3A). We'll see what's happening with recursive workspaces.

drigz:
Thanks for all the info - I've gone with a fork for now, so let's see what the Bazel devs recommend as the right way to do it.

PR #752, opened 2017-08-28
Isolate access to the cpp fragment

ianthehat:
This collects the information needed into the go toolchain, rather than
accessing the cpp fragment all over the place.
Theoretically this would make pure go cross compilation easier, in practice you
still have to set --cpu correctly, and bazel still sulks if it has no valid cc
toolchain for that cpu.

### DIFF
--- a/go/private/cgo.bzl
+++ b/go/private/cgo.bzl
@@ -32,12 +32,8 @@ def _mangle(ctx, src):
     return mangled_stem, src_ext 
 
 def _c_filter_options(options, blacklist):
-  filtered = []
-  for opt in options:
-    if any([opt.startswith(prefix) for prefix in blacklist]):
-      continue
-    filtered.append(opt)
-  return filtered
+  return [opt for opt in options
+        if not any([opt.startswith(prefix) for prefix in blacklist])]
 
 def _cgo_codegen_impl(ctx):
   go_toolchain = ctx.toolchains["@io_bazel_rules_go//go:toolchain"]
--- a/go/private/cgo.bzl
+++ b/go/private/cgo.bzl
@@ -14,7 +14,6 @@
 
 load("@io_bazel_rules_go//go/private:common.bzl", "dict_of", "split_srcs", "join_srcs", "pkg_dir")
 load("@io_bazel_rules_go//go/private:library.bzl", "go_library")
-load("@io_bazel_rules_go//go/private:binary.bzl", "c_linker_options")
 load("@io_bazel_rules_go//go/private:providers.bzl", "GoLibrary")
 
 def _cgo_select_go_files_impl(ctx):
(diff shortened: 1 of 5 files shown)


go/private/cgo.bzl
```diff
@@ -32,18 +31,28 @@ def _mangle(ctx, src):
     mangled_stem = ctx.attr.out_dir + "/" + src_stem.replace('/', '_')
     return mangled_stem, src_ext 
 
+def _c_filter_options(options, blacklist):
+  filtered = []
```
jayconrod: Shorter but marginally less readable. Feel free to ignore:

```
return [opt for opt in options
        if not any([opt.startswith(prefix) for prefix in blacklist])]
```

go/private/go_toolchain.bzl
```diff
@@ -57,6 +63,7 @@ go_toolchain_core_attrs = {
     "cgo_link_flags": attr.string_list(default=[]),
     "goos": attr.string(mandatory = True),
     "goarch": attr.string(mandatory = True),
+    "_external_linker": attr.label(default=_get_linker),
```
jayconrod: Why not inline `_get_linker`?

go/private/go_toolchain.bzl
```diff
@@ -93,4 +100,26 @@ go_toolchain_flags = rule(
...
         "compile_flags": attr.string_list(mandatory=True),
     },
-)
\ No newline at end of file
+)
+
+def _external_linker_impl(ctx):
+  cpp = ctx.fragments.cpp
+  features = ctx.features
+  options = cpp.compiler_options(features)
```
jayconrod: Nit: I think this would be more readable as one big statement using + instead of several statements with +=.

go/toolchain/toolchains.bzl
```diff
@@ -1,4 +1,4 @@
-load('//go/private:go_toolchain.bzl', 'go_toolchain')
+load('//go/private:go_toolchain.bzl', 'go_toolchain', "external_linker")
```
jayconrod: Micro-nit: alphabetize, consistent quotes.

go/private/cgo.bzl
ianthehat: Done

go/private/go_toolchain.bzl
ianthehat: The point of this change was for _external_linker to become a computed dependency, where it reduces to None if there is no external linker available, thus breaking the dependency on the cpp fragment where you are in pure go mode.

go/private/go_toolchain.bzl
ianthehat: Done

go/toolchain/toolchains.bzl
ianthehat: Done

PR #755, opened 2017-08-28
Fix tags in README.md

jayconrod:
[skip ci]
[ci skip]


jayconrod:
cc @ianthehat This is my fault, should have pointed out that these need to be updated.

PR #757, opened 2017-08-28
Merge the declaration of bootstrap and normal toolchains

ianthehat:
It will be possible to write rules that are agnostic as to the toolchain they
get this way. It was only needed when toolchains had to know their own type,
before the toolchain system changed the registration method.


go/private/go_toolchain.bzl
```diff
@@ -52,34 +51,62 @@ def _get_linker():
...
+        "root": attr.label(mandatory = True),
+        "go": attr.label(mandatory = True, allow_files = True, single_file = True, executable = True, cfg = "host"),
+        "tools": attr.label(mandatory = True, allow_files = True),
+        "stdlib": attr.label(mandatory = True, allow_files = True),
+        "headers": attr.label(mandatory = True),
+        "link_flags": attr.string_list(default=[]),
+        "cgo_link_flags": attr.string_list(default=[]),
+        "goos": attr.string(mandatory = True),
+        "goarch": attr.string(mandatory = True),
+        "bootstrap":attr.bool(mandatory = True),
```
jayconrod: Nit: space after `:`

Would also be good to add space around the `=` after `default` arguments below.

go/toolchain/toolchains.bzl
```diff
@@ -147,13 +146,13 @@ def _bootstrap(base):
   bootstrap["name"] = "bootstrap-" + base["name"]
   bootstrap["impl"] = "bootstrap-" + base["impl"]
   bootstrap["typ"] = "@io_bazel_rules_go//go:bootstrap_toolchain"
-  bootstrap["declare"] = go_bootstrap_toolchain
+  bootstrap["bootstrap"] = True
   return bootstrap
 
 def _default(base):
```
jayconrod: I don't think I understand the purpose of default toolchains. What uses them?

PR #758, opened 2017-08-28
Adding info command

ianthehat:
This adds a special target so that

`bazel run @io_bazel_rules_go//:go_info`

Will tell you information about the go toolchain being used by the rules.
This is useful for debugging, checking cross compile platform settings or just
replicatiing the internal commands for normal shell use.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,5 +1,6 @@
 load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_prefix", "go_path", "go_vet_test")
 load("@io_bazel_rules_go//go/private:lines_sorted_test.bzl", "lines_sorted_test")
+load("@io_bazel_rules_go//go/private:info.bzl", "go_info")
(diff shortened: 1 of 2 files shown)


krishicks:
Is it possible to use this to stamp a binary with the Go version it was compiled with?

jayconrod:
@krishicks Yes! There's some information on how to do this at [Defines and stamping](https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#defines-and-stamping).

krishicks:
I did see that section and have created a status script for outputting some git  information; did you have an idea how to integrate go info into that?

jayconrod:
I wouldn't recommend embedding `go_info` into a binary. It contains some absolute paths, so you wouldn't have a reproducible build.

Do you just need the Go version? [`runtime.Version()`](https://godoc.org/runtime#Version) would be a better option.

krishicks:
We embed the go version and the uname of the platform it was built with.

krishicks:
runtime.Version() sounds great; that'll work.

PR #772, opened 2017-08-30
Clean up go sdk rules

ianthehat:
They are going to become part of the public interface to allow custom
toolchains.
This simplifies that interface, and then adds the ability to specify either an
archive or an already unpacked goroot

### DIFF
--- a/go/private/toolchain.bzl
+++ b/go/private/toolchain.bzl
@@ -12,23 +12,66 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-def _go_sdk_repository_impl(ctx):
-  goroot = ctx.path(".")
-  ctx.template("BUILD.bazel", 
-    Label("@io_bazel_rules_go//go/private:BUILD.sdk.bazel"),
-    substitutions = {"{goroot}": str(goroot)}, 
-    executable = False,
-  )
+def _go_host_sdk_impl(ctx):
+  root = _detect_host_sdk(ctx)
+  _local_sdk(ctx, root)
+  _skd_build_file(ctx, root)
+
+go_host_sdk = repository_rule(_go_host_sdk_impl, environ = ["GOROOT"])
+"""go_host_sdk is a specialization of go_sdk just to add the GOROOT dependancy"""
+
+def _go_sdk_impl(ctx):
+  _skd_build_file(ctx, str(ctx.path(".")))
+  if ctx.attr.url:
+    if ctx.attr.root:
+      fail("url and root cannot both be set on go_sdk, got {} and {}".format(ctx.attr.url, ctx.attr.root))
+    _remote_sdk(ctx, ctx.attr.url, ctx.attr.strip_prefix, ctx.attr.sha256)
+  elif ctx.attr.root:
+    _local_sdk(ctx, ctx.attr.root)
+    _skd_build_file(ctx, ctx.attr.root)
+  else:
+    root = _detect_host_sdk(ctx)
+    _local_sdk(ctx, root)
+    _skd_build_file(ctx, root)
+    
+  # Build the standard library for valid cross compile platforms
+  #TODO: fix standard library cross compilation
+  if ctx.name.endswith("linux_amd64") and ctx.os.name == "linux":
+    _cross_compile_stdlib(ctx, "windows", "amd64")
+  if ctx.name.endswith("darwin_amd64") and ctx.os.name == "mac os x":
+    _cross_compile_stdlib(ctx, "linux", "amd64")
+
+go_sdk = repository_rule(
+    implementation = _go_sdk_impl, 
+    attrs = {
+        "root" : attr.string(),
+        "url" : attr.string(),
+        "strip_prefix" : attr.string(default="go"),
+        "sha256" : attr.string(),
+    },
+)
+"""
+    go_sdk is a rule for adding a new go SDK to the available set.
+    This does not make the sdk available for use directly, it needs to be exposed through a toolchain.
+    If you do not specify url or path, then it will attempt to detect the installed version of go.
+"""
+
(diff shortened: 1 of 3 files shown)


go/private/toolchain.bzl
```diff
@@ -12,24 +12,67 @@
...
+  ctx.download_and_extract(
+      url = ctx.attr.url,
+      stripPrefix = ctx.attr.strip_prefix,
+      sha256 = ctx.attr.sha256)
+
+def _local_sdk(ctx, root):
+  for entry in ["src", "pkg", "bin"]:
+    ctx.symlink(root+"/"+entry, entry)
+
+def _skd_build_file(ctx, goroot):
```
jayconrod: s/skd/sdk/

go/private/toolchain.bzl
```diff
@@ -12,24 +12,67 @@
...
+  #TODO: fix standard library cross compilation
   if ctx.name.endswith("linux_amd64") and ctx.os.name == "linux":
     _cross_compile_stdlib(ctx, "windows", "amd64")
   if ctx.name.endswith("darwin_amd64") and ctx.os.name == "mac os x":
     _cross_compile_stdlib(ctx, "linux", "amd64")
 
+go_sdk = repository_rule(
+    implementation = _go_sdk_impl, 
+    attrs = {
+        "root" : attr.string(),
```
jayconrod: I think this should be called `path` instead of `root` for consistency with `local_repository`.

Also, micro-nit: no space before `:`

go/private/toolchain.bzl
```diff
@@ -12,24 +12,67 @@
...
+def _go_host_sdk_impl(ctx):
+  root = _detect_host_sdk(ctx)
+  _local_sdk(ctx, root)
+  _skd_build_file(ctx, root)
+
+go_host_sdk = repository_rule(_go_host_sdk_impl, environ = ["GOROOT"])
+"""go_host_sdk is a specialization of go_sdk just to add the GOROOT dependancy"""
 
+def _go_sdk_impl(ctx):
+  _skd_build_file(ctx, str(ctx.path(".")))
```
jayconrod: Should this be inside the branch?

Also, will this end up being an absolute path? Not sure whether it will work with remote execution.

go/private/toolchain.bzl
```diff
@@ -12,24 +12,67 @@
...
+    attrs = {
+        "root" : attr.string(),
+        "url" : attr.string(),
+        "strip_prefix" : attr.string(default="go"),
+        "sha256" : attr.string(),
+    },
+)
+"""
+    go_sdk is a rule for adding a new go SDK to the available set.
+    This does not make the sdk available for use directly, it needs to be exposed through a toolchain.
```
jayconrod: It would be good to add a link to some documentation on how to do this, even if it's not written yet. Preferably an .md file on defining toolchains.

go/private/toolchain.bzl
```diff
@@ -12,24 +12,67 @@
...
+      stripPrefix = ctx.attr.strip_prefix,
+      sha256 = ctx.attr.sha256)
+
+def _local_sdk(ctx, root):
+  for entry in ["src", "pkg", "bin"]:
+    ctx.symlink(root+"/"+entry, entry)
+
+def _skd_build_file(ctx, goroot):
+  ctx.template("BUILD.bazel", 
+    Label("@io_bazel_rules_go//go/private:BUILD.sdk.bazel"),
```
jayconrod: Nit: 4-space indent

go/private/toolchain.bzl
```diff
@@ -12,24 +12,67 @@
...
 
+go_sdk = repository_rule(
+    implementation = _go_sdk_impl, 
+    attrs = {
+        "root" : attr.string(),
+        "url" : attr.string(),
+        "strip_prefix" : attr.string(default="go"),
+        "sha256" : attr.string(),
+    },
+)
```
jayconrod: Include `environ` with `GOROOT`.

go/private/toolchain.bzl
```diff
@@ -52,39 +95,15 @@ def _cross_compile_stdlib(ctx, goos, goarch):
...
-    implementation = _go_host_sdk_repository_impl, 
-    attrs = {},
-    environ = [
-      "GOROOT",
-    ],
-)
+    return ctx.os.environ["GOROOT"]
+  res = ctx.execute(["go", "env", "GOROOT"])
+  if res.return_code:
+      fail("Could not detect host go version")
```
jayconrod: Nit: 2-space indent

go/private/toolchain.bzl
ianthehat: Yes, it was safe but wrong.
Remote execution is a problem for another day (this does not add any new absolute paths).
I think it will be okay because under remote execution conditions you will have constructed an sdk that reports correct absolute paths for the execution container, not the local one.

go/private/toolchain.bzl
ianthehat: Done

go/private/toolchain.bzl
ianthehat: That's what go_host_sdk is for.
I did not want someone playing with their environment to invalidate the bazel cache when they are not even using it. If we don't care about that (maybe we should not) then I could merge the two rules.

go/private/toolchain.bzl
ianthehat: I have started, with an example, but it's right at the bottom of the change stack...

go/private/toolchain.bzl
ianthehat: Done

go/private/toolchain.bzl
ianthehat: Done

go/private/toolchain.bzl
ianthehat: Done

PR #774, opened 2017-08-31
Library as source

ianthehat:
This removes the concept of the "library" at the action level, instead allowing GoLibrary providers to be used as source entries.
This enables all the features by library, whilst also allowing other uses (in particular, code generators can use this mechanism to provide source and deps pairs to go library rules easily, looking at you proto...)
I would like to deprecate the library attribute in favour of this. For example
```
go_test(
    name = "go_default_test",
    size = "small",
    srcs = ["bindata_test.go"],
)
```
would become
```
go_test(
    name = "go_default_test",
    size = "small",
    srcs = [
      "bindata_test.go",
      ":go_default_library",
    ],
)
```

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -18,6 +18,7 @@ load("@io_bazel_rules_go//go/private:go_repository.bzl",
 load("@io_bazel_rules_go//go/private:providers.bzl",
     _GoLibrary = "GoLibrary",
     _GoBinary = "GoBinary",
+    _GoEmbed = "GoEmbed",
 )
 load("@io_bazel_rules_go//go/private:repositories.bzl",
     "go_rules_dependencies",
@@ -54,6 +55,14 @@ It provides the following fields:
 GoBinary = _GoBinary
 """
 This is the provider used to expose a go binary to other rules.
+It provides the following fields:
+  TODO: List all the provider fields here
+"""
+
+GoEmbed = _GoEmbed
+"""
+This is the provider used to provide paired source and deps to a go library.
+This should generally be the provider returned by code generators.
 It provides the following fields:
   TODO: List all the provider fields here
 """
--- a/go/private/actions/library.bzl
+++ b/go/private/actions/library.bzl
@@ -21,31 +21,32 @@ load("@io_bazel_rules_go//go/private:common.bzl",
 load("@io_bazel_rules_go//go/private:providers.bzl", 
     "GoLibrary", 
     "CgoLibrary",
+    "GoEmbed",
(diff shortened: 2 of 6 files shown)


pmbethe09:
Doesn't this deviate from how the internal go_test rule is written though?

ianthehat:
Yes, but that doesn't bother me very much...
It's a fairly easy migration or internal patch, and I would rather keep the skylark rules as clean as possible and migrate Google rather than make everybody do things our broken way or leave documented attributes in place that we don't want people to use.

pmbethe09:
sgtm, just wanted to make sure you had considered that issue.

ianthehat:
I would much rather you point out things that I have thought of than not mention things I failed to notice :)
Specifically my plan in this case is to remove all uses of library in the documentation/examples and gazelle generated files, leave the attribute in place and have a deprecation `fail` statement if it's set, but then also leave the code that adds it to the srcs in place. That way patching the rules for Google just involves disabling those fail statements. Once all of google3 is migrated at some point in the future, we can then remove it totally.
But first we have to discuss if we like the idea and whether this is the right interface for it.

jayconrod:
Discussed offline. My preference is to have another attribute for this. It's too much of a surprise to be able to pull in deps via srcs.

It's hard to come up with a name for this. `libraries` doesn't seem adequate. Here are some ideas (not sure I'm happier with any of these though).

* embeds (could be confused with `go_embed_data` though)
* imports (could be confused with Go imports, deps)
* using
* inherits
* compose
* combine
* extends
* includes (lots of baggage from C/C++)
* union

pmbethe09:
The current 'library' I always read as 'for_library' or 'of_library' or 'inside_library', implying an internal unit-test.  Not clear to me that any of the new names are any better.

ianthehat:
If you think about the proto use case (or other equivalent code generators), you want a thing that provides some source files and also some deps that are needed by those source files.
You also may need more than one of them.
library makes zero sense in that context (and would need to be pluralized anyway). the go_library->go_test link is one of many uses for this feature, and while it may be one of the more complicated ones, it's probably also the least interesting.

I am currently leaning towards embed.

ianthehat:
So this is the new proposal
I think it's a little clearer, and much easier to write code generators using it.

ianthehat:
@pmbethe09 I am going to submit this change, but I would still like to hear your thoughts on the newer version, and am still open to changing it if needed.

pmbethe09:
embed makes sense to me.  overall LGTM

PR #775, opened 2017-08-31
readme: fix string quote nit for external=vendored.

htuch:


bazel-io:

PR #778, opened 2017-08-31
Expose cgo_exports

### LINKED ISSUE #776: go_library(cgo=True, ...) emit .h file with exports as an output
I'm using a simple go file that looks something like this:

main.go
```
package main
import "C"

//export MyFunction
func MyFunction() {}

func main() {}
```

I'd like to use this as a src or dep in a cc_library, but the _cgo_export.h that I'm expecting to be emitted as an output doesn't seem to be propagating to the cc_library.

I know the _cgo_export.h file is being exported, because my work around right now is to manually copy it out of its generated location and commit it into my project. Any tips here about what I should do to ensure the file is properly exported?

ianthehat:
Fixes #776

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -113,6 +113,7 @@ def _cgo_codegen_impl(ctx):
       go_files = go_outs,
       main_c = depset([cgo_main]),
       cgo_deps = deps,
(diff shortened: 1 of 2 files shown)


ianthehat:
Jenkins test this please

PR #779, opened 2017-08-31
proto: allow shell vars to propagate to protoc.

htuch:
Similar to https://github.com/google/protobuf/pull/2508, when using
protoc with non-standard LD_LIBRARY_PATH, we need to be able to have it
pick up from the environment the correct LD_LIBRARY_PATH.

### DIFF
--- a/proto/go_proto_library.bzl
+++ b/proto/go_proto_library.bzl
@@ -163,7 +163,8 @@ def _go_proto_library_gen_impl(ctx):
       mnemonic="GoProtocGen",
       env = {"PATH": root_prefix + "/" + ctx.files.protoc_gen_go[0].dirname +
              ":/bin:/usr/bin"},  # /bin/sed for linux, /usr/bin/sed for macos.
-      executable=run)
+      executable=run,
+      use_default_shell_env=True)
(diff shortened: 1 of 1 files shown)


bazel-io:

ianthehat:
I think that use_default_shell_env is fundamentally incompatible with remote execution, this is probably not the right way to solve the problem.
Why is protoc special (compared to any other host binary), why does it need a custom LD_LIBRARY_PATH?

htuch:
It's built under Bazel using a C++ toolchain configured with CC/CXX/LD_LIBRARY_PATH. Other host binaries are built with whatever the host base image is using. FWIW, we're building on Goobuntu with CXX/LD_LIBRARY_PATH pointing at crosstools. Would anyone want to remotely execute this rule when building with a custom local toolchain?

htuch:
Closing this out based on offline discussion with @ianthehat. We have a bunch of needs (this PR, arbitrary plugins) that don't match well with the current `go_proto_library` rule, which is in a state of transition. The recommended approach is to fork `go_proto_library.bzl` in our repo (CC: @rodaine).

PR #793, opened 2017-09-05
Pass original go sources through to GoLibrary provider

ains:


bazel-io:

go/private/actions/library.bzl
```diff
@@ -127,7 +127,8 @@ def emit_library(ctx, go_toolchain, srcs, deps, cgo_object, embed, want_coverage
           importpath = importpath, # The import path for this library
           direct = direct, # The direct depencancies of the library
           transitive = transitive, # The transitive set of go libraries depended on
-          srcs = depset(srcs), # The original sources
+          srcs = depset(srcs), # The sources used for building (contains cgo generated outputs)
```
ianthehat: if srcs is the originals, I don't think we need the transformed ones returned in this provider at all, if they are really needed they can be picked out of the GoEmbed provider below.

go/private/actions/library.bzl
```diff
@@ -26,7 +26,7 @@ load("@io_bazel_rules_go//go/private:providers.bzl",
     "searchpath_attr",
 )
 
-def emit_library(ctx, go_toolchain, srcs, deps, cgo_object, embed, want_coverage, importpath, golibs=[]):
+def emit_library(ctx, go_toolchain, srcs, pure_go_srcs, deps, cgo_object, embed, want_coverage, importpath, golibs=[]):
```
ianthehat: I think I would probably swap these.
srcs should always be the original sources, the fact that we need to pass in cgo transformed ones is a temporary hack, and should be the extra parameter.

ains:
Thanks for the review @ianthehat PR still isn't ready for re-review, was just testing something on Travis.

However, it looks like things are going to get a little tricker if we always make srcs the "original" sources. The `setup_cgo_library` func does the work of filtering out non all .go sources and returning a struct with `go_srcs`.

If we pass the srcs directly from say, `go_library_macro` to `go_library` we have to allow all types of file in the srcs attribute and then filter later on for builds to continue to work.

Alternatively we can make setup_cgo_library return `original_go_srcs` which is just the original srcs list filtered to only contain go files (no generated code), but that seems a weird that it's now a subset of the original sources passed to the `go_library_macro` fn.

Last choice I can think of is similar to the original PR where we don't change too much, in which we introduce a new parameter which is all of the original sources, which we just thread through into a new field of the GoLibrary provider.

Thoughts?

ianthehat:
Where would the other file types cause a problem? `go_library` when `cgo=True` is going to get the transformed source list in the new `cgo_srcs` parameter and compile that, so it's not going to be a problem there, and I don't think much else will care.
The most important thing is that the `go_library` macro is supposed to hide the fact that we currently have to split into multiple rules. As soon as we can get to cc compile actions directly, it will become a single rule again, so it's inputs and outputs must match that desired end goal so we can still make that change in the future.
This means that the srcs of the `go_library` really will contain non go files, and things probably need to just deal with that.

ains:
Yeah, turns out it wasn't too big of a change after all, just needed to step away from the computer for a bit.

Just changing `allow_files = go_filetype` to `allow_files = True` for srcs and using `split_srcs` on either `cgo_srcs` or `srcs` depending on what's available seems to do the trick. 

Tests pass locally now, will check on Travis

ains:
@ianthehat Looks green now, can you take another look?

ianthehat:
I'm going to go ahead and submit this for you before I change stuff that would cause a merge conflict :)

PR #795, opened 2017-09-05
Simplify os_arch declarations

ianthehat:
The structs were more hassle than help, and can't be used in BUILD files where
you want to declare custom toolchains.

### DIFF
--- a/go/toolchain/toolchains.bzl
+++ b/go/toolchain/toolchains.bzl
@@ -58,7 +58,6 @@ def _generate_toolchains():
         full_name += ".0"
     version_constraints = [":go"+".".join(semver[:index+1]) for index, _ in  enumerate(semver)]
     for host in version.hosts:
-      "{}_{}".format(name, host)
       if hasattr(version, "sdk"):
         distribution = version.sdk
       else:
@@ -134,7 +133,11 @@ def declare_toolchains():
           "@io_bazel_rules_go//go/toolchain:" + goos,
           "@io_bazel_rules_go//go/toolchain:" + goarch,
       ]
-      exec_constraints = ["@io_bazel_rules_go//go/toolchain:" + part for part in toolchain["host"].split("_")]
+      host_goos, _, host_goarch = toolchain["host"].partition("_")
+      exec_constraints = [
+          "@io_bazel_rules_go//go/toolchain:" + host_goos,
+          "@io_bazel_rules_go//go/toolchain:" + host_goarch,
+      ]
       go_toolchain(
           name = toolchain["impl"],
           sdk = toolchain["sdk"],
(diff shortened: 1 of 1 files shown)


go/toolchain/toolchains.bzl
```diff
@@ -91,26 +58,22 @@ def _generate_toolchains():
         full_name += ".0"
     version_constraints = [":go"+".".join(semver[:index+1]) for index, _ in  enumerate(semver)]
     for host in version.hosts:
-      "{}_{}_{}".format(name, host.os, host.arch)
+      "{}_{}".format(name, host)
```
jayconrod: Should this be assigned to something?

go/toolchain/toolchains.bzl
```diff
@@ -166,20 +129,26 @@ def declare_toolchains():
   # Use the final dictionaries to create all the toolchains
   for toolchain in _toolchains:
     if "default" not in toolchain:
+      goos, _, goarch = toolchain["target"].partition("_")
+      target_constraints = [
+          "@io_bazel_rules_go//go/toolchain:" + goos,
+          "@io_bazel_rules_go//go/toolchain:" + goarch,
+      ]
+      exec_constraints = ["@io_bazel_rules_go//go/toolchain:" + part for part in toolchain["host"].split("_")]
```
jayconrod: Is there a difference in data format between `toolchain["target"]` and `toolchain["host"]`? They seem the same, but the code to build `exec_constraints` is different and seems to imply more constraints may be allowed. I guess it's just because `goos` and `goarch` are used below, but it might be good to build these lists the same way for readability.

go/toolchain/toolchains.bzl
```diff
@@ -3,81 +3,48 @@ load('//go/private:go_toolchain.bzl', 'external_linker', 'go_toolchain')
...
-  windows_amd64 = struct(os="windows", arch="amd64")
-  
   # The set of acceptable hosts for each of the go versions, this is essentially the
   # set of sdk's we know how to fetch
   versions = [
       struct(
           name = "host",
           sdk = "@go_host_sdk",
-          hosts = [darwin_amd64, linux_386, linux_amd64, windows_386, windows_amd64, freebsd_386, freebsd_amd64],
+          hosts = ["darwin_amd64", "linux_386", "linux_amd64", "windows_386", "windows_amd64", "freebsd_386", "freebsd_amd64"],
```
jayconrod: I was curious why we don't have the same set of hosts for all versions. I guess this is derived from `_sdk_repositories` in go/private/repositories.bzl. Could we build this list automatically from that table?

ianthehat: Not all versions have all host operating systems, new ones arrive and old ones stop being shipped, so it can't be just one list.
I did not bother adding hosts for old versions, I only added the ones we always supported, but I add all hosts that we ship every time we add a new version.

go/toolchain/toolchains.bzl
ianthehat: I originally wrote them both using the split form, but then I needed to cache the goos and goarch so I rewrote the first one.

PR #803, opened 2017-09-07
Go rules for new proto_library

### LINKED ISSUE #242: .proto files including root-repo .proto files fail to compile as valid Go
If you create a .proto file in the repo root and try including it from another .proto (whether also in the root or in a subdirectory) the including proto fails to compile via go_proto_library rule.

The .go file generated by the go_proto_library has an extra slash between the go_prefix and the dependency's go_proto_library name.  E.g. assume I have go_prefix 'my_prefix' and the following directory structure:
/BUILD
/WORKSPACE
/simple.proto
/subdir/BUILD
/subdir/includes_other.proto

Compiling includes_other.proto via go_proto_library produces the following error:
[...]includes_other.pb.go:19: non-canonical import path "my_prefix//simple_go_proto" (should be "my_prefix/simple_go_proto"

The go_proto_library should not be inserting a slash if the package is empty.

### LINKED ISSUE #403: correctly handle go_package option in go_proto_library
https://github.com/bazelbuild/rules_go/blob/master/proto/go_proto_library.bzl#L20
says:
> Gets confused if local protos use 'option go_package'

I am running into this issue. I am trying to use grpc.Status.Proto()
https://github.com/grpc/grpc-go/blob/master/status/status.go
which returns a proto with correct go package name which is based on the go_package option
"google.golang.org/genproto/googleapis/rpc/status;status"
according to:
https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto
but when I try to use it in a status proto that is embedded in my proto message I get a type mismatch error.

### LINKED ISSUE #440: Make proto rules more queryable
`go_proto_library` is a macro, which combines a private rule (`_go_proto_library_gen`) with `go_library`. This makes it hard to list proto libraries with `bazel query`. It should be converted into a rule.

We should consider exposing generated .pb.go files somehow in case people want to depend on those directly.

### LINKED ISSUE #457: go_proto_library does not find relative proto
I have the following `BUILD` content:

    load("@io_bazel_rules_go//proto:go_proto_library.bzl", "go_proto_library")

    go_proto_library(
        name = "go_default_library",
        srcs = [
            "digest.proto",
            "directory.proto",
            "nodes.proto",
        ],
    )

which gives me the error:

    Process exited with status 1 [sandboxed].
    nodes.proto: File not found.
    base_proto/directory.proto: Import "nodes.proto" was not found or had errors.

Which is strange, since all of the files exist.

### LINKED ISSUE #703: Replace com_github_google_protobuf with com_google_protobuf
Bazel itself references the latter, while these rules reference the former. It would be good to reuse the repos, Bazel is using.

### LINKED ISSUE #707: Update go_proto_library to use aspects referencing proto_library
bazel now has java_proto_library and cc_proto_library which do exactly that.  
go_proto_library should do the same, and in the case of gazelle finding .proto, it should generate both a proto_library for the srcs and a go_proto_library which references that.

ianthehat:
This works by adding a go_proto_library rule that takes a proto_library as it's
input and producing the same results as go_library would do using those
generated files.
It extracts the configuration to two toolchains, one called proto that supplies
protoc and one called go_proto that adds the go plugin and controls how it is
invokoked.
To add extra sources, you can make a go_library rule that uses the embed feature
to consume the go_proto_library.

Fixes #707
Fixes #703 
Fixes #457
Fixes #440
Fixes #403
Fixes #242


proto/toolchain.bzl
```diff
@@ -0,0 +1,58 @@
...
+    if outpath == None:
+        outpath = out.dirname[:-len(importpath)]
+  args= [
+      "--go_out=" + ",".join(go_proto_toolchain.options) + ":"+outpath,
+      "--plugin=protoc-gen-go=" + go_proto_toolchain.plugin.path,
+      "--descriptor_set_in", ":".join(
+          [s.path for s in lib.proto.transitive_descriptor_sets])
+  ]
+  args += [proto.short_path for proto in lib.proto.direct_sources]
+  ctx.action(
```
jayconrod: Why `command` and not `executable` + `arguments`?

WORKSPACE
```diff
@@ -3,6 +3,11 @@ workspace(name = "io_bazel_rules_go")
 load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")
 go_rules_dependencies()
 go_register_toolchains()
+register_toolchains(
```
jayconrod: Would it be worthwhile to wrap this in a `proto_register_toolchains` function? Developers should probably have something like this to call.

proto/toolchain.bzl
ianthehat: Because it was originally but I changed it so I could append `"&& tree"`, it's the only way to debug rules that don't produce the expected outputs, and I forgot to change it back again when I was done :)

WORKSPACE
ianthehat: Done

proto/def.bzl
```diff
@@ -0,0 +1,94 @@
...
+"""
+go_proto_library is a rule that takes a proto_library (in the proto 
+attribute) and produces a go library for it.
+"""
+
+go_grpc_library = rule(
+    _go_proto_library_impl,
+    attrs = {
+        "proto": attr.label(mandatory=True, providers=["proto"]),
+        "deps": attr.label_list(providers = [GoLibrary]),
```
pmbethe09: overall nice work.  Too bad about this though -- that you have to know your deps as opposed to the way aspects builds the shadow-graph for you.

cgrushko:
the more I think about it, the more I wish the new go_proto_library macro would be named slightly differently. It'll reduce confusion with the other LANG_proto_library rules.

Go's requirement of imports being paths to where the files reside is awesome - why is it being violated for protos? I understand that even ptypes violate this, but we can investigate fixing this.

ianthehat:
I don't think it should be renamed, it does the same job (builds a go library from a proto library), it just does not use an aspect to automatically build ones you did not add by hand.
I think the confusion of it not being called that when people with proto_libraries try to build go code will be worse than the confusion of the go version requiring you specify the deps by hand.

Go (the language) has no such requirement. The build system built into the go tool mostly has such a requirement (although it violates it internally in a couple of ways),  but we allow much more freedom in the world of Bazel.

Note that even the go build system cannot map from an import path to a file path in the presence of vendored code without examining the file system, so mapping from an import path to a label at analysis time is never viable even if we restricted ourselves to the subset of possibilities the go build system accepts.

ajbouh:
Do the main docs need to be updated to include telling people to add

```
load("@io_bazel_rules_go//proto:def.bzl", "proto_register_toolchains")
proto_register_toolchains()
```

To their WORKSPACE files?

Without that I get an IllegalStateException on Bazel 0.6.1.

jayconrod:
Yes. @ianthehat has been rewriting all of our documentation in #810. New proto documentation is not ready yet, but it will appear at [proto/core.rst](https://github.com/bazelbuild/rules_go/blob/master/proto/core.rst).

Issue #736, opened 2017-08-17
Not working on Windows 10.


abergmeier-dsfishlabs:
When trying to use `go_binary` on Windows I get:

```
ERROR: C:/projects/bazel_go_cas/BUILD.bazel:6:1: every rule of type go_binary implicitly depends upon the target '@io_bazel_rules_go_toolchain//:go_toolchain', but this target could not be found because of: no such package '@io_bazel_rules_go_toolchain//': Unsupported operating system: windows 10.
```

jayconrod:
Unfortunately, we don't support Windows builds yet, but we hope to add support soon.

The logic for this is in go/private/repository_tools.bzl if you want to experiment with this sooner.

ianthehat:
I have done most of the work that I can to get us to the point where we could support windows, and I would like to finish it, but I don't currently have a windows machine to finish it on. It should be really easy (less than a day of effort) to get the core rules working.
The only remaining issue should be things that are not expecting executables to end in .exe (the main toolchain sdk build file probably needs to be either select or glob to pick up the tools), we have purged all the uses of bash that would have caused issues.
Of course you never really know what else there is until you try it...

Xjs:
@ianthehat I would be interested in testing and possibly contributing in order to make the Go rules (and, if possible, gazelle) work on Windows 10. How can I help?

ianthehat:
Run a build on a windows machine and tell me what does not work :)
The first thing you will hit is `"Unsupported operating system:"` from [repository_tools.bzl](/bazelbuild/rules_go/blob/master/go/private/repository_tools.bzl#L43), it will need another else for windows, but I don't know what the ctx.os.name will report, which is why I have not added it.
The next thing you will hit is that the sdk rules assume the go binary name don't have an extension, on windows it will need .exe added. This means changing line 12 of [BUILD.sdk.bazel](/bazelbuild/rules_go/blob/master/go/private/BUILD.sdk.bazel#L12) to something like `srcs = [ "bin/go{exe_ext}" ],` and then adding another substitution in [toolchain.bzl] (/bazelbuild/rules_go/blob/master/go/private/toolchain.bzl#L73) to add it. It would be based on the based on the host os, just like the repository_tools sdk choice.
After that, it might all just work...

If you are not confident trying to make those changes, try a build and tell me the error message (it will say what the host os is reported as) and then I can make a PR for you to try.

Xjs:
[...]

PR #804, opened 2017-09-07
Drop hidden go_repository attributes

ianthehat:
Written as a contribution to the discussion in #736

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -49,10 +49,13 @@ def _go_repository_impl(ctx):
 
     # TODO(yugui): support submodule?
     # c.f. https://www.bazel.io/versions/master/docs/be/workspace.html#git_repository.init_submodules
+    _fetch_repo = "@io_bazel_rules_go_repository_tools//:bin/fetch_repo"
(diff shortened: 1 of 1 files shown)


go/private/go_repository.bzl
```diff
@@ -49,10 +49,13 @@ def _go_repository_impl(ctx):
 
     # TODO(yugui): support submodule?
     # c.f. https://www.bazel.io/versions/master/docs/be/workspace.html#git_repository.init_submodules
+    _fetch_repo = "@io_bazel_rules_go_repository_tools//:bin/fetch_repo"
```
jayconrod: Consider moving this logic into a function like `_maybe`. Could return a `Label`.

Also, rename to just `fetch_repo`, since this is local.

ianthehat: Not sure it is worth it, there exactly two cases, and one of them (the fetch_repo one) I am actively hoping to kill this month.

The name was mostly for similarity to the gazelle rule below, where the attribute was called `_gazelle` and there is already a local variable called `gazelle` that is the stringified path, so I could not rename it

Issue #792, opened 2017-09-05
Gazelle fix should explain what is wrong


jayconrod:
When a BUILD file needs to be fixed, `gazelle fix` just emits a warning without explaining what's wrong with it. This could be something as trivial as loaded symbols being out of order.

Our warning should explain what needs to be fixed. We should not emit a warning for trivial problems that won't interfere with later operations (loaded symbol ordering).

jayconrod:
Migrated to bazelbuild/bazel-gazelle#58

PR #818, opened 2017-09-08
Gazelle: Fix load statements when merging modified files

jayconrod:
fixLoads is now called in MergeWithExisting instead of FixLoads. This
will remove symbols that are no longer needed after empty rules are
deleted.

This also eliminates warnings that files need to be fixed if they only
need load changes.

Related #792


jayconrod:
Force submitted after multiple Travis internal failures (shell_session_update command not found).

PR #821, opened 2017-09-12
Some fixes for Windows (not working so far)

Xjs:
After discussion in #736, I'm submitting what I changed so far. The changes include:
* Check for .exe extension in a variety of places (I wrote a little helper function for that)
* Remove hidden attributes in go_repository, instead referencing gazelle und fetch_repo directly from the implementation
* Add TMP environment variable to execution contexts in some places because without it, go.exe tries to create temporary directories in C:\Windows, which it is typically not allowed to do
* In gazelle, replace the isDescendingDir implementation by one that uses filepath.Rel, thereby making paths containing slashes and paths containing backslashes comparable under Windows.


bazel-io:

ianthehat:
I think everything in this pull request is now in the master branch.

Xjs:
Looks good to me. I'm pulling back this request.

Issue #820, opened 2017-09-11
Checksum error downloading com_github_google_protobuf archive


nlacasse:
GitHub seems to have changed how they generate their archive.tar.gz files, resulting in them having different SHAs than previously.

As a result, using go_proto_library fails:

```
no such package '@com_github_google_protobuf//': Error downloading [https://github.com/google/protobuf/archive/v3.4.0.tar.gz] to /home/nlacasse/.cache/bazel/_bazel_nlacasse/db32ba9596c053450aeadce0671ee2d0/external/com_github_google_protobuf/v3.4.0.tar.gz: Checksum was f6600abeee3babfa18591961a0ff21e7db6a6d9ef82418a261ec4fee44ee6d44 but wanted cd55ee08e64a86cf12aaadd4672961813f592c194ed0c9ad94da0ec75acf219f 
```

See also https://github.com/google/protobuf/issues/3619

jayconrod:
Thanks for reporting! Are you still seeing this issue? I can't reproduce this, but it's late and maybe I'm missing something obvious.

Looks like this is affecting lots of repos. A [comment in libgit2](https://github.com/libgit2/libgit2/issues/4343#issuecomment-328631745) says this was due to an OS upgrade at GitHub, but that it's been reverted.

ianthehat:
I saw this earlier, but it's not happening now. I originally assumed they had just re-released 3.4.0 on us or something.

Directly accessing git tar-balls is always slightly fragile, it's just better than the alternatives.
What we should probably be doing is dumping specific versions into googlestorage and then getting them from there instead, and have a really easy way to add new ones.
We should talk to the Bazel team about doing this as a general service to users, and whether it would be allowed. If the guarantee was strong enough about keeping the files around forever people would be far more trusting about repeatable builds as well.

nlacasse:
This is still happening for me.  Try a `bazel clean` and then build something that depends on protos.

I think the revert mentioned in that issue was something that happened a while ago, the last time they tried to change how archives are constructed.

jayconrod:
Confirmed this is happening again now, in the same test workspace that was working last night. Apparently GitHub has rolled forward the change.

I don't think there's an easy solution to this. We should stop relying on tarballs served through codeload.github.com, since we have no guarantee that the sha256 is stable. So moving forward:

[...]

PR #823, opened 2017-09-12
Remove sha256 from dependencies fetched from codeload.github.com

jayconrod:
These archives are prepared on-the-fly by GitHub and are not
guaranteed to be the same every time (for example, the order of files
within the archive could change).

All archives are fetched over https; this should be as secure as
git_repository and still faster.

Related #820

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -93,7 +93,6 @@ def go_rules_dependencies():
       # master, as of 14 Aug 2017
       url = "https://codeload.github.com/bazelbuild/buildtools/zip/799e530642bac55de7e76728fa0c3161484899f6",
       strip_prefix = "buildtools-799e530642bac55de7e76728fa0c3161484899f6",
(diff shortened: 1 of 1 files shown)


jayconrod:
Once this is merged, I'll cherry-pick onto 0.5.4 and tag 0.5.5.

mattmoor:
> All archives are fetched over https; this should be as secure as git_repository and still faster.

Except if GitHub is compromised, so too are all `rules_go` users.  I think dropping the checksums is a mistake.

`git_repository` could at least depend on a commit, for which an attacker would at least have to break sha1.

jayconrod:
@mattmoor, If GitHub is compromised, then we lose security in several different ways. Attackers could change the released Bazel binaries for example. I'm not convinced that keeping `sha256` for GitHub URLs has significant security advantages, but moving to `git_repository` has significant performance and caching disadvantages.

When we tag releases in the future, we'll save a tarball with a known sha256, but we can't do this for all of our dependencies.

wteiken:
I agree that dropping the sha256 is a mistake, it adds a full trust dependency to github.  If there are already some other unnecessary full trust dependencies it would be good to remove them, not add more.

PR #824, opened 2017-09-12
Remove sha256 from dependencies fetched from codeload.github.com

jayconrod:
This is a cherry-pick of #823 onto a new branch, release/0.5.5 which is
based on tag 0.5.4.

----

These archives are prepared on-the-fly by GitHub and are not
guaranteed to be the same every time (for example, the order of files
within the archive could change).

All archives are fetched over https; this should be as secure as
git_repository and still faster.

Related #820

[skip ci]
[ci skip]


nlacasse:
You also need to remove the sha for go_proto_repositories from this file:

https://github.com/bazelbuild/rules_go/blob/0.5.4/proto/go_proto_library.bzl#L336

PR #825, opened 2017-09-12
Update README.md in anticipation of 0.5.5

jayconrod:
[skip ci]
[ci skip]


jayconrod:
Also, please take a look at the draft release text. The link to the archive won't work until the tag is created, but I believe it's in the correct format. I'll double-check that after tagging, before submitting this PR.

PR #828, opened 2017-09-13
Temporary directory handling

ianthehat:
The go toolchain on windows requires TMP to be a place it can work in.

Based on the work by Jannis Andrija Schnitzer (@xjs) in #821
The original approach did not quite work, and had some conflicts over TMP vs TMPDIR.
I am not really happy with this approach either, but it does seem to work.

### DIFF
--- a/go/private/toolchain.bzl
+++ b/go/private/toolchain.bzl
@@ -34,11 +34,15 @@ def _go_sdk_impl(ctx):
     _sdk_build_file(ctx, path)
     _local_sdk(ctx, path)
     
+  if "TMP" in ctx.os.environ:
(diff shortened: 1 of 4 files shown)


go/private/toolchain.bzl
```diff
@@ -36,10 +36,13 @@ def _go_sdk_impl(ctx):
     
   # Build the standard library for valid cross compile platforms
   #TODO: fix standard library cross compilation
+  tmp = ctx.os.environ['TMP']
```
jayconrod: Not sure if this is the cause of the Jenkins failure. Do you need to use `dict.get` here if it's not set?

PR #834, opened 2017-09-15
Gazelle: multiple fixes and refactorings before proto support

### LINKED ISSUE #611: Gazelle should add load statement at the top of a file
When Gazelle merges a `load` statement in an existing file, it checks for another `load` statement from the same file and attempts to merge with it. If no compatible `load` statement is found, Gazelle will append the the `load` to the end of the file. For example, say we have this existing file:

```
go_library(
    name = "go_default_library",
    srcs = ["lib.go"],
)
```

Gazelle will add the `load` statement after existing rules.

```
go_library(
    name = "go_default_library",
    srcs = ["foo.go"],
)

load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_prefix")

go_prefix("github.com/jr_hacker/example")
```

This seems wrong. `load` statements should be at the top of a file.

### LINKED ISSUE #822: Gazelle includes .cc files even when there's no cgo
To reproduce, put a trivial .go file (just a package declaration) and an empty .cc file in the same directory and run Gazelle. It will produce a BUILD.bazel file like the one below. We should ignore the .cc files.

```

go_library(
    name = "go_default_library",
    srcs = [
        "foo.cc",
        "foo.go",
    ],
    importpath = "github.com/jayconrod/scratch",
)

go_binary(
    name = "scratch",
    importpath = "github.com/jayconrod/scratch",
)
```

jayconrod:
* merger.FixLoads is now public and is called directly in
  main. rules.Generator no longer generates load statements at all
  (this functionality was redundant), so FixLoads must always be
  called before writing a file.
* Non-.go generated files (e.g., .s, .c if cgo is present) are now
  included in generated rules. .cc and other cgo files are no longer
  included if no cgo code is present.
* Removed unused parameters from buildPackage and newRule.

Fixes #822
Fixes #611


go/tools/gazelle/merger/fix.go
```diff
@@ -23,10 +23,15 @@ import (
 	"github.com/bazelbuild/rules_go/go/tools/gazelle/config"
 )
 
+// Much of this file could be simplified by using
```
ianthehat: I think it's worth filing an issue to see if there is a way to either remove the dependency, extract that proto to a repo you can depend on, or expose it in bazeltools or something.

jayconrod: I've filed bazelbuild/buildtools#143 for this. It seems like the simplest solution is to check in a copy of the proto file, then have a test (that downstream rules wouldn't depend on) to verify it matches the original. We might still need a dependency on `@io_bazel` to resolve the label in the test, but we shouldn't need to fetch it unless we actually run the test.

Issue #810, opened 2017-09-08
Full documentation overhaul


ianthehat:
The documentation is very out of date now for the rules.
They are mostly stabilizing, now we need to add more examples and document the new state.

AustinSchuh:
Food for thought: I've been trying to read the docs to sort out what your intended usage pattern is.  (I've got a bunch of binaries, tests, libraries in a repo with a bunch of other languages)  I see places where you document features of the tool.  I find myself fighting gazelle right now, and it would be very useful to know what your ideal setup would be so I know where I'm signing myself up for pain by disagreeing.  Maybe this exists, but I don't see links on the main doc pages that are getting me there.

jayconrod:
@AustinSchuh Gazelle's ideal set up is the same as `go build`. Although there are a few discrepancies, Gazelle takes a project that works with `go build` and generates build files for it. At some point, I'd like Gazelle to be more flexible (I'd especially like to be handle to handle multiple packages in the same directory), but for now, `go build` is the ideal layout.

The Bazel Go rules are more flexible: they allow pretty much arbitrary collections of rules, sources, and dependencies in arbitrary directories. So if you need something outside of what Gazelle supports, you can still express it by writing build files manually.

If you find yourself fighting Gazelle or if it's doing something that doesn't make sense, please let me know. Always happy to answer questions or take feature requests and bug reports.

jayconrod:
Closing old issues.

The reference documentation is now in a reasonable state. We're mostly lacking examples and tutorials, but there are many separate issues filed for those.

PR #840, opened 2017-09-19
Initial shell of new documentation

ianthehat:
The old documentation is still there so I can keep moving stuff
over, and is still the one shown by default.
I am switching to rst format because it's much more flexible, especially for when it comes to documenting the functions (table support is way way better)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,0 +1,274 @@
+Go rules for Bazel_
+=====================
+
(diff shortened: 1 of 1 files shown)


jayconrod:
Jenkins test this please

ianthehat:
This was the start of #810

PR #841, opened 2017-09-19
Gazelle: initial support for generalized directive comments

jayconrod:
Added config.ParseDirectives, which looks for comments of the form

    # gazelle:key value

Directives related to the configuration will be applied to a copy of
the current configuration, which is returned. This is passed down
through subdirectories by packages.Walk.

A full list of directives is returned. These may be processed by other
packages, for example gazelle:ignore and gazelle:exclude.

### DIFF
--- a/go/tools/gazelle/config/BUILD.bazel
+++ b/go/tools/gazelle/config/BUILD.bazel
@@ -5,13 +5,18 @@ go_library(
     srcs = [
         "config.go",
         "constants.go",
+        "directives.go",
     ],
+    deps = ["@com_github_bazelbuild_buildtools//build:go_default_library"],
     visibility = ["//visibility:public"],
 )
 
 go_test(
     name = "go_default_test",
     size = "small",
-    srcs = ["config_test.go"],
+    srcs = [
+        "config_test.go",
+        "directives_test.go",
+    ],
     library = ":go_default_library",
 )
--- a/go/tools/gazelle/config/config.go
+++ b/go/tools/gazelle/config/config.go
@@ -17,6 +17,7 @@ package config
 
 import (
 	"fmt"
+	"strings"
 )
 
 // Config holds information about how Gazelle should run. This is mostly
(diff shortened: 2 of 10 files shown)


jayconrod:
That's a good point. We should disallow directives that modify the configuration after the first rule for clarity.

I don't think we should do this for non-configuration directives though. I expect in the future there may be some that act as annotations on specific rules. `# keep` is already kind of like that. There are also things like `# gazelle:ignore` and `# gazelle:exclude` that clearly aren't rule-specific.

jayconrod:
Ok, I've been experimenting with this a bit. Trying to find something reasonable without being too complicated. Here's what I'm thinking:

* There will be two kinds of directives: top-level, and annotation (this change will only support top-level, but I'm expected that annotations will be needed in the future).
* Top-level directives apply to the whole package. Things like `prefix` and `build_tags`. These must be written before the first statement (including `load`).
* Annotation directives may be written above rules. They affect process of the rule they annotate.
* This file will have a list of all valid directives and which kind they are. It will print an error message about any unknown directives or anything that's out of place.

WDYT?

ianthehat:
Sounds good.

achew22:
Will annotations be able to apply to individual lines like in a srcs attribute having `# gazelle:keep` in

```
go_library(
    name = "demo",
    srcs = [
        "main.go", # gazelle:keep
    ],
)
```

The example is admittedly contrived since main.go would be trivial to detect, but you get the idea.

jayconrod:
@achew22 I'm not sure. Can you think of any uses? `# keep` will continue being supported, but it won't be processed by this module. That won't change.

There are some cosmetic directives supported by buildifier, like `keep sorted`, `do not sort`, and `buildifier: leave-alone`. I don't think Gazelle respects these, but we probably should. I can't think of others at the moment, but there are probably some.

jayconrod:
@ianthehat I've implemented checking for top-level directives. I also split `ParseDirectives` into a function that extracts directives (which these checks) and a separate function that applies them to the `Config`. PTAL.

achew22:
I just wanted to make sure that `# keep` still works. Thanks for taking input and being so open in your designs!

PR #842, opened 2017-09-19
Document gazelle and go_embed_data rules

ianthehat:
This is part of #810

### DIFF
--- a/go/extras.rst
+++ b/go/extras.rst
@@ -1,0 +1,107 @@
+Extra rules
+===========
+
+.. _`core go rules`: core.rst
+.. _`gazelle documentation`: tools/gazelle/README.md
+
+This is a collection of helper rules. These are not core to building a go binary, but are supplied to make life a little 
+easier.
+
+* gazelle_
+* go_embed_data_
+
+gazelle
+-------
+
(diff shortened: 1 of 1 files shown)


go/extras.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. code::
+  
+  bazel run //:gazelle
+
+will cause gazelle to run with the supplied options in the source tree at the root.
+See the `gazelle documentation`_ for more details.
+
++-----------------+-------------------+------------------------------------------------------------------------------------+
+| Name            | Type              | Values                                                                             |
```
jayconrod: I think the Values column should be renamed to Default Value, and only the default should be shown.

Attributes that accept specific values are well-described below, so I don't think they all need to be listed here.

ianthehat: I did that originally, and then thought having a short easy to read list was important, but the lists have become more compact so I think you are right. I think maybe I ought to style the values themselves so they stand out a little though in those lists. Bold italic or code?

How do you think we should indicate mandatory vs optional, currently I have been implying mandatory by not having a default value specified, is that enough?

PR #844, opened 2017-09-20
Gazelle: support loading symbols from multiple files

jayconrod:
This is needed to support proto rules, since they are loaded from a
different file. We must be careful not to override rules with the same
name loaded from a different file (pubref or our own legacy proto
rules).

### DIFF
--- a/go/tools/gazelle/merger/fix.go
+++ b/go/tools/gazelle/merger/fix.go
@@ -16,6 +16,7 @@ limitations under the License.
 package merger
 
 import (
+	"fmt"
 	"log"
 	"sort"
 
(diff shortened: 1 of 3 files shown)


go/tools/gazelle/merger/fix.go
```diff
@@ -325,16 +378,30 @@ func FixLoads(oldFile *bf.File) *bf.File {
...
+
+func init() {
+	knownKinds = make(map[string]string)
+	for _, f := range []struct {
+		file  string
+		kinds []string
+	}{
+		{"go:def.bzl", []string{"cgo_library", "go_binary", "go_library", "go_prefix", "go_test"}},
+		{"proto:def.bzl", []string{"go_grpc_library", "go_proto_library"}},
+	} {
```
ianthehat: I would declare this thing as a top level var (knownLoads?) near the two that are initialized from it, easy to find read and extend then.

go/tools/gazelle/merger/fix.go
```diff
@@ -325,16 +378,30 @@ func FixLoads(oldFile *bf.File) *bf.File {
...
+		filename := fmt.Sprintf("@%s//%s", config.RulesGoRepoName, f.file)
```
ianthehat: Is it really worth doing this in code rather than in the declaration?
I am just thinking of when we decide to add rules_cc and rules_proto for skylark versions of the native rules that come from separate repositories...

go/tools/gazelle/merger/fix.go
jayconrod: Done

go/tools/gazelle/merger/fix.go
jayconrod: Nope, not worth it. Done.

I don't think this architecture will make sense if we extend Gazelle beyond Go and proto. We'll probably want some kind of plugin architecture so that another tool can implement the analog of `packages` and `rules`. The fixing and merging logic is still pretty language-agnostic, but may need a few hints (what files and symbols to load from).

Let's cross that bridge when we come to it though.

PR #848, opened 2017-09-21
Document the core rules

ianthehat:
More for #810

Still need the x_ref section filled in, and some more about embedding and why we do it, but even without that it is strictly better and more accurate than what we have right now.

[skip ci]
[ci skip]

### DIFF
--- a/go/core.rst
+++ b/go/core.rst
@@ -1,0 +1,425 @@
+Core go rules
+=============
+
+.. _test_filter: https://bazel.build/versions/master/docs/bazel-user-manual.html#flag--test_filter
+.. _test_arg: https://bazel.build/versions/master/docs/bazel-user-manual.html#flag--test_arg
+.. _gazelle: tools/gazelle/README.md
+.. _build constraints: http://golang.org/pkg/go/build/
+.. _GoLibrary: providers.rst#GoLibrary
+.. _GoEmbed: providers.rst#GoEmbed
+.. _GoBinary: providers.rst#GoBinary
+.. _cgo: http://golang.org/cmd/cgo/
+.. _"Make variable": https://docs.bazel.build/versions/master/be/make-variables.html
+.. _Bourne shell tokenization: https://docs.bazel.build/versions/master/be/common-definitions.html#sh-tokenization
+.. _data dependencies: https://docs.bazel.build/versions/master/build-ref.html#data
+.. _static: modes.rst#using-the-race-detector
+.. _race: modes.rst#building-static-binaries
+
+.. role:: param(kbd)
+.. role:: type(emphasis)
+.. role:: value(code)
+.. |mandatory| replace:: **mandatory value**
+
+These are the core go rules, required for basic operation.
+The intent is that theses rules are sufficient to match the capabilities of the normal go tools.
+
+.. contents::
+
+-----
+
+Design
+------
+
+Defines and stamping
+~~~~~~~~~~~~~~~~~~~~
+
+**TODO**: More information
+
+Embedding
+~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


go/core.rst
```diff
@@ -0,0 +1,425 @@
...
+
+Defines and stamping
+~~~~~~~~~~~~~~~~~~~~
+
+**TODO**: More information
+
+Embedding
+~~~~~~~~~
+
+This is used for things like internal tests, where a library is recompiled with additional
```
jayconrod: Missing word after "additional".

Also s/dependancies/dependencies/g

go/core.rst
```diff
@@ -0,0 +1,425 @@
...
+Providers
+^^^^^^^^^
+
+* GoLibrary_
+* GoEmbed_
+
+Output groups
+^^^^^^^^^^^^^
+
+* default *: A library with the default build options.*
```
jayconrod: Would be better to have the group names emphasized and the descriptions in normal face.

go/core.rst
```diff
@@ -0,0 +1,425 @@
...
+| :param:`srcs`              | :type:`label_list`          | :value:`None`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| The list of Go source files that are compiled to create the package.                             |
+| Only :value:`.go` files are permitted, unless the cgo attribute is set, in which case the        |
+| following file types are permitted: :value:`.go, .c, .s, .S .h`.                                 |
+| The files may contain Go-style `build constraints`_.                                             |
++----------------------------+-----------------------------+---------------------------------------+
+| :param:`deps`              | :type:`label_list`          | :value:`None`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| List of GoLibrary_ providers this library imports directly.                                      |
```
jayconrod: I would rephrase this as "List of Go libraries this library imports directly. These may be `go_library` rules or compatible rules with the `GoLibrary` provider."

go/core.rst
```diff
@@ -0,0 +1,425 @@
...
++----------------------------+-----------------------------+---------------------------------------+
+| If :value:`True`, the package uses cgo_.                                                         |
+| The cgo tool permits Go code to call C code and vice-versa.                                      |
+| This does not support calling C++.                                                               |
+| When cgo is set, :param:`srcs` may contain C or assembly files; these files are compiled with    |
+| the normal c compiler and included in the package.                                               |
++----------------------------+-----------------------------+---------------------------------------+
+| :param:`cdeps`             | :type:`label_list`          | :value:`None`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| The list of other libraries that the c code depends on.                                          |
```
jayconrod: List of `cc_library` rules the cgo code depends on. Only valid if `cgo = True`.

(Though I don't think we actually check that).

go/core.rst
ianthehat: Done

go/core.rst
ianthehat: Done

go/core.rst
```diff
@@ -0,0 +1,425 @@
...
+
+.. role:: param(kbd)
+.. role:: type(emphasis)
+.. role:: value(code)
+.. |mandatory| replace:: **mandatory value**
+
+These are the core go rules, required for basic operation.
+The intent is that theses rules are sufficient to match the capabilities of the normal go tools.
+
+.. contents::
```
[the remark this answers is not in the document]
ianthehat: Yup, done.

go/core.rst
ianthehat: Done

go/core.rst
ianthehat: Done

PR #850, opened 2017-09-22
Cgo: eliminate partial link step, update some providers

### LINKED ISSUE #252: Cannot build cgo libraries on gcc-6 systems
I've encountered the same issue reported in https://github.com/kubernetes/kubernetes/issues/39819.

## Repro steps

I've checked out a Cgo project, apcera/gssapi@5fb4217df13b8e6878046fe1e5c10e560e1b86dc, into vendor/github.com/apcera/gssapi.

I've then used the following minimal rule:

```python
# vendor/BUILD
# -*-Python-*-
load("@io_bazel_rules_go//go:def.bzl", "go_library", "cgo_library")

cgo_library(
    name = "github.com/apcera/gssapi_c",
    srcs =
    glob([
        #"github.com/apcera/gssapi/*.go",
        "github.com/apcera/gssapi/buffer.go",
    ], exclude=[
        "github.com/apcera/gssapi/*_test.go",
        "github.com/apcera/gssapi/doc.go"])
     + [
        "bleh.c"
    ],
    clinkopts = [
        #"--no-pie", "-Wl,-no-pie",
        "-ldl",
    ],
)

go_library(
    name = "github.com/apcera/gssapi",
    library = ":github.com/apcera/gssapi_c",
)
```

Error received when building `//vendor:github.com/apcera/gssapi_c` is `/usr/bin/ld: -r and -pie may not be used together`.

Trying to pass `clinkopts = ["--no-pie", "-Wl,-no-pie"]` sometimes doesn't help, and sometimes results in `cannot find "-lgcc_s"`.

"bleh.c" may be needed and is an empty file. `bazel clean --expunge` may be required to reproduce `-lgcc_s`. Building without bleh.c and using *.go should result in the same output as long as --expunge was run between switching between GCC 4.9 and GCC 6.

## Cause

Quoting @mikedanese in https://github.com/kubernetes/kubernetes/issues/39819#issuecomment-275474974:

> this is caused by gcc 6 on debian which enable -pie by default (whether you pass -pie or not). This is different then gcc-5 behavior. To fix this we need to pass -no-pie explicitly or stop using -r.

## Versions

- GCC: I've used the Debian stable GCC 4.9, and Debian testing GCC 6.2.1-5 (package version 6.2.1-1).
- Bazel: 0.4.3 release
- Go rules: 81a7f69370a4917aee98c0b804626824f3203ad5

jayconrod:
Previously cgo compiled generated .c files, along with the other C
sources in a package, then partially linked them all into a single .o
file which was packed into the .a file for the library. This partial
linking caused problems on platforms where -pie is on by default
(newer Debian and Ubuntu) because the -r flag we used conflicted.

The partial link is not necessary and is being removed from
"go build", so we will remove it, too. The .a file generated by the
cgo cc_library rule is unpacked, and its contents are packed into the
.a file for the go_library.

Fixes #252


jayconrod:
This is preliminary -- not ready to submit yet. It need to verify it works with GCC 6 and with Darwin. It's a complicated change though, and early feedback would be helpful.

go/tools/builders/pack.go
```diff
@@ -80,19 +84,182 @@ func copyFile(inPath, outPath string) error {
...
+// there are no more files in the archive.
+//
+// Both BSD and GNU / SysV naming conventions are supported.
+func readMetadata(r *bufio.Reader, nameData *[]byte) (name string, size int64, err error) {
+retry:
+	// Each file is preceded by a 60-byte header that contains its metadata.
+	// We only care about two fields, name and size. Other fields (mtime,
+	// owner, group, mode) are ignored because they don't affect compilation.
+	var entry [entryLength]byte
+	if _, err := io.ReadFull(r, entry[:]); err != nil {
```
ianthehat: You probably need to process the entry anyway if it was read in full but returned an eof

jayconrod: `ReadFull` should work, even though it drops other errors.

jayconrod:
The merge commits above consolidate cgo information (including generated sources) into a single provider passed to a single attribute of `go_library` and friends. This should work with #793 (preserving original Go sources in the `GoLibrary` provider). PTAL.

PR #855, opened 2017-09-26
Make bzl files compatible with mutating behavior of += on lists

vladmos:
In the future versions of Bazel += will mutate lists instead of copying them. Because of that lists in the left-hand side can't be frozen, if they are they should be copied in the current mutability scope first.

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -43,7 +43,7 @@ def emit_link(ctx, go_toolchain, library, mode, executable, gc_linkopts, x_defs)
   extldflags = []
   if go_toolchain.external_linker:
     ld = go_toolchain.external_linker.compiler_executable
-    extldflags = go_toolchain.external_linker.options[:]
+    extldflags = list(go_toolchain.external_linker.options)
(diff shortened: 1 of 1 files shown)


vladmos:
Jenkins, retest this please.

(failed on a mac machine because the updated xcode license hadn't been manually accepted, not relevant to the pull request)

ianthehat:
This is going to collide with another change of mine, is it urgent?
I have been going through changing every single use of += in our code because it's too hard to reason about now (and depsets are going to stop supporting it anyway), and also fixing places where lists should really be tuples (like go_toolchain.external_linker.options).
If it's urgent and you really need to get it in it should use the list() syntax, [:] is much harder to read, and I believe no longer standard practice in the world of python either.

vladmos:
This prevents us from switching the behavior of += because it causes a breakage on our CI (and also some other repositories depend on rules_go). I'd like it to be merged asap if possible so that I can fix other repositories too. And even if this PR collides with something else it must be very easy to fix as it's just a one-line change.

I've replaced `[:]` with `list()`.

PR #864, opened 2017-09-29
Documentation for toolchains

ianthehat:
More work on #810

[skip ci]
[ci skip]

### DIFF
--- a/go/toolchains.rst
+++ b/go/toolchains.rst
@@ -2,23 +2,24 @@ Go toolchains
 =============
 
 .. _core: core.bzl
-.. _forked version of go: `Using a custom sdk`_
-.. _control the version: `Forcing the go version`_
-.. _installed sdk: `Using the installed go sdk`_
-.. _go website: https://golang.org/
+.. _forked version of Go: `Using a custom sdk`_
+.. _control the version: `Forcing the Go version`_
+.. _installed sdk: `Using the installed Go sdk`_
+.. _Go website: https://golang.org/
 .. _binary distribution: https://golang.org/dl/
 .. _cross compiling: crosscompile.rst
 .. _register: Registration_
 .. _register_toolchains: https://docs.bazel.build/versions/master/skylark/lib/globals.html#register_toolchains
-.. _modes: modes.rst
+.. _compilation modes: modes.rst#compilation-modes
+.. _go assembly: https://golang.org/doc/asm
 
 .. role:: param(kbd)
 .. role:: type(emphasis)
 .. role:: value(code)
 .. |mandatory| replace:: **mandatory value**
 
-The go toolchain is at the heart of the go rules, and is the mechanism used to
-customize the behavior of the core_ go rules.
+The Go toolchain is at the heart of the Go rules, and is the mechanism used to
+customize the behavior of the core_ Go rules.
 
 .. contents:: :depth: 2
 
@@ -27,24 +28,24 @@ customize the behavior of the core_ go rules.
 Design
 ------
 
-The go toolchain consists of two main layers, `the sdk`_ and `the toolchain`_.
+The Go toolchain consists of two main layers, `the sdk`_ and `the toolchain`_.
 
 The SDK
 ~~~~~~~
 
-At the bottom is the go sdk. This is the same thing you would get if you go to the main
-`go website`_ and download a `binary distribution`_.
+At the bottom is the Go SDK. This is the same thing you would get if you go to the main
+`Go website`_ and download a `binary distribution`_.
 
 The go_sdk_ rule is responsible for downloading these, and adding just enough of a build file
 to expose the contents to Bazel. It currently also builds the cross compiled standard libraries
 for specific combinations, although we hope to make that an on demand step in the future.
 
-SDK's are specific to the host they are running on, and the version of go they want to use,
-but not the target they compile for. The go SDK is naturally `cross compiling`_.
+SDKs are specific to the host they are running on and the version of Go they want to use
+but not the target they compile for. The Go SDK is naturally `cross compiling`_.
 
-The go rules already adds a go_sdk_ rule for all the host and go version pairs that are shipped
-on the main go language website, so there should be no need to declare one of these unless you
-need a `forked version of go`_, however you may want to `control the version`_ or use the
+The Go rules already adds a go_sdk_ rule for all the host and Go version pairs that are shipped
+on the main Go language website, so there should be no need to declare one of these unless you
+need a `forked version of Go`_\, however you may want to `control the version`_ or use the
 `installed sdk`_.
 
 The toolchain
@@ -54,15 +55,15 @@ Declaration
 ^^^^^^^^^^^
 
 Toolchains are declared using the go_toolchain_ macro. This actually registers two Bazel
-toolchains, the main go toolchain, and a special bootstrap toolchain. The bootstrap toolchain
+toolchains, the main Go toolchain, and a special bootstrap toolchain. The bootstrap toolchain
 is needed because the full toolchain includes tools that are compiled on demand and written in
 go, so we need a special cut down version of the toolchain to build those tools.
 
 Toolchains are pre-declared for all the known combinations of host, target and sdk, and the names
 are a predictable
-go\ **version**\ _\ **host**
+"<**version**>_<**host**>"
 for host toolchains and
-go\ **version**\ _\ **host**\ _cross\_\ **target**
+"<**version**>_<**host**>_cross\_<**target**>"
 for cross compilation toolchains. So for instance if the rules_go repository is loaded with
 it's default name, the following toolchain labels (along with many others) will be available
 
@@ -79,13 +80,14 @@ Registration
 
 Normally you would just call go_register_toolchains_ from your WORKSPACE to register all the
 pre-declared toolchains, and allow normal selection logic to pick the right one.
-It is fine to add more toolchains to the set available if you like, and because the normal
-toolchain matching mechaism prefers the first declared match you can also override individual
-toolchains by having another one with the same constraints that you register *before* calling
-the go_register_toolchains_.
+
+It is fine to add more toolchains to the available set if you like. Because the normal
+toolchain matching mechanism prefers the first declared match, you can also override individual
+toolchains by declaring and registering toolchains with the same constraints *before* calling
+go_register_toolchains_.
 
 If you wish to have more control over the toolchains you can instead just make direct
-calls to register_toolchains_ with only the toolchains you wish to install, you can see an
+calls to register_toolchains_ with only the toolchains you wish to install. You can see an
 example of this in `limiting the available toolchains`_.
 It is important to note that you **must** also register the boostrap toolchain for any other
 toolchain that you register, otherwise the tools for that toolchain cannot be built.
@@ -93,7 +95,7 @@ toolchain that you register, otherwise the tools for that toolchain cannot be bu
 Use
 ^^^
 
-If you are writing a new rule that wants to use the go toolchain, you need to do a couple of things.
+If you are writing a new rule that wants to use the Go toolchain, you need to do a couple of things.
 First, you have to declare that you want to consume the toolchain on the rule declaration.
 
 .. code:: bzl
@@ -122,6 +124,8 @@ Normal usage
 ~~~~~~~~~~~~
 
 This is an example of normal usage for the other examples to be compared against.
+This will download and use the latest Go SDK that was available when the version of rules_go
+you're using was released.
 
 WORKSPACE
 ^^^^^^^^^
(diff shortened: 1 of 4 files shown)


go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+^^^^^^^^^^^
+
+Toolchains are declared using the go_toolchain_ macro. This actually registers two Bazel
+toolchains, the main go toolchain, and a special bootstrap toolchain. The bootstrap toolchain
+is needed because the full toolchain includes tools that are compiled on demand and written in
+go, so we need a special cut down version of the toolchain to build those tools.
+
+Toolchains are pre-declared for all the known combinations of host, target and sdk, and the names
+are a predictable
+go\ **version**\ _\ **host**
```
jayconrod: Some underscores are missing here.

Consider formatting this as `go_<version>_<host>` instead though. I don't think the bold emphasizes the substitution enough.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+  @io_bazel_rules_go//go/toolchain:1.9.0_linux_amd64_cross_windows_amd64
+
+The toolchains are not usable until you register_ them.
+
+Registration
+^^^^^^^^^^^^
+
+Normally you would just call go_register_toolchains_ from your WORKSPACE to register all the
+pre-declared toolchains, and allow normal selection logic to pick the right one.
+It is fine to add more toolchains to the set available if you like, and because the normal
```
jayconrod: Couple small fixes:

It is fine to add more toolchains to the available set if you like. Because the normal toolchain matching mechanism prefers the first declared match, you can also override individual toolchains by declaring and registering toolchains with the same constraints *before* calling `go_register_toolchains`.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+toolchain matching mechaism prefers the first declared match you can also override individual
+toolchains by having another one with the same constraints that you register *before* calling
+the go_register_toolchains_.
+
+If you wish to have more control over the toolchains you can instead just make direct
+calls to register_toolchains_ with only the toolchains you wish to install, you can see an
```
jayconrod: Break sentence before "You can see..."

achew22:
Is it possible to change the toolchain using a flag on the CLI at build time? Would it be possible to include instructions for that in this CL?

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+
+
+Customizing
+-----------
+
+Normal usage
+~~~~~~~~~~~~
+
+This is an example of normal usage for the other examples to be compared against.
+
```
jayconrod: This will download and use the latest Go SDK that was available when the version of rules_go you're using was released.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+    go_sdk(name="my_linux_sdk", url="https://storage.googleapis.com/golang/go1.8.1.linux-amd64.tar.gz")
+    register_toolchains(
+        "@//:my_linux_toolchain", "@//:my_linux_toolchain-bootstrap",
+    )
+
+    go_rules_dependencies()
+    go_register_toolchains()
+
+
+Build.bazel
```
jayconrod: BUILD.bazel

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+WORKSPACE
+^^^^^^^^^
+
+.. code:: bzl
+
+    load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies")
+
+    go_rules_dependencies()
+    register_toolchains(
+        "@io_bazel_rules_go//go/toolchain:1.8.0_darwin_amd64",
```
jayconrod: 1.8.3 to match text above.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+| :param:`path`                  | :type:`string`              | :value:`""`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| The local path to a pre-installed go SDK.                                                        |
+|                                                                                                  |
+| If :param:`path` is set :param:`urls` must be left empty.                                        |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`urls`                  | :type:`string_list`         | :value:`[]`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| A list of mirror urls to the binary distribution of a go SDK.                                    |
+| You should generally also set the sha256 parameter when using :param:`urls`.                     |
```
jayconrod: :param:\`sha256\`

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+| They can be used to specify target specific flags that c linking actions generated by cgo        |
+| should apply when using this toolchain.                                                          |
++--------------------------------+-----------------------------+-----------------------------------+
+
+The toolchain object
+~~~~~~~~~~~~~~~~~~~~
+
+When you get a go toolchain from a context (see use_) it exposes a number of fields.
+
+The public interface is the actions member.
```
jayconrod: Instead of listing the emitting functions at the bottom of this section, I think it would be better to have a hierarchical list for the whole object here.

* go_toolchain
  * actions
    * asm
    * ...

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| The current rule context, used to generate the actions.                                          |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`go_toolchain`          | :type:`the go toolchain`    | |mandatory|                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| This must be the same go toolchain object you got this function from.                            |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`sources`               | :type:`list<File>`          | |mandatory|                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| An iterable of source code artifacts.                                                            |
```
jayconrod: Must be pure .go files. No assembly or cgo.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`go_toolchain`          | :type:`the go toolchain`    | |mandatory|                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| This must be the same go toolchain object you got this function from.                            |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`source`                | :type:`File`                | |mandatory|                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| A source code artifact to assemble.                                                              |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`hdrs`                  | :type:`list<File>`          | :value:`[]`                       |
```
jayconrod: I think these types should be written as "list of Files" or "iterable of Files". In most (all?) cases, any iterable is accepted. Tuples and `depset`s should be allowed.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+
+
+cover
+~~~~~
+
+The compile function adds an action that runs ``go tool cover`` on a set of source files
+to produce copies with cover instrumentation.
+
+Returns a tupe of the covered source list and the cover vars.
```
jayconrod: tuple

ianthehat:
There is not a simple answer to this, and it's all on the Bazel side not the rules_go side.
You don't select a toolchain, you select a platform, which is a bag of constraints, and then Bazel automatically picks toolchains that match those constraints.
You can set a platform on the command line using a temporary experimental flag, but it's mostly not helpful because too much of the way Bazel works is still driven by the cpu argument. Right now you can set the cpu argument, which will imply some constraints and thus pick a toolchain to match, but thats only helpful if the only thing you want to select on is the target os and arch.
I also don't want to document it here because it's part of Bazel, it's temporary/experimental and its not very useful...

If you tell me what you are trying to achieve I might be able to help you out.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| An iterable of source code artifacts.                                                            |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`importpath`            | :type:`string`              | :value:`""`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| The import path this package represents. This is passed to the -p flag.                          |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`golibs`                | :type:`list<GoLibrary>`     | :value:`[]`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| An iterable of all imported libraries.                                                           |
```
jayconrod: directly imported (transitive imports need not be listed)

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`mode`                  | :type:`string`              | :value:`NORMAL_MODE`              |
++--------------------------------+-----------------------------+-----------------------------------+
+| Controls the compilation setup affecting things like enabling profilers and sanitizers.          |
+| This must be one of the values in common.bzl#compile_modes.                                      |
```
jayconrod: Would be better to list the valid modes with short descriptions here. I know it's redundant, but common.bzl is private and may change, and it's not immediately clear what's allowed from modes.rst.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+| :param:`gc_goopts`             | :type:`string_list`         | :value:`[]`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| Additional flags to pass to the compiler.                                                        |
++--------------------------------+-----------------------------+-----------------------------------+
+
+
+cover
+~~~~~
+
+The compile function adds an action that runs ``go tool cover`` on a set of source files
```
jayconrod: s/The compile function/The cover function/

Note that this removes most comments, including cgo comments.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
+| An iterable of Go source files.                                                                  |
++--------------------------------+-----------------------------+-----------------------------------+
+
+library
+~~~~~~~
+
+This emits actions to compile Go code into an archive.
+It supports embedding, cgo dependencies, coverage, and assembling and packing .s files.
+
+It returns a tuple of GoLibrary and GoEmbed.
```
jayconrod: Link to provider documentation throughout.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| This must be the same go toolchain object you got this function from.                            |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`srcs`                  | :type:`list<File>`          | :value:`[]`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| An iterable of go source Files to be compiled.                                                   |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`deps`                  | :type:`list<GoLibrary>`     | :value:`[]`                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| The list of direct dependancies of this package.                                                 |
```
jayconrod: s/dependancies/dependencies/

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`cgo_info`              | :type:`CgoInfo`             | :value:`None`                     |
```
jayconrod: I wonder if we should leave this undocumented? I'd love to get rid of this if we ever gain the capability to produce actions that compile and link C files.

go/toolchains.rst
```diff
@@ -18,22 +27,568 @@ customize the behavior of the core_ go rules.
...
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`go_toolchain`          | :type:`the go toolchain`    | |mandatory|                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| This must be the same go toolchain object you got this function from.                            |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`library`               | :type:`GoLibrary`           | |mandatory|                       |
++--------------------------------+-----------------------------+-----------------------------------+
+| The library to link.                                                                             |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`mode`                  | :type:`string`              | :value:`NORMAL_MODE`              |
```
jayconrod: Same comment about `mode` as above.

PR #868, opened 2017-10-03
Gazelle: support for proto rules

### LINKED ISSUE #808: Generate new proto rules with gazelle
Needs to generate both proto_library and go_proto_library rules when it sees a proto file.

jayconrod:
Gazelle now generates proto_library, go_proto_library, and
go_grpc_library rules. proto_library will contain .proto files in a
directory. go_{proto,grpc}_library are buildable libraries, linked
with the proto_library. A go_library rules embeds the
go_{proto,grpc}_library and includes an extra .go sources (but .pb.go
files are excluded).

The new proto rule generation is turned on by default in packages that
don't have existing proto rules. If there are existing rules, proto
rule generation will either follow previous behavior (if
go_proto_library.bzl is loaded) or will be disabled (if
go_proto_library is loaded from some other repo). The proto mode can
be selected explicitly with the -proto command line flag or with the
'# gazelle:proto' directive. Legacy rules can be migrated with
'gazelle fix'.

Limitations:

* Gazelle still only allows one package per directory. Gazelle infers
  package name and import path from .proto files. These must match
  other files in the directory.
* Import resolution is fairly crude and is just a guess, based on
  the import string.
* There's no way to import protos from other repositories, except for
  Well Known Types.

Fixes #808


go/tools/gazelle/config/directives.go
```diff
@@ -107,10 +108,89 @@ func ApplyDirectives(c *Config, directives []Directive) *Config {
...
 		return c
 	}
 	return &modified
 }
+
+// InferProtoMode sets Config.ProtoMode, based on the contents of f.  If the
+// proto mode is already set to something other than the default, or if the mode
+// is set explicitly in directives, this function does nothing. If the legacy
+// go_proto_library.bzl is loaded, or if this is the Well Known Types
+// repository, legacy mode is used.  If go_proto_library is loaded from another
```
ianthehat: This part of the description implies it would not mantain new proto rules that it generated, but looking at the code below you stay in DefaultMode in that case.

go/tools/gazelle/packages/fileinfo.go
```diff
@@ -45,6 +34,10 @@ type fileInfo struct {
 	// "_test" suffix if it was present. It is empty for non-Go files.
 	packageName string
 
+	// importPath is the canonical import path for the package this file
+	// belongs to. Will be empty file files that don't specify this.
```
ianthehat: This sentence makes no sense.

go/tools/gazelle/packages/fileinfo_proto.go
```diff
@@ -0,0 +1,133 @@
...
+		default:
+			// Comment matched. Nothing to extract.
+		}
+	}
+	sort.Strings(info.imports)
+
+	if info.packageName == "" {
+		stem := strings.TrimSuffix(name, ".proto")
+		fs := strings.FieldsFunc(stem, func(r rune) bool {
+			return !('A' <= r && r <= 'Z' ||
```
ianthehat: You probably ought to be using unicode.IsLetter and unicode.IsDigit in here.

go/tools/gazelle/config/config.go
```diff
@@ -166,3 +173,33 @@ const (
 	// new_http_archive.
 	FlatMode
 )
+
+// ProtoMode determines how proto rules are generated.
+type ProtoMode int
+
+const (
+	// In DefaultProtoMode, Gazelle generates proto_library and new
```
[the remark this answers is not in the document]
jayconrod: Done

go/tools/gazelle/config/directives.go
jayconrod: Done. This was the result of a last-minute change.

go/tools/gazelle/gazelle/integration_test.go
```diff
@@ -745,6 +745,252 @@ go_library(
 	})
 }
 
+func TestMigrateProtoRules(t *testing.T) {
+	files := []fileSpec{
+		{path: "WORKSPACE"},
+		{
+			path: config.DefaultValidBuildFileNames[0],
+			content: `load("@io_bazel_rules_go//proto:go_proto_library.bzl", "go_proto_library")
```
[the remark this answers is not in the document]
jayconrod: Done. I need to clean up this test at some point and stick these files in testdata.

go/tools/gazelle/packages/fileinfo.go
jayconrod: Hopefully better now. Also added TODO to improve.

go/tools/gazelle/packages/fileinfo_proto.go
jayconrod: Done

PR #887, opened 2017-10-09
Add a reference to the proto rules in the Setup section.

jkinkead:
Without this, `gazelle` will generate unbuildable targets when you have `.proto` files.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,6 +105,9 @@ Setup
     load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")
     go_rules_dependencies()
     go_register_toolchains()
+    # The following are only needed if you have proto files in your repository.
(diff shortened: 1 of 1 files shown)


bazel-io:

jayconrod:
Thanks! I moved the code into a separate snippet. We haven't written documentation for the new proto rules yet, but the error (crash) you get when using them without registering the toolchain is extremely confusing, so it definitely makes sense for this to be in the main README.

I've cancelled Travis and skipped Jenkins, since this is a documentation-only change. Not sure why Travis was acting up earlier.

jkinkead:
You should probably include a blurb in the [`gazelle` docs](https://github.com/bazelbuild/rules_go/blob/master/go/tools/gazelle/README.md) as well, since it'll generate proto build targets for external dependencies as well.

jayconrod:
Yep. I'm planning to rewrite the Gazelle doc this week.

Issue #889, opened 2017-10-10
building packages with visibility is broken with rules_go 0.6.0 and bazel 0.5.4


nlacasse:
With rules_go version 0.6.0 and Bazel version 0.5.4, building a simple package with a visibility attribute fails.

Here is a repro case:
https://github.com/nlacasse/rules_go/tree/visibility_test/tests/visibility

```
bazel build //tests/visibility:foo                                                                                                                                                                                                        
...........
Unhandled exception thrown during build; message: Unrecoverable error while evaluating node 'CONFIGURED_TARGET://tests/visibility:foo 0f763d6a9b9fd42cc649394441b374f2 (1803222502 2081008356)' (requested by nodes )
INFO: Elapsed time: 3.308s
FAILED: Build did NOT complete successfully (15 packages loaded)
java.lang.RuntimeException: Unrecoverable error while evaluating node 'CONFIGURED_TARGET://tests/visibility:foo 0f763d6a9b9fd42cc649394441b374f2 (1803222502 2081008356)' (requested by nodes )
        at com.google.devtools.build.skyframe.ParallelEvaluator$Evaluate.run(ParallelEvaluator.java:475)
        at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:352)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
        at com.google.devtools.build.lib.analysis.ToolchainContext.lambda$findToolchains$0(ToolchainContext.java:131)
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
        at com.google.common.collect.CollectSpliterators$1.lambda$tryAdvance$0(CollectSpliterators.java:110)
        at java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1812)
        at com.google.common.collect.CollectSpliterators$1.tryAdvance(CollectSpliterators.java:109)
        at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)
        at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:498)
[...]

PR #891, opened 2017-10-10
Test for visibility

ianthehat:


jayconrod:
This is failing on Travis because of #889. Travis is testing Bazel 0.5.4. Perhaps we should bump the minimum version to 0.6.0?

jayconrod:
Closing in favor of #893, which includes this commit (since I can't push to the original branch).

PR #894, opened 2017-10-10
Allow configuring what type of protobuf generation to do in go_repository

andrew-d:
This allows configuring what form of protobuf generation to do in a `go_repository` rule.

The main reason that I'd like this feature is that it allows us to use `grpc-gateway` (and, I'm sure, other projects) that have already-generated protobuf code in their repositories.  By default, when using `go_repository`, Gazelle will generate a BUILD file that attempts to build the protobufs; since importing external protobufs isn't yet supported, this fails.

With these changes, it's now possible to manually create the correct `go_repository` rules with `--proto disable`, which allows builds to succeed.

I've created a repository that demonstrates this problem here: https://github.com/andrew-d/rules_go-gateway-demo

If you check it out and run `bazel build //example:greeter`, you'll get the following error:

```
$ bazel build //example:greeter
INFO: Found 1 target...
ERROR: /private/var/tmp/_bazel_adunham/901306667fe55f4a3c8aee5008d8180a/external/com_github_grpc_ecosystem_grpc_gateway/runtime/internal/BUILD.bazel:10:1: Generating into bazel-out/darwin_x86_64-fastbuild/bin/external/com_github_grpc_ecosystem_grpc_gateway/runtime/internal/internal_go_proto/github.com/grpc-ecosystem/grpc-gateway/runtime/internal failed (Exit 1).
../com_github_grpc_ecosystem_grpc_gateway/runtime/internal/stream_chunk.proto: No such file or directory
Target //example:greeter failed to build
INFO: Elapsed time: 64.845s, Critical Path: 0.07s
```

If you uncomment the marked lines in the [`WORKSPACE` file](https://github.com/andrew-d/rules_go-gateway-demo/blob/master/WORKSPACE), which uses the features in this PR, then the build will succeed.

Thoughts and feedback appreciated!

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -81,7 +81,8 @@ def _go_repository_impl(ctx):
     gazelle = ctx.path(Label(_gazelle))
     cmds = [gazelle, '--go_prefix', ctx.attr.importpath, '--mode', 'fix',
             '--repo_root', ctx.path(''),
-            "--build_tags", ",".join(ctx.attr.build_tags)]
+            "--build_tags", ",".join(ctx.attr.build_tags),
+            "--proto", ctx.attr.build_file_proto_mode]
     if ctx.attr.build_file_name:
         cmds += ["--build_file_name", ctx.attr.build_file_name]
     cmds += [ctx.path('')]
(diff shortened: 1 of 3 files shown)


bazel-io:

andrew-d:
I also just added a test for this behavior - it errors before I add the `build_file_proto_mode` line, and succeeds after.  Feedback appreciated on the test, too!

jayconrod:
Jenkins test this please

PR #898, opened 2017-10-11
Cgo: choose archive by extension

### LINKED ISSUE #896: cgo_c_lib must produce a single file error
Testing e254d73bf1181101ed82791fa5d204c4c5b1b105 with a larger repository, this error is shown. When I make a small repository with one go file linking org_golang_x_sys for testing, this error is not triggered. This seems to be caused by 567738b9d3b61b162c1eda5f7ed21751379122d6.

`ERROR: external/org_golang_x_sys/unix/BUILD.bazel:3:1: in lib attribute of _cgo_collect_info rule @org_golang_x_sys//unix:go_default_library.cgo_info: '@org_golang_x_sys//unix:go_default_library.cgo_c_lib' must produce a single file. Since this rule was created by the macro 'go_library_macro', the error might have been caused by the macro implementation in external/io_bazel_rules_go/go/private/rules/cgo.bzl:314:13`

jayconrod:
This fixes an error that occurred when cc_library generates multiple
archive files (.lo and .pic.lo) in some configurations (-c
opt). Rather than requiring a single file, _cgo_collect_info will now
choose a file based on extension.

Fixes #896

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -26,6 +26,30 @@ def _c_filter_options(options, blacklist):
   return [opt for opt in options
         if not any([opt.startswith(prefix) for prefix in blacklist])]
 
+def _select_archive(files):
+  """Selects a single archive from a list of files produced by a
+  static cc_library.
+
+  In some configurations, cc_library can produce multiple files, and the
+  order isn't guaranteed, so we can't simply pick the first one.
+  """
+  # list of file extensions in descending order or preference.
+  exts = [".pic.lo", ".lo", ".a"]
+  best_ext_index = len(exts) + 1
+  best_file = None
+  for f in files:
+    ext_index = len(exts)
+    for i, ext in enumerate(exts):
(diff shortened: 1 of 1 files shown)


go/private/rules/cgo.bzl
```diff
@@ -26,6 +26,30 @@ def _c_filter_options(options, blacklist):
...
 
+def _select_archive(files):
+  """Selects a single archive from a list of files produced by a
+  static cc_library.
+
+  In some configurations, cc_library can produce multiple files, and the
+  order isn't guaranteed, so we can't simply pick the first one.
+  """
+  # list of file extensions in descending order or preference.
+  exts = [".pic.lo", ".lo", ".a"]
```
ianthehat: I think the entire body of this function could just be

```
for ext in [".pic.lo", ".lo", ".a"]:
  for f in files:
    if f.basename.endswith(ext):
      return f
  fail("cc_library did not produce any files")
```

PR #917, opened 2017-10-14
Consistently apply visibility to toolchain implementations.

swsnider:
This was causing errors when building with custom toolchains.


bazel-io:

swsnider:
I'm operating within an organization where we don't want our builds to call out to the internet in general, so we vendor everything, essentially. As a result, I use the standard go 1.9.1 toolchain, but I register it myself so that it points to URLs inside our network instead of outside.

jayconrod:
Jenkins test this please.

(I'll submit this shortly).

PR #919, opened 2017-10-16
Abstract the environment setting

ianthehat:
The go toolchain now provides a helper for running a command with the correct go
environment.
This is a neccesary isolation for future improvements where we no longer hard
code the GOROOT and the GOOS and GOARCH can vary per rule.

### DIFF
--- a/go/private/actions/action.bzl
+++ b/go/private/actions/action.bzl
@@ -1,3 +1,17 @@
+# Copyright 2014 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


go/private/actions/action.bzl
```diff
@@ -0,0 +1,5 @@
+def action_with_go_env(ctx, go_toolchain, env=None, **kwargs):
```
jayconrod: Add copyright notice for new file.

go/private/go_toolchain.bzl
```diff
@@ -33,6 +34,7 @@ def _go_toolchain_impl(ctx):
           "TMP": tmp,
       },
       actions = struct(
+          env = action_with_go_env,
```
jayconrod: Will this be public? Should be documented in toolchains.rst.

ianthehat: It will be, but I don't want people using it just yet, so I will document once it's all in (there is a lot more changes to come on the way to cross compiling standard libraries on demand)

go/private/actions/action.bzl
ianthehat: Done

PR #929, opened 2017-10-17
Normalise go environment handling

ianthehat:
This pushes the handling of the environment into the helpers rather than
actually setting the environment in skylark.
This will allow us to do things like abs the paths during execution

### DIFF
--- a/go/tools/builders/cgo.go
+++ b/go/tools/builders/cgo.go
@@ -21,7 +21,6 @@ import (
 	"flag"
 	"fmt"
 	"go/ast"
-	"go/build"
 	"go/parser"
 	"go/token"
 	"io/ioutil"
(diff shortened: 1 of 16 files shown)


go/tools/builders/cgo.go
```diff
@@ -183,6 +177,9 @@ func run(args []string) error {
 		}
 	}
 	if pkgName == "" {
+		fmt.Fprintf(os.Stderr, "\n\n")
```
jayconrod: Not sure I understand this. Was it debug code?

go/tools/builders/env.go
```diff
@@ -0,0 +1,73 @@
...
+package main
+
+import (
+	"flag"
+	"fmt"
+	"go/build"
+	"strings"
+)
+
+// GoEnv hold the go environment as specified on the command line.
```
jayconrod: s/hold/holds/

PR #952, opened 2017-10-24
Change the bazel test methods

ianthehat:
We now run them one at a time, but all in the same folder.
This makes them much faster to run in total, and safer.

### DIFF
--- a/tests/bazel_tests.bzl
+++ b/tests/bazel_tests.bzl
@@ -1,6 +1,6 @@
-_bazelrc = """
-startup --batch
+load("@io_bazel_rules_go//go/private:go_repository.bzl", "env_execute")
 
+_bazelrc = """
 build --verbose_failures
 build --sandbox_debug
 build --test_output=errors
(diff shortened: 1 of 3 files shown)


tests/bazel_tests.bzl
```diff
@@ -13,83 +13,118 @@ build:isolate --fetch=False
...
+    workspace_content += register
+
+  workspace_file = ctx.actions.declare_file("WORKSPACE.in")
+  ctx.actions.write(workspace_file, workspace_content)
+  build_file = ctx.actions.declare_file("BUILD.in")
+  ctx.actions.write(build_file, ctx.attr.build)
+
+  targets = ["@" + ctx.workspace_name + "//" + ctx.label.package + t if t.startswith(":") else t for t in ctx.attr.targets]
+  output = "external/" + ctx.workspace_name + "/" + ctx.label.package
+  ctx.file_action(output=script_file, executable=True, content=_bazel_test_script_content.format(
```
jayconrod: I think this would be more readable if the `format` call were extracted into a separate statement.

tests/bazel_tests.bzl
```diff
@@ -151,20 +186,24 @@ def _test_environment_impl(ctx):
...
-    path = "{0}",
-    bazel = "{1}",
-    visibility = ["//visibility:public"],
-)
-""".format(execroot, bazel))
+
+  # Get a temporary directory to use as our scratch workspace
+  result = env_execute(ctx, ["mktemp", "-d"])
+  if result.return_code:
+        fail("failed to create temporary directory for bazel tests: {}".format(result.stderr))
```
jayconrod: Nit: 2-space indent.

tests/bazel_tests.bzl
```diff
@@ -151,20 +186,24 @@ def _test_environment_impl(ctx):
...
+  scratch_dir = result.stdout.strip()
+
+  # Work out where we are running so we can find externals
+  exec_root, _, _ = str(ctx.path(".")).rpartition("/external/")
+
+  # build the basic environment
+  ctx.file("WORKSPACE", 'workspace(name = "{}")'.format(ctx.name))
```
jayconrod: Do we actually need a WORKSPACE file in this repo?

tests/bazel_tests.bzl
```diff
@@ -1,6 +1,6 @@
-_bazelrc = """
-startup --batch
+load("@io_bazel_rules_go//go/private:go_repository.bzl", "env_execute")
 
+_bazelrc = """
```
jayconrod: Consider adding a suffixes to these variable names to indicate whether they are templates or final content.

Also, please add a brief comment for each one describing what it does and where it will go. It's not obvious to me from the context here.

tests/bazel_tests.bzl
```diff
@@ -176,19 +215,21 @@ _test_environment = repository_rule(
     ],
 )
 
-def test_environment():
-  _test_environment(name="test_environment")
-
-def _execroot_impl(ctx):
+def bazel_test_settings_impl(ctx):
```
jayconrod: Can this be private?

tests/bazel_tests.bzl
```diff
@@ -176,19 +215,21 @@ _test_environment = repository_rule(
...
   return struct(
-    path = ctx.attr.path,
     bazel = ctx.attr.bazel,
```
jayconrod: Nit: 4-space indent inside the struct.

tests/bazel_tests.bzl
ianthehat: If we don't make one bazel will do it for us...

PR #958, opened 2017-10-26
Structured compilation modes

### LINKED ISSUE #963: race build is broken at 3fd320621566c44aad0f7df5d0f5858efd2bbcf6
It seems that the stdlib building is busted when `--features=race` is added to `bazel test`. 

Getting different errors on different platforms, but one is:

```
ERROR: /Users/jmhodges/founding/src/founding/go/cmd/https-sidecar/BUILD:4:1: GoCompile go/cmd/https-sidecar/~normal~go_default_library~/founding/go/cmd/https-sidecar.a failed (Exit 1).
go/cmd/https-sidecar/main.go:6:2: can't find import: "flag"
```

ianthehat:
This changes the way we handle compilation modes to be a fully strucutred form.
It does not change (yet) the way we pick the mode, but it gives all the actions
a fully structured mode that is capapble of supporting all the combinations.

As a bonus we get full support for msan in this pr.

Fixes #963

### DIFF
--- a/go/private/actions/binary.bzl
+++ b/go/private/actions/binary.bzl
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@io_bazel_rules_go//go/private:common.bzl",
-    "link_modes",
+load("@io_bazel_rules_go//go/private:mode.bzl",
+    "mode_string",
+    "common_modes",
     "NORMAL_MODE",
     "RACE_MODE",
     "STATIC_MODE",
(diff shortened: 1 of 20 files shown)


go/private/mode.bzl
```diff
@@ -0,0 +1,74 @@
...
+  if link == None: link = base.link
+  return struct(
+      static = static,
+      race = race,
+      msan = msan,
+      pure = pure,
+      link = link,
+  )
+
+DEFAULT_MODE = mode(
```
jayconrod: I think it would be simpler to remove this name and and just assign `NORMAL_MODE` below.

go/private/mode.bzl
```diff
@@ -0,0 +1,74 @@
...
+    result.append("pure")
+  if not result or not mode.link == LINKMODE_NORMAL:
+    result.append(mode.link)
+  return "_".join(result)
+
+NORMAL_MODE = DEFAULT_MODE
+RACE_MODE = mode(base = DEFAULT_MODE, race=True)
+STATIC_MODE = mode(base = DEFAULT_MODE, static=True)
+
+common_modes = (NORMAL_MODE, RACE_MODE, STATIC_MODE)
```
jayconrod: Make this symbol all caps, too.

go/private/actions/binary.bzl
```diff
@@ -51,29 +52,34 @@ def emit_binary(ctx, go_toolchain,
...
-    if mode == "default":
-        # work out what the default mode should be
-        if "race" in ctx.features:
-            mode = RACE_MODE
-        elif "static" in ctx.features:
-            mode = STATIC_MODE
-        else:
-            mode = NORMAL_MODE
-
+    # work out what the default mode should be
```
jayconrod: Not sure I understand this part. This executable will be redundant with one of the files linked above, right? Would both be linked in the same build?

Also, can `get_mode` be used below? If not, nit: 2-space indent.

ianthehat: There is no change in here. This is a consequence of not being able to specify executable outputs, which means you need to distinct files one that is always the "race" build, and one that is the executable output that changes based on the feature flags.

go/private/mode.bzl
ianthehat: NORMAL_MODE and friends get deleted in a subsequent cl, I kept it separate so I could find and prune it's uses easily.

go/private/mode.bzl
ianthehat: This did not change, just moved files.
Also gets deleted with the switch to aspect based compilation

PR #962, opened 2017-10-26
added 'gofmt' to 'stdlib'

excavador:
I my skylark rules I need "gofmt" tool.
It different from "go fmt" (with space): accept more options

If I made it wrond or not optimal, please guide me, how to better integrate "gofmt" to rules_go

### DIFF
--- a/go/private/rules/stdlib.bzl
+++ b/go/private/rules/stdlib.bzl
@@ -44,10 +44,11 @@ stdlib(
 
 def _stdlib_impl(ctx):
   go = ctx.actions.declare_file("bin/go") # TODO: .exe
+  gofmt = ctx.actions.declare_file("bin/gofmt") # TODO: .exe
   src = ctx.actions.declare_directory("src")
   pkg = ctx.actions.declare_directory("pkg")
   root_file = ctx.actions.declare_file("ROOT")
-  files = [root_file, go, src, pkg]
+  files = [root_file, go, gofmt, src, pkg]
   goroot = root_file.path[:-(len(root_file.basename)+1)]
   sdk = ""
   for f in ctx.files._host_sdk:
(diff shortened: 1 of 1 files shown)


bazel-io:

jayconrod:
Jenkins test this please

jayconrod:
@ianthehat Is refactoring a lot of stuff around here right now. Not sure if this will conflict with any pending changes or not.

excavador:
@jayconrod for me is any way to access to gofmt would be good.

"go fmt" unfortunatelly has some predefined hardcoded options.
for some tasks, like "check format of code on pre-commit hook" or "format my code" ("run" targets) I need to access to "gofmt" ("go fmt" will not solve my problems).

If you guide me some better way for that -  I would be happy to re-do my patch.
So, before it works, because golang directory and goroot were exposed completely, but as far as I can understand from the actual rules_go code, you construct own clean "bin" with only allowed tools.

There is reason why I made this patch.

ianthehat:
You should not need this.
gofmt does not need to go in the stdlib, as it does not vary by build mode and does not get used in the build. You can just get it straight from the sdk, where it is already globbed and exposed as `@go_sdk//:bin/gofmt`

excavador:
@ianthehat could you explain, how to use it?
https://gist.github.com/excavador/a6bfb227fcfd1cc7e7e5a07355ab6669

ianthehat:
Your _gofmt attribute is wrong, you forgot to tag it as allow_single_file

excavador:
@ianthehat another problem
```
external/go_sdk/bin/gofmt: cannot open `external/go_sdk/bin/gofmt' (No such file or directory)
external/go_sdk/bin/gofmt -l -w exe lib
```

excavador:
Oh no, ploblem is solved. Wrong cwd was.

Thank you! You can reject the PR

PR #972, opened 2017-10-31
Use go_library for go_test

ianthehat:
This makes the go_test macro build a go_library rule for the package under test,
and then a go_test rule to actually test it.
Makes sure that the library is built the same way it would be for real, and also
cleans up the need for the "synthetic" library parameter

### DIFF
--- a/go/private/rules/wrappers.bzl
+++ b/go/private/rules/wrappers.bzl
@@ -53,20 +53,24 @@ def go_binary_macro(name, srcs=None, cgo=False, cdeps=[], copts=[], clinkopts=[]
       **kwargs
   )
 
-def go_test_macro(name, srcs=None, cgo=False, cdeps=[], copts=[], clinkopts=[], **kwargs):
+def go_test_macro(name, srcs=None, deps=None, importpath="", library=None, embed=[], gc_goopts=[], cgo=False, cdeps=[], copts=[], clinkopts=[], **kwargs):
   """See go/core.rst#go_test for full documentation."""
-  cgo_info = None
-  if cgo:
-    cgo_info = setup_cgo_library(
-        name = name,
-        srcs = srcs,
-        cdeps = cdeps,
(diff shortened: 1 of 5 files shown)


go/private/rules/wrappers.bzl
```diff
@@ -53,20 +53,24 @@ def go_binary_macro(name, srcs=None, cgo=False, cdeps=[], copts=[], clinkopts=[]
...
-        srcs = srcs,
-        cdeps = cdeps,
-        copts = copts,
-        clinkopts = clinkopts,
-    )
+  library_name = name + "~library~"
+  go_library(
+      name = library_name,
+      visibility = ["//visibility:private"],
+      srcs=srcs,
```
jayconrod: Nit: add space around = for all keyword arguments (here and in `go_test` below).

go/private/rules/wrappers.bzl
```diff
@@ -53,20 +53,24 @@ def go_binary_macro(name, srcs=None, cgo=False, cdeps=[], copts=[], clinkopts=[]
...
-  if cgo:
-    cgo_info = setup_cgo_library(
-        name = name,
-        srcs = srcs,
-        cdeps = cdeps,
-        copts = copts,
-        clinkopts = clinkopts,
-    )
+  library_name = name + "~library~"
+  go_library(
```
jayconrod: Should this call `go_library_macro` instead? The cgo-related arguments should be passed in, too.

ianthehat:
@jayconrod I have no idea why these tests are failing, it looks like the go_prefix rule is missing from generated build files, but I don't see how this change could have triggered it. Might just be I am too tired to spot something obvious, could you take a look for me?

ianthehat:
I think I worked it out, hopeful the test will pass while I go to sleep for a bit :)

ianthehat:
retest this please

PR #976, opened 2017-11-01
Transition to aspect based builds

ianthehat:
We can now build libraries in multiple modes using aspects.
This allows the mode to be controlled by the leaf rules, so you can specify
things like `pure = "on"` on a go_binary and have it build using the stdlib that
does not depend on libc.
This abandons all the previous output group behaviors.
This also moves the debug and strip controls in to the build mode. In theory
this means we can build stripped and unstripped variants at will in a single
build, or build tests in debug mode but binaries in release mode, all in a
single bazel invocation.
The documentation still needs more updates, but I will do that in another cl.


go/private/mode.bzl
```diff
@@ -51,24 +29,61 @@ def mode_string(mode):
...
   if not result or not mode.link == LINKMODE_NORMAL:
     result.append(mode.link)
   return "_".join(result)
 
-NORMAL_MODE = DEFAULT_MODE
-RACE_MODE = mode(base = DEFAULT_MODE, race=True)
-STATIC_MODE = mode(base = DEFAULT_MODE, static=True)
-
-common_modes = (NORMAL_MODE, RACE_MODE, STATIC_MODE)
+def _ternary_mode(values):
```
jayconrod: Several comments and questions:

* Is there any case where this would accept anything other than bools and the strings "on", "off", and "auto"? If not, let's not accept "true", "false", "yes", and "no".
* Consider renaming to just `_ternary` or `_tristate`. The current name is a little confusing, since this function doesn't operate directly on modes.
* Consider accepting `*args`.
* fail if the loop exits.

go/private/rules/binary.bzl
```diff
@@ -69,10 +67,12 @@ go_binary = rule(
...
         ),
         "srcs": attr.label_list(allow_files = go_filetype),
-        "deps": attr.label_list(providers = [GoLibrary]),
+        "deps": attr.label_list(providers = [GoLibrary], aspects = [go_archive_aspect]),
         "importpath": attr.string(),
-        "library": attr.label(providers = [GoLibrary]),
-        "embed": attr.label_list(providers = [GoEmbed]),
+        "library": attr.label(providers = [GoLibrary], aspects = [go_archive_aspect]),
+        "embed": attr.label_list(providers = [GoEmbed], aspects = [go_archive_aspect]),
+        "pure": attr.string(values=["on", "off", "auto"], default="auto"),
```
jayconrod: It's regrettable there's not a more canonical tristate attribute type. I don't feel comfortable with strings doing this job, but I can't think of anything better. I think some of the C++ rules use integers, and I definitely prefer strings over those.

go/private/mode.bzl
```diff
@@ -51,24 +29,61 @@ def mode_string(mode):
...
+  for v in values:
+    if v == None: continue
+    if type(v) == "bool": return v
+    if type(v) != "string": fail("Invalid value type {}".format(type(v)))
+    v = v.lower()
+    if v in ["true", "on", "yes"]: return True
+    if v in ["false", "off", "no"]: return False
+    if v in ["auto"]: continue
```
jayconrod: `if v == "auto"`

go/private/actions/archive.bzl
```diff
@@ -16,67 +16,117 @@ load("@io_bazel_rules_go//go/private:common.bzl",
...
+  return dep[GoArchive]
+
+def _go_archive_aspect_impl(target, ctx):
+  mode = get_mode(ctx)
+  goarchive = target[GoArchive]
+  if goarchive.mode == mode:
+    return [GoAspectArchive(archive = goarchive)]
+
+  direct = [get_archive(dep) for dep in ctx.rule.attr.deps]
+  direct.extend([get_archive(dep) for dep in ctx.rule.attr.embed])
```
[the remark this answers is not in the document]
ianthehat: done

go/private/mode.bzl
ianthehat: true and false were because the bools were getting stringified, but I fixed it a while ago, yes and no were because I kept changing my mind :)

go/private/rules/binary.bzl
ianthehat: Unfortunately, they *have* to be strings. Aspects are only allowed string type attributes, and those must be of a type that have a specified set of values...
I would rather have had mostly bool (true, false, None)
In fact, I would rather have been able to propagate the features string array, but I can't. And in fact, because only label type attributes are allowed function defaults, I can't even hide the string ones and calculate them from the features.

PR #997, opened 2017-11-08
Fixes for aspect based builds

### LINKED ISSUE #993: dlv problem
Hello,

I am trying to debug my golang program by dlv on OSX.
bazel 0.7.0, rules_go latest.

If I build program by bazel, I receive error
```
➜ dlv attach 74960
could not attach to pid 74960: decoding dwarf section info at offset 0x0: too short
```
I found same issue for dlv:
https://github.com/derekparker/delve/issues/79
But static linking does not work on OSX (according to rules_go documentation).

If I build program by
```
go build
```
dlv able to attach to my program

How to build golang binaries in bazel on OSX for working with dlv?
Thank you

ianthehat:
Use the deps, not the direct dependancies of the deps
Fixes reported in person issue with pure="on"

Get the toolchain_flags from the rule not the aspect
Fixes #993

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -34,14 +34,14 @@ def get_archive(dep):
   return dep[GoArchive]
 
 def _go_archive_aspect_impl(target, ctx):
-  mode = get_mode(ctx)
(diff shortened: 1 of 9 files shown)


ixdy:
Confirmed that this works on kubernetes/test-infra now. Thanks!

PR #1002, opened 2017-11-08
Mention Windows compatibility in README

cgbystrom:
Learned the hard way, and later through filed issues, that Go rules do not yet work on Windows.
Can hopefully save time for others if listed explicitly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,6 +82,7 @@ They currently do not support (in order of importance):
 * C/C++ interoperation except cgo (swig etc.)
 * coverage
 * test sharding
+* building on Windows
 
 :Note: The latest version of these rules (0.7.0) require Bazel ≥ 0.6.0 to
   work.


bazel-io:

jayconrod:
rules_go should actually work on Windows. What problems are you seeing?

We do have one big known issue, which is that go_binary and go_test can't add .exe file extensions. Bazel 0.8 (releasing in a week or so) will give us the ability to fix this. Until then, it's possible to work around this with a genrule.

jayconrod:
Closing this since we do intend to support Windows. Please open an issue if you're having any problems.

cgbystrom:
Sorry to be jumping to conclusions! I thought my errors + the issues I saw mean no Windows support.

I filed a separate issue about the snags I hit in https://github.com/bazelbuild/rules_go/issues/1007

PR #1004, opened 2017-11-09
cgo: remove -pthread flag on darwin

jayconrod:
clang keeps warning us that we don't need this, so let's remove it.


jayconrod:
Jenkins test this please

PR #1026, opened 2017-11-15
Go rules: several small fixes for analysis performance

jayconrod:
* Duplicated direct deps in emit_archive and srcs in emit_library. The
  former change caused #1023, though I don't fully understand why.
* In _go_archive_aspect_impl, direct now contains deps from library,
  not library itself.
* In emit_archive, transitive is built using the depset
  constructor. We no longer iterate over transitive.

TODO: understand depset performance and optimize all depset use.

Fixes #1023

PR #1029, opened 2017-11-16
Cleaning up lists, tuples and depsets

ianthehat:
Much less bouncing between them.
Also, no more ambiguious += statements
All in the name of performance,

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -93,10 +93,10 @@ def emit_archive(ctx, go_toolchain, mode=None, golib=None, goembed=None, direct=
     extra_objects.append(obj)
   archive = goembed.cgo_info.archive if goembed.cgo_info else None
 
-  transitive = direct
   for a in direct:
     if a.mode != mode: fail("Archive mode does not match {} is {} expected {}".format(a.library.label, mode_string(a.mode), mode_string(mode)))
-    transitive = sets.union(transitive, a.transitive)
+
+  transitive = sets.union(direct, *[a.transitive for a in direct])
 
   if len(extra_objects) == 0 and archive == None:
     go_toolchain.actions.compile(ctx,
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -55,7 +55,7 @@ def _go_archive_aspect_impl(target, ctx):
       mode = mode,
       golib = target[GoLibrary],
       goembed = target[GoEmbed],
-      direct = depset(direct),
+      direct = direct,
       importable = True,
   )
   return [GoAspectArchive(archive = goarchive)]
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -14,6 +14,8 @@
 
 load("@io_bazel_rules_go//go/private:common.bzl",
     "split_srcs",
+    "to_set",
+    "sets",
 )
 load("@io_bazel_rules_go//go/private:mode.bzl",
     "get_mode",
@@ -88,13 +90,13 @@ def emit_archive(ctx, go_toolchain, mode=None, golib=None, goembed=None, direct=
   for src in source.asm:
     obj = ctx.actions.declare_file("{}/{}.o".format(out_dir, src.basename[:-2]))
     go_toolchain.actions.asm(ctx, go_toolchain, mode=mode, source=src, hdrs=source.headers, out_obj=obj)
-    extra_objects += [obj]
+    extra_objects.append(obj)
   archive = goembed.cgo_info.archive if goembed.cgo_info else None
 
   transitive = direct
   for a in direct:
     if a.mode != mode: fail("Archive mode does not match {} is {} expected {}".format(a.library.label, mode_string(a.mode), mode_string(mode)))
-    transitive += a.transitive
+    transitive = sets.union(transitive, a.transitive)
 
   if len(extra_objects) == 0 and archive == None:
     go_toolchain.actions.compile(ctx,
--- a/go/private/actions/library.bzl
+++ b/go/private/actions/library.bzl
@@ -16,6 +16,7 @@ load("@io_bazel_rules_go//go/private:common.bzl",
     "split_srcs",
     "join_srcs",
     "structs",
+    "sets",
 )
 load("@io_bazel_rules_go//go/private:providers.bzl",
     "CgoInfo",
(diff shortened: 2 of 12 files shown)


go/private/actions/archive.bzl
```diff
@@ -88,13 +90,13 @@ def emit_archive(ctx, go_toolchain, mode=None, golib=None, goembed=None, direct=
...
     go_toolchain.actions.asm(ctx, go_toolchain, mode=mode, source=src, hdrs=source.headers, out_obj=obj)
-    extra_objects += [obj]
+    extra_objects.append(obj)
   archive = goembed.cgo_info.archive if goembed.cgo_info else None
 
   transitive = direct
   for a in direct:
     if a.mode != mode: fail("Archive mode does not match {} is {} expected {}".format(a.library.label, mode_string(a.mode), mode_string(mode)))
-    transitive += a.transitive
+    transitive = sets.union(transitive, a.transitive)
```
jayconrod: I think if we're setting Bazel 0.7.0 as the minimum, the new depset constructor is supported (with direct and transitive parameters). That would be more efficient here. My earlier attempt was in the first commit of #1026.

The depset constructor is O(m+n). Building a depset incrementally like this will be O(m*n).

go/private/actions/library.bzl
```diff
@@ -39,8 +40,8 @@ def emit_library(ctx, go_toolchain,
   dep_runfiles = [d.data_runfiles for d in deps]
   direct = depset()
   direct_archives = depset()
-  gc_goopts = tuple(ctx.attr.gc_goopts)
-  cover_vars = ()
+  gc_goopts = [] + ctx.attr.gc_goopts
```
jayconrod: `list(ctx.attr.gc_goopts)` would be a little more clear.

go/private/actions/library.bzl
```diff
@@ -49,13 +50,13 @@ def emit_library(ctx, go_toolchain,
     cgo_info_label = None
   for t in embed:
     goembed = t[GoEmbed]
-    direct_archives += get_archive(t).direct
+    direct_archives = sets.union(direct_archives, get_archive(t).direct)
```
jayconrod: Same comment as earlier: use the new depset constructor instead of union in a loop.

Also in the loops below.

go/private/actions/link.bzl
```diff
@@ -41,37 +45,38 @@ def emit_link(ctx, go_toolchain,
...
     fail("Link mode {} is not yet supported".format(mode.link))
 
-  libs = depset([archive.file])
+  libs = to_set([archive.file])
   link_opts = ["-L", "."]
   cgo_deps = depset()
   for a in archive.transitive:
-    libs += [a.file]
-    link_opts += ["-L", a.searchpath]
+    libs = sets.union(libs, [a.file])
```
jayconrod: Union loop again

go/private/common.bzl
```diff
@@ -137,5 +137,10 @@ def env_execute(ctx, arguments, environment = None, **kwargs):
   env_args = ["env", "-i"]
   if environment:
     for k, v in environment.items():
-      env_args += ["%s=%s" % (k, v)]
+      env_args.append("%s=%s" % (k, v))
   return ctx.execute(env_args + arguments, **kwargs)
+
+def to_set(v):
+  if type(v) == "depset":
+    fail("Do not pass a depset to to_depset")
```
jayconrod: Nit: "... to to_set"

proto/toolchain.bzl
```diff
@@ -1,11 +1,29 @@
+# Copyright 2014 The Bazel Authors. All rights reserved.
```
jayconrod: 2017

go/private/actions/archive.bzl
ianthehat: I thought that the cost of (a + b + c) was the same as ((a + b) + c) if all of a, b and c were depsets?
either way, I would prefer to stick with sets.union and rely on it to be implemented correctly (currently it uses += to add depsets together underneath)

go/private/actions/library.bzl
ianthehat: I am trying to avoid it because it can also convert types (depset to list for instance) 
This cl tries hard to remove all possible depset to list conversions, and not leave anything behind that could easily become one.

go/private/actions/archive.bzl
jayconrod: What you said is true, but it's faster to build a unified depset in a single step from a list of depsets. You can accumulate a list of depsets in the loop, or use a list comprehension.

```
transitive = depset(direct = list(direct), transitive = [d.transitive for d in direct])
```

The issue is that when you use `.union` or `+`, Bazel tries to de-duplicate all the direct and nested elements, which means creating two `CompactHashSets`, and calling `hashCode` and `equals` for each. If one of the sets has a lot of direct elements, this happens for all of them on every union.

go/private/actions/library.bzl
jayconrod: Got it. I was thinking list + depset was list, but it reports an error.

go/private/actions/archive.bzl
ianthehat: I thought that Bazel only did de-duplication when a depset is collapsed back to a list rather than on construction?

ianthehat: And even then, we ought to change union to use the new depset constructor rather than repeated +=
sets.union should *always* be the best answer for how to union a bunch of sets :)

jayconrod: > I thought that Bazel only did de-duplication when a depset is collapsed back to a list rather than on construction?

Sadly, that is not the case. See the constructor in [SkylarkNestedSet.java](https://github.com/bazelbuild/bazel/blob/17cb02dc4dad4a6bad59bc08d9789a3a0e5a17fe/src/main/java/com/google/devtools/build/lib/syntax/SkylarkNestedSet.java#L119). I was looking through this yesterday to understand why it was hashing things added to the set. Nothing in here worked the way I expected.

Note that this only de-duplicates the combined list of direct values and the combined list of transitive depsets. It doesn't look at the contents of the transitive depsets.

> And even then, we ought to change union to use the new depset constructor rather than repeated +=
> sets.union should always be the best answer for how to union a bunch of sets :)

Agreed, but I'm not sure what their minimum supported Bazel version is. The depset constructor was new in Bazel 0.7 (the first revision of my PR yesterday broke on 0.6). In any case, sets.union will still be slower than necessary when it's called in a loop. It will be better to build a list of depsets, then call sets.union on all of them. It should be implemented like this:

```
def _union(*args):
  for i, arg in enumerate(args):
    if type(arg) != "depset":
      args[i] = depset(args)
  return depset(transitive = args)
```

Issue #859, opened 2017-09-29
Gazelle should resolve labels using declared import paths


jayconrod:
Currently, Gazelle converts import paths into Bazel labels by following a few simple conventions. This requires `go_library` to be in specific packages with specific names (`go_default_library`) in order to work. This is causing an increasing amount of difficulty for several reasons.

* There is no warning when referencing a library that doesn't exist. (#388)
* There are multiple ways to structure a repository. Gazelle itself supports hierarchical mode and flat mode. We'd like to be able to mix these. (#608)
* We don't want to force a repository to have a single prefix (or any prefix at all). (#313, #597, #721)
* We don't want to force external repositories that have particular names. (#606)
* We should allow multiple vendor directories. (#486, #725)
* We should allow vendor directories and external repositories at the same time. (#404)
* We should respect `importpath` attributes in rules.
* With `go_proto_library`, there are multiple kinds of libraries with different naming conventions. (#808)
* We'd like to drop the `go_default_library` name at some point. (#265)

Smarter label resolution is a prerequisite for all of this.

------

### Proposal

When Gazelle starts, it will build a table that maps import paths to labels and directories. This table will be populated by several mechanisms:

* For external repositories that use flat build files generated by Gazelle stored in a third_party directory, Gazelle will scan these files.
* For packages in the current repository that Gazelle *won't* update, Gazelle will scan existing build files.
* For packages in the current repository that Gazelle *will* update, Gazelle will scan source files.

After build files and directories are scanned, Gazelle will generate rules for the packages it needs to update. Dependencies will be resolved using the table described above.

* If an import path corresponds to one or more labels, Gazelle will choose one using vendoring logic and visibility. If there are multiple valid labels, Gazelle will emit an error.
* If an import path points to a repository that could not be scanned for some reason, Gazelle will fall back to current conventions.
* If an import path can't be resolved (it points to a known prefix, and there's no visible label), Gazelle will emit an error.

ianthehat:
*"If there are multiple valid labels, Gazelle will emit an error"*
Will it write out the files anyway (using the first match with a comment) so you can resolve by hand?

jayconrod:
Yes. Gazelle errors are almost never fatal. When it fails to find a label for an import path, it won't emit a label for that import. So you'll need to manually add one with a `# keep` comment.

guoshimin:
Has this proposal been accepted? Has work started? If not, what's it blocked on?

jayconrod:
[...]

PR #1032, opened 2017-11-17
Gazelle: restructure run function in main to prepare for new phasing

jayconrod:
This will make it simpler to separate indexing and generating rules
and resolving dependencies into separate phases.

Related #859


jayconrod:
@ianthehat Any thoughts on this one? It should be a no-op change, just to make it easier to add the indexing stuff later.

PR #1033, opened 2017-11-20
Cleaning up transitive provider fields

ianthehat:
Reducing the information in the transitive parts to the minimum needed.
This is to prevent the pathalogical hashing that happens in complex graphs of
deps.

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -25,6 +25,7 @@ load("@io_bazel_rules_go//go/private:providers.bzl",
     "GoLibrary",
     "GoEmbed",
     "GoArchive",
+    "GoArchiveData",
 )
 
 
@@ -51,10 +52,11 @@ def _go_archive_aspect_impl(target, ctx):
   goarchive = go_toolchain.actions.archive(ctx,
       go_toolchain = go_toolchain,
       mode = mode,
-      golib = target[GoLibrary],
+      importpath = target[GoLibrary].package.importpath,
       goembed = target[GoEmbed],
       direct = direct,
       importable = True,
+      runfiles = target[GoLibrary].runfiles,
   )
   return [GoAspectArchive(archive = goarchive)]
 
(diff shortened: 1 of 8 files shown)


go/private/actions/archive.bzl
```diff
@@ -125,12 +128,19 @@ def emit_archive(ctx, go_toolchain, mode=None, golib=None, goembed=None, direct=
...
       file = out_lib,
+      importpath = importpath,
       searchpath = searchpath,
-      library = golib,
+  )
+  return GoArchive(
+      mode = mode,
+      data = data,
       embed = goembed,
       direct = direct,
```
jayconrod: `direct` is still a list of `GoArchive`s, right? Is `GoArchive` ever added to a depset? If so, we might still get bad hash performance for archives in the DAG that are reachable on multiple paths. This might affect other providers, too.

go/private/providers.bzl
```diff
@@ -17,6 +17,8 @@ load("@io_bazel_rules_go//go/private:mode.bzl", "mode_string")
 GoLibrary = provider()
 """See go/providers.rst#GoLibrary for full documentation."""
 
+GoPackage = provider()
```
jayconrod: Should `GoPackage` be named `GoLibraryData` for consistency with `GoArchiveData`? Or maybe they should be `SourceData` and `CompileData`?

Also, these are never returned as providers, right? Maybe they should be functions that return a `struct`. I suppose that's not really that different but it makes it a little clearer how they're expected to be used.

go/private/actions/archive.bzl
ianthehat: GoArchive is never allowed in a depset, only GoArchiveData

go/private/providers.bzl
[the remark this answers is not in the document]
ianthehat: providers give you better error messages than structs, so I am leaning towards using them for all user visible structs, which includes anything used in a provider even if it is not directly one itself.
Everything in here needs documenting in the rst, but it's not stable yet (still mangling GoEmbed and friends to fix cgo)

ianthehat: GoPackage came first (and I think it mostly makes sense, it's much more the fundamental values that represent a go package to the rest of Bazel, whereas GoLibrary is the output of the go_library rule, and has a lot more bazel specific stuff in it), and then I could not think of a good name for GoArchiveData.
Generally I think names ending in Data are just lazy, and it's that one that needs renaming, but I could not think of a good name and I really want to get this in to see if we can get k8s compiling so I threw it over the wall....

PR #1034, opened 2017-11-20
Unify srcs and deps into embed

ianthehat:
The srcs and deps are packaged into a GoEmbed entry in the embed list, to
simplify the handling.


go/private/actions/library.bzl
```diff
@@ -30,43 +32,35 @@ load("@io_bazel_rules_go//go/private:actions/archive.bzl",
 def emit_library(ctx, go_toolchain,
     mode = None,
     importpath = "",
-    srcs = (),
-    deps = (),
-    cgo_info = None,
-    embed = (),
+    embed = [],
```
jayconrod: I really like how much this simplifies the implementation of `emit_library`, but it feels a little awkward from the caller's perspective.

What do you think about keeping the same interface, and creating the `GoEmbed` object and prepending it to the list in here? Completely optional though, up to you.

ianthehat: I thinking keeping this interface simpler is better, I am planning on adding some more features into the embedded handling that make it even more compelling.

PR #1046, opened 2017-11-21
Gazelle: initial implementation of rule index

jayconrod:
(Unit tests will be in the PR to wire this into Resolver).

Related #859

### DIFF
--- a/go/tools/gazelle/config/constants.go
+++ b/go/tools/gazelle/config/constants.go
@@ -55,7 +55,10 @@ const (
 type Language int
 
 const (
+	// GoLang marks Go targets.
 	GoLang Language = iota
+
+	// ProtoLang marks protocol buffer targets.
 	ProtoLang
 )
 
@@ -69,8 +72,3 @@ func (l Language) String() string {
 		return "unknown"
 	}
 }
-
-const (
-	PublicVisibility  = "//visibility:public"
-	PrivateVisibility = "//visibility:private"
-)
--- a/go/tools/gazelle/config/constants.go
+++ b/go/tools/gazelle/config/constants.go
@@ -49,3 +49,28 @@ const (
 	// on generated rules. It is replaced with "deps" during import resolution.
 	GazelleImportsKey = "_gazelle_imports"
 )
+
+// Language is the name of a programming langauge that Gazelle knows about.
+// This is used to specify import paths.
+type Language int
+
+const (
+	GoLang Language = iota
+	ProtoLang
+)
+
+func (l Language) String() string {
+	switch l {
+	case GoLang:
+		return "go"
+	case ProtoLang:
+		return "proto"
+	default:
+		return "unknown"
+	}
+}
+
+const (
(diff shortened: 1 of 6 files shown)


go/tools/gazelle/config/constants.go
```diff
@@ -49,3 +49,28 @@ const (
 	// on generated rules. It is replaced with "deps" during import resolution.
 	GazelleImportsKey = "_gazelle_imports"
 )
+
+// Language is the name of a programming langauge that Gazelle knows about.
+// This is used to specify import paths.
+type Language int
+
+const (
+	GoLang Language = iota
```
ianthehat: Public symbols, trivial comments needed...

go/tools/gazelle/resolve/index.go
```diff
@@ -0,0 +1,358 @@
...
+//
+// Any number of rules may provide the same import. If no rules provide
+// the import, ruleNotFoundError is returned. If multiple rules provide the
+// import, this function will attempt to choose one based on visibility.
+// An error is returned if the import is still ambiguous.
+//
+// Note that a rule may be returned even if visibility restrictions will be
+// be violated. Bazel will give a descriptive error message when a build
+// is attempted.
+func (ix *RuleIndex) findRuleByImport(imp importSpec, lang config.Language, fromRel string) (*ruleRecord, error) {
```
ianthehat: In the presence of multiple vendor folders, we may need to be able to pick the "nearest" in the hierarchy here as well, but it's not clear to me that there is enough information to do so in the structures.

go/tools/gazelle/resolve/index.go
```diff
@@ -0,0 +1,358 @@
...
+		label = label.Abs("", buildRel)
+		if label.Repo != "" || label.Name != "__subpackages__" {
+			continue
+		}
+		visibility = append(visibility, label.Pkg)
+	}
+	return visibility
+}
+
+func isVisibleFrom(visibility visibilitySpec, defRel, useRel string) bool {
```
ianthehat: I am not convinced we should be handling visibility, I think it is going to be more surprising than helpful, especially as it's going to be hard to be sure we cover all the visibility cases.
I think the set of cases where visibility is going to help us pick the right one of multiple packages with the same import path are negligible to non existent, far more likely are the cases where we should generate the dep anyway so they get a nice "is not visible" error when they try to build the thing they really wanted.

go/tools/gazelle/resolve/index.go
jayconrod: I've added this logic for Go only (not proto). I'm using the Pkg part of the label as a proxy for the directory in the repository. This won't work in flat mode or with manually written rules, but I'm not sure how much vendoring matters for those cases.

go/tools/gazelle/resolve/index.go
jayconrod: You're right, I think I half-realized this when I allowed findRuleByImport to succeed when the only matches available were restricted by visibility.

I've removed all visibility parsing.

PR #1048, opened 2017-11-22
Move coverage

ianthehat:
We now do it as late as possible (just before compilation)
We also allow very fine grained control over which packages and source files are
covered.
We also properly obey configuration.coverage_enabled to disable all coverage.
This makes a few interfaces simpler, and is needed for us to handle the pure/cgo
source generation cleanly (which has to come before cover)


ianthehat:
Source changes are in, PTAL

PR #1056, opened 2017-11-22
Select original go sources in pure mode

### LINKED ISSUE #1009: experimental_platforms mac os to linux does not work
Cross compiling with release 0.7.0 --experimental_platforms does not work on mac to produce linux binaries.

I have create a sample repo that fails with the command `bazel build --experimental_platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //:logrusbazel`

https://github.com/margic/logrusbazel

ianthehat:
This adds a filtering mechanism to the new GoSourceList logic and uses it to
switch between the cgo processed sources and the unprocessed original sources
depending on whether the archive is being generated in pure mode.

Fixes #1009

### DIFF
--- a/go/private/providers.bzl
+++ b/go/private/providers.bzl
@@ -39,7 +39,7 @@ def _merge_runfiles(a, b):
   if not b: return a
   return a.merge(b)
 
-def _source_build_entry(srcs = [], deps = [], gc_goopts=[], runfiles=None, cgo_deps=[], cgo_exports=[], cgo_archive=None, want_coverage = False, source = None):
+def _source_build_entry(srcs = [], deps = [], gc_goopts=[], runfiles=None, cgo_deps=[], cgo_exports=[], cgo_archive=None, want_coverage = False, source = None, exclude = None):
   """Creates a new GoSource from a collection of values and an optional GoSourceList to merge in."""
   for e in (source.entries if source else []):
     srcs = srcs + e.srcs
@@ -62,6 +62,7 @@ def _source_build_entry(srcs = [], deps = [], gc_goopts=[], runfiles=None, cgo_d
       cgo_exports = cgo_exports,
       cgo_archive = cgo_archive,
       want_coverage = want_coverage,
+      exclude = exclude,
   )
 
 def _source_new(**kwargs):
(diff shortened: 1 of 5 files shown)


go/private/providers.bzl
```diff
@@ -62,6 +62,7 @@ def _source_build_entry(srcs = [], deps = [], gc_goopts=[], runfiles=None, cgo_d
       cgo_exports = cgo_exports,
       cgo_archive = cgo_archive,
       want_coverage = want_coverage,
+      exclude = exclude,
```
jayconrod: Consider renaming to indicate this checks the mode, and reversing the sense.

Maybe `_source_filter` below could be:

```
def _source_filter(ctx, source, mode):
  return GoSourceList(entries = [s for s in source.entries if (not s.mode_ok or s.mode_ok(ctx, mode))])
```

go/private/rules/cgo.bzl
```diff
@@ -149,6 +149,7 @@ def _cgo_codegen_impl(ctx):
       ),
       OutputGroupInfo(
           go_files = to_set(go_outs),
+          input_go_files = to_set(source.go),
```
jayconrod: Is it sufficient to return these? We should hide cgo sources from the compiler in pure mode, even if there's no build tag. "go build" seems to do this.

The best place is probably in the compile builder. We do tag filtering there, and we inspect the imports anyway for strict dependency checking.

ianthehat: In here this is all we can do, yes if we need to do more filtering it has to go in the compile wrapper, we should be setting the cgo bool correctly, I was presuming that was sufficient already.

go/private/providers.bzl
ianthehat: It does not have to check the mode, it can exclude for any reason it likes, checking feature flags on the ctx for instance.
I picked exclude because include is confusing in a world where c++ headers exist, and filter is ambiguous

ianthehat: I am going to submit this so I can keep going, but I am happy to do another PR to change the name (and invert the logic) if needed.

PR #1057, opened 2017-11-23
Be principled about generated file names

ianthehat:
We now direct all file generation through a function that works out an
appropriate name, including the rule label and build mode when needed.

### DIFF
--- a/go/private/tools/embed_data.bzl
+++ b/go/private/tools/embed_data.bzl
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@io_bazel_rules_go//go/private:common.bzl",
+    "declare_file",
+)
+load("@io_bazel_rules_go//go/private:providers.bzl",
+    "sources",
+)
+
 def _go_embed_data_impl(ctx):
   if ctx.attr.src and ctx.attr.srcs:
     fail("%s: src and srcs attributes cannot both be specified" % ctx.label)
@@ -32,10 +39,11 @@ def _go_embed_data_impl(ctx):
     if package == "":
       fail("%s: must provide package attribute for go_embed_data rules in the repository root directory" % ctx.label)
 
+  out = declare_file(ctx, ext=".go")
   args.add([
     "-workspace", ctx.workspace_name,
     "-label", str(ctx.label),
-    "-out", ctx.outputs.out,
+    "-out", out,
     "-package", package,
     "-var", ctx.attr.var,
   ])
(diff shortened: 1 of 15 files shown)


ianthehat:
jenkins test this please

ianthehat:
jenkins test this please

go/private/tools/embed_data.bzl
```diff
@@ -46,17 +54,20 @@ def _go_embed_data_impl(ctx):
...
   )
+  return [
+      DefaultInfo(files = depset([out])),
+      sources.new(srcs = [out], want_coverage = False),
+  ]
 
 go_embed_data = rule(
     implementation = _go_embed_data_impl,
     attrs = {
-        "out": attr.output(mandatory = True),
```
jayconrod: Let's keep this. Removing this attribute will break existing uses.

Also, Gazelle identifies rules with .go files in `out` and `outs` attributes and adds those files to `srcs`. Having this attribute makes it easier to use.

go/private/tools/path.bzl
```diff
@@ -13,6 +13,15 @@
 # limitations under the License.
 
 load("@io_bazel_rules_go//go/private:providers.bzl", "GoLibrary", "GoPath")
+load("@io_bazel_rules_go//go/private:common.bzl", "declare_file")
+
+
+def _tag(ctx, path, outputs):
```
jayconrod: Maybe add a doc string here. I'm not really clear on what this does.

go/private/tools/embed_data.bzl
ianthehat: It's the only remaining use of attr.output in our rules, it's the wrong way to do things and we ought to stop, mostly because people will copy this rule to write their own code generators.
Having implicit edges by filename (the only reason for the out attribute) is a broken design, and one the Bazel team are trying to steer people away from, if you depend on the outputs of a rule, you should depend on the rule, not a file it happens to produce.

It's not unreasonable to think that we add options like compression to the data embedding rule, and then maybe you want to build sometimes with, and sometimes without, so you want the generated filename to vary with build mode, or you need to add an automatic dependency on the compression library used to go along with the generated source (and thus use a go source to an embed attribute, rather than a file to a src attribute)

I could not find any uses on github right now, and the error produced is fairly obvious to fix (if there are non github users), so I figured a breaking change was acceptable.

jayconrod: Fair enough. Gazelle will need to use bazel query to discover generated files, but I think that's something we wanted anyway.

PR #1062, opened 2017-11-25
Finishing go_source rule and using it in a test

ianthehat:
This also fixes a double embedding case that was broken.

### DIFF
--- a/go/core.rst
+++ b/go/core.rst
@@ -95,7 +95,7 @@ Attributes
 +----------------------------+-----------------------------+---------------------------------------+
 | List of Go libraries this test library directly.                                                 |
 | These may be go_library rules or compatible rules with the GoSourceList_ provider.               |
-| These can provide both :param:`srcs` and param:`deps` to this library.                           |
+| These can provide both :param:`srcs` and :param:`deps` to this library.                          |
 | See Embedding_ for more information about how and when to use this.                              |
 +----------------------------+-----------------------------+---------------------------------------+
 | :param:`data`              | :type:`label_list`          | :value:`None`                         |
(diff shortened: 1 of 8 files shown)


go/core.rst
```diff
@@ -427,3 +427,57 @@ It references the library(s) being tested with :param:`deps`.
       srcs = ["lib_x_test.go"],
       deps = [":go_default_library"],
   )
+
+go_source
+~~~~~~~~~
+
+This declares a set of source files and related dependancies that can be embedded into on of the
```
jayconrod: s/dependancies/dependencies/
s/into on of/into one of/

go/core.rst
```diff
@@ -427,3 +427,57 @@ It references the library(s) being tested with :param:`deps`.
...
+| :param:`deps`              | :type:`label_list`          | :value:`None`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| List of Go libraries this source list imports directly.                                          |
+| These may be go_library rules or compatible rules with the GoLibrary_ provider.                  |
++----------------------------+-----------------------------+---------------------------------------+
+| :param:`embed`             | :type:`label_list`          | :value:`None`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| List of sources to directly embed in this list.                                                  |
+| These may be go_library rules or compatible rules with the GoSourceList_ provider.               |
+| These can provide both :param:`srcs` and param:`deps` to this library.                           |
```
jayconrod: Need a colon before param:\`deps\`

ianthehat: I did not even read this bit, just cut'n'paste it from one of the 3 other places this broken text is used...
all fixed now.

PR #1072, opened 2017-11-28
Fix `go_repository` links.

tiziano88:


bazel-io:

PR #1075, opened 2017-11-28
Fix cgo filtering

ianthehat:
This changes our filtering code to understand the implied build tags caused by
import "C" lines in go files that build.Context.Match does not handle.
It also changes our cgo processing code to share the same logic.

### DIFF
--- a/go/tools/builders/filter.go
+++ b/go/tools/builders/filter.go
@@ -15,8 +15,15 @@
 package main
 
 import (
+	"go/ast"
 	"go/build"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"log"
 	"path/filepath"
+	"strconv"
+	"strings"
 )
 
 // filterFiles applies build constraints to a list of input files. It returns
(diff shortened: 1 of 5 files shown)


go/tools/builders/filter.go
```diff
@@ -37,7 +44,57 @@ func filterFiles(bctx build.Context, inputs []string) ([]string, error) {
...
+	}
+	isCgo, pkg, err := testCgo(input, data)
+	if err != nil {
+		return false, false, "", err
+	}
+	// match if cgo is enabled or the file is not cgo
+	return match && (bctx.CgoEnabled || !isCgo), isCgo, pkg, err
+}
+
+func testCgo(src string, data []byte) (bool, string, error) {
```
jayconrod: Instead of requiring callers to pass in the full contents, let's just pass the filename to `ParseFile`. This should let us avoid reading the entire file, and will simplify calling code.

Also, name return values.

Also, since this more generally collecting metadata now, maybe rename to something like `readGoMetadata`?

Sorry I know this function was just moved verbatim. Feel free to tell me to do it in another PR. :)

ianthehat:
I agree with everything you wrote, filed #1081 to track doing it all.

PR #1076, opened 2017-11-28
control executable names

ianthehat:
This fixes executable extensions and also separates executables by build mode
for better cache hits.
This requires bazel 0.8 or higher to work


go/private/rules/binary.bzl
```diff
@@ -37,19 +37,19 @@ def _go_binary_impl(ctx):
     go_toolchain = ctx.toolchains["@io_bazel_rules_go//go:bootstrap_toolchain"]
   gosource = collect_src(ctx)
   executable = ctx.outputs.executable
```
jayconrod: This seems like the problem line.

ianthehat: Yeah, it was a rebase failure.
I have already fixed it, but not bothered pushing the fix because the test fixes in #1080 have to go in first anyway

PR #1079, opened 2017-11-28
Update //go/platform config_setting rules to use constraint_values

jayconrod:
* Minimum Bazel version is now 0.8
* GOOS_GOARCH (the list of platforms) is updated to include all
  platforms the Go SDK supports, using "go tool dist list".
* //go/platform: config_setting declarations moved to macro in list.bzl.
* declare_constraints macro moved to //go/toolchain:toolchains.bzl
* Gazelle's -experimental_platforms flag is removed. This behavior is
  always on now.
* Fixed a bug in Gazelle where files with negated architectures were
  not considered arch-specific (also for OS).


clnperez:
For ppc64le, this still pulls the amd64 go binary: 
`Downloading https://storage.googleapis.com/golang/go1.9.2.linux-amd64.tar.gz: 95,591,276 bytes`

ianthehat:
The stuff we have done was all to support cross compiling to all the possible goos/goarch pairs, we only support building on darwin_amd64, linux_amd64 and windows_amd64 right now.
If you look in [toolchain.bzl](https://github.com/bazelbuild/rules_go/blob/master/go/private/toolchain.bzl#L31) you can see how we pick the sdk to download. The ctx.os.name does not have sufficient information to pick between the unix variants, and there is not much else to go on. If you can find a way to improve the _go_download_sdk_impl function to pick the right sdk we would be grateful :)

clnperez:
Ah, I see. Well, definitely something we'd like to have working outside a cross-compile so we'll see what we can figure out.

Issue #1083, opened 2017-11-29
Built assembly code not reproducible


AustinSchuh:
For example, github.com/golang/snappy (for us) is being built with the following BUILD file

```
load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")

go_library(
    name = "go_default_library",
    srcs = [ 
        "decode.go",
        "decode_amd64.go",
        "decode_amd64.s",
        "encode.go",
        "encode_amd64.go",
        "encode_amd64.s",
        "snappy.go",
    ],  
    visibility = ["//visibility:public"],
)
```

The resulting .o file has a path built in.

```
$ xxd bazel-out/local_linux_k8-opt/bin/third_party/go/src/vendor/github.com/golang/snappy/~stripped~go_default_library~/decode_amd64.o
0000000: 676f 206f 626a 6563 7420 6c69 6e75 7820  go object linux 
0000010: 616d 6436 3420 676f 312e 392e 320a 210a  amd64 go1.9.2.!.
0000020: 0000 676f 3139 6c64 0100 fe12 2222 2e64  ..go19ld...."".d
0000030: 6563 6f64 6500 fe1e 7275 6e74 696d 652e  ecode...runtime.
0000040: 6d65 6d6d 6f76 6500 fe2e 2222 2e64 6563  memmove..."".dec
0000050: 6f64 652e 6172 6773 5f73 7461 636b 6d61  ode.args_stackma
0000060: 7000 fede 0267 6f66 696c 652e 2e2f 6465  p....gofile../de
0000070: 762f 7368 6d2f 6261 7a65 6c2d 7361 6e64  v/shm/bazel-sand
0000080: 626f 782e 3536 6130 3530 3434 6632 3834  box.56a05044f284
0000090: 3630 6631 6364 3235 3261 3534 3539 6439  60f1cd252a5459d9
00000a0: 6632 3437 2f35 3734 3238 3232 3137 3330  f247/57428221730
00000b0: 3239 3035 3639 3032 2f65 7865 6372 6f6f  29056902/execroo
00000c0: 742f 636f 6d5f 7065 6c6f 746f 6e5f 7465  t/com_peloton_te
00000d0: 6368 2f74 6869 7264 5f70 6172 7479 2f67  ch/third_party/g
00000e0: 6f2f 7372 632f 7665 6e64 6f72 2f67 6974  o/src/vendor/git
[...]

PR #1084, opened 2017-11-29
Strip out sandbox paths from built assembly code

AustinSchuh:
The full source filename gets put in the resulting binary by default.
According to 'go tool asm':

-trimpath string
        remove prefix from recorded source file paths

We can use the same trick as the C++ rules by setting PWD to be /proc/self/cwd, and then mapping that out.

Fixes: #1083

### DIFF
--- a/go/private/actions/asm.bzl
+++ b/go/private/actions/asm.bzl
@@ -40,10 +40,12 @@ def emit_asm(ctx, go_toolchain,
   add_go_env(asm_args, stdlib, mode)
   asm_args.add([source.path, "-o", out_obj])
   asm_args.add(includes, before_each="-I")
+  asm_args.add(['-trimpath', '/proc/self/cwd'])
   ctx.actions.run(
       inputs = inputs,
       outputs = [out_obj],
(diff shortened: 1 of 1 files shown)


bazel-io:

ianthehat:
Thanks for finding this, I did it for compile but forgot about asm
This cl is not the right way to fix it (it won't work in most environments, /proc/self/cwd is not universal)
If you look at what we did for compile, we just pass . and have the wrapper go program expand it at that point before invoking the underlying tool, which will work on all platforms (windows included)
If you want to change this PR to do it that way I will gratefully approve, or just let me know you don't have the time and I will do it sometime soon.

AustinSchuh:
Ian,

I'd love to make the change, but I'm significantly better at building golang than writing it at this point...  I assume I should be doing this conditional on the OS, but that's not something I can pattern match my way through.  Can you or Jay take it from here?

Thanks!!!

ianthehat:
See #1090

ianthehat:
Closing this now #1090 is merged

PR #1088, opened 2017-11-30
Run popular repo tests directly inside our build

### LINKED ISSUE #811: Migrate remaining non bazel tests
Most of the tests now run when you do bazel test //..., there are only 3 remaining, but one of them is the external repositories test which may be quite hard to port.

ianthehat:
Fixes #811


blah
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
achew22: blah?

ianthehat: yup, that should not be there, thanks!

tests/popular_repos/popular_repos.bzl
```diff
@@ -0,0 +1,165 @@
...
+_popular_repo_test_suites = repository_rule(
+    implementation = _popular_repo_test_suites_impl,
+)
+
+def popular_repos():
+  names = []
+  for repo in POPULAR_REPOS:
+    names.append(repo["name"])
+    repo = {k:v for k,v in repo.items() if not "excludes" in k}
+    go_repository(**repo)
```
jayconrod: Some of these repositories (`org_golang_x_tools`, proto stuff) will override dependencies of rules_go. This will only affect rules_go, but it could cause some weirdness. Let's wrap these declarations in `_maybe` to avoid this.

tests/popular_repos/popular_repos.bzl
```diff
@@ -0,0 +1,165 @@
...
+      fail("Invalid excludes found: {}".format(invalid_excludes))
+    content += """
+test_suite(
+    name = "{name}",
+    tests = {tests},
+)
+""".format(name=name, tests=[t for t in tests if t and not t in excludes])
+  content += """
+test_suite(
+    name = "all",
```
jayconrod: Won't this be confused with the `all` wildcard? Maybe rename to `all_tests`.

tests/popular_repos/update_test_lists.sh
```diff
@@ -0,0 +1,13 @@
+repos=(
```
jayconrod: We should probably have a copyright notice in here.

tests/popular_repos/update_test_lists.sh
```diff
@@ -0,0 +1,13 @@
...
+  org_golang_x_crypto
+  org_golang_x_net
+  org_golang_x_sys
+  org_golang_x_text
+  org_golang_x_tools
+  com_github_mattn_go_sqlite3
+  org_golang_google_grpc
+)
+
+for repo in ${repos[*]}; do
```
jayconrod: Rather than generating flat lists of tests, then combining them into a build file with Skylark, I'd argue that this script should generate the final BUILD.bazel file. It wouldn't be much extra string formatting, and it would be clearer what's happening. We could drop the `_popular_repos_test_suites` repo rule, too.

tests/test_chdir/remote.bzl
```diff
@@ -0,0 +1,14 @@
+def _test_chdir_remote_impl(ctx):
+  ctx.file("WORKSPACE", """workspace("test_chdir_remote")""")
+  ctx.file("BUILD.bazel", """""")
```
jayconrod: This is a very long way to write the empty string :)

tests/popular_repos/update_test_lists.sh
ianthehat: I would have to write all the exclusion logic, in bash, ug, but you might be right.
Originally I was trying to do this bit in skylark too, but it turned out to be not viable, and I moved the smallest part out that I could.

tests/test_chdir/remote.bzl
ianthehat: It wasn't always empty

tests/popular_repos/update_test_lists.sh
jayconrod: Maybe Python instead?

ianthehat:
jenkins test this please

ianthehat:
jenkins test this please

PR #1089, opened 2017-11-30
Clean out tmp handling

ianthehat:
In Bazel0.8 temp directories were fixed, so now we just rely on  them rather
than trying to make our own.

### DIFF
--- a/go/private/common.bzl
+++ b/go/private/common.bzl
@@ -123,7 +123,7 @@ def go_importpath(ctx):
     path = path[1:]
   return path
 
-def env_execute(ctx, arguments, environment = None, **kwargs):
+def env_execute(ctx, arguments, environment = {}, **kwargs):
   """env_executes a command in a repository context. It prepends "env -i"
   to "arguments" before calling "ctx.execute".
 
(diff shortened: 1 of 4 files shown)


go/private/common.bzl
```diff
@@ -123,7 +123,7 @@ def go_importpath(ctx):
     path = path[1:]
   return path
 
-def env_execute(ctx, arguments, environment = None, **kwargs):
+def env_execute(ctx, arguments, environment = {}, **kwargs):
```
Syeberman: Careful using mutable objects as default values: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

(I'm assuming this applies to .bzl the same as .py.)

ianthehat: It does apply, but because of "frozen" mutable objects it's much less dangerous (as in you tend to end up mutating a frozen object and then get a runtime error rather than silent bad behavior). In general I avoid functions that modify their arguments anyway, it's generally not a great idea without very explicit signaling, and only then when copying and returning is expensive enough to be worth the special case. This function for instance always makes full copy before modifying, so it's safe.
That is also btw a truly hideous piece of language design...

PR #1090, opened 2017-11-30
Trimpath for asm

ianthehat:
This uses the same solution as we did for compile, abs the path inside the
execution wrapper so that it works on all platforms.

Fixes: #1083


AustinSchuh:
I ran it on 2 copies of our repo, and it's working.  Thanks!

```
bazel test -c opt //... --stamp=no && find bazel-out/local_linux_k8-opt/bin/ bazel-out/local_linux_k8-opt/genfiles/ -type f | sort | xargs md5sum > /tmp/md5sums
```

And then diffed the outputs, ignoring manifests.

PR #1091, opened 2017-11-30
Gazelle: remove flat mode

jayconrod:
We no longer plan to use flat build files with external repositories
managed by Gazelle. Flat mode is impeding development of indexed
import resolution, so it's being removed immediately.


jayconrod:
FYI, I'll still send a few Gazelle PRs to rules_go until bazel-gazelle is completely set up.

PR #1099, opened 2017-12-02
Gazelle: split merging into two stages

jayconrod:
We need to merge some attributes on rules before indexing them. In
particular, we need to merge "importpath" and "srcs", since these
become keys in the index. It's important to account for '# keep'
comments and '# gazelle:ignore' directives so we need to use the
regular merging algorithm instead of something ad hoc in the indexer.
However, we also need to merge deps after they're resolved.

This means that we need to merge rules in two separate phases. Right
after rules are generated, we merge everything except "deps". Then we
resolve dependencies. Then we merge "deps" only.

This should also prevent us from indexing rules that are deleted.

Related #859


jayconrod:
Jenkins test this please

jayconrod:
Jenkins test this please

jayconrod:
The Travis CI failure is because run_non_bazel_tests.bash is missing. I believe Travis CI is reading the configuration from master instead of this branch.

Since the tests actually did run and pass, I'm force submitting.

Issue #1020, opened 2017-11-14
Cross compiling with cgo


ianthehat:
This is currently not possible, as we can only build the standard library in "pure" mode if we are cross compiling.
In theory there is no reason to limit this quite so much, but we need a way to pick the pure version of the standard library when cross compiling without having to compile *all* libraries in pure mode.

steeve:
Would it be possible to specify the toolchain itself?

The reason is I worry that relying on things like `CGO_CFLAGS, CGO_CPPFLAGS, CGO_CXXFLAGS, CGO_FFLAGS, CGO_LDFLAGS` or `CC`/`CXX` env vars might be a little overlapping with Bazel's way of doing, which in my view is far more powerful.

steeve:
hey @ianthehat, any good news for the new year ? :)

ianthehat:
Not even looked at this yet I am afraid

jayconrod:
(Re-triaging old issues)

This should work, at least theoretically. You can target a platform with `--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64_cgo` (note the `_cgo` suffix). An appropriate C/C++ toolchain must be configured.

Leaving this issue open because we don't have adequate test coverage or documentation yet.

devastating:
> (Re-triaging old issues)
> 
> This should work, at least theoretically. You can target a platform with `--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64_cgo` (note the `_cgo` suffix). An appropriate C/C++ toolchain must be configured.
> 
> Leaving this issue open because we don't have adequate test coverage or documentation yet.
Hello @jayconrod ,
Is this theoretically working for Linux only? I've tried it with "**--platforms @io_bazel_rules_go//go/toolchain:darwin_amd64_cgo**", and the binary built still shows "Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work. This is a stub"

These are the flags I was using:
CC=/usr/bin/gcc CXX=/usr/bin/g++ --features=pure --stamp --platforms @io_bazel_rules_go//go/toolchain:darwin_amd64_cgo

We are trying to cross-compile for darwin on Linux; my experiment is done on darwin and the result is the same.

Thanks in advance!

jayconrod:
@devastating The `--features=pure` flag disables cgo. Try without that.

devastating:
> @devastating The `--features=pure` flag disables cgo. Try without that.

[...]

PR #1101, opened 2017-12-04
ctx.fragments.cpp migration

calpeyser:
This change removes toolchain accesses that use ctx.fragments.cpp, instead using the find_cpp_toolchain macro.

The change will have to wait for a Bazel release that includes the find_cpp_toolchain macro before it is submitted.

### DIFF
--- a/go/private/rules/stdlib.bzl
+++ b/go/private/rules/stdlib.bzl
@@ -15,6 +15,7 @@
 load("@io_bazel_rules_go//go/private:providers.bzl",
     "GoStdLib",
 )
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 
 _STDLIB_BUILD = """
 load("@io_bazel_rules_go//go/private:rules/stdlib.bzl", "stdlib")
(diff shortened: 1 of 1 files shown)


jayconrod:
Looks good. When will `find_cpp_toolchain` in Bazel? Has 0.9 cut already, or will that be in 0.10?

steeve:
Could this help #1020?

ianthehat:
Nope, sorry. This is just about the cc team cleaning up, it moves something from one place to another, it does not make a fundamental difference in what is possible (not yet anyway)

jayconrod:
Closing this PR. It will need to be rewritten.

@juliexxia Ping me if this needs to go in soon. It looks like the object returned by `find_cpp_toolchains` doesn't contain everything that we need.

juliexxia:
@jayconrod, sorry having trouble seeing what is missing from find_cpp_toolchains?

jayconrod:
@juliexxia Sorry, I should have included that earlier. Here's a list of fields and methods we use in `ctx.fragments.cpp`:

* `target_gnu_system_name`
* `compiler_options(features)`
* `unfiltered_compiler_options(features)`
* `ld_executable`
* `compiler_executable`
* `c_options`

The object returned by `find_cpp_toolchain` only contains `*_executable` fields. Is there another way to get the options and the target system name?

PR #1102, opened 2017-12-04
Fix go proto library for once and for all

### LINKED ISSUE #173: Generating protos in sub directories fails
I'm trying to use a vendoring/third_party library with a flat BUILD file. The problem I have is that it contains a proto file deep in the tree. It seems fine to create a go_library with source files in sub directories, but protos don't seem to work.

I've made up a simple example. Something like:

```
~/tmp/go-bzl % tree lib 
lib
├── BUILD
└── sub
    └── bar.proto
```

```
~/tmp/go-bzl % cat lib/BUILD                                   
load("@io_bazel_rules_go//proto:go_proto_library.bzl", "go_proto_library")

go_proto_library(
    name = "sub/bar_proto",  # Also tried name="bar_proto"
    srcs = [
        "sub/bar.proto",
    ],
)
```

```
~/tmp/go-bzl % bazel build lib:sub/bar_proto                
INFO: Found 1 target...
ERROR: /home/y/tmp/go-bzl/lib/BUILD:20:1: Generating into bazel-out/local-fastbuild/bin/lib/sub/bar_proto/sub failed: linux-sandbox failed: error executing command /home/y/.cache/bazel/_bazel_y/1129aeaeb86f4311cfe08c57d10b52b4/execroot/go-bzl/_bin/linux-sandbox ... (remaining 6 argument(s) skipped).
bazel-out/local-fastbuild/bin/lib/sub/: No such file or directory
Target //lib:sub/bar_proto failed to build
```

I also inspected the sandbox, and it seems that just like the error says, there is no directory bazel-out/local-fastbuild/bin/lib/sub/, and protoc probably just fails when trying to write the proto. Should go_proto_library create the directories for output files? Any other ways to make this work?

Thanks,

### LINKED ISSUE #910: importpath must match path in repo for go_proto_library
For a go_proto_library, if I set importpath to something other than the path in the repo, go_proto_library fails with:

```
ERROR: /usr/local/google/home/rodrigoq/git/bazeltest/diff_importpath/foo/BUILD.bazel:16:1: output 'foo/lib_importpath_bar/bar/foo.pb.go' was not created
```

The rule expects the output at the declared importpath, but protoc appears to generate it based on the path within the repo.

Full repro (bazel 0.6.1):
```
> cat WORKSPACE 
git_repository(
    name = "io_bazel_rules_go",
    commit = "9cf23e2aab101f86e4f51d8c5e0f14c012c2161c",
    remote = "https://github.com/bazelbuild/rules_go.git",
)

load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains", "go_repository")

go_register_toolchains()

go_rules_dependencies()

load("@io_bazel_rules_go//proto:def.bzl", "proto_register_toolchains")

proto_register_toolchains()
> cat BUILD.bazel 
load("@io_bazel_rules_go//go:def.bzl", "go_prefix")

go_prefix("my_go_prefix")
> cat foo/foo.proto 
syntax = "proto3";

package foo;

message Foo {
  string very = 1;
  int32 useful = 2;
}
> cat foo/BUILD.bazel 

proto_library(
    name = "proto",
    srcs = ["foo.proto"],
)

go_proto_library(
    name = "lib_importpath_foo",
    importpath = "foo",
    proto = ":proto",
)

go_proto_library(
    name = "lib_importpath_bar",
    importpath = "bar",
    proto = ":proto",
)
> bazel build foo:lib_importpath_foo
INFO: Analysed target //foo:lib_importpath_foo (0 packages loaded).
INFO: Found 1 target...
Target //foo:lib_importpath_foo up-to-date:
  bazel-bin/foo/~normal~lib_importpath_foo~/foo.a
INFO: Elapsed time: 0.439s, Critical Path: 0.00s
INFO: Build completed successfully, 1 total action
> bazel build -s foo:lib_importpath_bar
INFO: Analysed target //foo:lib_importpath_bar (0 packages loaded).
INFO: Found 1 target...
SUBCOMMAND: # //foo:lib_importpath_bar [action 'Generating into bazel-out/local-fastbuild/bin/foo/lib_importpath_bar/bar']
(cd /usr/local/google/home/rodrigoq/.cache/bazel/_bazel_rodrigoq/42807b98f0f2cce20415d49167562dfe/execroot/__main__ && \
  exec env - \
  bazel-out/host/bin/external/com_github_google_protobuf/protoc '--go_out=:bazel-out/local-fastbuild/bin/foo/lib_importpath_bar/' '--plugin=protoc-gen-go=bazel-out/host/bin/external/com_github_golang_protobuf/protoc-gen-go/protoc-gen-go' --descriptor_set_in bazel-out/local-fastbuild/genfiles/foo/proto-descriptor-set.proto.bin foo/foo.proto)
ERROR: /usr/local/google/home/rodrigoq/git/bazeltest/diff_importpath/foo/BUILD.bazel:16:1: output 'foo/lib_importpath_bar/bar/foo.pb.go' was not created
ERROR: /usr/local/google/home/rodrigoq/git/bazeltest/diff_importpath/foo/BUILD.bazel:16:1: not all outputs were created or valid
Target //foo:lib_importpath_bar failed to build
INFO: Elapsed time: 0.531s, Critical Path: 0.07s
```

### LINKED ISSUE #1021: go_proto_library fails when go_import_path is not set.
bazel version 0.7.0
rules_go version 0.7.0.
platform: ubuntu-trusty

**Summary**: go_proto_library fails to locate output .pb.go files when go_package_path isn't set.

**Detail**:
There are a number of `go get` repositories which contain .proto files which define package, but not go_package_path.
When imported with go_repository, they fail to build, because `protoc` puts the `.pb.go` output in a different place from where the rule expects to find it.
When I control the source, a work around is setting go_package_path.
The combination makes it difficult to contribute upstream.

This can be reproduced by attempting to build a rule that depends on github.com/njt99/rules_go_missing_pb_go/tiny. I've also tested it at HEAD.

ianthehat:
This uses an aspect to collect up the full set of proto imports and what their
go import path should be, and then hands that to the proto compile line.
It then attempts to map the outputs to the expected outputs by basename.
All of this means that the option go_package lines are not needed for a correct
build.

Fixes #1021
Fixes #910
Fixes #173

### DIFF
--- a/go/tools/builders/protoc.go
+++ b/go/tools/builders/protoc.go
@@ -28,14 +28,14 @@ import (
 	"strings"
 )
 
-type FileInfo struct {
-	Base       string    // The basename of the path
-	Path       string    // The full path to the final file
-	Expected   bool      // Whether the file is expected by the rules
-	Created    bool      // Whether the file was created by protoc
-	From       *FileInfo // The actual file protoc produced if not Path
-	Unique     bool      // True if this base name is unique in expected results
-	Ambiguious bool      // True if there were more than one possible outputs that matched this file
+type genFileInfo struct {
+	base       string       // The basename of the path
+	path       string       // The full path to the final file
+	expected   bool         // Whether the file is expected by the rules
+	created    bool         // Whether the file was created by protoc
+	from       *genFileInfo // The actual file protoc produced if not Path
+	unique     bool         // True if this base name is unique in expected results
+	ambiguious bool         // True if there were more than one possible outputs that matched this file
 }
 
 func run(args []string) error {
@@ -75,21 +75,21 @@ func run(args []string) error {
 		return fmt.Errorf("error running protoc: %v", err)
 	}
 	// Build our file map, and test for existance
-	files := map[string]*FileInfo{}
-	byBase := map[string]*FileInfo{}
+	files := map[string]*genFileInfo{}
+	byBase := map[string]*genFileInfo{}
 	for _, path := range expected {
-		info := &FileInfo{
-			Path:     path,
-			Base:     filepath.Base(path),
-			Expected: true,
-			Unique:   true,
+		info := &genFileInfo{
+			path:     path,
+			base:     filepath.Base(path),
+			expected: true,
+			unique:   true,
 		}
-		files[info.Path] = info
-		if byBase[info.Base] != nil {
-			info.Unique = false
-			byBase[info.Base].Unique = false
+		files[info.path] = info
+		if byBase[info.base] != nil {
+			info.unique = false
+			byBase[info.base].unique = false
 		} else {
-			byBase[info.Base] = info
+			byBase[info.base] = info
 		}
 	}
 	// Walk the generated files
@@ -99,48 +99,48 @@ func run(args []string) error {
 		}
 		info := files[path]
 		if info != nil {
-			info.Created = true
+			info.created = true
 			return nil
 		}
-		info = &FileInfo{
-			Path:    path,
-			Base:    filepath.Base(path),
-			Created: true,
+		info = &genFileInfo{
+			path:    path,
+			base:    filepath.Base(path),
+			created: true,
 		}
 		files[path] = info
-		copyTo := byBase[info.Base]
+		copyTo := byBase[info.base]
 		switch {
 		case copyTo == nil:
 			// Unwanted output
-		case !copyTo.Unique:
+		case !copyTo.unique:
 			// not unique, no copy allowed
-		case info.From != nil:
-			copyTo.Ambiguious = true
-			info.Ambiguious = true
+		case info.from != nil:
+			copyTo.ambiguious = true
+			info.ambiguious = true
 		default:
-			copyTo.From = info
-			copyTo.Created = true
-			info.Expected = true
+			copyTo.from = info
+			copyTo.created = true
+			info.expected = true
 		}
 		return nil
 	})
 	buf := &bytes.Buffer{}
 	for _, f := range files {
 		switch {
-		case f.Expected && !f.Created:
-			fmt.Fprintf(buf, "Failed to create %v.\n", f.Path)
-		case f.Expected && f.Ambiguious:
-			fmt.Fprintf(buf, "Ambiguious output %v.\n", f.Path)
-		case f.From != nil:
-			data, err := ioutil.ReadFile(f.From.Path)
+		case f.expected && !f.created:
+			fmt.Fprintf(buf, "Failed to create %v.\n", f.path)
+		case f.expected && f.ambiguious:
+			fmt.Fprintf(buf, "Ambiguious output %v.\n", f.path)
+		case f.from != nil:
+			data, err := ioutil.ReadFile(f.from.path)
 			if err != nil {
 				return err
 			}
-			if err := ioutil.WriteFile(f.Path, data, 0644); err != nil {
+			if err := ioutil.WriteFile(f.path, data, 0644); err != nil {
 				return err
 			}
-		case !f.Expected:
-			fmt.Fprintf(buf, "Unexpected output %v.\n", f.Path)
+		case !f.expected:
+			fmt.Fprintf(buf, "Unexpected output %v.\n", f.path)
 		}
 		if buf.Len() > 0 {
 			fmt.Fprintf(buf, "Check that the go_package option is %q.", *importpath)
--- a/go/tools/builders/protoc.go
+++ b/go/tools/builders/protoc.go
@@ -16,6 +16,8 @@
 package main
 
 import (
+	"bytes"
+	"errors"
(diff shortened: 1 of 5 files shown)


go/tools/builders/protoc.go
```diff
@@ -27,21 +28,32 @@ import (
...
 	plugin := flags.String("plugin", "", "The go plugin to use.")
 	importpath := flags.String("importpath", "", "The importpath for the generated sources.")
+	flags.Var(&options, "option", "The plugin options.")
+	flags.Var(&descriptors, "descriptor_set", "The descriptor set to read.")
 	flags.Var(&expected, "expected", "The expected output files.")
+	flags.Var(&imports, "import", "Map a proto file to an import path.")
 	if err := flags.Parse(args); err != nil {
 		return err
 	}
+	plugin_base := filepath.Base(*plugin)
```
jayconrod: Nit: `pluginBase` and `pluginName`

go/tools/builders/protoc.go
```diff
@@ -57,22 +69,41 @@ func run(args []string) error {
 		}
 	}
 	if len(notFound) > 0 {
+		missing := []string{}
```
jayconrod: It doesn't seem like anything appends to this.

proto/compiler.bzl
```diff
@@ -22,37 +22,32 @@ load("@io_bazel_rules_go//go/private:providers.bzl",
...
     go_srcs.append(out)
     if outpath == None:
         outpath = out.dirname[:-len(importpath)]
-  plugin_base_name = compiler.plugin.basename
-  if plugin_base_name.startswith(_protoc_prefix):
-    plugin_base_name = plugin_base_name[len(_protoc_prefix):]
   args = ctx.actions.args()
-  args.add(["-protoc", compiler.protoc.path])
   args.add([
+      "-protoc", compiler.protoc,
```
jayconrod: Very small nit: `--protoc` for consistency with other options.

proto/def.bzl
```diff
@@ -18,12 +19,31 @@ load("@io_bazel_rules_go//proto:compiler.bzl",
     "GoProtoCompiler",
 )
 
+GoProtoImports = provider()
+
+def get_imports(attr):
+    imports = []
```
jayconrod: Style nit: 2-space indent for functions and if-statement.

go/tools/builders/protoc.go
ianthehat: Was missing a commit that I merged to the wrong branch that improved error checking
when I was merging it though I decided to re-write the whole loop to be more thorough and easier to understand (I hope)

go/tools/builders/protoc.go
ianthehat: That's what happens when you try to write skylark and go at the same time...

go/tools/builders/protoc.go
```diff
@@ -26,6 +28,16 @@ import (
 	"strings"
 )
 
+type FileInfo struct {
```
jayconrod: I think it's a little confusing this has the same name as `os.FileInfo`. It doesn't need to be public though, right? Maybe `genFileInfo` would be a better name.

Also, consider making the fields non-public.

go/tools/builders/protoc.go
```diff
@@ -62,47 +74,77 @@ func run(args []string) error {
...
-						}
-						if err := ioutil.WriteFile(matches[0], data, 0644); err != nil {
-							return err
-						}
-					}
-				}
-			}
+	// Walk the generated files
+	filepath.Walk(".", func(path string, f os.FileInfo, err error) error {
+		if !strings.HasSuffix(path, ".pb.go") {
```
jayconrod: Can't the suffix be something other than .pb.go? I thought the grpc gateway plugin generated something else.

go/tools/builders/protoc.go
ianthehat: Meh, this entire directory is "private", it's all built directly to binaries and not libraries, and it's all in package main, so it's not like it can be imported.

go/tools/builders/protoc.go
ianthehat: I have never run or seen the output of grpc-gateway, you may be right, who knows.
We can add an extra command line arg that passes in the expected extension from the compiler rule when we reach that bridge.

pcj: Drive-by comment...  Yes it can be `.pb.gw.go`.  I'm not sure there is an overall generalization.

https://github.com/pubref/rules_protobuf/blob/master/grpc_gateway/BUILD#L8

ianthehat: Okay, thanks!
I guess the real question is, do we even need to do this filtering, in theory there should not be any spare random files loafing around in this directory anyway... or maybe just .go is enough.

PR #1106, opened 2017-12-05
Gazelle: use rule index to resolve dependencies

jayconrod:
Related #859


jayconrod:
This is the last big PR for indexed dependency resolution. There may be a few small follow-up changes to close issues referenced in #859, but this is the big PR that turns everything on.

PR #1114, opened 2017-12-07
Allow pure go cross compilation within a single build

### LINKED ISSUE #1030: Provide a platform to the go_binary rule (via skylark)
I'd like to {have | be able to write} a skylark build rule that can assemble a set of cross-compiled go binaries.

My first issue is caching of outputs:

In my previous cross-compilation implementation for this repo, there was a rule `xgo_binary` that took an `os_arch` attribute, walked down the go_library deps via a parameterized aspect, and placed cross-compiled outputs under a subdirectory scoped to the os_arch attribute `bazel-bin/.../${os_arch}/foo.a`.  The benefit of this that bazel could cache all those outputs and not have to rebuild them every time you built a different "platform".  The design of this was from the go stdlib itself (cross-compiling the stdlib places it in a subdirectory like `windows_386/`).

From what I can tell thus far, every time I cross-compile a new platform now, bazel rebuilds everything down the aspect graph to that platform, eliminating the possibility of caching cross-compiled outputs.

My second issue is that I can't specify the platform (like `@io_bazel_rules_go//go/toolchain:linux_amd64`) to the rule `go_binary` rule itself.  This may or may not be possible by adding a 'platform(s)' attr to the go_binary rule.

This means that I have to script the assembly of those multiple cross-compiled outputs outside of bazel, and try to duplicate bazel's logic to figure out if any of the upstream deps have changed and rebuild.

It's a little frustrating because that rule did everything except cgo, and now we're back after a year-long loop with platforms and we still can't do cgo and the cache performance is much lower.

### LINKED ISSUE #1104: Cross compiling filegroups
Is it possible to use `filegroup`s `output_group` attribute to get a cross-compiled binary now that the preferred way of cross-compiling is not done through output groups?

ianthehat:
This adds goos and goarch attributes to a go_binary
If specified they force the binary to cross compile to the specified
architecture.
It also adds a test that uses this to cross compile and verify the binary was
correctly cross compiled.
This also required multiple rules for the same binary in different modes, which
meant we needed control over the binary file name (so not using the rule name)

Fixes #1104
Fixes #1030

### DIFF
--- a/go/private/actions/action.bzl
+++ b/go/private/actions/action.bzl
@@ -22,8 +22,8 @@ def add_go_env(args, stdlib, mode):
       "-cgo=" + ("0" if mode.pure else "1"),
   ])
 
-def bootstrap_action(ctx, go_toolchain, inputs, outputs, mnemonic, arguments):
-  stdlib = go_toolchain.stdlib.cgo
+def bootstrap_action(ctx, go_toolchain, mode, inputs, outputs, mnemonic, arguments):
+  stdlib = go_toolchain.stdlib.get(ctx, go_toolchain, mode)
   ctx.actions.run_shell(
     inputs = inputs + stdlib.files,
     outputs = outputs,
--- a/go/private/actions/binary.bzl
+++ b/go/private/actions/binary.bzl
@@ -39,7 +39,7 @@ def emit_binary(ctx, go_toolchain,
       importpath = importpath,
       importable = False,
   )
-  executable = declare_file(ctx, ext=go_toolchain.data.extension, mode=mode)
+  executable = declare_file(ctx, name=name, ext=go_toolchain.data.extension, mode=mode)
   go_toolchain.actions.link(ctx,
       go_toolchain = go_toolchain,
       archive=goarchive,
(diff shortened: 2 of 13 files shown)


pcj:
Woo hoo, long time coming on this, thanks guys.  👍 

I still think moving the platform targets to separate rule (I had called it `xgo_binary`) that takes a go_binary dependency as a list of platform strings like ["linux_amd64", "darwin_386"] and outputs a set of cross-compiled files (a non-runnable target) should be considered, because `go_binary is a runnable target (which obviously only makes sense on the host platform).

ianthehat:
I have thought about doing it that way, but there are all sorts of complications.
One is do you have to have the go_binary and then have the xgo_binary derive from it? If so, what do you do when the xgo_binary specifies the same mode as the current host?
What happens when you build a go_binary and specify a cross compiling toolchain on the command line that happens to match an xgo_binary?
Lists are bad to pick things out of, but easy to form from separate things, so going with a rule per target platform is generally more flexible (easier to consume in selects, or package in strange combinations)
Then there are things like pure and static, which do affect the host but also affect cross, and pure is required for cross right now.
If you can take the cross test I added in this cl, and show me how you think it should be, I am still open to change on all this stuff, and it will give us something concrete to discuss.
It's all really a temporary hack anyway, in the future the feature flags and/or toolchain transitions will make it redundant I hope.

PR #1121, opened 2017-12-08
fix(gazelle): fix nil pointer exception

codesuki:
#1099 introduced a nil pointer exception happening when BUILD files contain a `# keep` note.


bazel-io:

jayconrod:
Thanks for reporting and fixing this.

Could you post an example of a file where this happens with `# keep`? I see a path with `# gazelle:ignore`, but I'm surprised it happens with `# keep`.

jayconrod:
Jenkins test this please

codesuki:
Sorry to have confused you. It was `gazelle:ignore`.

PR #1123, opened 2017-12-09
Gazelle: support prefix directive

jayconrod:
Gazelle now supports '# gazelle:prefix <prefix>' directives in build
files. Such a directive will set the Go importpath prefix for rules in
that directory and in subdirectories.

Prefixes may be empty, but rules cannot be generated in a directory
where the prefix is set to the empty string (an error will be
reported).

Fixes bazelbuild/bazel-gazelle#6

### DIFF
--- a/go/tools/gazelle/config/config.go
+++ b/go/tools/gazelle/config/config.go
@@ -17,6 +17,7 @@ package config
 
 import (
 	"fmt"
+	"go/build"
 	"strings"
 )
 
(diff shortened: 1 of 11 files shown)


jayconrod:
One more thing I'd like to do before deprecating `go_prefix` is make `importpath` optional for `go_binary` and `go_test`. That attribute is still somewhat unexpected there. I can work on that this week.

PR #1126, opened 2017-12-11
Gazelle: don't set visibility in proto rules with default visibility

### LINKED ISSUE #1119: Gazelle does not respect proto visibility.
Gazelle auto generates visibility public for all proto files, however, often you want to restrict visibility for proto files that are internal to a project and not part of its external facing API.

jayconrod:
Fixes #1119


jayconrod:
Jenkins test this please

PR #1128, opened 2017-12-11
Add cpp flags to stdlib rule.

### LINKED ISSUE #985: cpp fragment options don't seem to be used during stdlib compilation
https://github.com/bazelbuild/rules_go/blob/f08376e2826201ad8dfe664b07b564262f79ecf4/go/private/rules/stdlib.bzl#L103

But they're exported on the line linked to above. Is this intended?

If these options are accidentally not wired to CGO_CXXFLAGS, etc. then I believe that using them *and* ensuring they include -fPIC is needed to make it possible to use go's c-shared buildmode. I say this because I reimplemented most of this logic by hand before realizing that this may be the problematic line.

nathantchan:
This change adds the flags from cpp fragments as environment variables
to the go install command.

Fixes #985

### DIFF
--- a/go/private/rules/stdlib.bzl
+++ b/go/private/rules/stdlib.bzl
@@ -51,13 +51,19 @@ def _stdlib_impl(ctx):
       cpp.unfiltered_compiler_options(features) +
       cpp.link_options +
       linker_options)
+  cleaned_compiler_options = []
+  for s in compiler_options:
+    if s == "-fcolor-diagnostics" or s == "-Wall":
+      continue
+    else:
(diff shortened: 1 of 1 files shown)


bazel-io:

jayconrod:
cc @calpeyser

jayconrod:
Jenkins test this please

nathantchan:
Hmm...looks like I need to strip out some more flags.

jayconrod:
Jenkins test this please

jayconrod:
Manually verified this is passing on my Mac. Jenkins is being flaky.

PR #1130, opened 2017-12-12
Gazelle: fix deletion of rules with deps

jayconrod:
Rules are deleted after they are merged with empty rules produced by
the generator when they have no attributes other than "name" and
"visibility".
    
An earlier change split merging into two phases: everything except
"deps", and then "deps".
    
With this change, both merge phases have a chance to delete
rules. This means that rules that had deps but are now empty will be
deleted.

Fixes bazelbuild/bazel-gazelle#22


jayconrod:
Jenkins test this please

PR #1140, opened 2017-12-14
Add GoContext

ianthehat:
This collects up all dependancies that were on the toolchain, but shoudld be in
target configuration (rather than host) and migrates them to a common
go_context_data dependancy.
It then adds a GoContext object, that holds the toolchain, build mode, standard
library and all mode specific attributes. go_context(ctx) will build one of
these and prepare to use it.
This simplifies a lot of the logic in the rules, and helps enforce the isolation
of action methods from ctx attributes.


extras/embed_data.bzl
```diff
@@ -12,18 +12,12 @@
...
-load("@io_bazel_rules_go//go/private:common.bzl",
-    "declare_file",
-)
-load("@io_bazel_rules_go//go/private:rules/helpers.bzl",
-    "new_go_library",
-    "library_to_source",
-)
-load("@io_bazel_rules_go//go/private:mode.bzl",
-    "get_mode",
+load("@io_bazel_rules_go//go/private:context.bzl", #TODO: This ought to be def
```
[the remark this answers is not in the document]
ianthehat: Causes a cycle, because we expose this through def.
If we are willing to break compatability, and not expose this from go:def.bzl we can fix it

go/private/context.bzl
```diff
@@ -0,0 +1,219 @@
...
+        package_list = ctx.file._package_list,
+    )
+
+go_context_data = rule(
+    _go_context_data,
+    attrs = {
+        "strip": attr.string(mandatory=True),
+        # Hidden internal attributes
+        "_stdlib_all": attr.label_list(default = _stdlib_all()),
+        "_crosstool": attr.label(default=Label("//tools/defaults:crosstool")),
```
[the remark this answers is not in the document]
ianthehat: I think so, it's the same thing we were doing before, except we used to do it on the toolchain.

go/toolchains.rst
```diff
@@ -56,6 +59,9 @@ repository before you call go_register_toolchains_.
 The toolchain
 ~~~~~~~~~~~~~
 
+This a wrapper over the sdk that provides enough extras to match, target and work on a specific
+platforms. It should be considered an opaqute type, you only ever use it through `the context`_.
```
jayconrod: s/opaqute/opaque/

go/toolchains.rst
```diff
@@ -97,6 +103,13 @@ example of this in `limiting the available toolchains`_.
 It is important to note that you **must** also register the boostrap toolchain for any other
 toolchain that you register, otherwise the tools for that toolchain cannot be built.
 
+
+
+The context
+~~~~~~~~~~~
+
+This is the type you use if you are writing custom rules that need
```
jayconrod: that need ... ?

go/toolchains.rst
```diff
@@ -12,6 +12,9 @@ Go toolchains
 .. _register_toolchains: https://docs.bazel.build/versions/master/skylark/lib/globals.html#register_toolchains
 .. _compilation modes: modes.rst#compilation-modes
 .. _go assembly: https://golang.org/doc/asm
+.. _GoLibrary: providers.rst#GoLibrary
```
jayconrod: I have no idea why, but GitHub is just showing the raw file instead of rendering it.

go/toolchains.rst
```diff
@@ -566,156 +626,191 @@ cover
...
+| :param:`go`                    | :type:`GoContext`           | |mandatory|                       |
 +--------------------------------+-----------------------------+-----------------------------------+
-| An additional archive whose objects will be appended to the output.                              |
-| This can be an ar file in either common form or either the bsd or sysv variations.               |
+| This must be the same GoContext object you got this function from.                               |
++--------------------------------+-----------------------------+-----------------------------------+
+| :param:`resolver`              | :type:`function`            | :value:`None`                     |
++--------------------------------+-----------------------------+-----------------------------------+
+| This is the function that gets invoked when converting from a GoLibrary to a GoSource.           |
+| See resolver_ for a                                                                              |
```
jayconrod: for a ... ?

PR #1141, opened 2017-12-14
Use bazel-gazelle instead of go/tools/gazelle everywhere

jayconrod:
* go_repository uses new Gazelle. @bazel_gazelle added to
  go_rules_dependencies() until go_repository is moved to
  @bazel_gazelle or removed.
* gazelle rule uses new Gazelle, but users should migrate to new
  gazelle rule in @bazel_gazelle//:def.bzl.
* README.rst and Vendoring.md point to new Gazelle.
* wtool and tests/bazel_tests.bzl also updated. Data dependency
  removed from build_with_old_sdk test (why was it there?).
* Old Gazelle left in place for now.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -25,6 +25,7 @@ lines_sorted_test(
 
 gazelle(
     name = "gazelle",
+    prefix = "github.com/bazelbuild/rules_go",
 )
 
 # This could be any file, used as an anchor point for the directory in tests
@@ -42,7 +43,6 @@ go_path(
         "//go/tools/builders:link",
         "//go/tools/builders:md5sum",
         "//go/tools/fetch_repo",
-        "//go/tools/gazelle/gazelle",
         "//go/tools/wtool",
     ],
 )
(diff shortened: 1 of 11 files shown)


BUILD.bazel
```diff
@@ -25,6 +25,7 @@ lines_sorted_test(
 
 gazelle(
     name = "gazelle",
+    prefix = "github.com/bazelbuild/rules_go",
```
ianthehat: I thought we were doing this with a directive rather than an attribute now?

jayconrod: Done. I'm a bit ambivalent about this though. Should we remove all attributes on the new `gazelle` rule (assuming they can be replaced by directives)?

ianthehat: If gazelle is doing the rewriting, yes, if not, lets add debug prints that strongly encourage people to switch to the directives.

jayconrod: Gazelle will at least rewrite the rule. It could do the attributes at the same time. bazelbuild/bazel-gazelle#27

PR #1146, opened 2017-12-16
Fixes for windows

ianthehat:
No env command
.exe extensions
c:/ is an absoulute path
some arguments need quoting

### DIFF
--- a/go/private/skylib/lib/paths.bzl
+++ b/go/private/skylib/lib/paths.bzl
@@ -65,7 +65,7 @@ def _is_absolute(path):
   Returns:
     `True` if `path` is an absolute path.
   """
-  return path.startswith("/")
+  return path.startswith("/") or path[1] == ":"
 
 
 def _join(path, *others):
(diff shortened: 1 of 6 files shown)


go/private/skylib/lib/paths.bzl
```diff
@@ -65,7 +65,7 @@ def _is_absolute(path):
   Returns:
     `True` if `path` is an absolute path.
   """
-  return path.startswith("/")
+  return path.startswith("/") or path[1] == ":"
```
jayconrod: Could cause problems with very short strings or with relative paths that coincidentally have a colon there.

Maybe translate `IsAbs` from $GOROOT/src/path/filepath/path_windows.go (skipping the UNC stuff).

ianthehat: Yes, this is far from correct, but it needs to be written properly up stream (bazelbuild/bazel-skylib#9). For now this lets us make progress, and none of the problems are relevant in practice for our only use of this function.

PR #1149, opened 2017-12-18
Handle windows-style paths and whitespace for C++ compiler path

droyo:


ianthehat:
#1146 should already fix these

PR #1151, opened 2017-12-18
Working binaries built on windows

ianthehat:
A basic build of a pure go binary on windows now fully works.

### DIFF
--- a/go/private/rules/stdlib.bzl
+++ b/go/private/rules/stdlib.bzl
@@ -17,7 +17,6 @@ load("@io_bazel_rules_go//go/private:providers.bzl",
 )
 load("@io_bazel_rules_go//go/private:common.bzl",
     "paths",
-    "goos_to_extension",
 )
 
 _STDLIB_BUILD = """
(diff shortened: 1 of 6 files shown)


go/private/rules/stdlib.bzl
```diff
@@ -39,12 +40,13 @@ def _stdlib_impl(ctx):
...
   for f in ctx.files._host_sdk:
-    prefix, found, extension  = f.path.partition("bin/go")
+    prefix, found, _  = f.path.partition("bin/go")
     if found:
       sdk = prefix
+      break
   if not sdk:
     fail("Could not find go executable in go_sdk")
-  go = ctx.actions.declare_file("bin/go" + extension)
+  go = ctx.actions.declare_file("bin/go" + goos_to_extension(ctx.attr.goos))
```
jayconrod: Is this right? Seems like this would add .exe if you were cross-compiling for Windows from Linux.

ianthehat: Nope, it's not right, realized as soon as I uploaded it.
The old code was broken though because it failed to actually hold on to the extension, and when I fixed that it was sometimes finding other binaries that start with the same string (gofmt).
I am finding a different solution...

ianthehat:
Jenkins test this please

PR #1153, opened 2017-12-18
Improve performance of gazelle --mode=diff

yasushi-saito:
Don't invoke the "diff" command if the contents haven't changed.
Don't call Sync() on every write. Just closing the file suffices.

Before this change, diffing a repo with 1600 BUILD files would take 60 seconds.
After the change, it takes 2 seconds.

### DIFF
--- a/go/tools/gazelle/gazelle/diff.go
+++ b/go/tools/gazelle/gazelle/diff.go
@@ -16,6 +16,7 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


bazel-io:

yasushi-saito:
Signed the CLA.

go/tools/gazelle/gazelle/diff.go
```diff
@@ -25,19 +26,31 @@ import (
 )
 
 func diffFile(c *config.Config, file *bf.File) error {
+	var oldContents []byte
```
jayconrod: This block can replaced with a call to `ioutil.ReadFile`.

go/tools/gazelle/gazelle/diff.go
```diff
@@ -25,19 +26,31 @@ import (
...
+		return nil
+	}
 	f, err := ioutil.TempFile("", c.DefaultBuildFileName())
 	if err != nil {
 		return err
 	}
 	defer os.Remove(f.Name())
-	defer f.Close()
-	if _, err := f.Write(bf.Format(file)); err != nil {
+	if _, err := f.Write(newContents); err != nil {
```
jayconrod: And this can be replaced with `ioutil.WriteFile`.

PR #1156, opened 2017-12-18
Check Bazel version in def.bzl

### LINKED ISSUE #1108: Check minimum Bazel version
We should check that the version of Bazel being used is at least the minimum supported version. Currently, error messages can be confusing when an earlier, unsupported version is used.

Parsing the version is non-trivial. Kubernetes does this at https://github.com/kubernetes/repo-infra/blob/master/defs/bazel_version.bzl (originally copied from TensorFlow). We should probably copy their implementation.

jayconrod:
When def.bzl is loaded, it will parse native.bazel_version and compare
against a defined minimum.

The implementation for this is borrowed from TensorFlow. k8s also uses
this.

Fixes #1108

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -57,8 +57,6 @@ load("@io_bazel_rules_go//go/private:tools/path.bzl",
 load("@io_bazel_rules_go//go/private:tools/vet.bzl",
     _go_vet_test = "go_vet_test",
 )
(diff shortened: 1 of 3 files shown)


jayconrod:
Force submitting. //tests/gc_opts_unsafe:check_unsafe_srcs_test timed out on Darwin, probably due to heavy contention on the Mac queue.

I wish I could re-run failing tests only.

Issue #1139, opened 2017-12-14
go_proto_library plugin implicitly depends on proctoc-gen-go


ashi009:
`go_proto_library` uses `tools/builders/go-protoc` to generate pb files.  Which adds `import_path` flag to the proto gen request.  Which is not a standard protoc plugin option, and leads to other plugins to complaint on this.  So the current status with `go_proto_library` is that it only support protoc-gen-go sub-plugins, more specifically, only grpc is supported atm.

Shall we make it part of the `go_proto_complier` options to avoid this?

ianthehat:
I have working prototypes using mwitkow/go-proto-validators and I made gogo work for the weaveworks project already.
If you have any more plugins you want me to add natural support for, please list them and I will look into it.
The plan is for the proto/ folder to have a .bzl file for all the reasonably common go plugins, and enough examples to make sure they keep working, the most recent commits in #1131 were related to fixing issues with these (notably allowing individual proto plugins to supply extra deps correctly)

ashi009:
Thanks for the quick reply. The one thing I’d love to see work is
protoc-gen-grpc-gateway. It has been a problem since day one, and I don’t
want to rely on pubref for gateway code generation any longer — it has
always been a blocker for me switching to the latest stable buildenv.

protoc-gen-grpc-gateway is a protoc plug-in that generates files that are
depend on go_libraries. The funny part of this generator is that, it’s
incapable of generating a fully functional go file, but instead it needs to
combine the pb.go file generated by protoc-gen-go in grpc mode.

I haven’t tried out the latest cl yet. Will take a look in the morning.
Ian Cottrell <notifications@github.com>于2017年12月15日 周五上午12:31写道：

achew22:
@ianthehat, FYI I'm a maintainer on grpc-gateway and would be more than happy to help adding support for it to rules_go

ianthehat:
If you want to look at what I did for the validator and put up a PR (with a test), it would be welcome, I was going to do the gogo ones next and then get to the gateway one sometime after that.
If you want to put the rules in the grpc-gateway repository, that's fine too, and should work just as well

ashi009:
@achew22 @ianthehat PTL.

tkellen:
Any chance someone subscribed to this issue could point me in the direction of an example implementation using these functionality?

jayconrod:
Fixed in #1165

Issue #992, opened 2017-11-07
Creating a new go proto toolchain


achew22:
With the advent of 0.7.0 it's beginning to seem like it might be feasible to write a set of toolchain based rules for [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gatweay). Do you think this code is in a good enough place to do that? Are things stable enough that I won't hate myself in 2 weeks or should I hold off for a bit longer till 1.0?

If things are getting close (or already there), would you be willing to point me around to some hints? I'm sort of flailing around in the dark right now on these efforts.

Most of my confusion revolves around what I think its probably a simple misunderstanding. I was expecting the `proto_library` rule to produce a `FileDescriptorSet` proto which was then passed into the `go_proto_library` which synthesized the files that were necessary based on the go-grpc/go plugins. However that understanding appears to be naive. Could you point me to any reference materials you have on how to interact with `proto_library`.

This work would probably also be appreciated by [gogo/protobuf](https://github.com/gogo/protobuf) and [go-proto-validators](https://github.com/mwitkow/go-proto-validators) so it will hopefully be more than just me.

jayconrod:
@ianthehat is the best person to answer this. We're at a conference this week, so we might be a bit slow to respond.

achew22:
Friendly ping

ianthehat:
The go_proto_library interface is probably mostly stable, the write your own toolchain interface is definitely not.
The big thing is that I mostly assumed that you only wanted to use a single plugin to translate from proto to go in any given build, but there seem to be people that want to use normal and gogo in the same tree, so we might need to rethink that part.
Currently I am still focused on cross compilation, then making it work on windows, after that I want to come back to proto and look again.

ashi009:
I've already gave grpc-gateway a try with the current toolchain implementation.  It turned out to be surpassingly easy:

- add googleapis as a new proto dependency, and gateway as a new go_repository
- add grpc-gateway tool chain
- create go_grpc_gateway_library rule
- add go_grpc_gateway_library next to the go_grpc_library rule, and embed the latter
- make go_default_library point to the gateway_library

It also comes with the benefit, that it will work with gogo generated package without any extra effort (I think).  I'll soon push the code to GitHub, and post the commit link here (probably not a PR as I'd love to hear some input from you guys first)

ashi009:
[...]

PR #1163, opened 2017-12-20
proto: Add grpc-gateway compiler

ashi009:
Add grpc-gateway compiler in //proto (#992 and #1139).

This CL is a working prototype.  I'll add tests if this direction looks good.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -17,8 +17,8 @@
 load("@io_bazel_rules_go//go/private:common.bzl", "check_version", "MINIMUM_BAZEL_VERSION")
 load("@io_bazel_rules_go//go/private:repository_tools.bzl", "go_repository_tools")
 load("@io_bazel_rules_go//go/private:go_repository.bzl", "go_repository")
-load('@io_bazel_rules_go//go/private:rules/stdlib.bzl', "go_stdlib")
-load('@io_bazel_rules_go//go/toolchain:toolchains.bzl', "go_register_toolchains")
+load("@io_bazel_rules_go//go/private:rules/stdlib.bzl", "go_stdlib")
+load("@io_bazel_rules_go//go/toolchain:toolchains.bzl", "go_register_toolchains")
 load("@io_bazel_rules_go//go/platform:list.bzl", "GOOS_GOARCH")
 load("@io_bazel_rules_go//proto:gogo.bzl", "gogo_special_proto")
 
@@ -118,6 +118,12 @@ def go_rules_dependencies():
   )
   _maybe(gogo_special_proto,
       name = "gogo_special_proto",
+  )
+  _maybe(go_repository,
+      name = "com_github_grpc_ecosystem_grpc_gateway",
+      importpath = "github.com/grpc-ecosystem/grpc-gateway",
+      tag = "v1.3.1",  # lastest as of 2018-01-10
+      build_file_proto_mode="disable",
   )
 
   # Only used by deprecated go_proto_library implementation
--- a/proto/BUILD.bazel
+++ b/proto/BUILD.bazel
@@ -45,6 +45,32 @@ go_proto_compiler(
     ],
 )
 
+GRPC_GATEWAY_GOOGLE_API_TYPE_REMAPS = [
+    "Mgoogle/api/{}.proto=google.golang.org/genproto/googleapis/api/annotations".format(t)
+    for t in [
+        "annotations",
+        "http",
+    ]
+]
+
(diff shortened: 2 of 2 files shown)


bazel-io:

bazel-io:

ashi009:
Rebased on the current master.   Will add an example later.

ashi009:
I just realized that, writing a working example is very difficult.   `google/api` repo cannot imported directly as gazelle is not able to generate proper build file for those files.  There is also no repository hosting generated go files . So the only option would be writing a `new_http_repository` rule with a local build file, which is probably not a good idea to serve the purpose of an example -- as we don't really want all the users to do this for their own repository.

Another issue with this approach is that, gazelle won't be able to generate build file "correctly" for a service proto file with http annotations.  So I'll leave this out of the example for now.

With the latest progress in grpc-gateway repo that import_path is added but the CI is failing.  I've put a big TODO on top to remove the import_path_option from the compiler option once it's released.

ianthehat:
My concern is that if this is added to this repository it becomes a rule I have to support, that other people will depend on, but I have no way to even run it, let alone check it works, so it will get broken.

Globegitter:
@ashi009 Yeah it would be great to get an example. I was just trying to use this branch and also did not quite manage. Also grpc gateway 1.3.1 has now been released, making it possible to fix one of your todos.

novabyte:
@ashi009 I've tried out the code on this pull request and have been unable to set it up either. The issue I believe is with how the "google/api/annotations.proto" and "google/api/http.proto" sources should be resolved. How were you able to make it work?

ashi009:
@Globegitter: The blocker is with google/api repo and gazelle generating build file for multiple proto files in the same directory ATM. I'll double check them this week.

@novabyte: It's tricky, but do-able. One simple solution is to put `google/api/{annotations,http}.proto` in the `${workspace}/google/api` , and throw a hand crafted `BUILD` file in it.  ie.

```
# gazelle:ignore

proto_library(
    name = "annotations_proto",
    srcs = [
        "annotations.proto",
        "http.proto",
    ],
    deps = ["@com_google_protobuf//:descriptor_proto"],
)

go_proto_library(
    name = "annotations_go_proto",
    importpath = "google.golang.org/genproto/googleapis/api/annotations",
    proto = ":annotations_proto",
    deps = ["@com_github_golang_protobuf//protoc-gen-go/descriptor:go_default_library"],
)

alias(
    name = "go_default_library",
    actual = "@org_golang_google_genproto//googleapis/api/annotations:go_default_library",
)
```

bazel-io:

Globegitter:
I added 

```
BUILD_FILE_CONTENT = """
proto_library(
  name = 'annotations_proto', 
  srcs = ['google/api/annotations.proto'], 
  deps=[
    ':http_proto',
    '@com_google_protobuf//:descriptor_proto'
  ],
  visibility = ['//visibility:public'],
)
proto_library(
  name = 'http_proto', 
  srcs = ['google/api/http.proto'], 
  visibility = ['//visibility:public'],
)
"""

new_http_archive(
    name = "com_github_googleapis_googleapis",
    build_file_content = BUILD_FILE_CONTENT,
    strip_prefix = "googleapis-94ba270676352628895b3cbb8da12a8fd3dcb88a",
    sha256 = "27eacc432c26408ab103ac1a689fd93dea2a81015f901c669d10738963c43e07",
    url = "https://github.com/googleapis/googleapis/archive/94ba270676352628895b3cbb8da12a8fd3dcb88a.tar.gz",
)
```

and now got it to work with the latest commit. Next step is getting the swagger generation set up.

Globegitter:
@ashi009 I know this is only partially related to this pr, but how do you get to generate the swagger documentations? Writing my own `go_proto_compiler` did not work. I could probably depend on `rules_protobuf` for this but would like to try and avoid that if possible.

ashi009:
You won’t solve it with go_proto_compiler rule. As the swagger gen
generates json file instead of go source code.

The short answer would be writing your own skylark extension, creating a
new rule that depends on proto_library. It should be fairly simple to
achieve that, though the docs is not very comprehensive.

Note that you won’t be able to use proto aspect atm, as the comments in the
proto files are stripped out before passing into the swagger gen as proto
descriptor set. (I have filed a bug with bazel team for tracking that.)
Markus Padourek <notifications@github.com>于2018年1月12日 周五下午4:47写道：

chaitanya9186:
When are we planning to merge this?

jayconrod:
Closed per discussion in #1495.

PR #1164, opened 2017-12-20
Run buildifier against bazel files.

ashi009:
The command I ran:

```
$ find . \( -name '*.bzl' -or -name '*.BUILD*' -or -name 'WORKSPACE' \) -exec buildifier {} +
```


bazel-io:

bazel-io:

jayconrod:
@ashi009 Thanks for sending this in. Sorry to be slow to respond.

In general, buildifier isn't sufficient for formatting .bzl file because it doesn't cover any of the code inside function definitions. So we don't want to make it a regular part of our workflow because of that. Some folks in Bazel are interesting in making a formatting for Skylark, so we'll use that as soon as it's ready.

That said, nearly all of the changes in this PR are improvements. When I get some time, I'll go through this and merge most of it, hopefully in the next few days.

bazel-io:

jayconrod:
I merged with master and re-ran buildifier on everything. Almost everything looked good, but there were a few things I rolled back:

* In rule definitions, the `implementation` argument should be first (buildifier wants to put it last). We could prevent buildifier from modifying it by making it a positional argument instead of a keyword argument (several rules are like this already), but personally, I like the keyword argument form better.
* There should be no newline between definitions and doc strings.
* Skylib is vendored, so I've rolled back modifications to those files.
* Entries in SDK_REPOSITORIES should be lined up.

@ianthehat PTAL to double-check.

jayconrod:
Jenkins test this please

ashi009:
LGTM.  Agree to @ianthehat.  We should update `buildifier` to support skylark rules, which I believe should not be difficult, but just a matter of time for writing formatting code and unit tests.   One thing I'd propose is to move `buildifer` to it's own repo like gazelle.  Which should help it to draw more attention from the community.

PR #1173, opened 2017-12-21
Delete old Gazelle

jayconrod:
New gazelle is at github.com/bazelbuild/bazel-gazelle


fejta:
Can we revert this?  It is breaking kubernetes builds

PR #1180, opened 2017-12-22
cgo: change long filenames in cgo archives

jayconrod:
The Go .a format does not allow long filenames. This is not a problem
for "go build", since all the names of files in the archive are
hard-coded. In Bazel, we extract .o files from .a archives produced by
cc_library, then append them to the .a archives produced by Go. If .o
files have long names with similar prefixes, multiple entries can end
up with the same name after truncaction. Linkers don't seem to mind
this, but it's weird and is probably going to bite us some day.

This change renames files as they are extracted from the .a archive
produced by cc_library so they have "safe" names (15 or fewer
characters; no conflicts with other files in the archive).

### DIFF
--- a/go/tools/builders/pack.go
+++ b/go/tools/builders/pack.go
@@ -108,6 +108,7 @@ func extractFiles(archive string) (files []string, err error) {
 	}
 
 	var nameData []byte
+	names := make(map[string]bool)
 	for {
 		name, size, err := readMetadata(r, &nameData)
 		if err == io.EOF {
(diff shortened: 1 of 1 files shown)


go/tools/builders/pack.go
```diff
@@ -258,6 +261,34 @@ func isObjectFile(name string) bool {
...
+		return name
+	}
+	var stem, ext string
+	if i := strings.LastIndexByte(name, '.'); i < 0 || len(name)-i >= 10 {
+		stem = name
+	} else {
+		stem = name[:i]
+		ext = name[i:]
+	}
+	for n := 0; n < 10000; n++ {
```
ianthehat: why not just use len(names)?

jayconrod: Done, seems like a better lower bound.

PR #1183, opened 2017-12-22
Clean up filtering logic

### LINKED ISSUE #1081: Clean up filtering
After the recent quick fixes to cgo filtering, we are in a state where some functions are badly organised, named and hard to read.
Also strict deps is replicating a lot of the work that filtering does, we should extract the common parts there as well.
see #1075

ianthehat:
Should also make compilation faster as we no longer reparse for strict deps

Fixes #1081

### DIFF
--- a/go/tools/builders/filter.go
+++ b/go/tools/builders/filter.go
@@ -19,64 +19,77 @@ import (
 	"go/build"
 	"go/parser"
 	"go/token"
-	"io/ioutil"
 	"log"
 	"path/filepath"
 	"strconv"
 	"strings"
 )
 
+type goMetadata struct {
+	Filename string
+	Matched  bool
+	IsCgo    bool
+	Package  string
+	Imports  []string
+}
+
+// readFiles collects metadata for a list of files.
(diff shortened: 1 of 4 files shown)


go/tools/builders/filter.go
```diff
@@ -19,64 +19,77 @@ import (
...
-		} else if match {
+		} else if m.Matched {
 			outputs = append(outputs, input)
 		}
 	}
 	return outputs, nil
 }
 
-// matchFile applies build constraints to an input file and returns whether
+// Match applies build constraints to an input file and returns whether
```
jayconrod: s/Match/readGoMetadata/

go/tools/builders/filter.go
```diff
@@ -19,64 +19,77 @@ import (
...
 	"go/token"
-	"io/ioutil"
 	"log"
 	"path/filepath"
 	"strconv"
 	"strings"
 )
 
+type goMetadata struct {
+	Filename string
```
jayconrod: Nit: make these private.

Doesn't make any difference since this is `main`, but it's better to default to private.

PR #1190, opened 2017-12-28
Add test to ensure go_source exposes GoLibrary provider.

abergmeier:


bazel-io:

jayconrod:
The providers have changed several times recently. Not sure if we're ready to commit to a stable interface there yet. @ianthehat WDYT?

PR #1192, opened 2017-12-29
Initial attempt on supporting Swagger.

abergmeier:


bazel-io:

abergmeier:
Ok, this should be good to merge.

abergmeier:
All tests now run successfully on my machine.

abergmeier:
@ianthehat Just noticed, that `GOROOT` is empty and the standard library cannot be found. Is there any way of setting `GOROOT` inside the wrapper script to the stdlib?

abergmeier:
So the current test now validates the usage of stdlib (and fails). As far as I read up on this topic I now need to create parallel `src` structure with `binary-only-package` annotation.

abergmeier:
So there are multiple stdlib tests now and they are working. Also added validation support.

abergmeier:
Any idea, why this fails on XCode? Runs fine on my Linux.

abergmeier:
Ping

bazel-io:

abergmeier:
Ok, rebased on 0.9 and added tests and support for non-sandbox.

jayconrod:
Sorry for not responding to this PR earlier. It slipped through the cracks during the holidays, and I never got around to it.

I'm closing this because, architecturally, I don't think swagger support belongs in rules_go. Ideally, this repository should contain support for the Go SDK and standard library and protos. To the extent possible, tools outside of the Go SDK should be supported through rules in other repositories.

This reduces the number of dependencies that may be pulled into projects using rules_go, and it simplifies maintenance for me. Personally, I know very little about Swagger, so I'm not the best person to fix bugs when they come up.

In the future, if there's a well supported set of go-swagger rules somewhere, I'm open to having Gazelle generate rules and dependencies. Also, for discoverability, I'm open to linking to external rules and documentation from the rules_go README.

PR #1195, opened 2018-01-02
Add _go_proto_aspect to the embed attribute of go_proto_library.

rogerhub:
This aspect provides the GoProtoImports provider, which is expected on both deps and embed. The deps attribute already specifies the aspect, but embed does not.

Let me know if you want the commit message formatted a certain way or if I should have filed an issue first.

### DIFF
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -93,7 +93,7 @@ go_proto_library = rule(
         "proto": attr.label(mandatory=True, providers=["proto"]),
         "deps": attr.label_list(providers = [GoLibrary], aspects = [_go_proto_aspect]),
         "importpath": attr.string(),
-        "embed": attr.label_list(providers = [GoLibrary]),
+        "embed": attr.label_list(providers = [GoLibrary], aspects = [_go_proto_aspect]),
         "gc_goopts": attr.string_list(),
         "compiler": attr.label(providers = [GoProtoCompiler]),
         "compilers": attr.label_list(providers = [GoProtoCompiler], default = ["@io_bazel_rules_go//proto:go_proto"]),


ianthehat:
What was the case that caused you to need this?
It was set to cascade across embed, but there should be no case where you need to trigger the initial propagation from embed.

rogerhub:
I'm using [grpc-ecosystem/grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/) (also see #1163), which contains a protoc plugin that generates "foo.pb.gw.go" files alongside the "foo.pb.go" files that are produced by protoc-gen-go. The two files (.pb.gw.go and .pb.go) are meant to go in the same Go package (the .pb.gw.go uses the same package name and references the message types which are only defined in the .pb.go file).

To make this work in bazel, I have rules that look like:

```bzl
# experimental/grpc_gateway_example/service/BUILD

proto_library(
    name = "service_proto",
    srcs = [
        "service.proto",
        "misc.proto",
    ],
    deps = [
        "@com_github_grpc_ecosystem_grpc_gateway//third_party/googleapis/google/api:annotations_proto",  # keep
    ],
)

proto_library(
    name = "service_http_proto",
    srcs = [
        "service.proto",
    ],
    deps = [
        ":service_proto",
        "@com_github_grpc_ecosystem_grpc_gateway//third_party/googleapis/google/api:annotations_proto",  # keep
    ],
)

go_grpc_library(
    name = "service_go_proto",
    importpath = "rogerhub/experimental/grpc_gateway_example/service",
    proto = ":service_proto",
    deps = [
        "@org_golang_google_genproto//googleapis/api/annotations:go_default_library",
    ],
)

go_grpc_gateway_library(
    name = "service_grpc_gateway",
    embed = [":service_go_proto"],
    importpath = "rogerhub/experimental/grpc_gateway_example/service",
    proto = ":service_http_proto",
    deps = [
        "@org_golang_google_genproto//googleapis/api/annotations:go_default_library",
    ],
)
```

The go_grpc_library rule comes from this repo (bazelbuild/rules_go). The go_grpc_gateway_library is a macro that wraps go_grpc_library (similar to go_grpc_library), but with protoc-gen-grpc-gateway as the compiler.

I'm using 2 distinct proto_library rules, because (last time I checked) protoc-gen-grpc-gateway doesn't like it when you feed it protos that don't have any service definitions (it produces no output files for those protos and bazel complains that it can't find the output file). In this case, service.proto defines a service, but misc.proto does not.

I'm not sure if embed is the right way to do this, but it seems to work (given that this PR is applied).

ianthehat:
The right way to do this is to have the two go_proto_library rules totally independent, and then embed them both into a normal go_library rule.

rogerhub:
That's a fine solution, except that the intermediate library (the go_proto_library rule that uses protoc-gen-grpc-gateway) will be unbuildable (the message types themselves aren't defined in the .gw.pb.go files). Is there any downside to accepting this patch? Or is it just technically unnecessary?

ianthehat:
unbuildable intermediates are not a problem, just tag them as 
```
tags = ["manual"],
```
so they never get built directly, we do this a lot.

We also already have similar cases (the validate plugin causes unbuildable libraries)
We solve it by allowing multiple compilers on a single go_proto_library rule, and combining all the resulting sources into a single library.you can see an example of this in [examples/proto/embed/BUILD.bazel](/bazelbuild/rules_go/blob/master/examples/proto/embed/BUILD.bazel)
This does require that you conform to the standard proto compiler pattern though, the embeddiing into a go_library is the escape hatch for things that can't.
I will accept the patch if it really is the best answer, I figured suggesting an alternative was friendlier than asking you to also write a test though :)

rogerhub:
Ah, I wasn't aware there was a way to exclude targets from wildcards. In that case, I'm happy to abandon this. Thanks!

ashi009:
Hey @rogerhub,

I think you should move directly to plugin approach, which is way cleaner than this.  Let me know if you  run into any problem.

PR #1196, opened 2018-01-02
remove proto_register_toolchains()

samarth:
As stated here, https://github.com/bazelbuild/rules_go/blob/master/proto/def.bzl#L116, this does nothing.


bazel-io:

samarth:
I signed it!

Issue #1186, opened 2017-12-27
go_rules fails to properly detect ppc64le


jbedorf:
My problem seems to be similar to #854  but as that one is closed and it is slightly different in the output I'm opening a new issue.

**Environment info**
```
$ uname -a
Linux host 3.10.0-693.11.1.el7.ppc64le #1 SMP Fri Oct 27 05:42:02 EDT 2017 ppc64le ppc64le ppc64le GNU/Linux
```
Bazel information
```
$ bazel version
build label: 0.9.0- (@non-git)
```

When running a simple test problem (same as given in the last comment of #854 ) with modifications to WORKSPACE to take the latest master version of `the rules_go` repository

WORKSPACE
````
 git_repository(
    name = "io_bazel_rules_go",
    remote = "https://github.com/bazelbuild/rules_go",
    commit = "bd1753da4c06bdd10c8bbe0086da18fcea055eea",
)

load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_register_toolchains")
go_rules_dependencies()
go_register_toolchains()
```` 

BUILD.bazel
```
load("@io_bazel_rules_go//go:def.bzl", "go_binary")

go_binary(
    name = "hello",
    srcs = ["hello.go"],
    importpath = "example.com/hello",
)
```

hello.go
```
package main

func main() {
}
```

When running `bazel build //:hello` it will attempt to download the linux_amd64 rules, this can be fixed by changing:
https://github.com/bazelbuild/rules_go/blob/master/go/private/sdk.bzl#L32
into:
` elif res.stdout.strip() == 'ppc64le':`
For some reason the stdout output contains a new line which causes the comparison to fail. 

Fixing that it will download the proper-sdk, but then during the building step it fails because it builds the sdk for `ppc64` instead of `ppc64le`: 

```
$ bazel build //:hello
INFO: Analysed target //:hello (163 packages loaded).
INFO: Found 1 target...
[5 / 13] 2 actions running
    GoStdlib external/go_stdlib_linux_ppc64_cgo/bin/go [for host]; 2s processwrapper-sandbox
    GoStdlib external/go_stdlib_linux_ppc64_cgo/bin/go; 2s processwrapper-sandbox
```

Which causes the build/compilation steps later on to fail:
```
$ external/io_bazel_rules_go/go/tools/builders/BUILD.bazel:22:1: 
[...]

PR #1197, opened 2018-01-02
Map @bazel_tools//platforms:ppc to GOARCH ppc64le instead of ppc64

jayconrod:
This constraint_value is selected for ppc64le machines, but we have
no ppc64 host toolchain, so it breaks the build.

Related #1186

### DIFF
--- a/go/platform/list.bzl
+++ b/go/platform/list.bzl
@@ -36,8 +36,8 @@ GOARCH = {
     "mips64": None,
     "mips64le": None,
     "mipsle": None,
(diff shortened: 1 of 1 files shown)


jayconrod:
@jbedorf Could you try out this patch and see if it fixes your build (without needing to modify Bazel)?

jbedorf:
@jayconrod This patch works, it now automatically builds the `ppc64le` toolchain without having to modify bazel. 👍

jayconrod:
Great! Thanks for verifying.

PR #1203, opened 2018-01-03
Make importpath optional on go_test and go_binary

### LINKED ISSUE #1125: Make importpath optional for go_binary and go_test
All Go rules have an `importpath` attribute, which is specified explicitly with the `importpath` attribute or is derived from `go_prefix`. `importpath` is used to determine output paths and is used by the experimental `go_path` rule.

We'd like to deprecate `go_prefix` (#721). In repositories without `go_prefix`, `importpath` is effectively mandatory, which is surprising for `go_binary` and `go_test`, since they can't be imported. When `importpath` isn't specified explicitly for these, we should create a dummy path. We should also report an error when rules without a real `importpath` are used in `go_path`.

ianthehat:
This allows rules to infer their import path from their embeds, and removes the
need for it on things that build binaries.

Fixes #1125


go/private/rules/wrappers.bzl
```diff
@@ -20,7 +20,7 @@ load("@io_bazel_rules_go//go/private:rules/cgo.bzl", "setup_cgo_library")
 def go_library_macro(name, srcs=None, embed=[], cgo=False, cdeps=[], copts=[], clinkopts=[], library=None, **kwargs):
   """See go/core.rst#go_library for full documentation."""
   if library:
-    #TODO: print("DEPRECATED: {}//{}:{} : the library attribute is deprecated. Please migrate to embed.".format(native.repository_name(), native.package_name(), name))
+    #TODO: print("\nDEPRECATED: {}//{}:{} : the library attribute is deprecated. Please migrate to embed.".format(native.repository_name(), native.package_name(), name))
```
jayconrod: Since this deprecation will be carried over a couple releases, it would probably be good to open an issue and reference it from these TODOs.

go/providers.rst
```diff
@@ -80,12 +80,20 @@ This is a non build mode specific provider.
...
-| GOPATH.                                                                                          |
+| Information about the source of the importpath.                                                  |
+| It's values can be                                                                               |
+| :value:`explicit`                                                                                |
+|     The importpath was explicitly supplied by the user and the library is importable.            |
+| :value:`inferred`                                                                                |
+|     The importpath was inferred from the directory structure and rule name. The library may be   |
+|     importable.                                                                                  |
+| :value:`export`                                                                                  |
+|     The importpath is used for generated file names, but the library should not be imported by   |
```
jayconrod: Mention main and test packages as examples.

go/private/rules/wrappers.bzl
ianthehat: These are still commented out, I tried to turn them on and then realized they are broken, so I turned them back off but left the \n I added to make them readable, I will be removing the TODO in another cl (hopefully today)

PR #1213, opened 2018-01-04
A better fix for prorto import paths

### LINKED ISSUE #1194: importpath seemed to be ignored
I have this use case that I want to to use `genrule` as a source of `proto_library`. If I use `genrule` import path will be wrong. I have something like this in `path/proto/lib`, Note that `importpath` has additional `repo`.
```
genrule(
  name = "lib_copy",
  srcs = ["lib.proto.orig"],
  outs = ["lib.proto"],
  cmd = "cp $(<) $(@)",
)

proto_library(
    name = "lib_proto",
    srcs = ["lib.proto"],
)

go_proto_library(
    name = "lib_go_proto",
    importpath = "repo/path/proto/lib",
    proto = ":lib_proto",
)
```
and another proto depends on this in `path/proto/dep`:
```
proto_library(
    name = "dep_proto",
    srcs = ["dep.proto"],
    deps = [
        "//path/proto/lib:lib_proto",
    ],
)

go_proto_library(
    name = "dep_go_proto",
    importpath = "repo/path/proto/dep",
    proto = ":dep_proto",
    deps = [
        "//path/proto/lib:lib_go_proto",
    ],
)
```
`dep_go_proto` failed with 
```
import of path/proto/lib, which is not a direct dependency
```
In generated `dep.pb.go`, `lib` is imported as  `path/proto/lib` instead of `repo/path/proto/lib`. Removing genrule and using a proto file fixes the problem.

ianthehat:
Fixes #1194

### DIFF
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -24,22 +24,15 @@ load("@io_bazel_rules_go//go/private:rules/prefix.bzl",
 )
 load("@io_bazel_rules_go//proto:compiler.bzl",
     "GoProtoCompiler",
+    "proto_path",
 )
 
 GoProtoImports = provider()
 
-def _trim_external_path(src):
-  path = src.path
-  if src.owner and src.owner.workspace_root:
-    prefix = src.owner.workspace_root+"/"
-    if path.startswith(prefix):
-      path = path[len(prefix):]
-  return path
-
 def get_imports(attr):
   imports = []
   if hasattr(attr, "proto"):
-    imports.append(["{}={}".format(_trim_external_path(src), attr.importpath) for src in attr.proto.proto.direct_sources])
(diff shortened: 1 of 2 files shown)


proto/def.bzl
```diff
@@ -24,22 +24,15 @@ load("@io_bazel_rules_go//go/private:rules/prefix.bzl",
 )
 load("@io_bazel_rules_go//proto:compiler.bzl",
     "GoProtoCompiler",
+    "proto_path",
```
jayconrod: Probably better import this as a private symbol.

ianthehat: It's in a private file an not exposed through def.bzl, I would rather not have multiple names for the same function, just makes it harder to navigate the code.

PR #1223, opened 2018-01-09
Fix go proto rules for Windows

pmuetschard:
- Pass the compiler path along
 - Strip .exe from plugin names
 - Use the system dependent path separator

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -187,7 +187,7 @@ func run(args []string) error {
 
 	cases := Cases{
 		Package: *pkg,
-		RunDir:  filepath.FromSlash(*runDir),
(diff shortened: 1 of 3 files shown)


go/tools/builders/generate_test_main.go
```diff
@@ -187,7 +187,7 @@ func run(args []string) error {
 
 	cases := Cases{
 		Package: *pkg,
-		RunDir:  filepath.FromSlash(*runDir),
+		RunDir:  strings.Replace(filepath.FromSlash(*runDir), "\\", "\\\\", -1),
```
ianthehat: Use "raw" strings for these 
```
`\`, `\\`,
```

pmuetschard: Done.

PR #1226, opened 2018-01-10
Fix version string parsing for Bazel 0.10.0rc1

jayconrod:
Fixes bazelbuild/bazel#4426


jayconrod:
I wish Skylark had regexp =P

PR #1227, opened 2018-01-10
don't use stdlib at all for bootstrap rules

ianthehat:
They just use the underlying sdk directly.
Also, we only ever use the original copy of the main go executable now.


go/private/context.bzl
```diff
@@ -204,7 +220,11 @@ def go_context(ctx, attr=None):
       # Fields
       toolchain = toolchain,
       mode = mode,
+      root = root,
```
jayconrod: Update toolchains.rst with new fields.

PR #1234, opened 2018-01-12
Moving stdlib compilation to a builder

ianthehat:
Now we have cleaned up all the context data cgo handling, we can build the
stdlib using the same builder environment, rather than a nasty && bash command
line.
This also means all the abs path handling will be fixed.
Also removed the _pure_race version that was not legal anyway

### DIFF
--- a/go/tools/builders/stdlib.go
+++ b/go/tools/builders/stdlib.go
@@ -1,0 +1,121 @@
+// Copyright 2018 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// stdlib builds the standard library in the appropriate mode into a new goroot.
+package main
+
+import (
+	"flag"
+	"fmt"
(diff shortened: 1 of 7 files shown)


go/tools/builders/stdlib.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		_ = os.Remove(to)
+		if err := os.Link(path, to); err != nil {
+			return fmt.Errorf("Failed link: %v", err)
+		}
+		return nil
+	})
+}
+
+func install_stdlib(goenv *GoEnv, target string, args []string) error {
```
jayconrod: Nit: camelCase

go/tools/builders/stdlib.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return err
+		}
+		to := filepath.Join(dst, relative)
+		dir := filepath.Dir(to)
+		if err := os.MkdirAll(dir, 0755); err != nil {
+			return fmt.Errorf("Failed to make %s: %v", dir, err)
+		}
+		_ = os.Remove(to)
+		if err := os.Link(path, to); err != nil {
```
jayconrod: I'm a little surprised by the hard link. Is there any situation where `src` and `dst` can be on different file systems? What if the host SDK is used? Will it work on Windows?

go/tools/builders/stdlib.go
```diff
@@ -0,0 +1,121 @@
...
+	_ = os.Remove(dst)
+	if err := os.Symlink(src, dst); err != nil {
+		return fmt.Errorf("Failed link: %v", err)
+	}
+	return nil
+}
+
+func copy(src, dst, common string) error {
+	return filepath.Walk(filepath.Join(src, common), func(path string, f os.FileInfo, err error) error {
+		if f.IsDir() {
```
jayconrod: Do empty directories not need to be copied?

PR #1235, opened 2018-01-12
Trim sandbox prefix from cgo generated line directives

### LINKED ISSUE #1069: Cgo paths in binary
I'm building a library which requires cgo.

I see things like

```
// Created by cgo - DO NOT EDIT

//line /dev/shm/bazel-sandbox.0484f5bd76075cd81cb4bc586a1d0a10/6359360015955161932/execroot/com_peloton_tech/third_party/go/src/vendor/github.com/zeromq/goczmq/auth.go:1
package goczmq

//line /dev/shm/bazel-sandbox.0484f5bd76075cd81cb4bc586a1d0a10/6359360015955161932/execroot/com_peloton_tech/third_party/go/src/vendor/github.com/zeromq/goczmq/auth.go:1
import _cgo_unsafe "unsafe"
```

in the generated files in bazel-genfiles.

That is non reproducible, and results in non-reproducible binaries downstream.

ianthehat:
Fixes #1069

### DIFF
--- a/go/tools/builders/cgo.go
+++ b/go/tools/builders/cgo.go
@@ -17,8 +17,6 @@
 package main
 
 import (
-	"bufio"
-	"bytes"
 	"errors"
 	"flag"
 	"fmt"
@@ -30,8 +28,6 @@ import (
 	"strings"
 	"unicode"
 )
-
-const linePrefix = "//line "
 
 func run(args []string) error {
 	sources := multiFlag{}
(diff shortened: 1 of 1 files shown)


go/tools/builders/cgo.go
```diff
@@ -29,6 +31,8 @@ import (
 	"unicode"
 )
 
+const linePrefix = "//line "
```
jayconrod: This is largely personal preference (feel free to ignore), but I would move this next to the code that uses it. It would help to have the actual string be visible in that code; if I were editing this, I'd definitely be scrolling back to check whether there's a space at the end. I don't expect the string will be relevant for anything else at global scope.

go/tools/builders/cgo.go
```diff
@@ -189,6 +195,34 @@ func run(args []string) error {
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("error running cgo: %v", err)
 	}
+	// Now we fix up the generated files
+	trim := linePrefix + abs(".")
+	for _, src := range cgoOuts {
+		body, err := ioutil.ReadFile(src)
```
jayconrod: Consider extracting the body of this loop into a function.

go/tools/builders/cgo.go
```diff
@@ -189,6 +195,34 @@ func run(args []string) error {
...
+		if err != nil {
+			return err
+		}
+		out, err := os.Create(src)
```
jayconrod: Instead of overwriting the original file here, write to a new file, then defer `os.Rename` / `os.Remove` at the end for commit / rollback. Better to keep the original file intact in case of error instead of a partial file.

go/tools/builders/cgo.go
```diff
@@ -189,6 +195,34 @@ func run(args []string) error {
...
+		if err != nil {
+			return err
+		}
+		buf := bufio.NewWriter(out)
+		s := bufio.NewScanner(bytes.NewReader(body))
+		for s.Scan() {
```
jayconrod: Scanner errors on long lines (over 64K IIRC). I guarantee we'll see that in the wild. You can increase the maximum length to an arbitrary limit, but I wonder if the transformation could be done with `bytes.Replace` instead?

Issue #1239, opened 2018-01-13
go_binary output paths contain a spurious os_arch component


creachadair:
Building a `go_binary` results in an output path that differs from the package path by the addition of $GOHOSTENV, $GOOS, and $GOARCH, e.g.,

    bazel build //path/to/my:binary

leaves the binary at

    bazel-bin/path/to/my/linux_amd64_stripped/binary

This makes it difficult to use Go binaries as part of `sh_test` rules: Other Bazel `*_binary` tools can be used by relative path without having to pre-bake these values into the script, e.g., if my `sh_binary` declares `//path/to/my:binary` as a `data` dependency, it should be executable by the path `path/to/my/binary` relative to the exec root when the test runs.

It works as described for everything but Go. Given that the Bazel output already reflects host and architecture constraints, the additional path decoration gets in the way of common practice. 

One can work around this by injecting rewrite rules, but that shouldn't be necessary.

jayconrod:
cc @ianthehat 

We added this to avoid cache poisoning when building the same targets in different modes. For example, if you ran a bunch of tests in race mode, then build a binary that depends on the same libraries as the tests, all the libraries have to be rebuilt. We work around this by including mode information in the output paths, which allows Bazel to cache files in multiple modes.

> Other Bazel *_binary tools can be used by relative path without having to pre-bake these values into the script

Is there another set of Skylark rules that work this way? The C / C++ / Objective C rules may have more flexibility than we do since they're part of Bazel itself.

creachadair:
> We added this to avoid cache poisoning when building the same targets in different modes.

I'm not sure what you mean by "modes" in this context, but it seems to me like the mode should land in the build configuration, thus changing (say) `k8-fastbuild` to `k8-fastbuild-race` or similar just like `-c opt` and `-c fastbuild` do.

ianthehat:
The way bazel works, that part of the path is constant for any given build, and not configurable in any way. It's affected by the existing options only (so we could not add race to it for instance). We also support building the same binary in multiple modes within a single build right now.
When bazel gets controllable toolchain transitions, and the active toolchain causes the root of the path to change, we will probably take it out of our generated path, but even then it does not help you.

If you hard code the path in a sh_test it's going to be wrong on windows (or any other platform where the binary name does not match the rule name). The same is true for all the languages, it's never safe to make this assumption. For the most common languages (c and, java), on linux and mac, the assumption holds, but it's not a good idea to rely on it.
[...]

PR #1257, opened 2018-01-19
Don't include compilation mode in binary output filenames.

jmillikin-stripe:
Fixes https://github.com/bazelbuild/rules_go/issues/1239

This should maintain nearly all of the caching benefits of the arch-specific filenames, because libraries are still built with the included mode. Binaries will need to be relinked if `-c` changes, but the filenames are consistent with other languages commonly used with Bazel.

### DIFF
--- a/go/private/actions/binary.bzl
+++ b/go/private/actions/binary.bzl
@@ -18,13 +18,14 @@ def emit_binary(go,
     gc_linkopts = [],
     linkstamp=None,
     version_file=None,
-    info_file=None):
+    info_file=None,
+    output_path=None):
   """See go/toolchains.rst#binary for full documentation."""
 
   if name == "": fail("name is a required parameter")
 
   archive = go.archive(go, source)
-  executable = go.declare_file(go, name=name, ext=go.exe_extension)
+  executable = go.declare_file(go, name=name, ext=go.exe_extension, output_path=output_path)
   go.link(go,
       archive=archive,
       executable=executable,
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -36,6 +36,7 @@ load(
     "structs",
     "goos_to_extension",
     "as_iterable",
+    "DEFAULT_OUTPUT_PATH",
 )
 
 GoContext = provider()
@@ -59,20 +60,22 @@ _LINKER_OPTIONS_BLACKLIST = {
 def _filter_options(options, blacklist):
   return [option for option in options if option not in blacklist]
 
-def _child_name(go, path, ext, name):
-  childname = mode_string(go.mode) + "/"
-  childname += name if name else go._ctx.label.name
-  if path:
-    childname += "~/" + path
-  if ext:
-    childname += ext
-  return childname
+def _child_name(go, path, ext, name, output_path):
+  if output_path == None:
+    output_path = DEFAULT_OUTPUT_PATH
 
-def _declare_file(go, path="", ext="", name = ""):
-  return go.actions.declare_file(_child_name(go, path, ext, name))
+  return output_path.format(
+    mode = mode_string(go.mode),
+    name = name if name else go._ctx.label.name,
+    path = "~/" + path if path else "",
+    ext = ext,
+  )
 
-def _declare_directory(go, path="", ext="", name = ""):
-  return go.actions.declare_directory(_child_name(go, path, ext, name))
+def _declare_file(go, path="", ext="", name="", output_path=None):
+  return go.actions.declare_file(_child_name(go, path, ext, name, output_path))
+
+def _declare_directory(go, path="", ext="", name="", output_path=None):
+  return go.actions.declare_directory(_child_name(go, path, ext, name, output_path))
 
 def _new_args(go):
   args = go.actions.args()
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -19,6 +19,7 @@ load(
 load(
     "@io_bazel_rules_go//go/private:common.bzl",
     "go_filetype",
+    "DEFAULT_OUTPUT_PATH",
 )
 load(
     "@io_bazel_rules_go//go/private:rules/prefix.bzl",
@@ -65,6 +66,7 @@ def _go_binary_impl(ctx):
       linkstamp=ctx.attr.linkstamp,
       version_file=ctx.version_file,
       info_file=ctx.info_file,
+      output_path=ctx.attr.output_path,
   )
   return [
       library, source, archive,
@@ -137,6 +139,7 @@ go_binary = go_rule(
         "linkstamp": attr.string(),
         "x_defs": attr.string_dict(),
         "linkmode": attr.string(values=LINKMODES, default=LINKMODE_NORMAL),
+        "output_path": attr.string(default = DEFAULT_OUTPUT_PATH),
     },
     executable = True,
 )
@@ -159,6 +162,7 @@ go_tool_binary = go_rule(
         "linkstamp": attr.string(),
         "x_defs": attr.string_dict(),
         "linkmode": attr.string(values=LINKMODES, default=LINKMODE_NORMAL),
+        "output_path": attr.string(default = DEFAULT_OUTPUT_PATH),
         "_hostonly": attr.bool(default=True),
     },
     executable = True,
--- a/tests/core/go_binary/BUILD.bazel
+++ b/tests/core/go_binary/BUILD.bazel
@@ -1,0 +1,48 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_test", "go_source")
+
+test_suite(
+    name = "go_binary",
+)
+
+go_binary(
+    name = "hello_plain",
+    srcs = ["hello_world.go"],
+    goarch = "amd64",
+    goos = "windows",
+    pure = "on",
+)
+
+go_binary(
+    name = "hello_no_mode",
+    srcs = ["hello_world.go"],
+    goarch = "amd64",
+    goos = "windows",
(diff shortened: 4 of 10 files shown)


bazel-io:

ianthehat:
Sorry, your example was not clear and I assumed that was a skylark fragment. When I was talking about the .exe extension to the name mattering, I was discussing the packaging rule when it is building the package, not the act of running the app on the target platform.

The packaging rule can either either do the renaming as a pre-step, as it builds the package, or it can write the name it was given into the text file that is part of the package, either way it's going to do the right thing if it knows the name, and $location is how you tell it the name. You can achieve any of these things either in skylark or inside your packaging app or a wrapper script or a genrule, there are lots of ways to fix it, but if the name is shared there is no possible work around for those people it affects.

Almost all packaging rules I have encountered for go so far have to do the renaming anyway, because they almost always want "./bin/myapp" where the input they get is "./project/subdir/cmd/myapp", the output file never matches the desired package directory structure.

jmillikin-stripe:
This is continuing to be a pain point for us. The number of hacky bash loops, genrules, and documentation with "if on mac run this command, on linux this command ..." increases daily.

Would you accept a PR that added an attribute to `go_binary` to turn off the build mode in filenames? I don't care if the platform-specific output paths are the default for rules_go, but for some of our binaries it's getting very frustrating to have `go_binary` produce filenames that are different (1) between Linux and MacOS and (2) between `-c dbg` and `-c opt`.

Something like `go_binary(..., mode_specific_filename=False)`.

ianthehat:
Jay and I talked about this at the time, and we agreed to add an extra attribute to go_binary that allowed you to be explicit about the name. If you specify it, the rule will just trust you and generate that file with *no* mutations to the name (it won't add extensions, or prefix with paths of any kind).
It is then up to the user to avoid collisions if the need to, or use select statements to get the right name.
I thought we had filed a new feature issue for it, but I can't find it anywhere, so I must have forgotten to do that.
It think that's the right level of escape hatch to add, if you want to try adding it feel free!

jmillikin-stripe:
OK, I've changed this PR so it adds an `output_path` attribute to `go_binary`. It can be used by the user for full control over the output filename. Portions of the default path can be used via template substitution.

jayconrod:
@jmillikin-stripe I'm most of the way through writing a PR for this. I'd rather just submit that. It will be simpler, just with an `out` attribute that you can set to a filename.

jmillikin-stripe:
I really really don't want to hardcode a filename in all our `go_binary` rules. Using a template lets me override it ~globally with a simple macro.

jayconrod:
Do you really need this level of flexibility for all `go_binary` rules?

I'm uncomfortable with how this exposes mode and path in our API. Those will very likely change or be removed in the future.

I think if the output filename is set, it should just be a filename in the package directory. We'll probably want to add a platform-specific extension (.exe, .so, etc), but I don't see the need for more configurability than that.

go/private/context.bzl
```diff
@@ -59,20 +60,22 @@ _LINKER_OPTIONS_BLACKLIST = {
...
 
-def _child_name(go, path, ext, name):
-  childname = mode_string(go.mode) + "/"
-  childname += name if name else go._ctx.label.name
-  if path:
-    childname += "~/" + path
-  if ext:
-    childname += ext
-  return childname
+def _child_name(go, path, ext, name, output_path):
```
jayconrod: See comments in `emit_binary`.

These functions should be left alone to generate mode-specific paths without user customization.

Non-mode-specific paths can still be declared with `ctx.actions.declare_file`.

go/private/rules/binary.bzl
```diff
@@ -137,6 +139,7 @@ go_binary = go_rule(
         "linkstamp": attr.string(),
         "x_defs": attr.string_dict(),
         "linkmode": attr.string(values=LINKMODES, default=LINKMODE_NORMAL),
+        "output_path": attr.string(default = DEFAULT_OUTPUT_PATH),
```
jayconrod: Nit: I'd prefer this attribute to be named `out`.

go/private/actions/binary.bzl
```diff
@@ -18,13 +18,14 @@ def emit_binary(go,
...
-    info_file=None):
+    info_file=None,
+    output_path=None):
   """See go/toolchains.rst#binary for full documentation."""
 
   if name == "": fail("name is a required parameter")
 
   archive = go.archive(go, source)
-  executable = go.declare_file(go, name=name, ext=go.exe_extension)
+  executable = go.declare_file(go, name=name, ext=go.exe_extension, output_path=output_path)
```
jayconrod: If `output_path` is not empty, use `go._ctx.actions.declare_file` instead of `declare_file`.

`declare_file` should be considered mode-specific.

tests/core/go_binary/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    goos = "windows",
+    pure = "on",
+)
+
+go_binary(
+    name = "hello_no_mode",
+    srcs = ["hello_world.go"],
+    goarch = "amd64",
+    goos = "windows",
+    output_path = "{name}{path}{ext}",
```
jayconrod: This would be `out = "hello_no_mode"`. The test can verify that the output file has an .exe extension.

jmillikin-stripe:
> I'd much rather have out = "foo", which would write foo in the package directory (without mode or name-specific prefixes). On Windows, it would write foo.exe. For c-shared mode, we may write libfoo.so.

That would require each target to declare its name twice, once as `name="foo"` and again as `out="foo"`. This isn't necessary for C++ binaries, and it's not obvious why Go should require it to get consistent behavior.

How about `out = attr.string(values=["mode", "no_mode"], default="mode")`? The user can set `basename` if they want to generate an output path unrelated to the target name.

jayconrod:
> That would require each target to declare its name twice, once as name="foo" and again as out="foo". 

You can hide this behind a macro if you want to apply it universally, but I hope it's not necessary for all binaries.

> This isn't necessary for C++ binaries, and it's not obvious why Go should require it to get consistent behavior.

C++ and Go support are implemented very differently in Bazel; it's not a fair comparison. C++ is natively supported and has much more powerful configuration capabilities via CROSSTOOL files. When you build C++, all the build products do go in a configuration-specific directory, but `bazel-bin` points inside that directory (e.g., bazel-out/k8-fastbuild/bin). Skylark rules don't have that power. We do have some configuration support with toolchains, but we're handling static, msan, pure, strip, etc. with aspects. Bazel won't create configuration-specific directories for us (other than what it does for C++). In order to keep files cached, we need to create mode-specific directories.

> How about out = attr.string(values=["mode", "no_mode"], default="mode")? The user can set basename if they want to generate an output path unrelated to the target name.

I'd rather have one attribute that provides flexibility without saying anything about modes. In my ideal world, configurations would be more explicitly exposed to Skylark rules, and modes are no longer necessary. I think of them as a temporary implementation detail, though "temporary" may be a long time.

jmillikin-stripe:
> You can hide this behind a macro if you want to apply it universally, but I hope it's not necessary for all binaries.

We'd certainly want it for all binaries in our repos, and probably would use Bazel 0.11's patch-external-http-archive feature to apply it to externals that we want to build as-is.

> C++ and Go support are implemented very differently in Bazel; it's not a fair comparison. C++ is natively supported and has much more powerful configuration capabilities via CROSSTOOL files. When you build C++, all the build products do go in a configuration-specific directory, but bazel-bin points inside that directory (e.g., bazel-out/k8-fastbuild/bin).

I'm not sure how this is true. Both C++ and rules_go place build products in a target-specific directory, where `bazel-bin` points to that directory. rules_go _additionally_ adds the target name and compilation mode to a subdirectory, which is something C++ doesn't do.

You can see that Go output paths have the target platform twice, once from bazel (`darwin-fastbuild`) and once from rules\_go (`darwin_amd64_stripped`):
```
[rules_go]$ bazel build //go/tools/fetch_repo
Target //go/tools/fetch_repo:fetch_repo up-to-date:
  bazel-bin/go/tools/fetch_repo/darwin_amd64_stripped/fetch_repo
[rules_go]$ shasum bazel-bin/go/tools/fetch_repo/darwin_amd64_stripped/fetch_repo bazel-out/darwin-fastbuild/bin/go/tools/fetch_repo/darwin_amd64_stripped/fetch_repo
6cdbc61fa32cb0fb7dc3b69b1b210b48393a9426  bazel-bin/go/tools/fetch_repo/darwin_amd64_stripped/fetch_repo
6cdbc61fa32cb0fb7dc3b69b1b210b48393a9426  bazel-out/darwin-fastbuild/bin/go/tools/fetch_repo/darwin_amd64_stripped/fetch_repo
[rules_go]$ 
```

> We do have some configuration support with toolchains, but we're handling static, msan, pure, strip, etc. with aspects. Bazel won't create configuration-specific directories for us (other than what it does for C++). In order to keep files cached, we need to create mode-specific directories.

I specifically want Bazel's default behavior, though. I understand that you want to let users have better caching behavior when building for multiple platforms, but the mechanism you've chosen to do this (changing the go_binary output path) is causing us pain and I would please like some way to turn that off.

jayconrod:
I'm planning to submit #1393 instead of this one. That PR adds an `out` parameter to `go_binary` that lets you set the output file to a custom name.

If you want to set `out` automatically for all binaries, I'd suggest wrapping `go_binary` in a macro. Something like this:

```
load("@io_bazel_rules_go//go:def.bzl", _go_binary = "go_binary")

def go_binary(**kwargs):
  kwargs.setdefault("out", kwargs["name"])
  _go_binary(**kwargs)
```

PR #1265, opened 2018-01-22
specify go version in beta SDK instructions

rickypai:
The README suggests using `go_download_sdk` to download a beta version, but does not actually specify the version in `go_register_toolchains`. This change should help users use beta versions easier.


bazel-io:

PR #1277, opened 2018-01-25
Fix comma instead of open bracket typo

BenElgar:


bazel-io:

Issue #1264, opened 2018-01-22
Enable transitive target CPU constraint


steeve:
Hi,

As part of implementing `c-archive` and `c-shared` support, I'm trying to properly propagate CPU selection to the Go toolchain when doing things such as building an iOS app or library.
See https://github.com/znly/rules_go/commit/344d1337b0350ea5e3b9e712437d38f33054d8a7 and https://github.com/znly/rules_go/commit/fce5ac2d1f8a18ae77712cd392c76b54cb03e2d8.

I'm am quite confused as to how to do that properly based on constraints.

I am able to properly `select` based on a `config_setting` such as:
```python
config_setting(
    name = 'ios_arm64',
    values = {"cpu": "ios_arm64"},
)
```

However, when trying to turn that into a constraint, in doing either:
```python
constraint_setting(name = ":cpu")
constraint_value(
    name = 'ios_arm64',
    constraint_setting = ":cpu",
)
```

Or even:
```python
constraint_value(
    name = 'ios_arm64',
    constraint_setting = "@bazel_tools//platforms:cpu",
)
```

And adding that constraints to the arch in `list.bzl`, it fails to properly select the correct target based on the `ios_application` I'm building, where `config_setting` works as expected.
I'm using the following command line
```
$ bazel build --ios_multi_cpus=arm64 //:sampleapp
```

This isn important because in the case of iOS one could have multiple versions of the `c-archive` built on different archs (armv7 and arm64).

Thanks!

ianthehat:
ios_arm64 is not a cpu...
I am not entirely sure what you want to achieve, but we already have all the constraints for every goos and goarch declared, along with platform declarations for all valid combinations, and then config_setting declarations to match. You should not need to declare any new ones.

I have no idea if what you are trying to do will work though, I think you are after toolchain transitions, which we would love to have but are not exposed to skylark right now.

steeve:
I'm trying to embed a `go_binary` built as `c-archive` as a dependency to a `cc_library`, which is then imported into a `ios_application` and have `goos` and `goarch` transitively selected based on `ios_application` architecture (or multi architecture, for that matter).

It does work fine when I specifically select the `goos/goarch` to match the `ios_application`'s arch via `platform` (in this case `darwin_arm64`).

Note that `select`ing (source files, for instance) on `config_setting` `ios_arm64` properly works when using `ios_application`, even though it's not specified on the command line.

[...]

Issue #539, opened 2017-06-14
support -buildmode=plugin linkopt


mikedanese:
I tried

```
 11 go_binary(
 12     name = "gcp",
 13     gc_linkopts = ["-buildmode=plugin"],
 14     library = ":go_default_library",
 15     visibility = ["//visibility:public"],
 16 )
```

And got:
```
$ bazel build //plugin/...
INFO: Found 4 targets...
ERROR: /usr/local/google/home/mikedanese/go/src/containers.google.com/idmgr/plugin/gcp/BUILD:11:1: error executing shell command: 'bazel-out/local-fastbuild/bin/plugin/gcp/gcp.a.GoLinkFile.params' failed: Process exited with status 1 [sandboxed].
.init: reloc 15 to non-elf symbol .initdone· (outer=.initdone·) 32
.init: reloc 15 to non-elf symbol .initdone· (outer=.initdone·) 32
Use --strategy=GoLink=standalone to disable sandboxing for the failing actions.
INFO: Elapsed time: 0.378s, Critical Path: 0.12s
```

jayconrod:
I'd like to do this eventually, but it won't be simple. When you build a plugin, `-buildmode=plugin` goes to `go build`. The compiler, assembler, and linker don't get this flag directly; they get extra flags from `go build` instead. Dependencies also need to be compiled with those flags. Not sure how this interacts with cgo.

I tried building a small binary with `-x`, and it looks like it actually needs to recompile the standard library. Support for this might end up being similar to support race, asan, or tsan.

ianthehat:
I have a plan to allow support for this kind of thing using [output groups](https://bazel.build/versions/master/docs/skylark/rules.html#output-groups) but it's stacked up behind platforms and cross compilation and I'm not 100% sure it will work without actually trying it.

f0rmiga:
@ianthehat Any progress on that? I'm very interested in this feature. :)

ianthehat:
No direct progress, but we are now in a good state to add more build modes. We build everything in multiple modes, including the standard library, so it's easy to add more, I have not done most of the `-buildmode` options only because I have no idea how they work or how people want to use them.
In the case of `-buildmode=plugin` for instance, should that option be on a go_binary rule? what should the output file name/type be on windows? How should another binary/library/test depend on a plugin? what should be put in runfiles?

f0rmiga:
Awesome!

[...]

Issue #54, opened 2016-07-23
buildmode shared


lingz:
I saw that using -buildmode=c-shared is currently unsupported here:
https://github.com/bazelbuild/rules_go/pull/10

Any plans to add this or ideas of how difficult it would be to add?

yugui:
It was discussed at https://groups.google.com/d/msg/bazel-dev/N_cS_OJocZI/-6vvvLUkAwAJ
But currently we don't have a good generic solution.

It turned out that my original proposal did not generally work fine because it instantiates go runtime for each `c-shared` library.  So if you link hundred number of such `c-shared` libraries into your binary, the binary will have hundred number of garbage collectors and their corresponding threads.
To avoid this issue, we need a way to provide `CcLinkParamProvider` from Skylark to native `CCLibrary`.

I plan to revisit this feature once we get a good support by Bazel.

lingz:
Ok I look forward to seeing how it goes. Are you requesting new features from Bazel before this can be implemented well?

ajbouh:
Any progress on support for -buildmode=c-shared?

ianthehat:
I was planning to add support for all the buildmode options following either the "static" or "race" patterns depending on if they are main or package level choices.
This means c-shared would become an outputgroup choice on go_binary.
All the supporting work was done and proven implementing those two modes, it's really easy to add the others, the only thing stopping me is I have no use case, no examples, and no way to test, and other things seemed more urgent.
If you need this, and are willing to test it for me or give me a github repo to work on, I can add it very quickly.

ajbouh:
I have hacked something together with a shell script to unblock myself, but I'd really like to get it working the "right" way.

I can certainly help you test this, though I should point out that I also need to use a shell script for swig support. So there's a chance that this is a larger bit of work than just -buildmode=shared support... Thoughts?

ajbouh:
@ianthehat Dug into it more and turns out only two things need to be done for c-shared outputs. A main.go file and gc_linkopts. :)

```
go_binary(
    name="example.so",
    deps=[
      ":example", # a go_library
    ],
    srcs=[
      "main.go",
    ],
    gc_linkopts=["-buildmode=c-shared"],
    visibility = ["//visibility:public"],
)
```

ajbouh:
Seems like there's more sorcery required to get this working linux. The above code works on Darwin, but on Linux I'm seeing:

```
[...]

PR #1278, opened 2018-01-25
Propagate linkmode into the binary rules

ianthehat:
Related to #1264 and #54 and #539

### DIFF
--- a/go/private/actions/compile.bzl
+++ b/go/private/actions/compile.bzl
@@ -16,6 +16,9 @@ load(
     "@io_bazel_rules_go//go/private:common.bzl",
     "sets",
 )
+load("@io_bazel_rules_go//go/private:mode.bzl",
+    "LINKMODE_PLUGIN",
+)
 
 def _importpath(l):
   return [v.data.importpath for v in l]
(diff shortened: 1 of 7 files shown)


go/private/actions/link.bzl
```diff
@@ -48,15 +48,19 @@ def emit_link(go,
   gc_linkopts, extldflags = _extract_extldflags(gc_linkopts, extldflags)
 
   # Add in any mode specific behaviours
-  if archive.source.mode.race:
+  linkExternal = False
```
jayconrod: Nit: Snake case

go/private/actions/compile.bzl
```diff
@@ -67,6 +70,8 @@ def emit_compile(go,
   args.add(go.toolchain.flags.compile)
   if go.mode.debug:
     args.add(["-N", "-l"])
+  if go.mode.link in [LINKMODE_PLUGIN]:
```
jayconrod: Maybe just `go.mode.link == LINKMODE_PLUGIN` until there are other modes that require this flag. I'm guessing other modes will need additional flags, so they'll probably have their own cases.

ianthehat: The intent was to have one if per flag rather than one if per mode, hence the list will include the other modes (C_SHARED and C_ARCHIVE will both do this flag).

steeve: i'll open a PR on top of this one for C_ARCHIVE and C_SHARED if you want (they also expose headers)

ianthehat:
jenkins test this please

go/private/actions/compile.bzl
ianthehat: I have been working mostly on plugin mode, but hit a wall with not being able to build the standard library in a way that is ready for it.
You may hit the same problem... but feel free to try!

PR #1279, opened 2018-01-25
Update to new version of Gazelle and increment version

jayconrod:
* Updated to 0.9 in go_rules_dependencies (for go_repository) and in
  README.rst setup instructions.
* Bump RULES_GO_VERSION to 0.10.0 for next release. This will be the
  minimum rules_go version for the next version of Gazelle (which will
  consolidate internal/external go_test rules).


jayconrod:
This breaks wtool. wtool can't be fixed without changing the visibility of some internal packages in Gazelle, and I'd rather delete than fix it anyway.

Before deleting wtool, I'm going to add equivalent functionality to Gazelle. This requires some refactoring. I'm also trying to make it a bit smarter: it should update existing repositories instead of adding redundant rules.

So, yak shaving...

PR #1289, opened 2018-01-29
remove unneeded toolchain from go_rule example

jmhodges:


bazel-io:

PR #1295, opened 2018-01-31
Build stdlib from inside the aspect

ianthehat:
This removes the hard coded list of standard libraries, and builds it per aspect
being propagated.
This is both cleaner and a pre-requisite for the additional link modes to work.

Related to #1264 and #54 and #539

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -45,7 +45,7 @@ INFERRED_PATH = "inferred"
 
 EXPORT_PATH = "export"
 
-def _declare_child(go, path, ext, name):
+def _child_name(go, path, ext, name):
   childname = mode_string(go.mode) + "/"
   childname += name if name else go._ctx.label.name
   if path:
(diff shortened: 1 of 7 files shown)


go/private/context.bzl
```diff
@@ -45,14 +45,20 @@ INFERRED_PATH = "inferred"
 
 EXPORT_PATH = "export"
 
-def _declare_file(go, path="", ext="", name = ""):
-  filename = mode_string(go.mode) + "/"
-  filename += name if name else go._ctx.label.name
+def _declare_child(go, path, ext, name):
```
jayconrod: Nit: `_child_name` or `_child_path` might be a better name for this function. It doesn't actually declare anything.

go/private/rules/rule.bzl
```diff
@@ -18,11 +18,14 @@ load(
 )
 
 def go_rule(implementation, attrs={}, toolchains=[], bootstrap=False, **kwargs):
+  attrs["_go_context_data"] = attr.label(default = Label("@io_bazel_rules_go//:go_context_data"))
+  aspects = []
+  if all([k in attrs for k in ["pure", "static", "msan", "race"]]):
```
jayconrod: I don't find this condition intuitive, but I can't think of a better way to express it. Maybe just a comment explaining would help.

PR #1300, opened 2018-02-04
Bazel 0.10.0

fkorotkov:
Upgraded Bazel version to 0.10.0


bazel-io:

jayconrod:
This field sets the minimum required version of Bazel that we test with in Travis CI. We plan to continue supporting 0.8 until we need to depend on a feature in a newer version, so we'll continue testing against that.

We still test against the tip version of Bazel in Jenkins, so we should have coverage with newer versions.

fkorotkov:
Thanks! Got it 👌

Issue #1307, opened 2018-02-07
go-bindata repo update


skymeyer:
The user https://github.com/jteeuwen has disappeared earlier from github.com hosting `go-bindata`. Meanwhile a hard fork has been put in place by https://github.com/a-urth. A better maintained fork is available at https://github.com/kevinburke/go-bindata.

jayconrod:
Closed via #1308

PR #1308, opened 2018-02-07
go-bindata repo change as jteeuwen went AWOL

skymeyer:
The user https://github.com/jteeuwen has disappeared earlier from github.com hosting go-bindata. Meanwhile a hard fork has been put in place by https://github.com/a-urth.
A better maintained fork is available at https://github.com/kevinburke/go-bindata which is referenced by this commit.

#1307

### DIFF
--- a/extras/bindata.bzl
+++ b/extras/bindata.bzl
@@ -61,7 +61,7 @@ bindata = go_rule(
         "_bindata": attr.label(
             allow_files = True,
             single_file = True,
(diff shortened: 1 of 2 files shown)


bazel-io:

jayconrod:
Jenkins test this please

Thanks for fixing this! Looks like the go-bindata repo is back, but it would be better to switch to something more actively maintained.

PR #1315, opened 2018-02-09
go_repository: raise timeout for fetch_repo

### LINKED ISSUE #1305: go_repository should have a 'timeout' parameter
Currently I have to use very slow cellular network and it makes an issue especially in attempts to fetch big repos like `github.com/docker/docker`. Every attempt fails with a timeout (current default is 600sec) error because of ~200MB must be fetched.

So, I think that it would be helpful if `go_repository` will have a `timeout` parameter that will be passed directly to `repository_ctx.execute`?

jayconrod:
Fixes #1305

### DIFF
--- a/go/private/go_repository.bzl
+++ b/go/private/go_repository.bzl
@@ -14,6 +14,9 @@
 
 load("@io_bazel_rules_go//go/private:common.bzl", "env_execute", "executable_extension")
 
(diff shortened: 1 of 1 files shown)


jayconrod:
Jenkins test this please

jayconrod:
Jenkins test this please

PR #1324, opened 2018-02-13
Convert go_repository rules in go_rules_dependencies to overlay rules

jayconrod:
* Copy overlay_repository.bzl from gazelle into go/private/tools. We
  need a copy of this to avoid a cyclic dependency.
* Replaced go_repository rules in go_rules_dependencies with
  git_repository or http_archive with overlaid build files checked
  into third_party. Proto directives are copied over to the build
  files where appropriate.


go/private/tools/overlay_repository.bzl
```diff
@@ -0,0 +1,76 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file was copied from github.com/bazelbuild/bazel-gazelle.
+# We need a copy here to avoid a circular dependency in go_rules_dependencies.
+# Do not edit directly.
```
ianthehat: Maybe a comment/issue number for a plan on how to fix this in the long term?

jayconrod: I added some explanation on why the copy is needed. Until we have recursive workspaces though, we won't be able to load any external .bzl files from any file transitively loaded from WORKSPACE.

PR #1326, opened 2018-02-14
Set strip value to false by default.

nathantchan:
This change will ensure that unstripped binaries will be built when
the --strip=never. By default, stripped binaries should still
be created, where --compilation_mode defaults to fastbuild and
--strip defaults to sometimes.


bazel-io:

jayconrod:
Jenkins test this please

PR #1328, opened 2018-02-15
Remove importpath from binary block

lexfrei:
Remove importpath from binary block in readme coz deprecated


bazel-io:

lexfrei:
I signed it!

jayconrod:
Travis CI got stuck and timed out. This change couldn't possibly cause a CI failure though, so force submitting.

jayconrod:
Thanks for fixing this by the way!

Issue #1237, opened 2018-01-13
inability to handle multiple different versions of vendored packages


guoshimin:
EDIT: please see https://github.com/bazelbuild/rules_go/issues/1237#issuecomment-357418520 instead.

Example repo: https://github.com/guoshimin/rules_go_test

Inside the example/ folder, there are two sets of go_{library,binary}, where the only difference is the order of deps in the respective go_library.

`//example:example` builds successfully, whereas `//example:example2` fails to build:
```
$ bazel build //example:example2
Loading: 
Loading: 0 packages loaded
INFO: Analysed target //example:example2 (0 packages loaded).
INFO: Found 1 target...
bazel: Entering directory `/home/shimin/.cache/bazel/_bazel_shimin/aa160757cc427c3c02795
fdc1f75cf88/execroot/__main__/'                                                        
[0 / 2] BazelWorkspaceStatusAction stable-status.txt
ERROR: /home/shimin/rules_go_test/example/BUILD.bazel:31:1: GoLink example/linux_amd64_s
tripped/example2 failed (Exit 1)                                                       
google.golang.org/grpc/transport.newFramer: relocation target type.golang.org/x/net/http
2.frameCache not defined                                                               
type.**golang.org/x/net/http2.frameCache: relocation target type.*golang.org/x/net/http2
.frameCache not defined                                                                
google.golang.org/grpc/transport.newFramer: undefined: "type.golang.org/x/net/http2.fram
eCache"                                                                                
type.**golang.org/x/net/http2.frameCache: undefined: "type.*golang.org/x/net/http2.frame
Cache"                                                                                 
2018/01/13 00:07:38 error running linker: exit status 1
bazel: Leaving directory `/home/shimin/.cache/bazel/_bazel_shimin/aa160757cc427c3c02795f
dc1f75cf88/execroot/__main__/'                                                         
Target //example:example2 failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.926s, Critical Path: 1.27s
FAILED: Build did NOT complete successfully
```

What's at play here?

Here's my bazel version:
```
$ bazel version
Build label: 0.8.1
Build target: bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/baz
el/BazelServer_deploy.jar                                                              
Build time: Tue Dec 5 14:29:45 2017 (1512484185)
Build timestamp: 1512484185
Build timestamp as int: 1512484185
```

[...]

PR #1330, opened 2018-02-15
Reverse sense of importpath and importmap

### LINKED ISSUE #1253: Document importmap
Once we are sure it solves the problem and does not need changing, document the importmap attribute added in #1250 and also add an example of it's use

jayconrod:
importpath is the source importpath. Packages can reference other
packages using import declarations with this string.

importmap is the actual importpath. The compiler and linker use this,
and it's visible in stack traces. This only needs to be set when
multiple packages with the same import path would otherwise be linked.

Related #1237 
Fixes #1253

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -38,7 +38,12 @@ def emit_archive(go, source=None):
   if go.cover:
     source, cover_vars = go.cover(go, source)
   split = split_srcs(source.srcs)
-  compilepath = source.library.importpath if source.library.importpath else source.library.name
+  if source.library.importmap:
+    compilepath = source.library.importmap
+  elif source.library.importpath:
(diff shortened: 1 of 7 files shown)


go/private/actions/archive.bzl
```diff
@@ -38,7 +38,12 @@ def emit_archive(go, source=None):
   if go.cover:
     source, cover_vars = go.cover(go, source)
   split = split_srcs(source.srcs)
-  compilepath = source.library.importpath if source.library.importpath else source.library.name
+  if source.library.importmap:
```
ianthehat: I think we should just *always* set importmap (in the library construction)
Then we could drop compilepath and just use 
```
lib_name = source.library.importmap + ".a"
```
and also not have to test it anywhere else

PR #1340, opened 2018-02-21
Various fixes and changes

steeve:
This PR is a mix of various changes and fixes we've made on https://github.com/znly/rules_go that I think are upstream-able first. Of course we can split and shrink if you guys want.

Here goes:
36bc7cc: Add runfiles to GoArchiveData and use them in the GoPath
Some binaries (such as https://github.com/golang/mobile/tree/master/cmd/gobind) rely on runtime files to be in the `GOPATH`

060c171: Expose GoPath provider
Title says it all. Useful for custom rules.

f59a93d: Add tags and environment to go_context
Title says it all.

e4a157e: Expose GoSource provider
Title says it all. Useful for custom rules.

7d1d956: Make the Go environment an update of the OS environment
This ensures that the Go environment also has the OS environment keys.

f789bcd: Propagate build tags when `go install`ing `stdlib`
`std` has a few tags here and there (for instance `crypto/x509` needs the `ios` tag to build on `darwin_arm64`). This makes sure they are propagated.

93ce009: Make stdlib building depend on `CROSSTOOL` (for `runtime/cgo`)
`runtime/cgo` needs the C compiler, and it's not declared as a dependency. As a result, when using a different `CROSSTOOL` (such as Android), it will fail to find the compiler (`clang`).

aad1ce3: extras/bindata: make it executable and run on host
Needed to use cross compilation (via `--platforms`) and bindata.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -23,6 +23,10 @@ load(
 load(
     "@io_bazel_rules_go//go/private:providers.bzl",
     _GoLibrary = "GoLibrary",
+    _GoSource = "GoSource",
+    _GoPath = "GoPath",
+    _GoArchive = "GoArchive",
+    _GoArchiveData = "GoArchiveData",
 )
 load(
     "@io_bazel_rules_go//go/private:repositories.bzl",
@@ -81,6 +85,18 @@ RULES_GO_VERSION = "0.10.0"
 
 GoLibrary = _GoLibrary
 """See go/providers.rst#GoLibrary for full documentation."""
+
+GoSource = _GoSource
+"""See go/providers.rst#GoSource for full documentation."""
+
+GoPath = _GoPath
+"""See go/providers.rst#GoPath for full documentation."""
+
+GoArchive = _GoArchive
+"""See go/providers.rst#GoArchive for full documentation."""
+
+GoArchiveData = _GoArchiveData
+"""See go/providers.rst#GoArchiveData for full documentation."""
 
 go_library = _go_library_macro
 """See go/core.rst#go_library for full documentation."""
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -268,6 +268,7 @@ def _go_context_data(ctx):
   compiler_options = [o for o in raw_compiler_options if not o in [
     "-fcolor-diagnostics",
     "-Wall",
+    "-g0",
   ]]
   linker_options = [o for o in raw_linker_options if not o in [
     "-Wl,--gc-sections",
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -46,6 +46,19 @@ INFERRED_PATH = "inferred"
 
 EXPORT_PATH = "export"
 
+_COMPILER_OPTIONS_BLACKLIST = {
+  "-fcolor-diagnostics": None,
+  "-Wall": None,
+  "-g0": None, # symbols are needed by Go, so keep them
+}
+
+_LINKER_OPTIONS_BLACKLIST = {
+  "-Wl,--gc-sections": None
+}
+
+def _filter_options(options, blacklist):
+  return [option for option in options if option not in blacklist]
+
 def _child_name(go, path, ext, name):
   childname = mode_string(go.mode) + "/"
   childname += name if name else go._ctx.label.name
@@ -259,20 +272,13 @@ def go_context(ctx, attr=None):
 def _go_context_data(ctx):
   cpp = ctx.fragments.cpp
   features = ctx.features
-  raw_compiler_options = cpp.compiler_options(features)
-  raw_linker_options = cpp.mostly_static_link_options(features, False)
-  options = (raw_compiler_options +
-      cpp.unfiltered_compiler_options(features) +
-      cpp.link_options +
-      raw_linker_options)
-  compiler_options = [o for o in raw_compiler_options if not o in [
-    "-fcolor-diagnostics",
-    "-Wall",
-    "-g0",
-  ]]
-  linker_options = [o for o in raw_linker_options if not o in [
-    "-Wl,--gc-sections",
-  ]]
+  compiler_options = _filter_options(
+    cpp.compiler_options(features) + cpp.unfiltered_compiler_options(features),
+    _COMPILER_OPTIONS_BLACKLIST)
+  linker_options = _filter_options(
+    cpp.link_options + cpp.mostly_static_link_options(features, False),
+    _LINKER_OPTIONS_BLACKLIST)
+
   env = {}
   tags = []
   if "gotags" in ctx.var:
@@ -292,7 +298,7 @@ def _go_context_data(ctx):
           ld_executable = cpp.ld_executable,
           compiler_options = compiler_options,
           linker_options = linker_options,
-          options = options,
+          options = compiler_options + linker_options,
           c_options = cpp.c_options,
       ),
   )
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -74,6 +74,9 @@ def _new_args(go):
       "-goarch", go.mode.goarch,
       "-cgo=" + ("0" if go.mode.pure else "1"),
   ])
+  if len(go.tags) > 0:
+    args.add("-tags")
+    args.add(go.tags, join_with = ",")
   if go.cgo_tools:
     args.add([
       "-compiler_path", go.cgo_tools.compiler_path,
@@ -231,6 +234,8 @@ def go_context(ctx, attr=None):
       pathtype = pathtype,
       cgo_tools = context_data.cgo_tools,
       builders = builders,
+      env = context_data.env,
+      tags = context_data.tags,
       # Action generators
       archive = toolchain.actions.archive,
       asm = toolchain.actions.asm,
@@ -267,6 +272,10 @@ def _go_context_data(ctx):
   linker_options = [o for o in raw_linker_options if not o in [
     "-Wl,--gc-sections",
   ]]
+  env = {}
+  tags = []
+  if "gotags" in ctx.var:
+    tags = ctx.var["gotags"].split(",")
   compiler_path, _ = cpp.ld_executable.rsplit("/", 1)
   return struct(
       strip = ctx.attr.strip,
@@ -274,6 +283,8 @@ def _go_context_data(ctx):
       package_list = ctx.file._package_list,
       sdk_files = ctx.files._sdk_files,
       sdk_tools = ctx.files._sdk_tools,
+      tags = tags,
+      env = env,
       cgo_tools = struct(
           compiler_path = compiler_path,
           compiler_executable = cpp.compiler_executable,
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -49,6 +49,25 @@ func abs(path string) string {
 	}
 }
 
+func envUpdate(oldEnv, newEnv []string) []string {
+	retEnvMap := map[string]string{}
+	retEnv := []string{}
+	for _, e := range oldEnv {
+		pair := strings.SplitN(e, "=", 2)
+		k, v := pair[0], pair[1]
+		retEnvMap[k] = v
+	}
+	for _, e := range newEnv {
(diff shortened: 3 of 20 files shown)


bazel-io:

zllak:
I am the author of one of the commits, it's part of a join effort with @steeve 👍

go/private/context.bzl
```diff
@@ -74,6 +74,7 @@ def _new_args(go):
       "-goarch", go.mode.goarch,
       "-cgo=" + ("0" if go.mode.pure else "1"),
   ])
+  args.add(go.tags, before_each = "-tags")
```
jayconrod: env.go expects `-tags` to be a comma separated list. It can't be repeated. I think `args.add(go.tags, join_with=",")` would be correct.

go/private/context.bzl
```diff
@@ -267,13 +270,17 @@ def _go_context_data(ctx):
   linker_options = [o for o in raw_linker_options if not o in [
     "-Wl,--gc-sections",
   ]]
+  env = {}
```
jayconrod: What would go here in the future? We generally want to avoid pulling anything out of the user's environment, since that makes builds less reproducible.

go/private/context.bzl
```diff
@@ -267,13 +270,17 @@ def _go_context_data(ctx):
...
+  tags = []
```
jayconrod: I think the best way to populate this is by reading `ctx.var["gotags"]` or something similar. Users could set tags with `--define=gotags=foo,bar,baz` on the command line or in an rc file.

go/def.bzl
```diff
@@ -23,6 +23,8 @@ load(
 load(
     "@io_bazel_rules_go//go/private:providers.bzl",
     _GoLibrary = "GoLibrary",
+    _GoSource = "GoSource",
+    _GoPath = "GoPath",
```
jayconrod: Feel free to expose `GoArchive` and `GoArchiveData`, too. They're documented and fairly stable at this point.

go/tools/builders/env.go
```diff
@@ -49,6 +49,29 @@ func abs(path string) string {
...
+	for _, e := range old {
+		pair := strings.SplitN(e, "=", 2)
+		k, v := pair[0], pair[1]
+		retEnvMap[k] = v
+	}
+	for _, e := range new {
+		pair := strings.SplitN(e, "=", 2)
+		k, v := pair[0], pair[1]
+		if preV := retEnvMap[k]; preV != "" {
+			retEnvMap[k] = preV + " " + v
```
jayconrod: This scares me. Seems like we could end up with `GOOS=linux linux` or something. Why is this needed?

go/tools/builders/stdlib.go
```diff
@@ -24,13 +24,14 @@ import (
 )
 
 func install_stdlib(goenv *GoEnv, target string, args []string) error {
+	if goenv.tags != "" {
+		args = append(args, "-tags", goenv.tags)
```
jayconrod: You mentioned `ios` in the description. I wonder if there's a way we can set that one automatically?

go/private/context.bzl
steeve: this is not about pulling the user's environment, but rather merging the rule environment with the builder's or `go install` based rules such as `stdlib`)

steeve:
thanks for the comments! i'll be updating it when I get the chance

go/private/context.bzl
steeve: good call! thanks

go/private/tools/path.bzl
```diff
@@ -75,7 +75,7 @@ Found {} in
     seen_libs[golib.importpath] = golib
     package_files = []
     prefix = "src/" + golib.importpath + "/"
-    for src in golib.srcs:
+    for src in as_iterable(golib.srcs) + tuple(as_iterable(golib.runfiles.files)):
```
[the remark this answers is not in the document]
steeve: sure thing

go/private/actions/archive.bzl
```diff
@@ -86,6 +86,7 @@ def emit_archive(go, source=None):
       file = out_lib,
       srcs = as_tuple(source.srcs),
       searchpath = searchpath,
+      runfiles = runfiles,
```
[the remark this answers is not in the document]
steeve: sounds good

go/private/context.bzl
steeve: this could be added later, but the main case for me to use it since the builders can too.
also, when building for ios, both tags and env need to be properly set up, and in that case I'm doing it from inside the rules themselves

steeve: actually `ctx.var["gotags"] or []` is rather easy to do, i'm sold

go/def.bzl
steeve: awesome

go/tools/builders/env.go
steeve: this is true. however, both ends (old and new) are controlled by rules_go. But in the case of `CGO_CFLAGS` for instance, we could get to the case when it would get overridden instead of merged

go/tools/builders/stdlib.go
steeve: At the moment the only way I found to reliably do it is by using `ctx.fragments.cpp.target_gnu_system_name`.

Here is the gist of the code, more or less (I plan to submit this later on):
```python
PLATFORMS = {
    'armv7-apple-ios': apple_common.platform.ios_device,
    'arm64-apple-ios': apple_common.platform.ios_device,
    'i386-apple-ios': apple_common.platform.ios_simulator,
    'x86_64-apple-ios': apple_common.platform.ios_simulator,
}

def apple_ensure_options(ctx, env, tags, compiler_options, linker_options):
    system_name = ctx.fragments.cpp.target_gnu_system_name
    platform = PLATFORMS.get(system_name)
    if platform == None:
        return
    if system_name.endswith("-ios"):
        tags.append("ios") # needed for stdlib building
    if platform in [apple_common.platform.ios_device, apple_common.platform.ios_simulator]:
        min_version = _apple_version_min(platform, "6.1")
        compiler_options.append(min_version)
        linker_options.append(min_version)
    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
    env.update(apple_common.target_apple_env(xcode_config, platform))
```

steeve: As you can see also, this method modifies the environment, and in this case to tell the compiler which SDK to use (iPhoneOS, iPhoneSimulator).

For instance:
```
    APPLE_SDK_PLATFORM=iPhoneOS
    APPLE_SDK_VERSION_OVERRIDE=11.2
    XCODE_VERSION_OVERRIDE=9.2.0
```

These also need to be set at stdlib bootstrapping.

go/tools/builders/env.go
steeve: by both end I mean the skylark and go builder side

go/private/tools/path.bzl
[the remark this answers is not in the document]
steeve: the thing is that i need the importpath to put them in the fake gopath properly, so keeping the archives is rather useful

steeve:
I've updated the PR with the gotags suggestions.

I have all the tests passing. I fixed them with https://github.com/bazelbuild/rules_go/pull/1340/commits/e42f021d5f890ce5777c6c1f4caef15f1ecdb401

steeve:
Changed the go-bindata commit message to be closer to the ones here.

steeve:
Added https://github.com/bazelbuild/rules_go/pull/1340/commits/940d41bdf9faef0c041041dff72fca0b040e6c6a which splits sources via a `dict` instead of using `any` + lists.
It improves splitting performance by about 8-9x. As this is kind of a hot path, hopefully it should yield noticeable improvements.

I benched it with this code: https://gist.github.com/steeve/56d3f107f24d8e67833f66ed3e06acdd

old: `[8.00760793685913, 8.148405075073242, 8.099946975708008]`
new: `[1.3760578632354736, 1.4115018844604492, 1.4008851051330566]`

steeve:
Added two more commits that ensure the proper dependency chain for CGo and makes sure that all compiler options are used (some were missed when cross compiling)

go/private/context.bzl
jayconrod: Ok, that makes sense.

go/private/tools/path.bzl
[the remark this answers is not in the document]
jayconrod: Hmm, makes sense. There needs to be some association between the import path and the data files, but we need to avoid placing `runfiles` in `GoArchiveData` to avoid a blowup in data structure size and time to iterate.

Instead of adding `runfiles` to `GoArchiveData`, I'd suggest adding a `data` field, which will be a tuple of data files just for that archive. This will be essentially the same format as `srcs`. You can get this tuple in `emit_archive` by iterating `source.runfiles` before it's merged.

steeve:
I'm going to leave out the gopath commit for now so as to not block the PR with this. I'll open it on another PR.

jayconrod:
+1 to moving gopath to a separate PR. It will be easier to review and merge in pieces.

My only remaining comment is this one from `d73858f`. GitHub's code review system confuses me, and I think it got buried in a rebase somewhere. I'll copy it here:

> Ok, I think I understand this better. I'm still concerned that it's prone to duplication. I don't think there's a situation where we actually need to merge flags in the same environment variable. Here's what I think we should do:
> 
> Any value that any builder needs to handle explicitly should be passed with a command line argument. For example, the builders perform filter on build tags, so they need to know GOARCH and GOOS. The builder will add these to the environment before executing an SDK tool.
> 
> Any value that all builders should simply pass through to executed programs should be an environment variable.
>
> There shouldn't be any overlap between these two categories, so env.go can continue doing what it did before.

steeve:
To be clear, what you're advising is that `cgo_ldflags` should be set specifically in the case of the stdlib bootstrap (it is set via a env now). SGTM

jayconrod:
My concern was about that we shouldn't do environment merging in env.go. It would be better to either set an environment variable to be passed through without modification to the toolchain, or to provide a command line argument to the builders (that maybe gets turned into an environment variable) but not both.

I guess `CGO_LDFLAGS` / `-ld_flag` is actually both right now. That's something I'm not happy with, but there's no need to fix that in this PR.

steeve:
I'm going to do with the `-ld_flags` from `env.go` because, well, it's there and setting `CGO_LDFLAGS` from the rule doesn't really make sense actually, imho

steeve:
Ok so what I'm going to do is if there is a conflict between the old (`os.Environ()` and the new `env.env()`), then the new will take precedence.
Effectively working like `dict.update()`

steeve:
I've updated the branch with your suggestions in https://github.com/bazelbuild/rules_go/pull/1340/commits/840fda363fd78b2e982051320a83777481099a73 and https://github.com/bazelbuild/rules_go/pull/1340/commits/5cda48b4b942651580274674f1e2f178db02b312, so far so good

steeve:
I've added https://github.com/bazelbuild/rules_go/pull/1340/commits/112739d98e91c27bed0759bff3f7efbcec0081ed for a very quick fix

steeve:
Also add https://github.com/bazelbuild/rules_go/pull/1340/commits/aa59cc99eeba0af422cfa7e0212ba796d3c64826 which ensures go-bindata doesn't use the file's modtime unless otherwise specified. This would make our builds no hermetic when using it.

steeve:
And rebased on `master`

steeve:
Added https://github.com/bazelbuild/rules_go/pull/1340/commits/dc60119711df3be578f7777b61d4688a101c129d to add some parameters and `extra_args` to the go-bindata rule

steeve:
Added https://github.com/bazelbuild/rules_go/pull/1340/commits/31222eac4db6cd9174ab1eba142cbcf8f5f7f6a3 which add the `field_mask` well known type to gogo remaps

Sorry I keep adding commits but I figure given their scopes that now is a good time to do it

jayconrod:
Ok, I think everything looks good at this point.

PR #1344, opened 2018-02-22
Remove previously copied files before building stdlib

### LINKED ISSUE #1343: stdlib fails after Go version is changed and sandboxing is disabled
The stdlib builder copies stdlib sources into a temporary directory before building them. If a different version of Go is built later, the new sources will be copied over the old sources, which can result in redundant definitions after files are renamed.

jayconrod:
Fixes #1343


jayconrod:
Jenkins test this please

PR #1354, opened 2018-02-26
Add importmap attribute to go_proto_library

jayconrod:
Also: bump version number, since this is the first API change
since 0.10

### DIFF
--- a/tests/core/go_proto_library_importmap/importmap_test.go
+++ b/tests/core/go_proto_library_importmap/importmap_test.go
@@ -11,6 +11,6 @@ func TestImportMap(t *testing.T) {
 	got := reflect.TypeOf(foo.Foo{}).PkgPath()
 	want := "never/gonna/give/you/up"
 	if got != want {
-		t.Errorf("got %q ; want %q", got, want)
+		t.Errorf("got %q; want %q", got, want)
 	}
 }
--- a/tests/core/go_proto_library_importmap/importmap_test.go
+++ b/tests/core/go_proto_library_importmap/importmap_test.go
@@ -1,0 +1,16 @@
+package importmap_test
+
+import (
+	"reflect"
+	"testing"
+
+	foo "github.com/bazelbuild/rules_go/tests/core/go_proto_library_importmap"
+)
+
+func TestImportMap(t *testing.T) {
+	got := reflect.TypeOf(foo.Foo{}).PkgPath()
+	want := "never/gonna/give/you/up"
+	if got != want {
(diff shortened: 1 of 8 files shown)


jayconrod:
Notes for review

* The importmap attribute controls the -importmap parameter on the command line for libraries. We added it last month for go_library but neglected to do so for go_proto_library. This is needed to avoid conflicts with multiple vendored versions of the same library.
* This is a one-line change in proto/def.bzl. The attribute is read in go/private/context.bzl in _new_library.
* Tables of contents in test documentation are generated with tests/update.py. Looks like some entries were out of order.

achew22:
How does this interact with having multiple `go_proto_libraries`-esque rules that generate in the same importpath? For example, if you're generating a gRPC proto along with [go-proto-validators](https://github.com/mwitkow/go-proto-validators) or [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway)? Is there any risk that having two `go_proto_library`-esque rules will conflict?

jayconrod:
@achew22 `go_proto_library` will behave the same way as `go_library` with respect to import paths.

The Go toolchain makes a distinction between *source* import paths (those used in source code to import packages) and *actual* import paths (those used by the compiler and linker to refer to packages). The packages linked into a binary must have unique actual import paths. Normally source and actual import paths are the same, but they can be changed with the `-importmap` or `-importcfg` flags to `go tool compile` and `go tool link`. `go build` uses these automatically for vendored packages to avoid conflicts when linking multiple packages with the same source import path. The `importmap` attribute here lets you do the same with Bazel. I'm working on a change for Gazelle to generate `importmap` attributes for libraries in vendor directories.

You asked specifically about multiple Go proto libraries in the same import path. If these are linked into the same binary, they'll need to have different actual import paths (`importmap` will need to be set). You can also use `embed` to combine libraries into a single package.

tests/core/go_proto_library_importmap/importmap_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"testing"
+
+	foo "github.com/bazelbuild/rules_go/tests/core/go_proto_library_importmap"
+)
+
+func TestImportMap(t *testing.T) {
+	got := reflect.TypeOf(foo.Foo{}).PkgPath()
+	want := "never/gonna/give/you/up"
+	if got != want {
+		t.Errorf("got %q ; want %q", got, want)
```
alandonovan: no space before semicolon

PR #1358, opened 2018-02-28
proto: implemented full end-to-end support for well-known-types

teh-cmc:
This PR includes complete support for well-known-types; which are now compiled directly via the official `proto_library` rules over at github.com/google/protobuf.

This has been tested on several repositories of us that make heavy use of both protobuf & grpc; everything appears to be working properly for both golang/protobuf & gogo/protobuf.

Thanks for your time!

cc @steeve

### DIFF
--- a/proto/well_known_types.bzl
+++ b/proto/well_known_types.bzl
@@ -1,0 +1,40 @@
+load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
+
+WELL_KNOWN_TYPES_MAP = {
+    "any": ("github.com/golang/protobuf/ptypes/any", []),
+    "api": ("github.com/golang/protobuf/ptypes/api", ["source_context", "type"]),
+    "compiler_plugin": ("github.com/golang/protobuf/protoc-gen-go/plugin", ["descriptor"]),
+    "descriptor": ("github.com/golang/protobuf/protoc-gen-go/descriptor", []),
+    "duration": ("github.com/golang/protobuf/ptypes/duration", []),
+    "empty": ("github.com/golang/protobuf/ptypes/empty", []),
+    "field_mask": ("google.golang.org/genproto/protobuf/field_mask", []),
+    "source_context": ("google.golang.org/genproto/protobuf/source_context", []),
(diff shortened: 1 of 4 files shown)


steeve:
Basically the main idea is not to use the already-generated `.pb.go` files from both repos (which is rather recent for gogo anyway) but have the `go_proto_library` generate them directly from `com_google_protobuf`.

teh-cmc:
I signed it

jayconrod:
LGTM, but I'd like @ianthehat to take a look at this tomorrow before merging.

proto/well_known_types.bzl
```diff
@@ -0,0 +1,40 @@
+load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
```
jayconrod: Actually I do have one change to request: could you create a new `wkt` subdirectory (`//proto/wkt`) and move this bzl file there, along with the call to `gen_well_known_types`? I think it makes sense for them to be in a separate package from the compilers.

proto/well_known_types.bzl
```diff
@@ -0,0 +1,40 @@
...
+    for wkt, (go_package, _) in WELL_KNOWN_TYPES_MAP.items() if "protoc-gen-go" not in go_package
+] + [
+    "Mgoogle/protobuf/descriptor.proto=github.com/gogo/protobuf/protoc-gen-gogo/descriptor",
+    "Mgoogle/protobuf/compiler_plugin.proto=github.com/gogo/protobuf/protoc-gen-gogo/plugin",
+]
+
+def gen_well_known_types():
+    rules = []
+    for wkt, (go_package, deps) in WELL_KNOWN_TYPES_MAP.items():
+        name = "wkt_{}_proto".format(wkt)
```
jayconrod: Since these will be in the `//proto/wkt` package, they won't need `wkt` in the name. Also, since they're `go_proto_library` rules, they should probably be named `{}_go_proto`, which is what Gazelle conventionally names `go_proto_library` rules.

teh-cmc:
I've updated the implementation based on your recommendations @jayconrod.

PR #1366, opened 2018-03-05
Split C and C++ compilation

steeve:
## What it does
This PR splits C and C++ CGo compilation into separate steps. The main idea behind this is for specifying per language options (such as `std=c++14` for C++).

For that, the `pack` builder has been modified to take multiple archives (C and C++) and fuse them.

This is also a required step for adding ObjC support.

I have also added a commit which tries to make relative `CC` options absolute (via whitelisting)


jayconrod:
I took a quick look, but I won't be able to do a detailed review until tomorrow. I have a couple concerns before then though:

* Could you explain a bit more about what the ObjC support will look like? I guess some sources would need to be compiled by `objc_library` or something like that.
* Bazel treats the C / C++ toolchain as mostly one thing, configured with CROSSTOOL. I think Objective C / C++ is also configured with CROSSTOOL, but I don't know how the flags are separated. In general, I think CROSSTOOL is the right place for language-specific flags. `go_library` `copts` should be reserved for library-specific flags like `-D`, `-I`, `-framework`.
* If C / C++ / ObjC sources and headers are separated into different `cc_library` / `objc_library` rules, how will they compile if there are dependencies between sources in different languages?
* I mentioned this in the other PR, but there is ongoing work to expose C++ toolchains to Skylark. I'm not sure whether ObjC is part of that effort, but that would let us avoid `cc_library` and `objc_library` rules altogether. This is a big enough change that it might be better to wait until that work is complete.

steeve:
Thanks for being so fast.

Here are some answers:

1. For ObjC support, you can check out the early work on this at https://github.com/znly/rules_go/commit/30ee77332fe195d6819c3b8ff6b9233b20ae985a, but meat is is at https://github.com/znly/rules_go/commit/30ee77332fe195d6819c3b8ff6b9233b20ae985a#diff-1d1e845efd5e9ec6b7a69dcecb01ed38R446
2. You're right, but if you're building something else than C, setting some flags will contaminate the CGoCodeGen pure C compilation flags and fail the whole thing. In our project, for instance, we're using C++14 in cpp files in Go packages, and setting `CXXFLAGS` is the only way to make it work. However, if we add `-std=c++14` to a pure C compilation, `clang` will fail.
3. For now I'm propagating the deps to all the languages. I don't think it's that bad really? Wether or not one uses the deps is up to them but at least Bazel makes them available to the library.
4. I'd love for this to happen! It would be the cleanest solution for sure. Last I checked though the spec is expected to land for Q1, then there is the implementation. Wet finger estimate point to at least may-june before it's in `master`. Or maybe you have some insider information ? 😄

jayconrod:
For c++14, it looks like the [recommended way](https://stackoverflow.com/a/43388168/1891) is to pass `--cxxopt='-std=c++14'` on the command line. You can put this in your tools/bazel.rc file so Bazel picks it up automatically. That will also apply to other `cc_library` rules.

About C/C++ sandwich: @mhlopko, how's it going? Will ObjC be supported, too? For context, it would be useful to build mixed C/C++/ObjC sources that appear alongside cgo code. Right now, we build C/C++ as a `cc_library`.

steeve:
> For c++14, it looks like the recommended way is to pass --cxxopt='-std=c++14' on the command line. You can put this in your tools/bazel.rc file so Bazel picks it up automatically. That will also apply to other cc_library rules.
Indeed, I saw that flag but I figure it was easier just splitting and doing it package by package and also because Go supports setting `CFLAGS` and `CXXFLAGS` separately, too.

jayconrod:
I'll wait until #1365 is merged before diving into this.

steeve:
Sure thing!

mhlopko:
Sorry for the slow reply. We are working on 2 things in C++ rules:

* Skylark API to C++ toolchain ([tracking issue](https://github.com/bazelbuild/bazel/issues/4571)) - so you can ask cc_toolchain to give you the same command line that cc_library/binary would use
* Skylark API to C++ rules (C++ sandwich, [tracking issue](https://github.com/bazelbuild/bazel/issues/4570)) - so you can access C++ providers and interoperate with c++ rules from other Skylark rules.

I don't know much about open source go rules, but I assume you'll want to use the sandwich. That will allow you to consume C++ providers, dig objects out, and use them with cgo. @oquenchil  can comment on timeframes.

Re ObjC, they will use their own providers, but tbh I don't know much about that. Maybe @c-parsons has more information.

steeve:
@mhlopko thanks for the reply

Indeed we'd like to be able to return `cc` providers to be consumed by `cc_` rules. Right now we rely on macros and filegroups to do so, and in the end are forced to name them differently.

See for instance at https://github.com/bazelbuild/rules_go/pull/1365/files#diff-dfaef6484291891479c192a6169bc232R90

steeve:
Changed `pack` to ensure object names are unique among the multiple archives in https://github.com/bazelbuild/rules_go/pull/1366/commits/dd11da3fa63a739abfdf16c2784b5eab1dcd541d

jayconrod:
@mhlopko I'm actually most interested in the toolchain being exposed in Skylark. How far along is that? #4571 doesn't have much information. We'd like to be able to invoke a C / C++ / ObjC compiler and linker on some generated code during execution without creating `cc_library` / `objc_library` rules. I think we can get the compiler command and flags right now, but I'm guessing the includes and libraries and other toolchain files won't be visible in the sandbox.

The sandwich is also useful for other things, but I'm mainly interested in the toolchain right now.

steeve:
Correct me if I'm wrong, but why not do a pure skylark `cc_` rules when #4571 lands ?

jayconrod:
I think that's the long-term plan, but the C/C++ rules are really complicated, having accumulated a decade's worth of features and idiosyncrasies within Google. A full migration will take years. I hope a simplified version that serves most common cases will be available sooner than that though.

mhlopko:
Steeve: that's the end goal, to migrate existing rules into Skylark. But there are things that C++ rules do that are not supported in Skylark yet, plus we want to use the same rules internally and externally, and for that we need to do small incremental steps, making sure google doesn't break.

Jay: Since I still know very little about external go rules let me make sure at least you have enough information to decide what's best :) Part of the sandwich (spoiler alert :) are functions cc_common.compile(...) and cc_common.link(...). AFAIK this is what you need, give C++ rules sources, compile them, make a .a of them, and propagate providers. Toolchain api will only enable you to get the command line flags that C++ rules would use. That's definitely something you need for internal go rules though.

The work is ongoing, optimistically EOQ1 for test runs behind a whitelist. I'll keep you updated.

oquenchil:
> The work is ongoing, optimistically EOQ1 for test runs behind a whitelist. I'll keep you updated.

This timeframe sounds correct. This should be tested internally first where we have more control on who's using the API and we can change it more easily if necessary.

steeve:
Please hold on this one, I'm going to properly rebase it on top of `master`.

steeve:
I've properly rebased it and added tests.
@jayconrod your move

go/private/rules/wrappers.bzl
```diff
@@ -25,6 +25,8 @@ _CGO_ATTRS = {
     "srcs": None,
     "cdeps": [],
     "copts": [],
+    "cxxopts": [],
+    "cppopts": [],
```
jayconrod: Please update attribute tables in go/core.rst to include the new attributes.

This may break people that have C++ code that currently depends on options in `copts`. Hopefully this won't affect many people. I'll file an issue on Gazelle to separate these.

tests/core/cxx/BUILD.bazel
```diff
@@ -0,0 +1,19 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
```
jayconrod: Several test comments:

1. In terms of test organization, I'd prefer this to be tests/core/cgo, since this is really testing cgo functionality. Could be divided further into language subdirectories (tests/core/cgo/cxx).
2. We really want to verify that the right set of options get passed to each language. `copts` should be C-only, `cxxopts` should be C++-only, `cppopts` should be both. Easiest way to do that is probably with `-D` flags, then checking at build time with `#if` and `#error`. This would also fix the error on Travis CI due not having a C++14 compiler available.
3. Please create a README.rst describing the tests, and run tests/update.py.

steeve:
I'm in a bit of a dilemma over `copts` vs `cppopts`. When one wants to use `-D` they now need to put it into `cppopts` or else it won't work. While is this correct, this is rather not friendly imho...

steeve:
PR updated with most of the advices.
See https://github.com/bazelbuild/rules_go/pull/1366/commits/8ee185c352236a6f7bfe1eecc131fa6ed1180c53 for what I'm describing

steeve:
Never mind I think I solved it by https://github.com/bazelbuild/rules_go/pull/1366/commits/a41c767517fe0f6f3623c57ee5bc68f4873afd70

go/private/rules/cgo.bzl
```diff
@@ -78,20 +78,23 @@ def _c_filter_options(options, blacklist):
   return [opt for opt in options
         if not any([opt.startswith(prefix) for prefix in blacklist])]
 
-def _select_archive(files):
+def _select_archives(files):
```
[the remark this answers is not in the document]
steeve: i still need to check that one out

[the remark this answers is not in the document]
steeve: ok got it, i've pushed a fix

go/private/rules/cgo.bzl
```diff
@@ -91,14 +91,28 @@ def _select_archive(files):
...
+def _select_archives(libs):
+  """Selects a one per item in a cc_library, if needed.
+
+  If no archive can be extracted from all the libraries, this will fail.
+  """
+  # list of file extensions in descending order or preference.
+  outs = []
+  for lib in libs:
+    archive = _select_archive(lib.files)
+    if archive:
```
jayconrod: When would a `cc_library` not produce any usable archives? Does that happen if there are no source files?

go/private/rules/cgo.bzl
```diff
@@ -91,14 +91,28 @@ def _select_archive(files):
...
+    fail("cc_library(s) did not produce any files")
+  return outs
 
 def _cgo_codegen_impl(ctx):
   go = go_context(ctx)
   if not go.cgo_tools:
     fail("Go toolchain does not support cgo")
   linkopts = ctx.attr.linkopts[:]
-  copts = go.cgo_tools.c_options + go.cgo_tools.compiler_options + ctx.attr.copts
+  cppopts = go.cgo_tools.c_options + go.cgo_tools.compiler_options + ctx.attr.copts + ctx.attr.cppopts
```
jayconrod: Let's keep `cppopts` and `copts` separate here for clarity.

```
cppopts = go.cgo_tools.compiler_options +ctx.attr.cppopts
copts = go.cgo_tools.c_options + ctx.attr.copts
```

go/private/rules/cgo.bzl
```diff
@@ -161,10 +181,10 @@ def _cgo_codegen_impl(ctx):
   # The first -- below is to stop the cgo from processing args, the
   # second is an actual arg to forward to the underlying go tool
   args.add(["--", "--"])
-  args.add(copts)
+  args.add(cppopts)
```
jayconrod: Both `cppopts` and `copts` can be added here.

tests/core/cgo/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    cgo = True,
+    copts = ["-DRULES_GO_C"],
+    cxxopts = ["-DRULES_GO_CPP"],
+    srcs = [
```
jayconrod: Define something in `cppopts` and check that define is visible in both add.c and add.cpp.

go/core.rst
```diff
@@ -392,6 +392,18 @@ Attributes
 | Subject to `"Make variable"`_ substitution and `Bourne shell tokenization`_.                     |
 | Only valid if :param:`cgo` = :value:`True`.                                                      |
 +----------------------------+-----------------------------+---------------------------------------+
+| :param:`cxxopts`           | :type:`string_list`         | :value:`[]`                           |
```
jayconrod: Looks good, but copy to `go_library` and `go_test`.

go/private/rules/cgo.bzl
steeve: This can happen if there are only header files (which happens if C++ files are missing, for instance).

go/private/rules/cgo.bzl
steeve: got it

tests/core/cgo/BUILD.bazel
steeve: 👍

go/private/rules/cgo.bzl
steeve: I realize I could have been clearer. In the case of a regular C compilation, the `cc_library` responsible of building the C++ files will not produce any archive.

steeve:
I've updated the PR, thank you for the feedback!

steeve:
@jayconrod do you want me to rebase instead ?

jayconrod:
@steeve No need, I was just waiting for CI to run again after resolving the merge conflict.

Thanks again for working on this.

PR #1382, opened 2018-03-15
Handle Windows long file names in a few more places.

pmuetschard:
Go can handle long file names on Windows if they are not relative. This change makes some file paths absolute so that go can handle them. See https://github.com/golang/go/blob/b86e76681366447798c94abb959bb60875bcc856/src/os/path_windows.go#L133.

The `protoc` binary cannot handle long filenames either, so this change generates go protos in the temporary folder and then moves the contents into place.

### DIFF
--- a/go/tools/builders/compile.go
+++ b/go/tools/builders/compile.go
@@ -49,6 +49,7 @@ func run(args []string) error {
 	if err := goenv.update(); err != nil {
 		return err
 	}
+	absOutput := abs(*output) // required to work with long paths on Windows
 
 	var matcher func(f *goMetadata) bool
 	switch *testfilter {
@@ -80,7 +81,7 @@ func run(args []string) error {
 		}
 	}
 	if len(files) <= 0 {
-		return ioutil.WriteFile(*output, []byte(""), 0644)
+		return ioutil.WriteFile(absOutput, []byte(""), 0644)
 	}
 
 	goargs := []string{"tool", "compile"}
@@ -102,7 +103,7 @@ func run(args []string) error {
 		goargs = append(goargs, "-importmap", mapping)
 		strictdeps = append(strictdeps, source)
 	}
-	goargs = append(goargs, "-pack", "-o", *output)
+	goargs = append(goargs, "-pack", "-o", absOutput)
 	goargs = append(goargs, flags.Args()...)
 	for _, f := range files {
 		goargs = append(goargs, f.filename)
--- a/go/tools/builders/filter.go
+++ b/go/tools/builders/filter.go
@@ -37,7 +37,7 @@ type goMetadata struct {
 func readFiles(bctx build.Context, inputs []string) ([]*goMetadata, error) {
 	outputs := []*goMetadata{}
 	for _, input := range inputs {
-		if m, err := readGoMetadata(bctx, input, true); err != nil {
+		if m, err := readGoMetadata(bctx, abs(input), true); err != nil {
 			return nil, err
 		} else if m.matched {
 			outputs = append(outputs, m)
--- a/go/tools/builders/protoc.go
+++ b/go/tools/builders/protoc.go
@@ -57,6 +57,17 @@ func run(args []string) error {
 	if err := flags.Parse(args); err != nil {
 		return err
 	}
+
+	// Output to a temporary folder and then move the contents into place below.
+	// This is to work around long file paths on Windows.
+	tmpDir, err := ioutil.TempDir("", "go_proto")
+	if err != nil {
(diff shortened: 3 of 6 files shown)


go/tools/builders/compile.go
```diff
@@ -49,6 +49,7 @@ func run(args []string) error {
 	if err := goenv.update(); err != nil {
 		return err
 	}
+	absOutput := abs(*output)
```
jayconrod: Without any comment here, we'll probably forget why this is here and remove it in six months. :)

It would also be good to have a more detailed comment on `abs` itself that references `path_windows.go` as you did in the commit message.

go/tools/builders/protoc.go
```diff
@@ -95,28 +104,46 @@ func run(args []string) error {
...
+		}
+		if relPath == "." {
 			return nil
 		}
-		info := files[path]
-		if info != nil {
-			info.created = true
+
+		if f.IsDir() {
+			if err := os.Mkdir(abs(filepath.Join(*outPath, relPath)), f.Mode()); !os.IsExist(err) {
```
jayconrod: Move the `abs` call to `*outPath` next to `tmpDir` above and use the result here. The comment can explain why both are needed.

go/tools/builders/protoc.go
```diff
@@ -57,14 +57,23 @@ func run(args []string) error {
...
 	}
+
+	// Output to a temporary folder and then move the contents into place below.
+	// This is to work around long file paths on Windows.
+	tmpDir, err := ioutil.TempDir("", "go_proto")
+	if err != nil {
+		return err
+	}
+	tmpDir = abs(tmpDir)
+
```
jayconrod: `defer os.RemoveAll(tmpDir)` instead of calling it at the end.

go/tools/builders/protoc.go
```diff
@@ -95,28 +104,46 @@ func run(args []string) error {
...
+		}
+
+		info := &genFileInfo{
 			path:    path,
 			base:    filepath.Base(path),
 			created: true,
 		}
-		files[path] = info
+
+		foundInfo := files[relPath]
```
jayconrod: `if foundInfo, ok := files[relPath]; ok { ...`

go/tools/builders/filter.go
```diff
@@ -37,7 +37,7 @@ type goMetadata struct {
 func readFiles(bctx build.Context, inputs []string) ([]*goMetadata, error) {
 	outputs := []*goMetadata{}
 	for _, input := range inputs {
-		if m, err := readGoMetadata(bctx, input, true); err != nil {
+		if m, err := readGoMetadata(bctx, abs(input), true); err != nil {
```
jayconrod: Looks like tests are failing because `//go/builders:filter_test` does not include env.go, which it now needs.

go/tools/builders/compile.go
pmuetschard: Done.

go/tools/builders/filter.go
pmuetschard: Done.

go/tools/builders/protoc.go
pmuetschard: Done.

go/tools/builders/protoc.go
pmuetschard: Done.

go/tools/builders/protoc.go
pmuetschard: Done.

pmuetschard:
> So if I understand correctly, in order to be safe on Windows, we need to make every path absolute before calling any file API on Windows?

Yes.

> Is this a common problem, or does it affect a few things (protoc) that have particularly long paths?

The path lenght limitation is a Windows global thing. Most tools work around it, by using the `\\?\c:\foo\bar` form of paths. Golang does, too, but, as mentioned in the comment of `path_windows.go`, bails if the input path is relative (the `\\?\` form has some processing limitations, including only accepting absolute paths). So, the absolute path vs relative path, is a golang thing. It is indeed more of a problem with protobuf and go with bazel due to paths like this:
`bazel-out/x86_windows-fastbuild/bin/external/io_bazel_rules_go/proto/wkt/windows_amd64_stripped/source_context_go_proto~/google.golang.org/genproto/protobuf/source_context/source_context.pb.go`. Although as is, this path is just ~200 chars, the length check also has to pass on the abosolute form, which includes the `c:\users\user\appdata\local\temp\_bazel_user\xxxxxxxx\execroot\workspace\` prefix...

Issue #1150, opened 2017-12-18
Supporting Code Coverage


jeb2239:
I noticed that coverage is not supported , but there seems to be some experimental support. How can I contribute to the progression of this feature?

ianthehat:
All the things needed to correctly run coverage are done, the issues are how to collect the results and how to display them.
We either need to massage the results into a form that Bazel can understand, or provide an easy way to use the normal go tooling. Maybe even both.

jeb2239:
Is there a way to make a profile `go test -coverprofile=c.out` currently in bazel.

jayconrod:
You can pass arguments to tests on the command line with `--test_arg=arg`.

Note that tests run in a sandbox by default, and Bazel will clean up their output files after they run. You can disable the sandbox with `--test_strategy=standalone`.

jeb2239:
the --test_arg method doesn't work is there another way to get this coverage profile besides using the go_path method? This profile is needed to do anything related to code coverage.

jayconrod:
The way to do this is with `--test_arg -test.coverprofile=out`. However, I just tried this and got this message:

```
==================== Test output for //:go_default_test:
testing: cannot use -test.coverprofile because test binary was not built with coverage enabled
```

I think that used to work, but it doesn't anymore. It looks like we need to set the `Mode` field in `Cover` objects.

jeb2239:
Hello has any progress been made with this?

jayconrod:
I just checked this again with rules_go on master and Bazel 0.10.0rc6, and I'm getting a different error. Not sure if something changed, or I was running it wrong before.

```
$ bazel coverage :go_default_test --test_arg -test.coverprofile=$PWD/cover.out
...
testing: open /usr/local/google/home/jayconrod/go/src/example.com/repo/cover.out: read-only file system
```

I think this fails because of sandboxing; the test can't write anything outside of the sandbox. Running with `--test_strategy=standalone` did not work. You can write to /tmp though.

```
$ bazel coverage :go_default_test --test_arg -test.coverprofile=/tmp/cover.out
...
//:go_default_test                                                       PASSED in 0.1s

$ go test -coverprofile=/tmp/cover.out
go test -coverprofile=/tmp/cover.out
PASS
coverage: 100.0% of statements
ok  	example.com/repo	0.011s
```

[...]

PR #1387, opened 2018-03-16
Save test coverage data to bazel-testlogs.

jmillikin-stripe:
This lets `bazel coverage //some:test` produce a Go coverage data file
in `bazel-testlogs/some/test/coverage.dat`. This file can be processed
by normal Go coverage tooling to produce a coverage report.

The change to `go/private/rules/test.bzl` needs some explanation --
Bazel only looks for coverage data if the test target has an
`InstrumentedFilesProvider`, but this provider can currently only be
created using "legacy" provider syntax. Old and new provider syntaxes
can be combined by putting new-style providers in a `providers` field
of the old-style struct.

If the provider is found and at least one source file is present, Bazel
will set the `COVERAGE_OUTPUT_FILE` environment variable during tests
and will save that file to the build events + test outputs.

Related to https://github.com/bazelbuild/rules_go/issues/1150 and https://github.com/bazelbuild/rules_go/issues/140


jeb2239:
Do you know of any coverage tools which work with this dat file or ones that this was tested with?

jmillikin-stripe:
This is the plain Go coverage data (same as generated by `go test -coverprofile`), so any tools based on `golang.org/x/tools/cover` will be able to process it.

`go tool coverage` needs a little bit of help because it assumes all your sources are under `src/`, but this is pretty easy to work around with `ln -s . src` if you want to use the unmodified upstream SDK.

jayconrod:
The `InstrumentedFilesProvider` thing is super weird. I wish coverage was getting more attention from the Bazel folks.

I'm guessing we'll need to break this when Bazel has full coverage support. Blaze (Google internal version) expects coverage data in the lcov format, so we may need to write that instead of Go's format.

For now though, this is welcome. Thanks for submitting this.

jmillikin-stripe:
I've been talking with the Bazel devs about coverage as part of the Envoy project, and one of the sticking points is whether coverage should be Blaze-style (unified for all languages), or toolchain-specific.

Being able to gather code coverage in the native format from Go will be useful data for that decision, and I'm hopeful we can expand this approach to other Skylark rulesets.

PR #1388, opened 2018-03-18
proto: embed doesn't have GoProtoImports provider

kellycampbell:
In go_proto_library rule definition, embed is declared as
  "embed": attr.label_list(providers = [GoLibrary]),

This causes the following error when get_imports runs:
  (rule 'go_proto_library') doesn't contain declared provider 'GoProtoImports'

### DIFF
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -42,7 +42,6 @@ def get_imports(attr):
   if hasattr(attr, "proto"):
     imports.append(["{}={}".format(proto_path(src), attr.importpath) for src in attr.proto.proto.direct_sources])
   imports.extend([dep[GoProtoImports].imports for dep in getattr(attr, "deps", [])])
-  imports.extend([dep[GoProtoImports].imports for dep in getattr(attr, "embed", [])])
   return sets.union(*imports)
 
 def _go_proto_aspect_impl(target, ctx):
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -49,10 +49,7 @@ def _go_proto_aspect_impl(target, ctx):
 
 _go_proto_aspect = aspect(
     _go_proto_aspect_impl,
-    attr_aspects = [
-        "deps",
-        "embed",
-    ],
+    attr_aspects = ["deps"],
 )
 
 def _proto_library_to_source(go, attr, source, merge):
--- a/tests/core/README.rst
+++ b/tests/core/README.rst
@@ -9,6 +9,7 @@ Contents
 .. Child list start
 
 * `Cross compilation <cross/README.rst>`_
+* `Basic go_proto_library functionality <go_proto_library/README.rst>`_
 * `Import maps <importmap/README.rst>`_
 * `Basic go_test functionality <go_test/README.rst>`_
 * `go_proto_library importmap <go_proto_library_importmap/README.rst>`_
--- a/tests/core/go_proto_library/BUILD.bazel
+++ b/tests/core/go_proto_library/BUILD.bazel
@@ -1,0 +1,29 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
+
+proto_library(
+    name = "embed_proto",
+    srcs = ["embed.proto"],
+)
+
+go_proto_library(
+    name = "embed_go_proto",
+    proto = ":embed_proto",
+    importpath = "github.com/bazelbuild/rules_go/tests/core/go_proto_library/foo",
+    embed = [":extra_lib"],
+)
+
+go_library(
+    name = "extra_lib",
+    srcs = ["extra.go"],
+    importpath = "github.com/bazelbuild/rules_go/tests/core/go_proto_library/foo",
+)
+
+go_test(
+    name = "embed_test",
+    srcs = ["embed_test.go"],
+    deps = [
+        ":embed_go_proto",
+        "@com_github_golang_protobuf//proto:go_default_library",
+    ],
+)
--- a/tests/core/go_proto_library/README.rst
+++ b/tests/core/go_proto_library/README.rst
@@ -1,0 +1,13 @@
+Basic go_proto_library functionality
+====================================
+
+.. _go_proto_library: /proto/core.rst#_go_proto_library
+
+Tests to ensure the basic features of `go_proto_library`_ are working.
+
+.. contents::
+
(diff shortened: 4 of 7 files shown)


kellycampbell:
I signed it.

jayconrod:
Thanks for submitting this! I think the intent of this code was to propagate the aspect on both `deps` and `embed` attributes, not just `deps`. I'll submit a separate PR to fix this problem.

kellycampbell:
Ok. I have an example repo where I encountered the problem here:
https://github.com/kellycampbell/grpc_examples/blob/master/cats_n_pups/protos/BUILD

On Mar 19, 2018 12:26 PM, "Jay Conrod" <notifications@github.com> wrote:

Thanks for submitting this! I think the intent of this code was to
propagate the aspect on both deps and embed attributes, not just deps. I'll
submit a separate PR to fix this problem.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/bazelbuild/rules_go/pull/1388#issuecomment-374274084>,
or mute the thread

jayconrod:
I was wrong, I don't think it actually makes sense for a `go_proto_library` to embed another `go_proto_library`.

If multiple `go_proto_library` rules compile different protos in the same package, there can be self-imports, which we'll reject. They should be combined into a single `go_proto_library` or embedded into a single `go_library`.

If multiple `go_proto_library` rules compile the same proto, as is the case in your example, they should be combined into a single `go_proto_library` rule with multiple entries in the `compilers` attribute or embedded into a single `go_library`.

cc @ianthehat to confirm

jayconrod:
I added a test to your original PR and also stopped the aspect propagating on `embed`.

jayconrod:
Force submitting; test failure is due to timeout in a slow test which is working on other configurations. Probably due to contention on Travis CI macOS cluster.

kellycampbell:
> If multiple go_proto_library rules compile different protos in the same package, there can be self-imports, which we'll reject. They should be combined into a single go_proto_library or embedded into a single go_library.

I definitely ran into this case while trying out different build rules. I also ran into a case where there was an empty import in the .go, e.g. 

`import _ ""`

This was difficult to debug and I ended up adding extra comments to the go generator to figure it out, but ultimately it was just due to not specifying an `importpath`. So there could be some improvements here.

> If multiple go_proto_library rules compile the same proto, as is the case in your example, they should be combined into a single go_proto_library rule with multiple entries in the compilers attribute or embedded into a single go_library.

What about the case of wanting the first go_proto_library (compiler = go_grpc) for e.g. implementing the backend without wanting the 2nd one (compiler = go_grpc_gateway) since the gateway code is unrelated and not needed for the backend? 

FYI, the updated rules_go works fine as-is on my example with two go_proto_library's and embedding one in the other.

Thanks!

jayconrod:
> What about the case of wanting the first go_proto_library (compiler = go_grpc) for e.g. implementing the backend without wanting the 2nd one (compiler = go_grpc_gateway) since the gateway code is unrelated and not needed for the backend?

The best way to do that is to have two `go_proto_library` rules with different `compilers`, and a `go_library` that embed them both. The `go_library` needs to have a matching `importpath`, but it doesn't need any of its own `srcs` or `deps`.

kellycampbell:
That doesn't work in this case because the `go_grpc_gateway` generated code depends on the prior `go_grpc` code and they have to be in the same importpath. `deps` didn't work for that so `embed` seemed like the only way.

jayconrod:
Ah ok, so in your example `catpic_proto_go` and `catpic_proto_gw` need to be separate targets (different things depend on them), and `catpic_proto_gw` needs to include `catpic_proto_go`. 

`embed` does seem like the best way to do this.

kellycampbell:
Found another way that pattern of targets is required. The `go_grpc_gateway` plugin doesn't create output files for src .proto files which don't include any services, so bazel errors out on that case. 

The python protoc plugin just creates empty _pb2_grpc.py for this situation. I'm pretty sure an empty go file is not allowed by the go compiler.

Is there a possibility of adding some kind of way to specify which outputs are expected/not expected to go_proto_compiler ?

jayconrod:
Bazel requires all output files to be specified during analysis (without reading the files). Maybe the best way to handle that is for protoc.go (our wrapper binary that invokes protoc) to create .go files that the plugin did not create? The Go compiler won't accept empty files, but we could create files with `// +build ignore` or something.

kellycampbell:
Yup, I know about the blaze phases and such. I worked on the build tools team in NYC around 9 years ago. I didn't know about the build ignore pragma. Is that new, or just something bazel looks for?

I can update grpc_gateway to output that empty file if that makes more sense?

jayconrod:
Ah, small world.

Comments that start with `// +build` before the package declaration are recognized both by Bazel and "go build". They're usually used to specify which goos / goarch a file should be built for, but custom tags are allowed, too. `ignore` is normally used to prevent a file from being compiled altogether.

Now that I look closer though, there still needs to be a package declaration or a parse error will be reported (this is why the compiler doesn't accept empty files. But we can potentially copy the package declaration from another source file.

I think it makes sense for `go_proto_library` to generate ignored files automatically if the plugins don't do it. We assumed plugins generated one .go file per .proto source file, but I guess that's not true for all plugins.

ianthehat:
There should never be any reason to embed a go_proto_library into another go_proto_library, in fact it should not be possible to embed anything into a go_proto_library.
If you need to add multiple go_proto_libraries together (which itself should be very rare) then you do it by embedding into a go_library.
In general for a single package you should process all the proto files in a single rule, and run all the compilers at once on the same rule.

kellycampbell:
@ianthehat the use case here is for the `grpc_gateway` protoc plugin. It generates a `.pb.gw.go` file in the same package as the `go_grpc` plugin generated `.pb.go` file which uses the services code in the .pb.go.

ianthehat:
Yes, but can't it be done in the same go_proto_library rule, that's the use case we were thinking of when I made it accept multiple compilers

ianthehat:
Okay, so try it that way and let us know if there are still issues, it's an important use case and we would like to make sure it is clean and easy.

kellycampbell:
I'll try it out. We're trying to do something similar to the googleapis protos ( https://github.com/googleapis/googleapis) or envoy v2 api protos (https://github.com/envoyproxy/data-plane-api) using the http annotations to generate a REST api from gRPC services.

I was surprised googleapis didn't have full BUILD files throughout.

PR #1393, opened 2018-03-20
Add go_binary out attribute for custom file names

### LINKED ISSUE #1239: go_binary output paths contain a spurious os_arch component

jayconrod:
go_binary now has an "out" attribute which allows users to set the
output filename for the generated executable. When set, go_binary will
write this file without mode-specific directory prefixes, without
linkmode-specific prefixes, and without platform-specific extensions.

Fixes #1239


jmillikin-stripe:
How is this different from the `basename` attribute? Does it replace it?

jayconrod:
`basename` lets you use a custom stem instead of the rule name, but the binary will still be written into a mode-specific prefix directory, and the final file name may have prefixes and extensions like `lib` or `.exe`.

`out` lets you completely specify the output file name.

I don't think this completely replaces `basename`, but I probably wouldn't add `basename` if we already had `out`.

jmillikin-stripe:
It seems pretty awkward to have executables on Windows written without `.exe`, unless each user of this new param sets up appropriate `select` calls for all the platforms that might need a suffix on binaries. Are there any examples of when a user might want to create a binary without a suffix?

I guess I don't understand why any value of `out` other than `attr.name` would be useful. Maybe it'd be good to add an example or two of how this would be used?

jayconrod:
So just to throw out a random example, files in Linux /lib directories are supposed to be named with a version after the extension, e.g., `libfoo.so.1.2.3`. `basename = "foo"` would just give us `libfoo.so` without the version. If `out` added an extension, you'd get `libfoo.so.1.2.3.so`.

`out` is a general purpose escape hatch for when our naming scheme doesn't work. I don't expect it to be commonly used, but when it's needed, we should let users do exactly what they want with it.

jmillikin-stripe:
Isn't the version usually part of the pre-suffix name? Looking at my `/lib` and `/usr/lib` I see names like `libc-2.23.so`, `libdb-5.3.so `, `libelf-0.165.so` that have symlinks pointing at them with names like `libc.so.6`.

ianthehat:
basename was meant as an escape hatch, but it's not really sufficient for any useful case, we should probably deprecate it's use.
out lets you specify anything you like, if it still did anything automatically that would not be true, and we would be back having this conversation again with someone else that wanted different behaviour.

Issue #1309, opened 2018-02-07
Define go_proto_compiler with optional output files


jayconrod:
Original post: [How to define a go_proto_compiler with optional output file](https://groups.google.com/forum/#!topic/bazel-go-discuss/Hq2XEUDi7DA) on bazel-go-discuss.

Summary: some proto plugins may or may not produce output files depending on the contents of their inputs. We should ensure these files are created, possibly just by creating files with `// +build ignore`.

PR #1394, opened 2018-03-20
protoc: create trivial files for missing outputs

jayconrod:
The grpc_gateway plugin only creates output files for .proto files
that includes service definitions. We still need to declare output .go
files, since we don't know which .proto files have services during
analysis, but we don't need to report errors.

Related #1388


kellycampbell:
How will this tell the difference between a file that the plugin doesn't generate vs. a real failure where the file didn't get generated?

jayconrod:
I'm assuming protoc will exit with a non-zero status if there's a failure in a plugin. Our protoc wrapper detects that and will also exit with a non-zero status.

kaoet:
Does it fix https://github.com/bazelbuild/rules_go/issues/1309 ?

jayconrod:
@kaoet Yes I think that's the same issue.

PR #1409, opened 2018-03-26
Rewrite go_path rule

### LINKED ISSUE #1329: go_path rule does not allow go_test dependencies
I have a small test project that demonstrates that [HERE](https://github.com/iamneal/bazel_go_test).

if you have a bazel build file like this:
```bazel
load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test", "go_prefix", "go_path")
go_path(
    name = "path",
    deps = [
        ":go_default_library",
        ":go_default_xtest",
    ],
)

go_library(
    name = "go_default_library",
    srcs = ["main.go"],
    importpath = "bazel_go_test",
    deps = [
        "//lib:go_default_library",
        "@com_github_burntsushi_toml//:go_default_library",
        "@com_github_sirupsen_logrus//:go_default_library",
        "@org_golang_google_grpc//:go_default_library",
        "@org_golang_x_net//context:go_default_library",
    ],
)
go_test(
    name = "go_default_xtest",
    srcs = ["main_test.go"],
    data = ["//:Gopkg.lock"],
    importpath = "bazel_go_test_test",
    deps = [":go_default_library"],
)
```
and you run: ```bazel build //:path```

you will see this error:
```sh
ERROR: /Users/<user>/go/src/github.com/iamneal/bazel_go_test/BUILD.bazel:17:12: in deps attribute o
f go_path rule //:path: '//:go_default_xtest' does not have mandatory providers: 'GoLibrary'
ERROR: /Users/<user>/go/src/github.com/iamneal/bazel_go_test/BUILD.bazel:15:1: in go_path rule //:p
ath: non-test target '//:path' depends on testonly target '//:go_default_xtest' and doesn't have testonly attribute set
ERROR: Analysis of target '//:path' failed; build aborted: Analysis of target '//:path' failed; build a
borted
INFO: Elapsed time: 0.241s
FAILED: Build did NOT complete successfully (1 packages loaded)
```
It would be nice if we could use the go_path rule for any go_*  dependency, since test files can import packages that libraries don't need.

my bazel version output is this:
```sh
Build label: 0.10.0-homebrew
Build time: Wed Jan 10 00:01:03 +50057 (1517480006463)
Build timestamp: 1517480006463
Build timestamp as int: 1517480006463
```
my go rules were received from this rule in my WORKSPACE file:
```bazel
http_archive(
    name = "io_bazel_rules_go",
    sha256 = "4d8d6244320dd751590f9100cf39fd7a4b75cd901e1f3ffdfd6f048328883695",
    url = "https://github.com/bazelbuild/rules_go/releases/download/0.9.0/rules_go-0.9.0.tar.gz",
)
```

### LINKED ISSUE #1370: go_path: still produces files which are no longer needed
Suppose we have a `go_path` rule that includes a program `//cmd` and a library which isn't otherwise depended on, `//baz:go_default_library`:

```
go_path(
    name = "go_path",
    deps = [
        "//cmd",
        "//baz:go_default_library",
    ],
)
```

When we build this, `baz` will be in the tree.

If we remove `baz` and build, `baz` will still be in the tree. We have to `bazel clean` and rebuild for it to go away.

I don't think `go_path` is actually placing files there. This might be because we're producing individual files instead of directory outputs.

### LINKED ISSUE #1371: go_path: include cgo sources
`go_path` does not pick up C / C++ source files stored alongside cgo packages.

This may be difficult to implement without the C/C++ sandwich. Currently, these sources are segregated into a separate `cc_library` rule by `go_library_macro`.

### LINKED ISSUE #1372: go_path: support vendoring
go_path should write packages into a vendor directory if they were originally defined in a vendor directory.

This should be controlled with a new attribute like `vendor = True`. Probably off by default.

rules_go generally doesn't recognize vendoring at all. Gazelle knows about vendor directories in order to set `importpath` and resolve dependencies, but directory names aren't communicated directly to `go_path`. Maybe examining package names in labels will be good enough though.

### LINKED ISSUE #1373: go_path: cgo generated files are being included
Steps to reproduce:

* Start with an empty repository with Gazelle boilerplate
* Vendor `github.com/coreos/pkg/dlopen` (for example). This is a cgo library with no dependencies.
* `bazel run //:gazelle`
* Create a `go_path` rule like below:

```
load("@io_bazel_rules_go//go:def.bzl", "go_path")

go_path(
    name = "go_path",
    deps = ["//vendor/github.com/coreos/pkg/dlopen:go_default_library"],
)
```

* Observe that the built `go_path` contains `dlopen.cgo1.go`, `_cgo_import.go`, and `_cgo_gotypes.go`. We want the original source files, not generated files.

jayconrod:
* Implementation moved to //go/tools/builders:go_path
* Added "archive" mode, which produces a zip file.
* "copy" mode now produces a tree artifact.
* go_path accepts a "data" attribute, for additional files.
* Added documentation for the go_path rule and GoPath provider.

Fixes #1370
Fixes #1371
Fixes #1372
Fixes #1329
Fixes #1373


go/tools/builders/go_path.go
```diff
@@ -0,0 +1,200 @@
...
+		w, err := outZip.Create(entry.Dst)
+		if err != nil {
+			srcFile.Close()
+			return err
+		}
+		if _, err := io.Copy(w, srcFile); err != nil {
+			srcFile.Close()
+			return err
+		}
+		srcFile.Close()
```
ianthehat: You are ignoring the close errors on the normal path (the ones above are already in the error path)

jayconrod: Done. Is it important to check for errors when closing a file that's being read? I thought it was only important when writing files.

PR #1413, opened 2018-03-28
use go_context in some go_rule docs

jmhodges:


go/toolchains.rst
```diff
@@ -126,8 +128,7 @@ And then in the rule body, you need to get the toolchain itself and use it's act
 .. code:: bzl
 
   def _my_rule_impl(ctx):
-    go_toolchain = ctx.toolchains["@io_bazel_rules_go//go:toolchain"]
-    srcs, vars = go_toolchain.actions.cover(ctx, go_toolchain, ctx.files.srcs)
+    go_ctx = go_context(ctx)
```
jayconrod: Nit: we call this variable `go` in all our rules. Best to do the same here.

PR #1423, opened 2018-03-31
Allow gazelle extra args to be passed through in go_repository

iancmcc:
I've needed this in order to pass the `known_import` arg to gazelle, although it's a generally useful pass-through. 

I could also see wanting to declare each possible argument individually, but since gazelle was moved into its own repo that seemed like too tight a coupling.


iancmcc:
I signed it!

jayconrod:
Thanks for submitting this! I've moved this over to the Gazelle repo as bazelbuild/bazel-gazelle#173 and added documentation. The primary implementation of `go_repository` is there. We're planning to deprecate the `go_repository` in rules_go after the next release so that we can break the dependence of rules_go on Gazelle.

PR #1424, opened 2018-04-02
Slight clarification in strict dependencies error message

jayconrod:


jayconrod:
I added another commit to print the known dependencies, too, so it's easier to tell what the mismatch is. PTAL.

jayconrod:
It would be nice to add a "did you mean" message after each missing import instead of listing all the known imports using edit distance. I'm not aware of a simple way to do that without going beyond the standard library.

PR #1426, opened 2018-04-03
Add shebang to .pb.go.run generated file.

AustinSchuh:
Without this, not all interpreters know how to run the .run files.


ianthehat:
I would strongly suggest you switch to using the new proto library rules rather than fixing these ones, they have been deprecated for quite a while, and will be deleted really soon.
If you really need to fix this rule, the right way is to change it to using [ctx.actions.run_shell](https://docs.bazel.build/versions/master/skylark/lib/actions.html#run_shell) which already knows how to do it right without the shebang line. and will work correctly on mac and windows as well.

jayconrod:
+1 to what Ian said. I expect we'll remove the legacy `go_proto_library.bzl` in the next release (about a month from now).

If you're not able to move to the next `go_proto_library`, please let us know and we'll try to fill in any missing features.

PR #1434, opened 2018-04-09
Optionnally add compiled binaries in virtual GOPATH

steeve:
It makes all programs that use `go/importer` work properly. I'm not too much satisfied with the way the aspect is propagated, however.

### DIFF
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -32,10 +32,25 @@ load(
     "@io_bazel_rules_go//go/private:rules/rule.bzl",
     "go_rule",
 )
+load(
+    "@io_bazel_rules_go//go/private:rules/aspect.bzl",
+    "go_archive_aspect",
+)
+load(
+    "@io_bazel_rules_go//go/platform:list.bzl",
+    "GOOS",
+    "GOARCH",
+)
+load(
+    "@io_bazel_rules_go//go/private:mode.bzl",
+    "LINKMODE_NORMAL",
+    "LINKMODES",
+)
 
 def _go_path_impl(ctx):
   # Gather all archives. Note that there may be multiple packages with the same
   # importpath (e.g., multiple vendored libraries, internal tests).
+  go = go_context(ctx)
   direct_archives = []
   transitive_archives = []
   for dep in ctx.attr.deps:
@@ -51,11 +66,14 @@ def _go_path_impl(ctx):
     if importpath == "":
       continue  # synthetic archive or inferred location
     out_prefix = "src/" + pkgpath
+    pkg_out_prefix = "pkg/" + go.mode.goos + "_" + go.mode.goarch + "/"
     pkg = struct(
         importpath = importpath,
         dir = out_prefix,
+        pkgdir = pkg_out_prefix + pkgpath,
         srcs = as_list(archive.orig_srcs),
         data = as_list(archive.data_files),
+        file = archive.file,
     )
     if pkgpath in pkg_map:
       _merge_pkg(pkg_map[pkgpath], pkg)
@@ -72,6 +90,12 @@ def _go_path_impl(ctx):
           dst = pkg.dir + "/" + f.basename,
       ))
       inputs.append(f)
+    if ctx.attr.with_binaries:
+      manifest_entries.append(struct(
+          src = pkg.file.path,
+          dst = pkg.pkgdir + "." + pkg.file.extension,
+      ))
+      inputs.append(pkg.file)
   for f in ctx.files.data:
     manifest_entries.append(struct(
         src = f.path,
@@ -132,10 +156,13 @@ def _go_path_impl(ctx):
       ),
   ]
 
-go_path = rule(
+go_path = go_rule(
     _go_path_impl,
     attrs = {
-        "deps": attr.label_list(providers = [GoArchive]),
+        "deps": attr.label_list(
+            providers = [GoArchive],
+            aspects = [go_archive_aspect],
+        ),
         "data": attr.label_list(
             allow_files = True,
             cfg = "data",
@@ -148,6 +175,48 @@ go_path = rule(
                 "link",
             ],
         ),
+        "pure": attr.string(
+            values = [
+                "on",
+                "off",
+                "auto",
+            ],
+            default = "auto",
+        ),
+        "static": attr.string(
+            values = [
+                "on",
+                "off",
+                "auto",
+            ],
+            default = "auto",
+        ),
+        "race": attr.string(
+            values = [
+                "on",
+                "off",
+                "auto",
+            ],
+            default = "auto",
+        ),
+        "msan": attr.string(
+            values = [
+                "on",
+                "off",
+                "auto",
+            ],
+            default = "auto",
+        ),
+        "goos": attr.string(
+            values = GOOS.keys() + ["auto"],
+            default = "auto",
+        ),
+        "goarch": attr.string(
+            values = GOARCH.keys() + ["auto"],
+            default = "auto",
+        ),
+        "linkmode": attr.string(values=LINKMODES, default=LINKMODE_NORMAL),
+        "with_binaries": attr.bool(default = False),
         "_go_path": attr.label(
             default = "@io_bazel_rules_go//go/tools/builders:go_path",
             executable = True,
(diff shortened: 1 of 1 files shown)


jayconrod:
Sorry for not responding to this earlier.

I'm hesitant to accept this as-is because I don't think the mode attributes belong in `go_path`. These already cause a lot of problems: `goos` and `goarch` interact very poorly with `select` expressions and platform-specific dependencies.

I'm open to including compiled binaries, and possibly even archives in `go_path` though. A simpler interface may be to add `include_bin` and `include_pkg` Boolean attributes, similar to `include_data`.

I wonder if this is necessary though for `go/importer`? We are working on a static analysis framework that implements `types.Importer` with information gather from the command line. See [checker_main.go](https://github.com/bazelbuild/rules_go/blob/1ced0fe0a52b425c4d2a38ef74c3c0e9d61c6613/go/tools/builders/checker_main.go#L171) (note this is on a feature branch, not on master yet). `go_path` is lossy, so for tools, it's better to adapt to Bazel instead of using `go_path` whenever possible.

steeve:
Hey, sorry for the long wait, I will rebase it.

steeve:
I have rebased and updated the PR.

steeve:
Also, we found that this PR would enable the generated GoPath to be compatible with `gocode` and the like, enabling proper code completion to work.

steeve:
Please do not merge, as this version somehow makes the action execute twice when run with `include_binaries`.

go/private/tools/path.bzl
```diff
@@ -144,10 +169,13 @@ def _go_path_impl(ctx):
...
     ]
 
-go_path = rule(
+go_path = go_rule(
     _go_path_impl,
     attrs = {
-        "deps": attr.label_list(providers = [GoArchive]),
+        "deps": attr.label_list(
+            providers = [GoArchive],
+            aspects = [go_archive_aspect],
```
jayconrod: I think `go_path` should accept archives in multiple modes. Ideally, you'd be able to point to multiple binaries and libraries with their own modes in `deps`, and `go_path` would include them all in the appropriate directories. That may not actually work at this point (the same archive in multiple modes will probably cause conflicts right now), and that's fine for the PR, but I don't want to commit to a single mode in this API.

In order the make this work, I think you'll need to copy `go.mode` into `GoArchive`. It shouldn't need to be in `GoArchiveData`, since anything an archive depends on will be in the same mode. You won't need an aspect or any mode attributes in this rule.

go/private/tools/path.bzl
```diff
@@ -161,6 +189,41 @@ go_path = rule(
             ],
         ),
         "include_data": attr.bool(default = True),
+        "include_binaries": attr.bool(default = False),
```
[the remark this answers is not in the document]
steeve: Sounds good

go/private/tools/path.bzl
steeve: How do you feel this working file layout wise for two libraries in different mode?

jayconrod: If you have the same library in multiple modes, then there should be multiple elements in `archives` for that library.

For the sources and data, I don't think we'll actually get conflicts because `_add_manifest_entry` checks whether file paths are different before reporting an error.

For the archives, .a files built in different modes will go in different `pkg` directories. For example, you might have `pkg/linux_amd64/foo.a` and `pkg/linux_amd64_race/foo.a`.

steeve: That won't work since pkg has a pretty rigid layout and expects GOOS_GOARCH, and likely most tools that depend on it such as gocode.

Also, when setting up code completion we run the target once and once in a while refresh it to update it. Then we point our editor's go_path to the go_path inside bazel-out.

steeve: Unfortunately most ides (vscode in our case) will boot gocode with a predefined goos and goarch but I don't think there more to it.

Also, I think it would be simpler to have multiple targets, should one need it?

jayconrod: > Unfortunately most ides (vscode in our case) will boot gocode with a predefined goos and goarch but I don't think there more to it.

What problems does that cause? This would cause issues if you are building a `go_path` for one target but gocode expects another. If `go_path` depends on a `go_binary` with explicit `goos` and `goarch` attributes, that could cause issues though.

> Also, I think it would be simpler to have multiple targets, should one need it?

You mean multiple directories in `pkg`? I agree, we should support this as much as possible. However, some of the mode attributes (`static`, `pure`, `debug`, `stripped`) don't correspond to any conventional install suffix. So when we're generating the install suffix under pkg, we can either leave them out (lossy), or we can include them but some tools won't expect those parts of the suffix.

steeve: True...
Well, since this rule is precisely for those tools, the lossy option looks fine to me, as they don't expect a suffix.
Is there a case for bazel tools to use the generated gopath save for rules wrapping some go tools?
In our case we are wrapping `gomobile bind` with success.

steeve:
I've updated the PR, changing to `include_pkg`

go/private/tools/path.bzl
jayconrod: Sorry I meant to answer this earlier. `go_path` has a couple different purposes.

One is wrapping tools that expect a GOPATH. We have an ongoing project to build a workspace abstraction layer (not Bazel-specific) that will address these kinds of problems for tools in general. So at some point we won't need `go_path` for that.

The other important use case is building source archives. For example, if you have a Bazel project with a non-standard layout or a lot of generated code, you could package everything up with a `go_path` and drop it in a vendor directory in another project.

steeve:
You mean by having `go_binary` in deps?

jayconrod:
Yep, `go_test` works, too. That's the most natural way to include test sources.

steeve:
Following the discussion, I have removed the mode attributes

go/private/tools/path.bzl
steeve: Going on the subject, I'm not sure I'm getting what to change here

steeve:
Oh wow. Alright I'm going to test it but I'm a bit worried about the `installsuffix` part

go/private/tools/path.bzl
```diff
@@ -70,6 +72,13 @@ def _go_path_impl(ctx):
         for f in pkg.srcs:
             dst = pkg.dir + "/" + f.basename
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
+    if ctx.attr.include_pkg:
+        for pkg in pkg_map.values():
+            for mode, f in pkg.pkgs.items():
+                # TODO(jayconrod): include other mode attributes, e.g., race.
```
steeve: it should be possible to retain the default mode (no install suffix) for tools that don't understand it

jayconrod:
I just pushed a commit which adds support for multiple copies of archives in different modes. Feel free to change / roll back if it's not what you want.

* `GoArchive` now includes a `mode` field. This is only in `GoArchive`, not `GoArchiveData`, so we'll only have the mode for direct dependencies. Transitive dependencies should always have the same mode.
* In `go_path`, we build a `depset` for each distinct mode, rather than one `depset` for everything.
* When we enumerate and merge packages, we keep track of archive files in a `pkgs` dict which maps mode to archive file. We may see multiple copies of a package in different modes, but these should merge cleanly.
* Added tests.

go/private/tools/path.bzl
jayconrod: I'm not sure what you mean. We always need something in `pkg/<something>/path/to/lib.a`. By default, that will be `<goos>_<goarch>` for the target platform, which I think is what most tools will look for.

`installsuffix` might not actually be the right term for that string. Not sure what to call it.

steeve: Ah, I thought you wanted to open the door for a `GOOS_GOARCH_MODE` kind of thing

jayconrod: Ideally yes, but I think we should leave that as an enhancement for the future, unless you need it now.

But in that scenario, by default, I think it should match whatever go build would have put in there by default (without an explicit `-installsuffix` anywhere). So no stripped, debug, pure... but probably still `race` or `shared`.

Not sure whether that would work for everything. Need to think about it more.

PR #1453, opened 2018-04-18
go_path: improve handling of data files

### LINKED ISSUE #1449: data deps not handled correctly in go_path
This is how data deps is handled:

https://github.com/bazelbuild/rules_go/blob/102eab428f4bfda4833d58fcc1cd2246dd1dbb68/go/private/tools/path.bzl#L70-L72

This means if a package has two data deps that have the same basename, such as `//data1:data.json` and `//data2:data.json`, go_path will fail to build.

jayconrod:
* Data files referenced by go rules are placed in their package
  directories with their basenames UNLESS they have testdata in their
  path, in which case, the part of the path below testdata/ is
  preserved.
* New attribute: include_data. Can be set to False to prevent data
  files from appearing in the output. Will be useful for tools when
  data files don't matter or cause problems.

Fixes #1449

### DIFF
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -67,9 +67,19 @@ def _go_path_impl(ctx):
   manifest_entries = []
   manifest_entry_map = {}
   for pkg in pkg_map.values():
-    for f in pkg.srcs + pkg.data:
-      _add_manifest_entry(manifest_entries, manifest_entry_map, inputs,
-                          f, pkg.dir + "/" + f.basename)
+    for f in pkg.srcs:
+      dst = pkg.dir + "/" + f.basename
+      _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
+  if ctx.attr.include_data:
+    for pkg in pkg_map.values():
+      for f in pkg.data:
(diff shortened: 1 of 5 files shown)


go/private/tools/path.bzl
```diff
@@ -67,9 +67,19 @@ def _go_path_impl(ctx):
...
-    for f in pkg.srcs + pkg.data:
-      _add_manifest_entry(manifest_entries, manifest_entry_map, inputs,
-                          f, pkg.dir + "/" + f.basename)
+    for f in pkg.srcs:
+      dst = pkg.dir + "/" + f.basename
+      _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
+  if ctx.attr.include_data:
+    for pkg in pkg_map.values():
+      for f in pkg.data:
+        parts = f.path.split("/")
```
ianthehat: wouldn't it be easier to just split on /testdata/

jayconrod: In this case yes. As a habit though, when I'm looking for a specific path component, I prefer to split the path and look at the parts. It avoids corner cases with the first and last components.

PR #1440, opened 2018-04-10
Make the stdlib reproducible by silencing the build id

steeve:
When go install'ing the stdlib, leverage `-toolexec` to remove the `-buildid` flag from `compile` invocations.

To run the test:
```
$ ./tests/core/stdlib/test_reproduce.sh
```

Note that this will expunge the current bazel workspace.

At the moment the everything is reproducible, save for `runtime/cgo` on Linux only. I'm investigating.

Looking at the hexdump diff, it looks like the sandbox path is built into the archive:
```
< 000076e0  2f 62 61 7a 65 6c 2d 73  61 6e 64 62 6f 78 2f 37  |/bazel-sandbox/7|
< 000076f0  33 37 34 35 37 31 39 37  32 36 32 34 30 33 31 36  |3745719726240316|
< 00007700  32 32 2f 65 78 65 63 72  6f 6f 74 2f 69 6f 5f 62  |22/execroot/io_b|
< 00007710  61 7a 65 6c 5f 72 75 6c  65 73 5f 67 6f 2f 62 61  |azel_rules_go/ba|
< 00007720  7a 65 6c 2d 6f 75 74 2f  6b 38 2d 66 61 73 74 62  |zel-out/k8-fastb|
< 00007730  75 69 6c 64 2f 62 69 6e  2f 65 78 74 65 72 6e 61  |uild/bin/externa|
< 00007740  6c 2f 69 6f 5f 62 61 7a  65 6c 5f 72 75 6c 65 73  |l/io_bazel_rules|
< 00007750  5f 67 6f 2f 6c 69 6e 75  78 5f 61 6d 64 36 34 5f  |_go/linux_amd64_|
< 00007760  73 74 72 69 70 70 65 64  2f 73 74 64 6c 69 62 7e  |stripped/stdlib~|
< 00007770  2f 73 72 63 2f 72 75 6e  74 69 6d 65 2f 63 67 6f  |/src/runtime/cgo|
< 00007780  00 47 4e 55 20 41 53 20  32 2e 32 38 00 01 80 01  |.GNU AS 2.28....|
< 00007790  11 00 10 06 11 01 12 01  03 08 1b 08 25 08 13 05  |............%...|
< 000077a0  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
< 000077b0  00 00 2c 00 00 00 02 00  00 00 00 00 08 00 00 00  |..,.............|
[...]

Issue #1391, opened 2018-03-19
Build id seems to prevent stdlib from being reproducible


jayconrod:
Original comment by @steeve https://github.com/bazelbuild/rules_go/issues/1357#issuecomment-374068333

steeve:
Relevant source code seems to live at:
https://github.com/golang/go/blob/master/src/cmd/go/internal/work/buildid.go#L23-L31

Although one quick fix would be to overwrite the build id in the objects files, kind of like what we do with `.a` archives.

jayconrod:
If the build id is a hash of the inputs, it should be reproducible. I suspect that it's hashing absolute paths that vary across different runs. If that's the case, we can try to build it at a deterministic location.

Modifying the .a files may be difficult to support. They don't have a stable format across Go releases.

steeve:
Adding https://go-review.googlesource.com/c/go/+/73291

steeve:
And https://github.com/golang/go/issues/22491

steeve:
Reading buildid.go, it seems the build is divided into several parts by a `/`. Indeed, when looking at the diff, for instance:
```diff
37,38c37,38
< 00000240  49 69 78 62 6e 54 46 35  61 53 30 4e 61 59 50 45  |IixbnTF5aS0NaYPE|
< 00000250  49 53 78 73 2f 73 71 38  38 4e 51 53 47 58 6a 6b  |ISxs/sq88NQSGXjk|
---
> 00000240  58 59 7a 63 77 6e 63 4e  4f 4b 4e 59 7a 35 57 53  |XYzcwncNOKNYz5WS|
> 00000250  44 70 43 44 2f 73 71 38  38 4e 51 53 47 58 6a 6b  |DpCD/sq88NQSGXjk|
```

We can see that the second part is the same, only the first changes.

steeve:
My guess is that the compile action id is calculated at https://github.com/golang/go/blob/ad0ebc3994fc7f74434d922b80401e680162c7d1/src/cmd/go/internal/work/exec.go#L177

steeve:
If we could print the value of the input buffer at https://github.com/golang/go/blob/ad0ebc3994fc7f74434d922b80401e680162c7d1/src/cmd/go/internal/work/exec.go#L295 we could see what changes, I guess.

steeve:
[...]

PR #1454, opened 2018-04-18
Test that the stdlib rule does not insert build ids in archives

jayconrod:
Related #1440

### DIFF
--- a/tests/core/stdlib/buildid_test.go
+++ b/tests/core/stdlib/buildid_test.go
@@ -27,10 +27,19 @@ import (
 )
 
 func TestEmptyBuildID(t *testing.T) {
-	// Locate fmt.a (any .a file in the stdlib will do) and the buildid tool.
-	// The path may vary depending on the platform and architecture, so we
-	// just do a search.
-	var fmtPath, buildidPath string
+	// Locate the buildid tool and several archive files to check.
+	//   fmt.a - pure go
+	//   crypto/aes.a - contains assembly
+	//   runtime/cgo.a - contains cgo
+	// The path may vary depending on platform and architecture, so just
+	// do a search.
+	var buildidPath string
+	pkgPaths := map[string]string{
+		"fmt.a": "",
+		"aes.a": "",
+		"cgo.a": "",
+	}
+	n := len(pkgPaths)
 	done := errors.New("done")
 	var visit filepath.WalkFunc
 	visit = func(path string, info os.FileInfo, err error) error {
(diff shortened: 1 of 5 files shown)


tests/core/stdlib/stdlib_files.bzl
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@io_bazel_rules_go//go/private:providers.bzl", "GoSource")
+
+def _stdlib_files_impl(ctx):
+  files = ctx.attr._stdlib[GoSource].stdlib.files
+  runfiles = ctx.runfiles(files = files)
```
steeve: ah, that's what I missed.
wouldn't it be simpler to add it to the `//:stdlib` rule ?

tests/core/stdlib/buildid_test.go
```diff
@@ -1,9 +1,80 @@
...
+			return err
+		}
+		if (info.Mode() & os.ModeType) == os.ModeSymlink {
+			path, err = filepath.EvalSymlinks(path)
+			if err != nil {
+				return err
+			}
+			return filepath.Walk(path, visit)
+		}
+		if filepath.Base(path) == "fmt.a" {
```
steeve: unfortunately `fmt` is not a good candidate because pure go packages where producing the same build ids.

that said, we should check 3 types of packages:
- the ones pure go (`fmt` looks like one)
- the ones with asm (`crypto/*` are in that case I think)
- the ones with cgo (`os/user` comes to mind)

You can find the list at https://github.com/bazelbuild/rules_go/issues/1391#issuecomment-377991295, perhaps check `fmt` and the ones from the list ?

steeve: it's likely it'll be the same but for the sake of correctness and future proofing...

steeve: also, we should probably check for the sandbox path by binary search for `bazel-sandbox` for cgo packages because of https://github.com/bazelbuild/rules_go/pull/1440/commits/f4201a501c6ca245054035fc54f8037715f241c0

but that may out of the scope of this PR though

jayconrod: Very good advice. I've expanded the test to check `fmt.a`, `crypto/aes.a`, and `runtime/cgo.a`. That should cover those cases.

About searching for `bazel-sandbox`, `tests/legacy/reproducible_binary` checks for common reproducibility problems in compiled binaries, but I wrote it before `bazel-sandbox` appeared it paths, so it should be updated (in another PR).

tests/core/stdlib/stdlib_files.bzl
jayconrod: I'd rather keep this in the `GoStdLib` provider for now, but I don't have a strong opinion about it, since this isn't a public interface. Fewer things to keep in sync though.

Issue #1256, opened 2018-01-19
generate_test_main is given incorrect --package flag when using go_prefix


jmillikin-stripe:
Given the following build file:

```
load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test", "go_prefix")

go_prefix(prefix = "go-testing")

go_library(name = "foo", srcs = ["foo.go"])
go_library(name = "bar", srcs = ["bar.go"])

go_test(
    name = "bar_test",
    srcs = ["bar_test.go"],
    embed = [":bar"],
    deps = [":foo"],
)
```

Doing a `bazel build //:bar_test` runs `generate_test_main --package go-testing/foo_test`, which generates a testmain.go that imports the wrong package as `undertest`.

```
$ grep undertest bazel-out/darwin-fastbuild/bin/darwin_amd64_stripped/bar_test~/testmain.go
	undertest "go-testing/foo_test"
	{"TestStuff", undertest.TestStuff },
```

A bit of testing indicates that the go_test rule is calculating this as `deps[0].package_name + "_test"` which will probably never do the right thing, considering that the library being tested is usually placed in either `embed=` or `library=`.

jmillikin-stripe:
Looking at the source of `_infer_importpath()` in `go/private/context.bzl`, I think this could be fixed by allowing inferred paths from targets listed in `ctx.attr.embed` when building a test library. Currently embedded libs can only be considered if they have an explicit path, and inferred paths in tests can only be found from `deps`.

jmillikin-stripe:
This is working now (at rules_go 0.11.1).

PR #1462, opened 2018-04-24
Refactor coverage support

### LINKED ISSUE #1456: Clean up cgo flags
In particular, see [cgo.bzl:164](https://github.com/bazelbuild/rules_go/blob/1b366507438cfa0f78871e410a3ec9a9f1116c5b/go/private/rules/cgo.bzl#L164).

We're accumulating options from `cdeps` dependencies.

* Include paths may be redundant across libraries, especially those defined in the same directory.
* I don't think we actually need to set link options for cgo code generation. Internally, it runs the compiler on generated code to determine types of symbols referenced in cgo comments, but it shouldn't link that code (we have a separate `cc_binary` rule for that).
* If builders don't need to process these options themselves, it may be better to set`CGO_CPPFLAGS` directly in analysis.

jayconrod:
Previously, we tracked variables added by `go tool cover` in
GoArchive, then registered them all in the generated testmain. This
required tracking a lot of extra information, and it required the test
archive to import all transitive dependencies.

This change introduces a new library, coverdata. Covered packages now
import this library and call coverdata.RegisterFile in an init
function added by the cover action. testmain is much simpler, and
coverage information is removed from GoArchive.

Fixes #1456


jmillikin-stripe:
I've verified this fixes coverage runs, and the resulting coverage data looks reasonable. It seems like something's wrong with instrumentation filter handling, though, because the coverage data for my tests looks like it's the full transitive dependency set (including external dependencies).

I'm also seeing some filenames that don't exist, for example a `client/service_test.go` is getting mapped to `client/service_test/service.go`. I think this is an effect of https://github.com/bazelbuild/rules_go/issues/1256.

jayconrod:
@jmillikin-stripe Filed #1466 and #1467 for the issues you mentioned. I think I know what's happening with both, but I might ping you for more information. Should have a fix for both tomorrow.

By the way, do you want these fixes backported to 0.11.x or 0.10.x? This PR changed some interfaces, so I can't cherry-pick it without modification.

jmillikin-stripe:
It would be nice to get them in a stable release, and I think it would make sense to have in 0.11.1 since .0 was the release that broke coverage.

Issue #1467, opened 2018-04-26
More libraries are covered than are specified with the instrumentation filter


jayconrod:
Mentioned in [this comment](https://github.com/bazelbuild/rules_go/pull/1462#issuecomment-384782057).

> It seems like something's wrong with instrumentation filter handling, though, because the coverage data for my tests looks like it's the full transitive dependency set (including external dependencies).

jmillikin-stripe:
The above merge fixed everything except cgo files, which are still being picked up. It looks like `_cgo_resolve_source` always copies srcs to `source["cover"]`.

With the following patch, coverage data looks correct in all my tests. Not sure if this is the correct approach (is `library.resolve` part of the public API?).
```
diff --git a/go/private/context.bzl b/go/private/context.bzl
index 17d97f4..8366e8e 100644
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -147,7 +151,7 @@ def _library_to_source(go, attr, library, coverage_instrumented):
     x_defs[k] = v
   source["x_defs"] = x_defs
   if library.resolve:
-    library.resolve(go, attr, source, _merge_embed)
+    library.resolve(go, attr, source, _merge_embed, coverage_instrumented)
   return GoSource(**source)
 
 def _infer_importpath(ctx):
diff --git a/go/private/rules/cgo.bzl b/go/private/rules/cgo.bzl
index 1888bbd..420fab5 100644
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -312,7 +312,7 @@ def _pure(ctx, mode):
 def _not_pure(ctx, mode):
     return not mode.pure
 
-def _cgo_resolve_source(go, attr, source, merge):
+def _cgo_resolve_source(go, attr, source, merge, coverage_instrumented):
   library = source["library"]
   cgo_info = library.cgo_info
 
@@ -321,10 +321,12 @@ def _cgo_resolve_source(go, attr, source, merge):
   if source["mode"].pure:
     split = split_srcs(cgo_info.orig_srcs)
     source["srcs"] = split.go + split.asm
-    source["cover"] = source["srcs"]
+    if coverage_instrumented:
+      source["cover"] = source["srcs"]
   else:
     source["srcs"] = cgo_info.transformed_go_srcs + cgo_info.gen_go_srcs
-    source["cover"] = cgo_info.transformed_go_srcs
+    if coverage_instrumented:
[...]

PR #1478, opened 2018-05-01
Fix coverage issues around cgo

jayconrod:
* Instrumentation filter is now applied to cgo sources. Previously,
  they were always instrumented if coverage was enabled.
* In coverage data, cgo sources are recorded with their original
  basenames instead of the .cgo1.go generated names.
* If the final binary is linked externally, the --coverage flag is
  passed to the linker. This fixes link errors with gcov functions
  called from instrumented C files.

Also:

* GoSource has a new field: orig_src_map, which is used by coverage to
  find original basenames.

Related #1467

### DIFF
--- a/go/private/actions/cover.bzl
+++ b/go/private/actions/cover.bzl
@@ -30,23 +30,30 @@ def emit_cover(go, source):
     return source
 
   covered = []
+  covered_src_map = dict(source.orig_src_map)
   for src in source.srcs:
     if not src.basename.endswith(".go") or src not in source.cover:
       covered.append(src)
       continue
+    orig = covered_src_map.get(src, src)
+    _, pkgpath = effective_importpath_pkgpath(source.library)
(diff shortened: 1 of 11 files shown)


go/private/actions/cover.bzl
```diff
@@ -30,23 +30,30 @@ def emit_cover(go, source):
...
 
   covered = []
+  covered_src_map = dict(source.orig_src_map)
   for src in source.srcs:
     if not src.basename.endswith(".go") or src not in source.cover:
       covered.append(src)
       continue
+    orig = covered_src_map.get(src, src)
+    _, pkgpath = effective_importpath_pkgpath(source.library)
+    srcname = pkgpath + "/" + orig.basename if pkgpath else orig.path
```
ianthehat: We probably ought to get in the habit of using skylibs paths.join...

PR #1479, opened 2018-05-01
Support Objective C libraries in cdeps

jayconrod:
* Inputs and options are now added from cdeps targets with the objc
  provider. There isn't a perfect mapping between cc and objc
  providers, so there may still be some gaps in support.
* Tags are once again passed to `go install` by the GoStdlib
  builder. This is important for iOS.
* Added a test case that uses cdeps.


steeve:
This is very very cool.

One thing it's missing is the -framework option (via objc library sdk frameworks).

I need to test this but it looks great.

go/private/rules/cgo.bzl
```diff
@@ -176,35 +176,49 @@ def _cgo_codegen_impl(ctx):
...
+    elif hasattr(d, "objc"):
+      cppopts.extend(['-D' + define for define in d.objc.define.to_list()])
+      for inc in d.objc.include:
+        _include_unique(cppopts, "-I", inc, seen_includes)
+      for inc in d.objc.iquote:
+        _include_unique(cppopts, "-iquote", inc, seen_quote_includes)
+      for inc in d.objc.include_system:
+        _include_unique(cppopts, "-isystem", inc, seen_system_includes)
+      # TODO(jayconrod): do we need to link against dynamic libraries or
+      # frameworks? We link against *_fully_linked.a, so maybe not?
```
steeve: sdk_framework would be great

steeve: Also if modules are enabled you might want to add -fmodules

jayconrod: Would you be willing to do a PR to add these? I'm not familiar enough with Objective C to know if I'm doing it right. :)

steeve: i'll try ! thanks for doing this :D

PR #1487, opened 2018-05-05
 Dynamically generate and execute check libraries

stjj89:
Link all the check libraries passed to the go_checker rule into the
generated checker binary and run the analyses these libraries register
at build-time. Each check library must register its analysis using the
newly added package analysis.

First pass. Pending questions:
- Is there a need for `go_check_library`? Should checks be given as `go_tool_library`?
- The checker code is getting complex. Should we keep that much code in a template, or are there cleaner alternatives?
- How should we write integration tests for this? We will probably need some way to dynamically change the WORKSPACE file in the test environment.
- Package analysis, gcexportdata, and gcimporter15 need to be go_tool_library to solve the bootstrapping problem. Is there any cleaner way to solve this problem?

### DIFF
--- a/go/private/actions/compile.bzl
+++ b/go/private/actions/compile.bzl
@@ -30,6 +30,9 @@ def _searchpath(l):
 def _importmap(l):
   return ["{}={}".format(v.data.importpath, v.data.importmap) for v in l]
 
+def _archivefile(l):
+  return ["{}={}".format(v.data.importpath, v.data.file.path) for v in l]
+
 def emit_compile(go,
     sources = None,
     importpath = "",
@@ -71,6 +74,7 @@ def emit_compile(go,
   args.add(archives, before_each="-dep", map_fn=_importpath)
   args.add(archives, before_each="-I", map_fn=_searchpath)
   args.add(archives, before_each="-importmap", map_fn=_importmap)
+  args.add(archives, before_each="-archivefile", map_fn=_archivefile)
   args.add(["-o", out_lib, "-trimpath", ".", "-I", "."])
   if testfilter:
     args.add(["--testfilter", testfilter])
--- a/go/private/rules/library.bzl
+++ b/go/private/rules/library.bzl
@@ -87,7 +87,7 @@ go_tool_library = go_rule(
     },
 )
 """
-This is used instead of `go_library` for packages that are depended on
-implicitly by code generated within the Go rules. This avoids a
-bootstrapping problem.
+This is used instead of `go_library` for dependencies of the go_checker rule and
+packages that are depended on implicitly by code generated within the Go rules.
+This avoids a bootstrapping problem.
 """
--- a/go/private/rules/source.bzl
+++ b/go/private/rules/source.bzl
@@ -47,6 +47,7 @@ def _go_source_impl(ctx):
 
 go_source = go_rule(
     _go_source_impl,
+    bootstrap = True,
     attrs = {
         "data": attr.label_list(
             allow_files = True,
--- a/go/tools/builders/checker_main.go
+++ b/go/tools/builders/checker_main.go
@@ -1,0 +1,207 @@
+/* Copyright 2018 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Loads and runs registered analyses on a well-typed Go package.
+
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"go/ast"
+	"go/build"
+	"go/parser"
+	"go/token"
+	"go/types"
+	"log"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/bazelbuild/rules_go/go/tools/analysis"
+	"golang.org/x/tools/go/gcexportdata"
+)
+
+// run returns an error only if the package is successfully loaded and at least
+// one analysis fails. All other errors (e.g. during loading) are logged but
+// do not return an error so as not to unnecessarily interrupt builds.
+func run(args []string) error {
+	archiveFiles := multiFlag{}
+	flags := flag.NewFlagSet("checker", flag.ContinueOnError)
+	flags.Var(&archiveFiles, "archivefile", "Archive file of a direct dependency")
+	stdlib := flags.String("stdlib", "", "Root directory of stdlib")
+	if err := flags.Parse(args); err != nil {
+		log.Println(err)
+		return nil
+	}
+	if *stdlib == "" {
+		log.Printf("missing stdlib root directory")
+		return nil
+	}
+	importsToArchives := make(map[string]string)
+	for _, a := range archiveFiles {
+		kv := strings.Split(a, "=")
+		if len(kv) != 2 {
+			continue // sanity check
+		}
+		importsToArchives[kv[0]] = kv[1]
+	}
+	fset := token.NewFileSet()
+	imp := &importer{
+		fset:              fset,
+		packages:          make(map[string]*types.Package),
+		importsToArchives: importsToArchives,
+		stdlib:            *stdlib,
+	}
+	apkg, err := load(fset, imp, flags.Args())
+	if err != nil {
+		log.Printf("error loading package: %v\n", err)
+		return nil
+	}
+
+	c := make(chan result)
+	for _, a := range analysis.Analyses() {
+		go func(a *analysis.Analysis) {
+			defer func() {
+				// Prevent a panic in a single analysis from interrupting other analyses.
+				if r := recover(); r != nil {
+					c <- result{err: fmt.Errorf("recovered from panic during analysis %q: %v", a.Name, r)}
+				}
+			}()
+			res, err := a.Run(apkg)
+			if err != nil {
+				if err != analysis.ErrSkipped {
+					c <- result{err: fmt.Errorf("analysis failed: %v", err)}
+				}
+				c <- result{}
+			}
+			c <- result{findings: res.Findings}
+		}(a)
+	}
+	var allFindings []*analysis.Finding
+	for i := 0; i < len(analysis.Analyses()); i++ {
+		result := <-c
+		if result.err != nil {
+			log.Println(result.err)
+			continue
+		}
+		allFindings = append(allFindings, result.findings...)
+	}
+	if len(allFindings) != 0 {
+		sort.Slice(allFindings, func(i, j int) bool {
+			return allFindings[i].Pos < allFindings[j].Pos
+		})
+		errMsg := "errors found during build-time code analysis:\n"
+		for _, f := range allFindings {
+			errMsg += fmt.Sprintf("%s: %s\n", fset.Position(f.Pos), f.Message)
+		}
+		return errors.New(errMsg)
+	}
+	return nil
+}
+
+func main() {
+	log.SetFlags(0) // no timestamp
+	log.SetPrefix("GoChecker: ")
+	if err := run(os.Args[1:]); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type result struct {
+	findings []*analysis.Finding
+	err      error
+}
+
+// load parses and type checks the source code in each file in filenames.
+// On failure, it reports errors to Error and returns (nil, nil, nil).
+func load(fset *token.FileSet, imp types.Importer, filenames []string) (*analysis.Package, error) {
+	if len(filenames) == 0 {
+		return nil, errors.New("no filenames")
+	}
+	var files []*ast.File
+	for _, file := range filenames {
+		f, err := parser.ParseFile(fset, file, nil, parser.ParseComments)
+		if err != nil {
+			return nil, err
+		}
+		files = append(files, f)
+	}
+
+	config := types.Config{
+		Importer: imp,
+		Error: func(err error) {
+			e := err.(types.Error)
+			msg := fmt.Sprintf("%s", e.Msg)
+			posn := e.Fset.Position(e.Pos)
+			if posn.Filename != "" {
+				msg = fmt.Sprintf("%s: %s", posn, msg)
+			}
+			fmt.Fprintln(os.Stderr, msg)
+		},
+	}
+	info := &types.Info{
+		Types:     make(map[ast.Expr]types.TypeAndValue),
+		Uses:      make(map[*ast.Ident]types.Object),
+		Defs:      make(map[*ast.Ident]types.Object),
+		Implicits: make(map[ast.Node]types.Object),
+	}
+	pkg, err := config.Check(files[0].Name.Name, fset, files, info)
+	if err != nil {
+		// Errors were already reported through config.Error.
+		return nil, nil
+	}
+	return &analysis.Package{Fset: fset, Files: files, Types: pkg, Info: info}, nil
+}
+
+type importer struct {
+	fset     *token.FileSet
+	packages map[string]*types.Package
+	// importsToArchives maps import paths to the path to the archive file representing the
+	// corresponding library.
+	importsToArchives map[string]string
+	// stdlib is the root directory containing standard library package archive files.
+	stdlib string
+}
+
+func (i *importer) Import(path string) (*types.Package, error) {
+	archive, ok := i.importsToArchives[path]
+	if !ok {
+		// stdlib package.
+		ctxt := build.Default
+		archive = filepath.Join(i.stdlib, "pkg", ctxt.GOOS+"_"+ctxt.GOARCH, path+".a")
+	}
+	// open file
+	f, err := os.Open(archive)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		f.Close()
+		if err != nil {
+			// add file name to error
+			err = fmt.Errorf("reading export data: %s: %v", archive, err)
+		}
+	}()
+
+	r, err := gcexportdata.NewReader(f)
+	if err != nil {
+		return nil, err
+	}
+
+	return gcexportdata.Read(r, i.fset, i.packages, path)
+}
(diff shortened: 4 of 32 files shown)


go/private/rules/library.bzl
```diff
@@ -91,3 +91,24 @@ This is used instead of `go_library` for packages that are depended on
...
+        ),
+        "srcs": attr.label_list(allow_files = True),
+        "deps": attr.label_list(providers = [GoLibrary]),
+        "embed": attr.label_list(providers = [GoLibrary]),
+    },
+)
+"""See go/checks.rst#go_check_library for full documentation."""
+
+"""TODO(samueltan): wrap returned providers in a new provider type?"""
+"""TODO(samueltan): merge with go_tool_library?"""
```
jayconrod: Let's do this. These rules are conceptually very similar.

go/tools/builders/generate_checker_main.go
```diff
@@ -31,15 +31,206 @@ import (
 var codeTpl = `
 package main
 
+import (
```
jayconrod: As you mentioned in the PR comment, this has gotten pretty large, and it seems like the only parts that need to be generated are the blank imports of the checker libraries.

I think it's probably cleaner to move this code to a new file, perhaps `//go/tools/builders:checker_main.go`. You can define a `go_source` rule in `//go/tools/builders` that includes this code, as well as flags.go (so you don't need to redefine `multiFlag`). In the `go_checker` rule, you can list that target in `embed` in the `attr` struct passed to `library_to_source`. The generated file would then just contain the blank imports.

go/tools/builders/generate_checker_main.go
```diff
@@ -31,15 +31,206 @@ import (
...
+	return fmt.Sprint(*m)
+}
+
+func (m *multiFlag) Set(v string) error {
+	(*m) = append(*m, v)
+	return nil
+}
+
+var (
+	archiveFiles multiFlag
```
jayconrod: Style nit: we've been following this pattern for builders:

* `main` should just call `log.SetFlags(0)` and `log.SetPrefix("GoCheck: ")`, then a `run` function that takes `args []string` and returns `error`. `main` should call `log.Fatal(err)` if `run` fails.This makes it at least theoretically possible to test the `run` function (though we don't usually do that).
* Flags should be defined locally within the `run` function using `flag.NewFlagSet`. That's just helpful for reducing global state.

go/tools/builders/generate_checker_main.go
```diff
@@ -31,15 +31,206 @@ import (
...
+func init() {
+	flag.Var(&archiveFiles, "archivefile", "Archive file of a direct dependency")
+	flag.StringVar(&stdlib, "stdlib", "", "Root directory of stdlib")
+}
+
 func main() {
-	// This is a stub.
+	flag.Parse()
+	if stdlib == "" {
+		fmt.Fprintln(os.Stderr, "checker: missing stdlib root directory")
```
jayconrod: Nit: Use `log.Printf` instead of using `fmt.Fprintf(os.Stderr, ...)` directly. It will automatically have a prefix, so you won't need to write `checker: ` here.

go/tools/builders/generate_checker_main.go
```diff
@@ -31,15 +31,206 @@ import (
...
+	var allFindings []*analysis.Finding
+	for i := 0; i < len(analysis.Analyses()); i++ {
+		result := <-c
+		if result.err != nil {
+			fmt.Fprintf(os.Stderr, "checker: %v\n", result.err)
+			continue
+		}
+		allFindings = append(allFindings, result.findings...)
+	}
+	if len(allFindings) != 0 {
```
jayconrod: Sort by source position before displaying.

go/tools/builders/generate_checker_main.go
```diff
@@ -31,15 +31,206 @@ import (
...
+	}
+	return &analysis.Package{Fset: fset, Files: files, Types: pkg, Info: info}, nil
+}
+
+type importer struct {
+	fset     *token.FileSet
+	packages map[string]*types.Package
+	// pkgs maps import paths to the path to the archive file representing the
+	// corresponding library.
+	pkgs map[string]string
```
jayconrod: `importMap` or `importsToArchives` might be better names. `pkgs` is too similar to `packages`.

go/tools/builders/generate_checker_main.go
```diff
@@ -31,15 +31,206 @@ import (
...
+	// stdlib is the root directory containing standard library package archive files.
+	stdlib string
+}
+
+func (i importer) Import(path string) (*types.Package, error) {
```
jayconrod: Nit: receiver should be `*importer`. It seems like this type isn't mutated here, so maybe it doesn't matter, but `importer` doesn't seem like something that should be passed by value.

analysis/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
+# load("@io_bazel_rules_go//go:def.bzl", "go_library")
```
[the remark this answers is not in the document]
stjj89: Done.

analysis/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
...
+load("@io_bazel_rules_go//go/private:rules/library.bzl", "go_tool_library")
+
+go_tool_library(
+    name = "analysis",
+    srcs = ["analysis.go"],
+    importpath = "github.com/bazelbuild/rules_go/analysis",
+    # TODO(samueltan): maybe limit to only :default_checker
```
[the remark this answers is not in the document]
stjj89: Right. I have removed this TODO.

analysis/analysis.go
```diff
@@ -0,0 +1,169 @@
+// The analysis package defines a uniform interface for static checkers
```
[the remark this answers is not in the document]
stjj89: Done.

analysis/analysis.go
```diff
@@ -0,0 +1,169 @@
...
+import (
+	"errors"
+	"flag"
+	"go/ast"
+	"go/token"
+	"go/types"
+	"log"
+	"sort"
+	"sync"
+	// No google3 dependencies!  We plan to open-source this.
```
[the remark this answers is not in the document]
stjj89: Done.

analysis/analysis.go
```diff
@@ -0,0 +1,169 @@
...
+
+// A Package describes a package of Go source code to be analyzed.
+type Package struct {
+	// syntax trees
+	Fset        *token.FileSet // file position information
+	Files       []*ast.File    // the abstract syntax tree of each file
+	ParseErrors bool           // whether there were errors during parsing
+
+	// type information
+	Types      *types.Package // type information about the pakage
```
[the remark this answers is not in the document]
stjj89: Done.

analysis/analysis.go
```diff
@@ -0,0 +1,169 @@
...
+// For the kind, callers should use a private type to avoid namespace collisions.
+// For a fine-grained extension that describes, for example, a single
+// function, the key should incorporate the identity of that function.
+//
+// Not all extensions are provided in all frameworks.
+// Analyses should gracefully handle the absence of a given extension.
+//
+// Example extensions:
+//
+//   google3/third_party/golang/analysis/ssaext/ssaext
```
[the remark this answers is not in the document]
stjj89: Done.

analysis/analysis.go
```diff
@@ -0,0 +1,169 @@
...
+	return ext
+}
+
+// A Result describes the results of applying an analysis to a package.
+type Result struct {
+	Findings []*Finding
+
+	// TODO(adonovan):
+	// - diffs (refactorings or cleanups)?
+	//   Look at http://cs/errorprone/fixes/Fix.java for inspiration.
```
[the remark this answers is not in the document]
stjj89: Done.

go/private/actions/compile.bzl
```diff
@@ -71,6 +74,8 @@ def emit_compile(go,
   args.add(archives, before_each="-dep", map_fn=_importpath)
   args.add(archives, before_each="-I", map_fn=_searchpath)
   args.add(archives, before_each="-importmap", map_fn=_importmap)
+  args.add(archives, before_each="-archivefile", map_fn=_archivefile)
+  args.add(["-stdlib_rootfile", go.stdlib.root_file])
```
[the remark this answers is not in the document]
stjj89: Done.

jayconrod:
I don't see any new commits. Could you double-check the push to `stjj89 check`?

We're in a summit this week, so code review may be a bit delayed.

go/tools/builders/generate_checker_main.go
jayconrod: So as you said, since everything in `importer` is either a reference type or is never mutated, it doesn't actually matter in this case. However, it requires some thought to verify that invariant, and it would be easy to change this type accidentally so that invariant is no longer true.

There are some guidelines at [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type) on the golang wiki. I don't believe this falls directly under any of those cases, so "Finally, when in doubt, use a pointer receiver."

go/tools/builders/checker_main.go
```diff
@@ -0,0 +1,207 @@
...
+// one analysis fails. All other errors (e.g. during loading) are logged but
+// do not return an error so as not to unnecessarily interrupt builds.
+func run(args []string) error {
+	archiveFiles := multiFlag{}
+	flags := flag.NewFlagSet("checker", flag.ContinueOnError)
+	flags.Var(&archiveFiles, "archivefile", "Archive file of a direct dependency")
+	stdlib := flags.String("stdlib", "", "Root directory of stdlib")
+	if err := flags.Parse(args); err != nil {
+		log.Println(err)
+		return nil
```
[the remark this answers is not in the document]
stjj89: In the internal Go code analysis framework (nogo), only legitimate code analysis failures should block the build. We surface error messages for other internal errors, but do not block builds for them. We do not want a mistake on the check-writer's part to interfere with Go developer productivity.

I think that the same policy should apply here. A build-blocking check failure can be particularly high impact if the developer cannot turn off the check. (i.e. in a setup where the configuration file is only writable by a subset of developers at the organization.)

go/private/rules/source.bzl
```diff
@@ -59,3 +59,27 @@ go_source = go_rule(
     },
 )
 """See go/core.rst#go_source for full documentation."""
+
+go_tool_source = go_rule(
```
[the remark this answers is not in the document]
stjj89: Done. I have removed go_tool_source.

go/tools/builders/checker_main.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if *stdlib == "" {
+		log.Printf("missing stdlib root directory")
+		return nil
```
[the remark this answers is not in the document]
stjj89: See above comment about fatal errors.

go/tools/builders/checker_main.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	importsToArchives := make(map[string]string)
+	for _, a := range archiveFiles {
+		kv := strings.Split(a, "=")
+		if len(kv) != 2 {
+			continue // sanity check
```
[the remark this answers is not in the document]
stjj89: See above comment about fatal errors.

go/tools/builders/compile.go
```diff
@@ -136,7 +140,13 @@ func run(args []string) error {
...
 	}
 	if *checker != "" {
-		checkerCmd := exec.Command(*checker)
+		var checkerargs []string
+		for _, a := range archiveFiles {
+			checkerargs = append(checkerargs, "-archivefile", a)
+		}
+		checkerargs = append(checkerargs, "-stdlib", filepath.Dir(goenv.rootFile))
+		checkerargs = append(checkerargs, filenames...)
+		checkerCmd := exec.Command(*checker, checkerargs...)
```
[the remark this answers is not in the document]
stjj89: Done. The checker now writes to a buffer that will only be printed if the compiler succeeds.

go/tools/builders/generate_checker_main.go
```diff
@@ -26,20 +26,22 @@ import (
...
-// TODO(samueltan): ensure that panics caused by check libraries are differentiated
-// from legitimate check errors.
 var codeTpl = `
 package main
 
-func main() {
-	// This is a stub.
-}
+import (
+{{range $i, $importPath := .}}
```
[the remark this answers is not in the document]
stjj89: Done.

go/private/rules/builders.bzl
```diff
@@ -23,7 +23,7 @@ def _builders_impl(ctx):
           link = ctx.executable._link,
           cgo = ctx.executable._cgo,
           checker_generator = ctx.executable._checker_generator,
-          test_generator = ctx.executable._test_generator,
+          test_generator = ctx.files._test_generator,
```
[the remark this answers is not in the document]
stjj89: Whoops, this was a mistake. I have reverted this change.

jayconrod:
Changes look good. A couple `bazel_tests` are failing, but I'm not really sure why yet. Looking into it.

stjj89:
I'm almost done with integration tests. Expect another push tomorrow.

go/tools/builders/checker_main.go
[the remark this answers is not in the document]
rogerhub: In that case, why not return ([]*analysis.Findings, error)? It should make this code easier to test, and you can preserve the desired behavior by doing the logging/exiting in main().

stjj89:
I have added some basic integration tests. PTAL, thanks.

tests/core/checks/custom/BUILD.bazel
```diff
@@ -0,0 +1,150 @@
...
+if [[ result -eq 0 ]]; then
+  echo "error: expected build error" >&2
+  result=1
+else
+  {check_err}
+fi
+"""
+
+_check_err_tmpl = """
+  grep -q '{err}' bazel-output.txt
```
jayconrod: Nit: simplify to `if ! grep -q '{err}' bazel-output.txt; then`

tests/core/checks/custom/BUILD.bazel
```diff
@@ -0,0 +1,150 @@
...
+    name = "single_checks_no_errors",
+    build = _single_check_build,
+    workspace = _workspace,
+    check = _no_error_check,
+    command = "build",
+    targets = [":no_errors"],
+    extra_files = [":foo_func_check.go"],
+)
+
+bazel_test(
```
jayconrod: Do the single check tests give much more coverage than the multiple check tests? `bazel_tests` are expensive enough to run that I think we should just have the two multiple-check cases.

I'll create a TODO for myself (#1508) to allow multiple tests in the same workspace. Then we could consolidate the `has_errors` and `no_errors` cases into a single test.

tests/core/checks/custom/BUILD.bazel
```diff
@@ -0,0 +1,150 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//tests:bazel_tests.bzl", "bazel_test")
+
+_workspace = """
```
jayconrod: Style nit: these variables can be `ALL_CAPS` since they are constants within a build file. The leading underscore is only needed in .bzl files to prevent variables from being imported by other files.

tests/core/checks/custom/foo_func_check.go
```diff
@@ -0,0 +1,36 @@
...
+	var findings []*analysis.Finding
+	for _, f := range p.Files {
+		ast.Inspect(f, func(n ast.Node) bool {
+			switch n := n.(type) {
+			case *ast.FuncDecl:
+				if n.Name.Name == "Foo" {
+					findings = append(findings, &analysis.Finding{
+						Pos:     n.Pos(),
+						End:     n.End(),
+						Message: "function must not be named Foo",
```
jayconrod: Hah, this is excellent.

vendor/golang.org/x/tools/go/gcexportdata/example_test.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2016 The Go Authors. All rights reserved.
```
jayconrod: Let's remove the `_test.go` files and the `testdata` directory. No need to keep these.

tests/bazel_tests.bzl
```diff
@@ -105,11 +106,17 @@ def _bazel_test_script_impl(ctx):
   build_file = go.declare_file(go, path="BUILD.in")
   ctx.actions.write(build_file, ctx.attr.build)
 
+  extra_files = ""
+  for extra in ctx.attr.extra_files:
```
jayconrod: Quote filenames using `shell.quote` in `go/private/skylib/lib/shell.bzl`. Also, join paths using `paths.join` in the same directory.

It's best to build a list of strings (or a list of files that are converted to strings in a list comprehension), then join them with `" ".join(...)`.

tests/bazel_tests.bzl
```diff
@@ -44,6 +44,7 @@ mkdir -p {work_dir}
 mkdir -p {cache_dir}
 cp -f {workspace} {work_dir}/WORKSPACE
 cp -f {build} {work_dir}/BUILD.bazel
+cp -f {extra_files} {work_dir}/
```
jayconrod: Wrap this in an `if` in case there are no files. Currently, this prints an error (but proceeds).

tests/core/checks/custom/BUILD.bazel
stjj89: Done.

tests/core/checks/custom/BUILD.bazel
stjj89: Done. I thought that having both single and multiple checks would help us inductively reason about the behavior of the checking framework, but based on my knowledge of the implementation, running only the latter is probably fine.

tests/core/checks/custom/BUILD.bazel
stjj89: Done.

vendor/golang.org/x/tools/go/gcexportdata/example_test.go
stjj89: Done.

tests/bazel_tests.bzl
stjj89: Done.

tests/bazel_tests.bzl
stjj89: Done.

stjj89:
`//tests/legacy/build_with_old_sdk:build_with_old_sdk` and `//tests/legacy/custom_go_toolchain:custom_go_toolchain` fail because they use Go 1.8. The generated checker binary imports a version of package gcimporter that is built for Go 1.9, so this binary will fail to buikd when the host is using Go 1.8.

Since Go 1.8 support will be removed soon, this is fine.

PR #1493, opened 2018-05-09
Propagate restricted_to and compatible_with attributes in cgo rules

### LINKED ISSUE #1488: Platform constraints don't work with cgo
I have a project which is built for multiple platforms (x86 and ARM). In this project there's a `go_binary` with `cgo = True` which I want to build only for x86. For example:
```
load("@io_bazel_rules_go//go:def.bzl", "go_binary")

go_binary(
    name = "hello",
    srcs = ["main.go"],
    cgo = True,
    restricted_to = ["//:k8"],
)
```

I expect this target to be completely skipped when building for a platform other than x86_64 (k8). Instead, I see that e.g. `hello.cgo_codegen` is built regardless of platform constraints. Perhaps the constraints need to be propagated to the targets created by `go_binary` macro.

Without `cgo = True` this works as expected.

`//:k8` is defined like this:
```
environment(name = "k8")  # x86_64
environment(name = "arm")

environment_group(
    name = "cpus",
    defaults = [
        ":k8",
        ":arm",
    ],
    environments = [
        ":k8",
        ":arm",
    ],
)
```

iley:
This fixes #1488.

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -397,7 +397,7 @@ _cgo_collect_info = go_rule(
 """No-op rule that collects information from _cgo_codegen and cc_library
 info into a GoSourceList provider for easy consumption."""
 
-def setup_cgo_library(name, srcs, cdeps, copts, cxxopts, cppopts, clinkopts, objc, objcopts, tags):
+def setup_cgo_library(name, srcs, cdeps, copts, cxxopts, cppopts, clinkopts, objc, objcopts, **common_attrs):
   # Apply build constraints to source files (both Go and C) but not to header
   # files. Separate filtered Go and C sources.
 
@@ -426,8 +426,8 @@ def setup_cgo_library(name, srcs, cdeps, copts, cxxopts, cppopts, clinkopts, obj
       cxxopts = cxxopts,
       cppopts = cppopts,
       linkopts = clinkopts,
-      tags = tags,
       visibility = ["//visibility:private"],
+      **common_attrs
   )
 
   select_go_files = name + ".select_go_files"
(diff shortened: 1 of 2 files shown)


iley:
I'll have a look at the failing build.

go/private/rules/cgo.bzl
```diff
@@ -523,6 +523,8 @@ def setup_cgo_library(name, srcs, cdeps, copts, cxxopts, cppopts, clinkopts, obj
 
   if objc:
     cgo_objc_lib_name = name + ".cgo_objc_lib"
+    kwargs = extra_args
+    kwargs.update(objcopts)
```
jayconrod: I think you need to update `objcopts` with `extra_args` here instead of updating `extra_args`. This statement modifies `extra_args`, which is used in other rules below.

go/private/rules/cgo.bzl
```diff
@@ -397,7 +397,7 @@ _cgo_collect_info = go_rule(
 """No-op rule that collects information from _cgo_codegen and cc_library
 info into a GoSourceList provider for easy consumption."""
 
-def setup_cgo_library(name, srcs, cdeps, copts, cxxopts, cppopts, clinkopts, objc, objcopts, tags):
+def setup_cgo_library(name, srcs, cdeps, copts, cxxopts, cppopts, clinkopts, objc, objcopts, **extra_args):
```
jayconrod: One small nit: rename `extra_args` to `common_attrs`.

go/private/rules/cgo.bzl
iley: Oh, right! Thanks for catching this.

go/private/rules/cgo.bzl
iley: Makes sense. Done.

jayconrod:
Thanks!

The Travis failure appears to be a problem with Travis CI itself, not this change. Tests pass locally and on Buildkite, so force submitting.

PR #1497, opened 2018-05-11
Enable use of FreeBSD SDK

rescrv:
This allows a simple repo to be used on FreeBSD (amd64).  I manually
tested the change on a toy project and bazel build/test both worked as I
expected them to work based upon what I've come to expect on Linux.

Signed-off-by: Robert Escriva <robert@rescrv.net>


iphydf:
Nice, thanks, I'll be testing this regularly myself. Currently not with ci, but hopefully that'll change some time this year.

PR #1526, opened 2018-06-05
fix ios cross compilation with cgo

bennyscetbun:
we were unable to compile cgo for ios due to 2 issues:
- the ios minimum version should be higher than 6.1 or an error occurs at compile time with the arguments
- when cgo is called pure should always be set to off to setup the environment the right way


bennyscetbun:
I signed it!

PR #1536, opened 2018-06-08
turn relative path to absolute path in environment's cgo's flags

bennyscetbun:
This pr can be merge as it, but it s main focus is to discuss a "rules_go" way to fix the problem.

Problem:
cgo flags will sometime be setup with relative path in the environment but Go will make its own sandbox and wont be able to use those path.

Solution:
make all the path absolute in cgo environment variable.

I have no better solution than to embbed shlex in rules_go and use it on the environmnet variables.

### DIFF
--- a/go/tools/builders/asm.go
+++ b/go/tools/builders/asm.go
@@ -52,7 +52,7 @@ func run(args []string) error {
 	// Build source with the assembler.
 	goargs := goenv.goTool("asm", toolArgs...)
 	goargs = append(goargs, source)
-	absArgs(goargs, []string{"I", "o", "trimpath"})
+	absArgs(goargs, []string{"-I", "-o", "-trimpath"})
 	return goenv.runCommand(goargs)
 }
 
--- a/go/tools/builders/compile.go
+++ b/go/tools/builders/compile.go
@@ -118,7 +118,7 @@ func run(args []string) error {
 	for _, f := range files {
 		goargs = append(goargs, f.filename)
 	}
-	absArgs(goargs, []string{"I", "o", "trimpath"})
+	absArgs(goargs, []string{"-I", "-o", "-trimpath"})
 	return goenv.runCommand(goargs)
 }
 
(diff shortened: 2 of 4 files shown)


bennyscetbun:
something went wrong with travis... Could you restart it ?

go/tools/builders/cgo.go
```diff
@@ -180,6 +180,12 @@ func run(args []string) error {
 		}
 	}
 
+	// Modifying CGO flags to use only absolute path
+	// because go is having its own sandbox, all CGO flags must use absolute path
+	if err := absEnv([]string{"CGO_CFLAGS", "CGO_CXXFLAGS", "CGO_CPPFLAGS", "CGO_LDFLAGS"}, []string{"-I", "-L", "-isysroot", "-isystem", "-iquote", "-include", "-gcc-toolchain", "--sysroot"}); err != nil {
```
steeve: perhaps they should be defined in a global variable up in `env.go`

jayconrod:
Change looks good, but some tests are failing on BuildKite. Specifically `//tests/core/cgo:dylib_test` seems to be broken. Mind taking a look?

```
external/go_sdk/pkg/tool/darwin_amd64/link: running external/local_config_cc/cc_wrapper.sh failed: exit status 1
ld: warning: directory not found for option '-L/private/var/tmp/_bazel_buildkite/be4b3196a21ee19508aa8fdc35451a6f/sandbox/darwin-sandbox/2028/execroot/io_bazel_rules_go/bazel-out/darwin-fastbuild/bin/_solib_darwin_x86_64/_U_S_Stests_Score_Scgo_Cdarwin_Uimported_Udylib___Utests_Score_Scgo'
ld: library not found for -limported
clang: error: linker command failed with exit code 1 (use -v to see invocation)
GoLink: error running subcommand: exit status 2
```

bennyscetbun:
on it

bennyscetbun:
i ve remove absEnv from cgo.go. It s only useful in stdlib.go which is build with go build.

jayconrod:
Merged. Thanks for submitting this.

I guess it makes sense this wouldn't be needed in cgo -- only the stdlib builder changes the working directory internally. Other builders should be able to tolerate relative paths.

PR #1543, opened 2018-06-12
Use absolute path to the compiler dir in stdlib builder

siddharthab:
Needed to fix a regression after https://github.com/bazelbuild/rules_go/pull/1503.

### DIFF
--- a/go/tools/builders/stdlib.go
+++ b/go/tools/builders/stdlib.go
@@ -21,6 +21,7 @@ import (
 	"go/build"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
@@ -55,6 +56,13 @@ func run(args []string) error {
 	// Make sure we have an absolute path to the C compiler.
 	// TODO(#1357): also take absolute paths of includes and other paths in flags.
 	os.Setenv("CC", abs(os.Getenv("CC")))
+
+	// Ensure paths are absolute.
+	absPaths := []string{}
+	for _, path := range filepath.SplitList(os.Getenv("PATH")) {
+		absPaths = append(absPaths, abs(path))
+	}
+	os.Setenv("PATH", strings.Join(absPaths, string(os.PathListSeparator)))
 
 	// Build the commands needed to build the std library in the right mode
 	installArgs := goenv.goCmd("install", "-toolexec", abs(*filterBuildid))


siddharthab:
[This](https://github.com/grailbio/bazel-toolchain/blob/a964325290a98766305ec2d8963b1749fe612533/toolchain/cc_wrapper.sh.tpl#L56,L58) is how I call the actual compiler through a wrapped compiler when being called from rules_go tools.

jayconrod:
Thanks for submitting! Several comments though:

1. What exactly regressed in #1503? I don't think any part of that change touched `PATH`. I'd like to understand the issue before accepting this change.
2. `PATH` is a list of paths separated by `os.PathListSeparator`. Need to split them up, absolutize them all, then join again.
3. Is this something that can be tested by building something? I'm guessing that fully testing this would require setting up an alternate C++ toolchain, which is something I won't ask for in this PR. Without tests, I can't promise it won't break in the future.

siddharthab:
Thank you for your prompt response.

1. Sorry, it's actually https://github.com/bazelbuild/rules_go/pull/1465. Before this change, we used to set the supplied path to be absolute, but stopped doing so after that.

https://github.com/bazelbuild/rules_go/blob/7ca15c8102c8783f96e055390ac9cb15830b9e1d/go/tools/builders/env.go#L127

2. In this particular instance, PATH is always set to compilerPath. And hence, merely converting to an absolute path works. Happy to work on any suggestions you may have. Whether it is converting to absolute paths after splitting the string, or to introduce a new environment variable for better readability on what the value represents.

https://github.com/bazelbuild/rules_go/blob/e397b7bb9bce199a7dff08318c5b73402a564932/go/private/context.bzl#L243

3. I tested this using my toolchain on macOS which needs a wrapped compiler. I believe you are now part of bazel CI, so you should have builds running on macOS. If you want, I can add my LLVM toolchain to your WORKSPACE file, and then we can run the tests again with a crosstool_top flag.

jayconrod:
1. Cool, that makes more sense.
2. Right, at the moment, I think we only have one entry in `PATH`, but that could easily change in the future. Splitting the string, converting the individual paths to absolute, then joining should be fine.
3. At some point, we're going to need to test against a public toolchain; I have my eye on github.com/vsco/bazel-toolchains for this purpose. Let's leave that aside for this PR though.

siddharthab:
1. Ack.
2. Done.
3. I will recommend https://github.com/grailbio/bazel-toolchain. I have made sure that the toolchain works with rules_go. If it does not meet some standards, let me know and I can make improvements.

go/tools/builders/stdlib.go
```diff
@@ -56,6 +56,13 @@ func run(args []string) error {
 	// TODO(#1357): also take absolute paths of includes and other paths in flags.
 	os.Setenv("CC", abs(os.Getenv("CC")))
 
+	// Ensure paths are absolute.
+	absPaths := []string{}
+	for _, path := range strings.Split(os.Getenv("PATH"), ":") {
```
jayconrod: Looks good except `os.PathListSeparator` instead of `":"` for splitting and joining. The separator is `";"` on Windows.

siddharthab: Awesome! Done.

siddharthab: Sorry, now done for real. I found `filepath.SplitList` to get the split.

siddharthab:
Thank you!

jayconrod:
Thanks for fixing it!

Issue #1554, opened 2018-06-17
Xcode is mandatory after objc support


siddharthab:
Posting here so that others facing this issue are able to find a solution easily. Please feel free to close this issue if you decide that this is WAI.

https://github.com/bazelbuild/rules_go/pull/1437 made an Xcode install mandatory on macOS if the user does not specify `--macos_sdk_version` command line flag. Before this change, merely having the Xcode command line tools installed was sufficient.

It appears that the `apple_ensure_options` function adds the following two environment variables:
https://github.com/bazelbuild/rules_go/blob/bd1c095bdf04385a412a57c2feb9cd6e8fbe89c0/go/private/context.bzl#L308

```
APPLE_SDK_PLATFORM=MacOSX
APPLE_SDK_VERSION_OVERRIDE=10.10
```

When these environment variables are present, bazel ([XcodeLocalEnvProvider](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/exec/apple/XcodeLocalEnvProvider.java)) will use xcrun to determine the location of Apple SDK when sandboxing the build. The SDK at this specific version is available only when Xcode (app) is installed, but not when Xcode Command Line tools are installed. For the latter, only the SDK for the current OS version will be available, e.g. 10.13 is available on High Sierra.

In such a case, every rules_go action will fail with this error:
```
ERROR: /.../external/io_bazel_rules_go/BUILD.bazel:8:1: GoStdlib external/io_bazel_rules_go/darwin_amd64_stripped/stdlib~/pkg failed: I/O exception during sandboxed execution: xcrun failed with code 1.
This most likely indicates that SDK version [10.10] for platform [MacOSX] is unsupported for the target version of xcode.
Process exited with status 1
stdout: stderr: xcrun: error: SDK "macosx10.10" cannot be located
xcrun: error: SDK "macosx10.10" cannot be located
xcrun: error: unable to lookup item 'Path' in SDK 'macosx10.10'
```

A quick solution to this is for users to specify `build --macos_sdk_version=10.13` for High Sierra, and similar for other OS versions in their bazelrc.

A desired solution at rules_go level would be to introduce these environment variables only when objc files are present in sources, as people not writing Apple specific code are not familiar with these bazel command line flags.

jayconrod:
[...]

PR #1555, opened 2018-06-17
Set Xcode configuration only when building objc

siddharthab:
Resolves https://github.com/bazelbuild/rules_go/issues/1554.

@steeve Can you also take a look and see if this change works for you?

### DIFF
--- a/go/platform/apple.bzl
+++ b/go/platform/apple.bzl
@@ -19,11 +19,9 @@ PLATFORMS = {
     "arm64-apple-ios": apple_common.platform.ios_device,
     "arm64-apple-tvos": apple_common.platform.tvos_device,
     "i386-apple-ios": apple_common.platform.ios_simulator,
-    "i386-apple-macosx": apple_common.platform.macos,
     "i386-apple-tvos": apple_common.platform.tvos_simulator,
     "i386-apple-watchos": apple_common.platform.watchos_simulator,
     "x86_64-apple-ios": apple_common.platform.ios_simulator,
-    "x86_64-apple-macosx": apple_common.platform.macos,
     "x86_64-apple-tvos": apple_common.platform.ios_simulator,
     "x86_64-apple-watchos": apple_common.platform.watchos_simulator,
 }


steeve:
Unfortunately that won't work, since those configurations are needed across all build targets (`stdlib` even).
However, since I believe the issue is only on MacOS, we can try to disable it from being recognized as an Apple Platform needing configuration. Just remove it from https://github.com/bazelbuild/rules_go/blob/master/go/platform/apple.bzl#L26 and that may solve your problem.

That will prevent from accessing `apple_common.XcodeVersionConfig`.

siddharthab:
Are you suggesting I disable it behind a feature flag? I.e. a flag on the command line stops the function from being called.

steeve:
I meant try to comment out the line I mentioned, it will then ignore mac OS as an Apple platform.

siddharthab:
Oh yes, that definitely works. And so does providing macos_sdk_version flag. I am trying to get a fix merged here so my organisation does not have to maintain its own fork.

siddharthab:
I see. I updated the PR.

Will having objc code still work for macOS platform, or are we saying we should not support it?

steeve:
It should still work, although you're right that needs testing.
These env vars are here for cc_library. I think objc_library will work fine, but that needs testing.

jayconrod:
I'm not sure I fully understand the impact of this change. Will this break builds with C/C++/ObjC code on macOS?

siddharthab:
I am not sure. @steeve says they should work, and the Travis tests pass. Is there some more testing I can do?

jayconrod:
Travis and BuildKite all seem to be passing, but I doing think we have a CI configuration that tests macOS with only Xcode command line tools; I think all CI machines have Xcode installed.

On my laptop, I tried uninstalling Xcode and reinstalling just the command line tools. Builds that include objc fail with the same message with and without this change. Is that intentional, or is everything supposed to pass after this change?

siddharthab:
Yes, I think Objective C definitely needs Xcode. We want to ensure two things with this change:

1. C/C++ builds continue to work without Xcode for macOS platform.
2. ObjC builds continue with with Xcode for macOS platform.

I think that Travis checked for condition 2, and you and I checked for condition 1. So we should be good now. 😄

jayconrod:
I hope so :) I'm not familiar enough with ObjC and Xcode support in Bazel to be confident in this change, but it seems like it's working so...

Issue #1470, opened 2018-04-28
c-archive mode should output PIC


siddharthab:
Is there a way to specify the equivalent of GOGCCFLAGS?

I want to create an an archive with 'c-archive' link mode and use it as a .pic.a. Unfortunately, the .a output from the go_binary rule does not have PIC.

When I build using `go build -buildmode c-archive`, then the .a has PIC. I am guessing this is because my GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build763117176=/tmp/go-build -gno-record-gcc-switches". Any help would be appreciated.

jayconrod:
Are you trying to compile Go code as position independent, or associated C code? I'm assuming the latter.

I'm not sure Go code compiled with `-buildmode=c-archive` is position independent. I don't think it is. It's definitely not influenced by `GOGCCFLAGS`.

For C code, there are several ways to set flags:

* You can pass `--copt` arguments on the command line (see `bazel help build` for info). These affect all C targets. You can stick these in a [bazelrc file](https://docs.bazel.build/versions/master/user-manual.html#bazelrc) in your project (tools/bazel.rc) so you don't have to type them out. This sounds like your best option.
* For `cc_library` and `go_library` targets, you can set the `copts` attribute. That will only affect those specific targets. `go_binary` also has `clinkopts`.
* You can write a CROSSTOOL file. This is very difficult but gives you complete control over the C/C++ toolchain. Consider this a last resort.

siddharthab:
Thank you for your prompt response. I apologize for not providing enough details before. I intend to use cgo exports from my go code as part of a dynamically loaded library in R packages. For this, I need the code to be position independent.

I have now created an example project demonstrating the two approaches (one using `go build` and another using rules_go) at https://github.com/siddharthab/rgo-example.

We have a rapidly growing base of go code for bioinformatics at https://github.com/grailbio, and we would like to expose some of this through interfaces in R language, which is a popular language in the bioinformatics community.

If you will be so kind to take a look and suggest changes to get position independent code, I will be grateful.

Thank you.

siddharthab:
[...]

PR #1556, opened 2018-06-17
build c-archive as PIC

siddharthab:
Resolves https://github.com/bazelbuild/rules_go/issues/1470.

For rationale, see [discussion on golang-dev](https://groups.google.com/forum/#!topic/golang-dev/lI82ONL2ALg), and the [subsequent patch](https://go-review.googlesource.com/c/go/+/24180).

Tested with https://github.com/siddharthab/rgo-example.


steeve:
This looks great, however before merging I'd like to test it with an iOS target if that's ok with y'all.

PR #1566, opened 2018-06-21
Add srcs attribute to GoStdLib provider

siddharthab:
The src directory is not added to the files attribute so it needs to be
explicitly depended upon. Therefore, this change does not have any
effect on downstream rules, and does not add to the sandboxing overhead.

The src directory is useful for some linter invocations which can now
depend on sandboxed stdlib instead of expecting the users to have an
installed stdlib.

### DIFF
--- a/go/private/rules/stdlib.bzl
+++ b/go/private/rules/stdlib.bzl
@@ -32,6 +32,7 @@ load(
 
 def _stdlib_library_to_source(go, attr, source, merge):
     pkg = go.declare_directory(go, "pkg")
+    src = go.declare_directory(go, "src")
     root_file = go.declare_file(go, "ROOT")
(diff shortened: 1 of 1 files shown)


jayconrod:
As a heads up, `GoStdLib` is not part of our [documented provider interface](https://github.com/bazelbuild/rules_go/blob/master/go/providers.rst), and it will likely change in the future. Hopefully not too much.

I've been working on a change that makes use of the prebuilt standard libraries in the Go SDK but still allows other variants of the SDK to be built as needed. So there will be a set of rules that provide `GoStdLib`, and at that point it will be public. It will likely still be accessible through the `go_context` object.

siddharthab:
Thank you! I am aware that the interface might change, and I will look out for it.

PR #1572, opened 2018-06-26
Windows fixes

pmuetschard:
- Don't expect a `cp` command on Windows
- `syscall.Exec` on Windows always returns an error
- Fix path comparison in the linker
- Allow passing deps to the linker in a file rather than the command line, to allow for complex projects with lots of packages
- Fix a race condition in pack.go (doesn't affect other platforms due to sandboxes)

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -132,6 +132,8 @@ def emit_link(
     if extldflags:
         tool_args.add(["-extldflags", " ".join(extldflags)])
 
+    builder_args.use_param_file("@%s")
+    builder_args.set_param_file_format("multiline")
     go.actions.run(
         inputs = sets.union(
             archive.libs,
--- a/go/private/tools/overlay_repository.bzl
+++ b/go/private/tools/overlay_repository.bzl
@@ -93,7 +93,7 @@ def _apply_overlay(ctx, overlay):
 
     # TODO(jayconrod): sanitize destination paths.
     for src_path, dst_rel in overlay:
-        _check_execute(ctx, ["cp", src_path, dst_rel], "failed to copy file from %s" % src_path)
+        ctx.template(dst_rel, src_path)
 
 def _check_execute(ctx, arguments, message):
     res = ctx.execute(arguments)
--- a/go/tools/builders/link.go
+++ b/go/tools/builders/link.go
@@ -98,7 +98,7 @@ This will be an error in the future.`, pkgPath, label, conflictLabel)
 			return fmt.Errorf("package file name %q must have searchable suffix %q", pkgFile, pkgSuffix)
 		}
 		searchPath := pkgFile[:len(pkgFile)-len(pkgSuffix)]
-		goargs = append(goargs, "-L", searchPath)
+		goargs = append(goargs, "-L", abs(searchPath))
 	}
 	for _, xdef := range xstamps {
 		split := strings.SplitN(xdef, "=", 2)
--- a/go/tools/builders/link.go
+++ b/go/tools/builders/link.go
@@ -94,7 +94,7 @@ This will be an error in the future.`, pkgPath, label, conflictLabel)
 		depsSeen[pkgPath] = label
 
 		pkgSuffix := string(os.PathSeparator) + filepath.FromSlash(pkgPath) + ".a"
-		if !strings.HasSuffix(pkgFile, pkgSuffix) {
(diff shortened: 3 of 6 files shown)


go/private/actions/link.bzl
```diff
@@ -83,7 +83,9 @@ def emit_link(
         "{}={}={}".format(d.label, d.importmap, d.file.path)
         for d in test_archives
     ])
-    builder_args.add(dep_args, before_each = "-dep")
+    depsFile = go.actions.declare_file(executable.basename + ".deps", sibling = executable)
```
jayconrod: I think there are two better potential solutions for this:

1. Ideal long term solution: we should write an importcfg file. We no longer support Go 1.8, so nothing is blocking this. The main issue is that for error reporting, we should include the Bazel label for each package. I guess we could write that in a comment within the importcfg file.
2. Short term solution: we could use [Bazel's support for param files](https://docs.bazel.build/versions/master/skylark/lib/Args.html#use_param_file). Some common functionality could be added in go/tools/builders/env.go so that all of the builders would benefit from this.

go/private/tools/overlay_repository.bzl
```diff
@@ -93,7 +93,7 @@ def _apply_overlay(ctx, overlay):
 
     # TODO(jayconrod): sanitize destination paths.
     for src_path, dst_rel in overlay:
-        _check_execute(ctx, ["cp", src_path, dst_rel], "failed to copy file from %s" % src_path)
+        ctx.symlink(src_path, dst_rel)
```
jayconrod: I'm not confident a symlink will always work here: I've been told they may require administrative privilege to create on NTFS. Also, we may need to replace files, and I don't think `ctx.symlink` does that. I don't have my Windows machine with me at the moment, so I can't test it out right now.

I was tinkering with this a few days ago, and the best thing I came up with was using the `BAZEL_SH` environment variable to locate msys2 bash, then looking in the same directory for cp. However, it seems like the requirement to set `BAZEL_SH` has been removed from the installation instructions, so maybe I shouldn't rely on that.

Maybe the best thing to do is create a .bat file with `copy` commands, then execute that?

In any case, this file was copied from `github.com/bazelbuild/bazel-gazelle/internal/overlay_repository.bzl`. It should be modified there first, and the changes should be backported here.

go/tools/builders/link.go
```diff
@@ -94,7 +105,7 @@ This will be an error in the future.`, pkgPath, label, conflictLabel)
 		depsSeen[pkgPath] = label
 
 		pkgSuffix := string(os.PathSeparator) + filepath.FromSlash(pkgPath) + ".a"
-		if !strings.HasSuffix(pkgFile, pkgSuffix) {
+		if !strings.HasSuffix(filepath.FromSlash(pkgFile), pkgSuffix) {
```
jayconrod: Is this needed? It looks like this should come from `d.file.path` in link.bzl:78. I would expect that to have the correct separator, but I can't verify at the moment.

go/private/tools/overlay_repository.bzl
dwatley: Yes. Specifically, SeCreateSymbolicLinkPrivilege is a right given by default only to users in the Administrators group.

pmuetschard: `rules_go` uses `repository_ctx.symlink` already in a few places. Bazel uses it all over the place. I believe bazel actually creates junctions for directories and copies for files. Either way, without  `repository_ctx.symlink` nothing would work.

go/tools/builders/link.go
pmuetschard: Yes it is needed. It comes in with `/` on Windows, but `pkgSuffix` uses the OS specific separator.

go/private/actions/link.bzl
pmuetschard: Done. Updated the change to use bazel's params files.

go/private/tools/overlay_repository.bzl
pmuetschard: I've sent https://github.com/bazelbuild/bazel-gazelle/pull/246, making the same change to bazel-gazelle.

pmuetschard:
Since the change to `bazel-gazelle` is submitted, and I've made the mirror change here, would it be possible to get this submitted? or is there some other change you would like me to make?

jayconrod:
Merged. Thanks again for doing this!

PR #1550, opened 2018-06-14
Arm64 support on Linux

ArielleA:
Arm64 support fixes #1506  requires pull request bazelbuild/bazel#5360

Issue #1506, opened 2018-05-17
arm64 support 


vielmetti:
As noted in https://github.com/lyft/protoc-gen-validate/issues/75#issuecomment-389946862 , the operating system automatic selection system at

https://github.com/bazelbuild/rules_go/blob/master/go/private/sdk.bzl#L29-L38

misidentifies an linux_arm64 system as linux_amd64 (because there is no case that properly matches the `uname -p` call). As it is written in that mentioned issue by @akonradi:

> The issue is that if repository_ctx.os.name is 'linux', there's no way to set the host as linux_arm64 - it defaults to linux_amd64. See
https://github.com/bazelbuild/rules_go/blob/3b103d6954cd4508e1f63c8e2ea57bda18d98bcf/go/private/sdk.bzl#L31-L37

vielmetti:
The similar issue for ppc64le was resolved at https://github.com/bazelbuild/rules_go/issues/854 and https://github.com/bazelbuild/rules_go/issues/1186 with a note also of changes near https://github.com/bazelbuild/rules_go/blob/master/go/platform/list.bzl#L34

jayconrod:
At the moment, I don't have an arm64 machine to develop or verify this. I'm happy to accept a PR though.

vielmetti:
Thanks @jayconrod - @moderation is working on this at https://github.com/moderation/rules_go/commit/20b05a8b4a635dd5ad822c4b8f5bbbb4a448d877 and has a test environment of some sort set up.

I'm happy to provide short or long term access to arm64 hardware via the Works on Arm cluster, http://worksonarm.com/cluster or http://github.com/worksonarm/cluster/issues/new .

vielmetti:
There are notes on a port of Bazel to arm64 here http://zhiyisun.github.io/2017/02/15/Running-Google-Machine-Learning-Library-Tensorflow-On-ARM-64-bit-Platform.html but unfortunately they were never upstreamed.

PR #1575, opened 2018-06-27
Allow go_download_sdk work on ARM.

likan999:


vielmetti:
Reference #1506 "arm64 support"

jayconrod:
I think this change is a subset of #1550, but please correct me if I'm wrong.

In any case, that change was blocked on the release of Bazel 0.15.0 (which is now out). I'd like to do one more major release of rules_go (likely early next week) before making Bazel 0.15.0 the minimum supported version.

likan999:
Oh, that’s nice. I didn’t realize the existence of the other PR. I’ll close this one.

PR #1585, opened 2018-07-04
Move legacy reproducibility test to new setup

siddharthab:
Also included:
1. Adds an LLVM toolchain to the WORKSPACE file that can be used to
manually test changes against clang on linux.
2. Adds os.PathSeparator to stripped absolute paths to be consistent
with stripping being done everywhere else.
3. Adds a test to check the string "bazel-sandbox" in the binary.
4. Uses the cgo binary instead of a pure binary to check for
reproducibility in the go_test.
5. Tags the target "collect_digests" as manual.

To see how binaries are not reproducible with clang and `-g`, use
```
bazel test --copt=-g --crosstool_top=@llvm_toolchain//:toolchain \
  //tests/reproducibility:go_default_test
```


jayconrod:
Thanks for submitting!

I'm traveling for the rest of the week, so I won't have time to evaluate the LLVM toolchain part of this, probably until Monday. The rest looks good.

FYI, the reproducibility test is very slow. The Travis CI failure is a timeout on that. I expect a lot of this is Bazel slowness in a new workspace, but I may temporarily disable that test until I have time to investigate whether it can be sped up.

PR #1598, opened 2018-07-19
Update feature/check branch to master

jayconrod:
This merges feature/check with master in several steps. This is a bit more complicated than a normal merge because I ran buildifier on all .bzl files after this branch diverged.

1. `617a25d Merge commit '6ed1cd1' into check-fmt` - Update to the commit right before buildifier was run.
2. `108c0d4 default_checker: use GOROOT instead of relying on -stdlib flag` - Fix an issue where the checker was reading .a files for the host platform instead of the target platform (still not completely resolved, but necessary to proceed).
3. `d0dcad2 Format modified .bzl files with buildifier` - Prepare for buildifier merge. The only differences at this point should be the changes made on `feature/check`, and whitespace changes won't matter.
4. `5f7a859 Merge commit '0b731fe' into check-fmt` - Merge with `master` right after buildifier was run.
5. `c26ea7b Merge branch 'master' into check-fmt` - Merge with tip of `master`. The only differences at this point should be the changes made on `master` after buildifier was run.


jayconrod:
cc @stjj89

PR #1610, opened 2018-07-27
Optimize args and inputs construction

ash2k:
Applied advice from https://docs.bazel.build/versions/master/skylark/performance.html to some of the files. WDYT?


go/private/actions/link.bzl
```diff
@@ -24,6 +24,30 @@ load(
     "LINKMODE_PLUGIN",
 )
 
+def _format_archive(d):
+    return "{}={}={}".format(d.label, d.importmap, d.file.path)
+
+def _map_archive(x):
```
jayconrod: I'd rather keep this the way it was. `test_archives` will pretty much always be 0 or 1 elements, so the filtering using `any` before should actually be very fast (probably faster than building a `dict` and doing hash table lookups). I think it was shorter and more readable before, so I prefer that.

I like having `_format_archive` as a separate function called with `map_each` though.

go/private/actions/link.bzl
```diff
@@ -24,6 +24,30 @@ load(
...
+        t.importmap: None
+        for t in x.test_archives
+    }
+    return [
+        _format_archive(d)
+        for d in deps.to_list()
+        if d.importmap not in test_importmaps
+    ]
+
+def _map_test_archive(d):
```
jayconrod: Any reason to have this as a separate function?

go/private/actions/compile.bzl
```diff
@@ -53,9 +53,9 @@ def emit_compile(
             fail("compile does not accept deps in bootstrap mode")
         return _bootstrap_compile(go, sources, out_lib, gc_goopts)
 
-    inputs = (sources + [go.package_list] +
-              [archive.data.file for archive in archives] +
-              go.sdk_tools + go.stdlib.files)
+    inputs = sets.union(sources, [go.package_list],
```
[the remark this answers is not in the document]
ash2k: It is totally possible that my understanding is incorrect, but I believe that there are these benefits of using a depset in such a situation:
1. Memory savings - list is not concatenated when the rule is executed and hence the memory is not allocated to hold it. Instead, depset is just holding references to parts (lists/dpsets).
2. Execution time savings - I assume that constructing a depset is cheap, faster than list concatenation.

Eventually, of course, depset needs to be expanded into a (deduplicated) list and that's when the price is paid - both for deduplication and for memory allocation. However, that price is only paid if that action needs to be executed and never otherwise. And most of the time many actions do not need to run - caching or the output is not required. That is why I believe it is much better to use depsets - memory/cpu cost of analysis (or loading? not sure) phase is reduced and cost is postponed to the execution phase, but only for those actions where it needs to be paid (i.e. where the output is needed).

I may be operating under a wrong assumption here, but I don't see how __constructing__ a depset can be expensive - it would not make any sense, I think. [Performance](https://docs.bazel.build/versions/master/skylark/depsets.html#performance) section does not say it directly, but I read it as if it says that `to_list()` is expensive, not construction of a depset.

go/private/actions/compile.bzl
```diff
@@ -101,12 +101,15 @@ def emit_compile(
...
+    args.add_all(gc_goopts)
+    args.add_all(sources)
     go.actions.run_shell(
-        inputs = sources + go.sdk_files + go.sdk_tools,
+        inputs = sets.union(sources, go.sdk_files, go.sdk_tools),
         outputs = [out_lib],
+        arguments = [args],
         mnemonic = "GoCompile",
-        command = "export GOROOT=$(pwd)/{} && export GOROOT_FINAL=GOROOT && {} {}".format(go.root, go.go.path, " ".join(args)),
+        command = "export GOROOT=$(pwd)/{} && export GOROOT_FINAL=GOROOT && {} \"$@\"".format(go.root, " ".join(cmd)),
```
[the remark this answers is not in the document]
ash2k: This also passes arguments properly - no chance that an unescaped character breaks the script.

go/private/actions/link.bzl
ash2k: Ok, makes sense - I'll change it back to `any(...)`.
The main motivation to extract this code into a function it to postpone the cost (depset expansion and execution of this code) to the execution phase and, likely, not pay it at all if the output is not needed. See https://docs.bazel.build/versions/master/skylark/performance.html#use-ctx-actions-args-for-command-lines
Basically Args is a way to postpone and avoid the cost of constructing the arguments list. The more processing is delegated to it the better. Currently each instance of the rule where this function is used pays the cost. After this refactoring only the ones where the argument list is actually needed will pay the cost. Does it make sense?

go/private/actions/link.bzl
ash2k: No, I just thought it is more readable that way. Happy to just use `_format_archive` directly.

go/private/actions/link.bzl
```diff
@@ -67,23 +91,9 @@ def emit_link(
...
-        if not any([d.importmap == t.importmap for t in test_archives])
-    ]
-    dep_args.extend([
-        "{}={}={}".format(d.label, d.importmap, d.file.path)
-        for d in test_archives
-    ])
-    builder_args.add_all(dep_args, before_each = "-dep")
+    builder_args.add_all([struct(archive = archive, test_archives = test_archives)], before_each = "-dep",
+        map_each = _map_archive)
+    builder_args.add_all(test_archives, before_each = "-dep", map_each = _map_test_archive)
```
ash2k: Instead of doing the work inline here, we postpone it. It will only be done if the action needs to be run.

ash2k:
I've rebased on master and got the following warning when running `bazel test`:
```console
INFO: From GoLink tests/core/cgo/objc/darwin_amd64_stripped/objc_test:
ld: warning: URGENT: building for OSX, but linking in object file (/var/folders/fh/7zp6kdcj3rgdd_7_nmw439_st0zyrt/T/go-link-075800598/000003.o) built for iOS. Note: This will be an error in the future.
ld: warning: object file (/var/folders/fh/7zp6kdcj3rgdd_7_nmw439_st0zyrt/T/go-link-075800598/000003.o) was built for newer iOS version (11.4) than being linked (10.13)
ld: warning: URGENT: building for OSX, but linking in object file (/var/folders/fh/7zp6kdcj3rgdd_7_nmw439_st0zyrt/T/go-link-075800598/000004.o) built for iOS. Note: This will be an error in the future.
ld: warning: object file (/var/folders/fh/7zp6kdcj3rgdd_7_nmw439_st0zyrt/T/go-link-075800598/000004.o) was built for newer iOS version (11.4) than being linked (10.13)
ld: warning: ObjC object file (/var/folders/fh/7zp6kdcj3rgdd_7_nmw439_st0zyrt/T/go-link-075800598/000003.o) was compiled for iOS Simulator, but linking for MacOSX
```
I'm on latest macOS, using bazel 0.15.2. Is it my branch or there is a bug in master or something else?

go/private/actions/compile.bzl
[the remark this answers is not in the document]
jayconrod: Sadly, implementation of `depset` is inefficient. It's good to use if you know the asymptotic complexity is strictly better, but compared with something that with the same complexity (list concatenation in thise case), it's almost always worse by a constant factor.

The underlying Java class is [NestedSet](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/collect/nestedset/NestedSet.java;l=43?q=NestedSet&ss=bazel). That one is used directly within Bazel. [SkylarkNestedSet](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/syntax/SkylarkNestedSet.java;l=88?q=NestedSet&ss=bazel) is a layer on top of that.

When you create a `NestedSet`, the constructor will de-duplicate the direct elements and the transitive sets using a hash table, so construction is O(len(direct) + len(transitive)) in time and space. Several copies are made, and all the direct elements are hashed at least once, so it's a pretty high constant factor, too.

The crucial thing is that the len(transitive) above is the number of transitive sets passed to the constructor, not the number of elements in those sets. So when you're building a big subgraph of values for linking, it's much more efficient than list concatenation.

go/private/actions/link.bzl
jayconrod: I have a mild preference for using `_format_archive` directly.

I was planning to migrate these arguments into files that could be passed to `-importcfg` directly. Given what you've said about Args though, maybe it makes sense keep this and do the work in the builder, since it might not need to be done.

jayconrod:
The iOS error is a bug in master. I've seen it before, but I don't think there was an issue, so I filed #1616 to track it.

go/private/actions/compile.bzl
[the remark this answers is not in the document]
ash2k: Thanks for going into details and explaining! I appreciate that very much. I'll undo the change here and in other places.

ash2k:
I've made the requested change but one of the tests has failed without printing the reason. I think it might be just a network timeout - that test passes on my computer. It also did pass in the other build (on linux).

jayconrod:
I restarted Travis CI on macOS. I think I've seen that test flake in the past though, and I don't think it's because of this change.

PR #1614, opened 2018-07-28
Pass tests if rules_go is an external repository

Yannic:
Some tests failed if they were ran with rules_go as external repository
(`bazel test @io_bazel_rules_go//...` instead of `bazel test //...`).
Passing all tests if ran like that is useful for users who import
rules_go into their repository and have local modifications they want
to test.


ash2k:
To run tests in rules_go why not just do `bazel test //...` when inside the directory of the fork? To test modifications to rules_go when used from some other project, why not use [`local_repository`](https://docs.bazel.build/versions/master/be/workspace.html#local_repository) to import the fork? I've done that a few times and it work well =)

Yannic:
That`s what we currently do. rules_go is vendored as `third_party/io_bazel_rules_go` and imported with `local_repository`. However, this workflow isn't great because changes to rules_go can break other (vendored) repositories (seen this in the past for other repositories). `cd`-ing to all of them, update their version of rules_go to the local version, and `bazel test //...` them just doesn't feel like a good solution to me.

.bazelci/presubmit.yml
```diff
@@ -3,15 +3,21 @@ platforms:
   ubuntu1404:
     build_targets:
     - "..."
+    - "@io_bazel_rules_go//..."
```
jayconrod: Let's test `@io_bazel_rules_go//...` in here and `//...` in Travis CI.

I'd like to have coverage for both, but I don't want to make the presubmit 2x slower.

Ideally Bazel would recognize these as the same targets :(

extras/bindata.bzl
```diff
@@ -65,6 +68,7 @@ bindata = go_rule(
             cfg = "data",
         ),
         "package": attr.string(mandatory = True),
+        "experimental_strip_extenal": attr.bool(default = False),
```
jayconrod: This exposes too much implementation detail in the API. Maybe a more general attribute, `prefix` would be better?

go/tools/bazel/bazel_test.go
```diff
@@ -70,7 +70,7 @@ func createPaths(paths []string) error {
 }
 
 func TestRunfile(t *testing.T) {
-	file := "go/tools/bazel/README.md"
+	file := "README.md"
```
jayconrod: I think if this is changed to a file in the test's working directory, it won't cover the case where `Runfile` needs to look in the runfiles directory, which is what we really want to test.

I'm not sure whether `Runfile` itself will need to change in order for tests to use it when run from the local workspace and from "remote", named workspaces.

tests/bazel_tests.bzl
```diff
@@ -142,8 +142,7 @@ def _bazel_test_script_impl(ctx):
     register = ""
     if any([ext.label.workspace_root.endswith("/go_sdk") for ext in ctx.attr.externals]):
         register += "register_toolchains(\n{}\n)\n".format(
-            "\n".join(['    "@go_sdk//:{}",'.format(name)
-                       for name in generate_toolchain_names()]))
+            "\n".join(['"@go_sdk//:{}",'.format(name) for name in generate_toolchain_names()]))
```
jayconrod: Why remove the indentation in the generated call to `register_toolchains`? It's not strictly necessary, but it makes it more readable.

tests/bazel_tests.bzl
```diff
@@ -169,32 +168,27 @@ def _bazel_test_script_impl(ctx):
     build_file = go.declare_file(go, path = "BUILD.in")
     ctx.actions.write(build_file, ctx.attr.build)
 
-    output = "external/" + ctx.workspace_name + "/" + ctx.label.package
-    targets = ["@" + ctx.workspace_name + "//" + ctx.label.package + t if t.startswith(":") else t for t in ctx.attr.targets]
+    output = "/".join([ctx.label.workspace_root if ctx.label.workspace_root else "external/{}".format(ctx.workspace_name), ctx.label.package])
```
jayconrod: In general, I'd prefer to keep the string concatenation for paths where there are a known number of elements. I think `"/".join([...])` is less readable.

tests/bazel_tests.bzl
```diff
@@ -218,8 +214,8 @@ _bazel_test_script = go_rule(
                 "run",
             ],
         ),
-        "args": attr.string_list(default = []),
-        "targets": attr.string_list(mandatory = True),
+        "args2": attr.string_list(default = []),
```
jayconrod: Why `args2`?

tests/bazel_tests.bzl
```diff
@@ -204,10 +198,12 @@ def _bazel_test_script_impl(ctx):
             [workspace_file, build_file] + ctx.files.extra_files,
             collect_data = True,
         ),
+        executable = script_file,
     )
 
-_bazel_test_script = go_rule(
+_bazel_test_script_test = go_rule(
```
jayconrod: Please don't rename this or set `test = True` or return `executable = script_file` above. This is a rule that produces a script that may be run in a test. It is not a test itself because it doesn't produce an executable.

Sorry for the annoying distinction: on Windows, files can't be marked executable, and shell scripts by themselves are not executable. This is why we need to wrap the `native.sh_test` around the generated script in the `bazel_test` macro below.

tests/core/cross/cross_test.go
```diff
@@ -57,10 +57,15 @@ var checks = []check{
 
 func TestCross(t *testing.T) {
 	for _, c := range checks {
-		if _, err := os.Stat(*c.file); os.IsNotExist(err) {
-			t.Fatalf("Missing binary %v", *c.file)
+		file := *c.file
```
jayconrod: I think you should be able to use `bazel.Runfile(*c.file)` instead of adding this logic (`bazel` is `github.com/bazelbuild/rules_go/go/tools/bazel`).

tests/core/go_path/go_path_test.go
```diff
@@ -84,6 +84,10 @@ func TestArchivePath(t *testing.T) {
 	}
 	defer os.RemoveAll(dir)
 
+	if strings.HasPrefix(archivePath, "external") {
```
jayconrod: Same comment about `bazel.Runfile`.

tests/core/race/race_on_test.go
```diff
@@ -23,6 +25,9 @@ func TestRaceTag(t *testing.T) {
 }
 
 func checkRaceBinary(t *testing.T, bin string) {
+	if strings.HasPrefix(bin, "external") {
```
jayconrod: Same comment about `bazel.Runfile`

tests/legacy/go_embed_data/embed_test.go
```diff
@@ -82,6 +83,10 @@ func TestUnpack(t *testing.T) {
 }
 
 func checkFile(t *testing.T, path string, data []byte) {
+	if strings.HasPrefix(path, "external/") {
```
jayconrod: Same comment about `bazel.Runfile`.

tests/reproducibility/reproducible_binary_test.go
```diff
@@ -37,7 +39,11 @@ func TestMain(m *testing.M) {
 		log.Fatalf("usage: %s <binary>\n", os.Args[0])
 	}
 
-	binaryData, err := ioutil.ReadFile(os.Args[1])
+	binary := os.Args[1]
```
jayconrod: Same comment about `bazel.Runfile`.

extras/bindata.bzl
Yannic: While it would be possible to add a `prefix` attribute, this solves a more general problem not limited to `rules_go` and the scope of this PR. I'm not sure if `bindata` works without `prefix` for external repositories and I don't think `prefix` would be used by a lot of users, leaving consumers of these rules with broken artifacts.

I hoped we could turn this on by default eventually and remove it afterward. WDYT?

extras/bindata.bzl
```diff
@@ -43,7 +43,10 @@ def _bindata_impl(ctx):
         arguments.add_all(["-modtime", "0"])
     if ctx.attr.extra_args:
         arguments.add_all(ctx.attr.extra_args)
-    arguments.add_all(ctx.files.srcs)
+    srcs = [f.path for f in ctx.files.srcs]
+    if ctx.attr.experimental_strip_external and [f for f in srcs if f.startswith("external/")]:
+        arguments.add_all(["-prefix", "/".join([ctx.label.workspace_root, ctx.label.package])])
```
ash2k: This all can be done lazily to avoid constructing `srcs` in the analyzis phase. Something like:
```bzl
def _maybe_add_prefix(t):
    if [f for f in t.srcs if f.path.startswith("external/")]:
        # use paths.join() from skylib to construct path
        return ["-prefix", paths.join(t.workspace_root, t.package)]
    else:
        return None
if ctx.attr.experimental_strip_external:
    arguments.add_all([struct(srcs = ctx.files.srcs, workspace_root = ctx.label.workspace_root, package = ctx.label.package)], map_each = _maybe_add_prefix)
arguments.add_all(ctx.files.srcs)
``

File: .bazelci/presubmit.yml (hunk 1 above)
Comment (jayconrod) [reply .bazelci/presubmit.yml]: It looks like this is still going to test everything twice in BuildKite.

File: extras/bindata.bzl [hunk 14]
```diff
@@ -65,6 +68,7 @@ bindata = go_rule(
             cfg = "data",
         ),
         "package": attr.string(mandatory = True),
+        "experimental_strip_external": attr.bool(default = False),
```
jayconrod: Let's just make this `strip_external`.

extras/bindata.bzl
```diff
@@ -43,7 +43,10 @@ def _bindata_impl(ctx):
         arguments.add_all(["-modtime", "0"])
     if ctx.attr.extra_args:
         arguments.add_all(ctx.attr.extra_args)
-    arguments.add_all(ctx.files.srcs)
+    srcs = [f.path for f in ctx.files.srcs]
+    if ctx.attr.experimental_strip_external and [f for f in srcs if f.startswith("external/")]:
```
jayconrod: Readability nit:

```
if ctx.attr.experimental_strip_external and any([f.startswith("external/") for f in srcs]):
    arguments.add("-prefix", ctx.label.workspace_root + "/" + ctx.label.package)
```

extras/bindata.bzl
jayconrod: This would make sense if were were accumulating a depset of sources over a potentially large tree, but I think it's overkill in this situation.

tests/bazel_tests.bzl
```diff
@@ -219,7 +223,7 @@ _bazel_test_script = go_rule(
             ],
         ),
         "args": attr.string_list(default = []),
-        "targets": attr.string_list(mandatory = True),
+        "targets": attr.label_list(mandatory = True),
```
jayconrod: Is there a way to avoid making this `label_list`? That will cause the targets to be analyzed in the workspace where `bazel test` is invoked, which isn't strictly necessary.

Yannic: Unfortunately not. We need the absolute name of the targets, which we can only get if we make this `label_list`.

Constructing the absolute name as `[Label(t) for t in targets]` in `bazel_test` does not work.

jayconrod:
Closing this PR since it has been inactive for a long time.

Yannic:
@jayconrod This PR was inactive because I was waiting for another review. I think I addressed all your comments in 8e66889. PTAL

jayconrod:
@Yannic Sorry I thought some tests were still broken. I'll take another look.

jayconrod:
@Yannic It looks like `//tests/core/coverage:coverage_test_test` is broken on multiple platforms. This may be related to the change in `bazel_test`. Would you mind updating to `master` and taking a look?

Yannic:
Updated to master.

I'm not sure whether the failure is related to this PR because the error seems to occur during execution of `collect_coverage.sh` inside `@bazel_tools`, and I'm unable to reproduce it locally. Could we just re-run the test suite?
```
+ exec --coverage_dir=/private/var/tmp/_bazel_buildkite/ff412c23b472f1e3662dfa0fb82a066c/execroot/bazel_test/_coverage/external/io_bazel_rules_go/tests/core/coverage/coverage_test/test --output_file=/private/var/tmp/_bazel_buildkite/ff412c23b472f1e3662dfa0fb82a066c/execroot/bazel_test/bazel-out/darwin-fastbuild/testlogs/external/io_bazel_rules_go/tests/core/coverage/coverage_test/coverage.dat
external/bazel_tools/tools/test/collect_coverage.sh: line 132: exec: --: invalid option
exec: usage: exec [-cl] [-a name] file [redirection ...]
Target @io_bazel_rules_go//tests/core/coverage:coverage_test up-to-date:
  bazel-bin/external/io_bazel_rules_go/tests/core/coverage/darwin_amd64_stripped/coverage_test
INFO: Elapsed time: 1,904s, Critical Path: 0,38s
INFO: 11 processes: 11 local.
INFO: Build completed, 1 test FAILED, 16 total actions
@io_bazel_rules_go//tests/core/coverage:coverage_test                    FAILED in 0.1s
  /private/var/tmp/_bazel_buildkite/ff412c23b472f1e3662dfa0fb82a066c/execroot/bazel_test/bazel-out/darwin-fastbuild/testlogs/external/io_bazel_rules_go/tests/core/coverage/coverage_test/test.log
  /private/var/tmp/_bazel_buildkite/ff412c23b472f1e3662dfa0fb82a066c/execroot/bazel_test/bazel-out/darwin-fastbuild/testlogs/external/io_bazel_rules_go/tests/core/coverage/coverage_test/coverage.dat

Executed 1 out of 1 test: 1 fails locally.
INFO: Build completed, 1 test FAILED, 16 total actions
```

jayconrod:
Looks like everything is working. Thanks, and sorry for the long delay.

PR #1638, opened 2018-08-08
Add CI config to test on RBE.

xingao267:


jayconrod:
@xingao267 It looks like there are a lot of failures. Is that expected?

I'm not able to see the results.

xingao267:
It's because those tests are not RBE compatible yet. The main purpose of the PR is to onboard rules to start testing with RBE. So for now we'll just skip them. I'm working on it.

xingao267:
You may be able to the test results in the buildkite link?

Some tests fail due to a flag being improperly set, I'm fixing that as well.

jayconrod:
Thanks for fixing, looks good.

Could you give me access to the BES page though? If something fails with RBE in the future, I may not be able to debug it without logs. The BuildKite page shows me the Bazel invocation logs, but not the individual test logs.

xingao267:
Thanks for reviewing!

@philwo or @buchgr should be able to give you access to the BES page. Actually, @philwo or @buchgr could you also give me and @nlopezgi access to the BES page as well? We don't have access either, so we can't see the logs of CI on RBE for other repos such as rules_docker or bazel-toolchains.

philwo:
Just to clarify: With "BES page" do you mean the Build Status result page of e.g. "bazel test (Ubuntu 14.04)"? Like this one: https://source.cloud.google.com/results/invocations/946b578a-0bb7-42ae-b80c-e737e23e0c41/targets?

I think that's supposed to "just work" as our Build Status results are enabled for public anonymous access, but somehow the results for the RBE run are missing. I'll investigate tomorrow. We probably just have to tweak some flags.

xingao267:
Yes. The Build Status result page for RBE runs.

That link shows a "Oops, page not found." to me as well.

Sounds good. Thank you @philwo .

nlopezgi:
@philwo, yes, the result store UI is what Xin refers to. We do want it to be public anonymous so that all rule owners (and folks who contribute PRs) can debug why their builds are failing on RBE. Please let us know, Thanks!

WORKSPACE
```diff
@@ -39,3 +40,12 @@ llvm_toolchain(
     name = "llvm_toolchain",
     llvm_version = "6.0.0",
 )
+
+http_archive(
+    name = "bazel_toolchains",
+    sha256 = "7938215442cc96b1a16809f4643db14af7d91c3ef1e98d39593bb46862204537",
+    strip_prefix = "bazel-toolchains-809c13c3a96966f63c7a43c586cb4e72589184b6",
+    urls = [
+        "https://github.com/bazelbuild/bazel-toolchains/archive/809c13c3a96966f63c7a43c586cb4e72589184b6.tar.gz",
```
nlopezgi: can we pin them to the latest release instead of this commit?

xingao267: Will do that in a separate PR.

philwo:
@xingao267 @nlopezgi I've enabled BES / Build Status reporting for rbe_ubuntu1604 - there were a couple of flags missing. For new PRs (or when tests get re-run in existing PRs) it should just work from now on. Let me know if you see any more issues!

nlopezgi:
Thanks, tested successfully with https://github.com/bazelbuild/bazel-toolchains/pull/158 (https://source.cloud.google.com/results/invocations/375e325c-0a05-47af-87bd-fed1363e0333/targets)

PR #1663, opened 2018-08-15
Remove uses of deprecated dictionary concatenation

jjudd:
This removes the need for --incompatible_disallow_dict_plus


jjudd:
I was updating our internal version of Bazel to 0.16 and these errors showed up with `--all_incompatible_changes` enabled, so I fixed them.

PR #1664, opened 2018-08-16
Force absolute paths in builders

Xjs:
This is a follow-up to https://github.com/bazelbuild/rules_go/pull/1647. `GOROOT` is still injected as relative path, which is considered bad practice anyway elsewhere [1]. This patch converts the importmap paths to absolute paths even if `GOROOT` is relative. (Will this break anything on any platform? E. g. OSX?)

[1] https://github.com/golang/go/issues/26917#issuecomment-412814805

### DIFF
--- a/go/tools/builders/compile.go
+++ b/go/tools/builders/compile.go
@@ -186,6 +186,7 @@ func buildImportcfgFile(archives []archive, stdImports []string, installSuffix, 
 	if !ok {
 		return "", errors.New("GOROOT not set")
 	}
+	goroot = abs(goroot)
 	for _, imp := range stdImports {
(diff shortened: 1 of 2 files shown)


jayconrod:
Looks, good thanks.

I don't expect absolute paths in importcfg files to break anything. It may have a tiny overhead if the tools need to read larger files and traverse more of the file system, but I expect that's negligible.

About the comment for `GOROOT` being an absolute path, it's required to be an absolute path by `go build` and `go installl`, which we only use to build the standard library. The compiler and linker don't read `GOROOT`, so its value doesn't matter for them. We must set `GOROOT` to a relative path when invoking the builders, since analysis code can't produce absolute paths (we don't know where actions will run at that point), but builders can absolutize it if needed.

PR #1673, opened 2018-08-20
Allow the checker to run vet

stjj89:
Add a go_checker attribute that causes the generated checker
binary to run 'go tool vet'.

Just like in the official Go build toolchain, only a subset of
vet checks are run. Any vet findings will fail compilation. To avoid
breaking users, vet is disabled by default.

While there:
- Fix issues with the importer used to load analysis packages.
- Silence type-checking errors generated by the checker binary. It will
  now fail silently if type-checking errors occur.
- Refactor the checker binary and unit tests to improve readability.


go/tools/builders/vet.go
```diff
@@ -0,0 +1,184 @@
...
+			res = append(res, line)
+		}
+	}
+	return res
+}
+
+// parseArchivesAndImportMap parses the -archivefile and -importmap arguments
+// created by the compile rule into the packageFiles and importMap maps used
+// by the vet and checker importers.
+func parseArchivesAndImportMap(archiveFiles, importMapIn []string) (packageFiles map[string]string, importMap map[string]string, err error) {
```
jayconrod: This function should probably be in checker_main.go, since its output is also used for the importer.

go/tools/builders/vet.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		packageFiles[path] = kv[1]
+	}
+	return
+}
+
+// makeVetCfg creates a vet.cfg file and returns its file path. It is the
+// caller's responsibility to remove this file when it is no longer needed.
+func makeVetCfg(packageList, stdLib string, archiveFiles, importMap, files []string) (string, error) {
+	pf, im, err := parseArchivesAndImportMap(archiveFiles, importMap)
```
jayconrod: `parseArchivesAndImportMap` does a fair amount of work, but it's called twice when vet is enabled. Can its output be passed into `makeVetCfg` instead?

go/tools/builders/vet.go
```diff
@@ -0,0 +1,184 @@
...
+// parseArchivesAndImportMap parses the -archivefile and -importmap arguments
+// created by the compile rule into the packageFiles and importMap maps used
+// by the vet and checker importers.
+func parseArchivesAndImportMap(archiveFiles, importMapIn []string) (packageFiles map[string]string, importMap map[string]string, err error) {
+	packageFiles, importMap = make(map[string]string), make(map[string]string)
+	for _, mapping := range importMapIn {
+		i := strings.Index(mapping, "=")
+		if i < 0 {
+			err = fmt.Errorf("invalid importmap %v: no = separator", mapping)
+			return
```
jayconrod: `return nil, nil, err` or `return nil, nil, fmt.Errorf...` when errors occur.

go/tools/builders/vet.go
```diff
@@ -0,0 +1,184 @@
...
+		actual := mapping[i+1:]
+		if source == "" || actual == "" {
+			continue
+		}
+		importMap[source] = actual
+	}
+	for _, a := range archiveFiles {
+		kv := strings.Split(a, "=")
+		if len(kv) != 2 {
+			err = fmt.Errorf("invalid archivefile %v: no = separator", a)
```
jayconrod: Same comment about errors.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +43,54 @@ import (
...
-		log.Fatalf("GOROOT not set")
-	}
-	importsToArchives := make(map[string]string)
-	for _, a := range archiveFiles {
-		kv := strings.Split(a, "=")
-		if len(kv) != 2 {
-			continue // sanity check
+
+	if enableVet {
+		vcfgFile, err := makeVetCfg(*packageList, *stdLib, archiveFiles, importMapIn, filenames)
```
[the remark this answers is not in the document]
stjj89: One problem is that analysis.Finding values are specific by token positions which we cannot accurately extract from vet's command-line output. Another problem is that the analysis.Package that each analysis is given is not enough to construct the vet.cfg file that vet needs to run.

The first problem I can hack around by leaving the token positions blank and adding special cases to the filtering and error-printing logic to account for these empty tokens. Fixing the second second problem is trickier. Even if I create the vet.cfg outside of the analysis-loop, the analysis API doesn't give me a way to hand this file path to a specific analysis.

Another route I explored was special-casing in the main analysis loop so that it runs in parallel with all the other analyses. However, this made the implementation of runAnalyses() and checkAnalysisResults() a lot messier, so I decided against it. This option might be worth revisiting if vet becomes a big bottleneck in the checker program.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +43,54 @@ import (
...
+		if err != nil {
+			log.Printf("error creating vet config: %v", err)
+			return nil
+		}
+		defer os.Remove(vcfgFile)
+		findings, err := runVet(*vetTool, vcfgFile)
+		if err != nil {
+			log.Printf("error running vet:\n%v\n", err)
```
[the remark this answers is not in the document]
stjj89: The idea is that only legitimate analysis findings should fail builds. Other internal errors (e.g. type-checking, loading) should merely emit a warning message. Developers shouldn't be blocked by errors they have no control over.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +43,54 @@ import (
 // do not return an error so as not to unnecessarily interrupt builds.
 func run(args []string) error {
 	archiveFiles := multiFlag{}
+	importMapIn := multiFlag{}
+	filenames := multiFlag{}
 	flags := flag.NewFlagSet("checker", flag.ContinueOnError)
 	flags.Var(&archiveFiles, "archivefile", "Archive file of a direct dependency")
+	flags.Var(&importMapIn, "importmap", "Import maps of a direct dependency")
```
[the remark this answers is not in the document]
stjj89: Thanks for the pointer. This change has not been merged into feature/check yet.

I should be able to reconstruct the information I need from the importcfg file. However, I noticed that the importcfg files uses the [host-specific GOROOT](https://github.com/bazelbuild/rules_go/blob/master/go/tools/builders/compile.go#L185) as the base path for standard library packages. My implementation uses `go.toolchain.sdk.root_file.dirname` instead. Is one of these paths preferred to the other?

Also, having the checker binary read importcfg will add some I/O overhead to the program. I'm not sure how whether this will affect performance noticeably.

go/private/rules/checker.bzl
```diff
@@ -33,12 +33,14 @@ def _go_checker_impl(ctx):
     go = go_context(ctx)
     checker_main = go.declare_file(go, "checker_main.go")
     checker_args = ctx.actions.args()
-    checker_args.add(["-output", checker_main])
+    checker_args.add_all(["-output", checker_main])
```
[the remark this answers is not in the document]
stjj89: Done.

go/private/rules/checker.bzl
```diff
@@ -33,12 +33,14 @@ def _go_checker_impl(ctx):
...
     check_archives = [get_archive(dep) for dep in ctx.attr.deps]
     check_importpaths = [archive.data.importpath for archive in check_archives]
-    checker_args.add(check_importpaths, before_each = "-check_importpath")
+    checker_args.add_all(check_importpaths, before_each = "-check_importpath")
+    if ctx.attr.vet:
+        checker_args.add("-vet")
     if ctx.file.config:
-        checker_args.add(["-config", ctx.file.config])
+        checker_args.add_all(["-config", ctx.file.config.path])
```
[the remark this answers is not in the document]
stjj89: Done.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +43,54 @@ import (
...
+	flags.Var(&filenames, "filename", "A source file being compiled")
```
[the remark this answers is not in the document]
stjj89: Done.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +43,54 @@ import (
...
+	}
+	return nil
+}
 
-	c := make(chan result)
-	// Perform analyses in parallel.
+// runAnalyses runs all analyses, filters results, and writes findings to the
+// given channel.
+func runAnalyses(c chan result, fset *token.FileSet, archiveFiles, importMapIn, srcFiles []string, stdLib string) error {
+	apkg, err := newAnalysisPackage(fset, archiveFiles, importMapIn, srcFiles, stdLib)
```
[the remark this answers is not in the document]
stjj89: Done.

go/tools/builders/vet.go
stjj89: Done.

go/tools/builders/vet.go
stjj89: Done.

go/tools/builders/vet.go
stjj89: Done.

go/tools/builders/vet.go
stjj89: Done.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +44,93 @@ import (
...
-	goroot, ok := os.LookupEnv("GOROOT")
-	if !ok {
-		log.Fatalf("GOROOT not set")
+
+	packageFile, importMap, err := parseArchivesAndImportMap(archiveFiles, importMapIn)
+	if err != nil {
+		log.Printf("error parsing arguments: %v", err)
+		return nil
+	}
+	if enableVet {
```
jayconrod: Continuing the thread about wrapping vet as an `Analysis`, I asked Alan about this, and he wants to keep findings they way they are. It's important to be able to programmatically consume findings alongside the other data produced by the API.

Vet as it stands doesn't fit that model (though we plan to do some refactoring so it will). I expect in the future we'll want to wrap external tools in general, so we may want another abstraction, but no change needed for now.

go/tools/builders/checker_main.go
```diff
@@ -44,39 +44,93 @@ import (
 // do not return an error so as not to unnecessarily interrupt builds.
 func run(args []string) error {
 	archiveFiles := multiFlag{}
+	importMapIn := multiFlag{}
+	srcs := multiFlag{}
 	flags := flag.NewFlagSet("checker", flag.ContinueOnError)
 	flags.Var(&archiveFiles, "archivefile", "Archive file of a direct dependency")
+	flags.Var(&importMapIn, "importmap", "Import maps of a direct dependency")
```
jayconrod: Continuing the thread about importcfg files, Bazel will set `GOROOT` to the *target* standard library root, which is what we need. That's what the code being checked will compile and link against.

I don't think the I/O overhead will be bad. The file will be freshly written by compile.go, so it will definitely be in the kernel cache. I'd be mildly surprised if that data ever actually hits the disk.

If you want to do this in another PR after `master` is merged to `feature/check`, that's fine, but please add a TODO here.

stjj89: Ack. I've added a TODO here which I will address after merging `master` into `feature/check`.

go/tools/builders/checker_main.go
stjj89: Ack.

go/private/rules/checker.bzl
```diff
@@ -36,9 +36,11 @@ def _go_checker_impl(ctx):
     checker_args.add(["-output", checker_main])
     check_archives = [get_archive(dep) for dep in ctx.attr.deps]
     check_importpaths = [archive.data.importpath for archive in check_archives]
-    checker_args.add(check_importpaths, before_each = "-check_importpath")
+    checker_args.add_all(check_importpaths, before_each = "-check_importpath")
+    if ctx.attr.vet:
+        checker_args.add("-vet")
     if ctx.file.config:
-        checker_args.add(["-config", ctx.file.config])
+        checker_args.add(["-config", ctx.file.config.path])
```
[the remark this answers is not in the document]
stjj89: Done. Thanks for pointing this out.

go/tools/builders/checker_main.go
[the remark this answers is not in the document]
stjj89: I was thinking of the third kind of error. The vet wrapper is a hack that might break if the contract between vet and and the go build program changes in an upcoming version of Go. I don't think we should fail users if our implementation fails to handle some of these edge cases. OTOH, failing hard will ensure that these bugs are fixed in a timely manner.

Since users have to opt-in to use vet and can easily disable it, the impact on developer productivity will probably be minimal. I'll make all vet errors hard errors then.

stjj89:
Sorry about the latest forced push. I read your meta-comment too late and went along with my commit-amending workflow. I'll make sure to use separate commits in future PRs.

stjj89:
The only extra change I have planned is a big renaming to brand this feature as "nogo". I also have the "bottom-up" and "top-down" checks on to-do list, but that can wait till after the launch.

Let's see if I can get the importcfg change in by the end of this week. Otherwise, I don't mind waiting till after Sept 4. Thanks again for reviewing all of this code.

PR #1680, opened 2018-08-22
Create vet.cfg from importcfg

stjj89:
The importcfg file generated by the compile program contains all the
information we need to generate vet.cfg, so pass that file to the
checker instead and remove all redundant flags.


go/tools/builders/checker_main.go
```diff
@@ -90,44 +84,52 @@ func run(args []string) error {
...
 }
 
-// parseArchivesAndImportMap parses the -archivefile and -importmap arguments
-// created by the compile rule into the packageFiles and importMap maps used
-// by the vet and checker importers.
-func parseArchivesAndImportMap(archiveFiles, importMapIn []string) (packageFiles map[string]string, importMap map[string]string, err error) {
+// parseImportCfg parses the importcfg file generated by the compile program
+// and produces the packageFiles and importMap maps used by the vet and checker
+// importers.
+func parseImportCfg(importcfg string) (packageFiles map[string]string, importMap map[string]string, err error) {
```
jayconrod: You may want to copy `readImportCfg` from `$GOROOT/src/cmd/compile/internal/gc/main.go`. This parser works fine for the files generated by our compile.go, but it doesn't handle comments or extra whitespace, so it's possible we may be break it in the future.

stjj89: Done.

go/tools/builders/compile.go
```diff
@@ -244,6 +235,11 @@ func buildImportcfgFile(archives []archive, stdImports []string, installSuffix,
 		return "", errors.New("GOROOT not set")
 	}
 	goroot = abs(goroot)
+	// UGLY HACK: The vet tool called by compile program expects the vet.cfg file
```
stjj89: By the way, do you have any thoughts about this hack? I'm not exactly sure how 'go build' gets around this (surely it has to handle cases where you run tests that do not import "fmt").

jayconrod: :man_shrugging: 

It looks like go build uses the [same hack](https://github.com/golang/go/blob/master/src/cmd/go/internal/work/exec.go#L987).

My only thought about this is the `-stdimport` flags are regrettable, but I like that better than checking whether `packagefile` strings start with `GOROOT`, which is the other thing I thought of. Explicit is better than implicit.

PR #1696, opened 2018-08-28
Disable GOCACHE and only enable -shared/-dynlink on selected platforms

steeve:
Go 1.11 has its own build cache, which is enabled by default. Disable it because it doesn't make sense with bazel. See https://golang.org/cmd/go/#hdr-Build_and_test_caching

Also, only enable `-shared`/`-dynlink` on selected platforms, mimicking what `go build` does.

### DIFF
--- a/go/private/mode.bzl
+++ b/go/private/mode.bzl
@@ -150,3 +150,61 @@ def mode_tags_equivalent(l, r):
             l.goarch == r.goarch and
             l.race == r.race and
             l.msan == r.msan)
+
+# Ported from https://github.com/golang/go/blob/master/src/cmd/go/internal/work/init.go#L76
+_LINK_C_ARCHIVE_PLATFORMS = {
+    "darwin/arm": None,
+    "darwin/arm64": None,
+}
+
+_LINK_C_ARCHIVE_GOOS = {
+    "dragonfly": None,
+    "freebsd": None,
+    "linux": None,
+    "netbsd": None,
+    "openbsd": None,
+    "solaris": None,
+}
+
+_LINK_C_SHARED_PLATFORMS = {
+    "linux/amd64": None,
+    "linux/arm": None,
+    "linux/arm64": None,
+    "linux/386": None,
+    "linux/ppc64le": None,
+    "linux/s390x": None,
+    "android/amd64": None,
+    "android/arm": None,
+    "android/arm64": None,
+    "android/386": None,
+}
+
+_LINK_PLUGIN_PLATFORMS = {
+    "linux/amd64": None,
+    "linux/arm": None,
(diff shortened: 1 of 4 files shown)


go/private/context.bzl
```diff
@@ -241,6 +241,7 @@ def go_context(ctx, attr = None):
     env.update({
         "GOARCH": mode.goarch,
         "GOOS": mode.goos,
+        "GOCACHE": "off", # bazel has it's own caching
```
jayconrod: As was mentioned in another comment, this will probably break in Go 1.12. I think they want to integrate the cache more tightly into `go build`.

`GoStdLib` should be the only action that needs this. The other actions all read inputs and write outputs directly. So maybe `GOCACHE` could just be set in stdlib.go?

I share your concern about interacting with the user's cache outside of Bazel's workspace. It's probably okay as long as sandboxing or remote execution are in use, but that's not guaranteed. It's probably better to create a temp directory, set `GOCACHE` to that, then delete it at the end of the action.

It's probably best for all that to be in a separate PR though. Or feel free to file an issue, and I can fix it next week.

go/private/mode.bzl
```diff
@@ -129,3 +129,62 @@ def installsuffix(mode):
...
+    "linux/s390x": None,
+    "linux/ppc64le": None,
+    "android/amd64": None,
+    "android/arm": None,
+    "android/arm64": None,
+    "android/386": None,
+    "darwin/amd64": None,
+}
+
+def add_link_arg(go, args):
```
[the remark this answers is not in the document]
steeve: sold!

go/private/context.bzl
steeve: sure thing, I'll remove it

steeve:
I'm going to rebase on master

go/private/mode.bzl
```diff
@@ -150,3 +150,61 @@ def mode_tags_equivalent(l, r):
...
+    "android/386": None,
+    "darwin/amd64": None,
+}
+
+def link_mode_args(mode):
+    platform = mode.goos + "/" + mode.goarch
+    args = []
+    if mode.link == LINKMODE_C_ARCHIVE:
+        if platform in _LINK_C_ARCHIVE_PLATFORMS or mode.goos in _LINK_C_ARCHIVE_GOOS:
+            if platform == "linux/ppc64":
```
jayconrod: This `return` could cause problems. Can the condition be folded into the `if` above? `... and platform != "linux/ppc64"`

steeve: Sure!

steeve:
Sorry I got sidetracked, thank you for making the change!

PR #1711, opened 2018-09-04
Allow bazel version to be empty if bazel is built from source

jgavris:
Follow-up to https://github.com/bazelbuild/rules_go/pull/1703. Sorry @jayconrod, I tested this slightly better this time.


jgavris:
Much better, thank you!

PR #1712, opened 2018-09-05
Don't check bazel version on development builds

jayconrod:
Reverted #1703. This should not have modified the vendored versions.bzl.

We now just avoid calling `versions.check` if no version is defined.

Related #1711.


mhlopko:
Just saw rules_go are red and hopefully this PR fixes it :) Thanks! If anybody is interested: https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/425#48c9dc0d-08f4-4fd9-801a-5b53d7ab8ffa

PR #1715, opened 2018-09-07
Propagate testonly flag in go_vet_test

akshayjshah:
To run `go_vet_test`, the user must supply a `GoPath`. However, the
generated rule doesn't propagate the `testonly` attribute from the
supplied `GoPath`, which makes it difficult to vet test code. To verify
that this change works as intended, this commit also introduces some
*very* basic unit tests.

Please forgive me if this is way off-track - this is my first experience with
Bazel and Starlark, so it's likely that I've done many things wrong.


tests/core/go_vet_test/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+go_path(
+    name = "lib_srcs",
+    deps = [":lib"],
+)
+
+go_path(
+    name = "all_srcs",
+    deps = [":internal_test"],
+    testonly = True,
+)
```
akshayjshah: Should `go_path` automatically add the `testonly` attribute if any of the deps have it? Or is that too magical?

jayconrod: Too magical.

akshayjshah: Fair enough.

akshayjshah:
Thanks for the quick review - I'll take a look at the build failure early next week.

I'm really glad that you're working to integrate static analysis into the build pipeline! I'm not attached to this code at all, and it probably makes sense to delete this rule completely once you have something better available. I'll keep an eye on that branch.

jayconrod:
The build failure is happening with remote execution. I'll take a look when I have a chance, probably not until next week though. This might have something to do with passing a directory as input (though I think that's supposed to work).

akshayjshah:
Okay, sounds good. Let me know if there's anything I can do to debug further.

jayconrod:
@akshayjshah Sorry to take so long to get back to you on this.

It looks like the mac failure is happening because the macOS builder is timing out fetching something from the remote cache. I've verified it works locally.

The failure in the RBE configuration seems like a bug (something about a hash being the wrong size). I'll follow up internally on that.

To proceed, I'd suggest just removing the test altogether. The change is easy enough to verify by inspection, and we already have basic coverage of `go_vet_test` in `//:vet`.

jayconrod:
The RBE configuration seems to be working smoothly now. I'm told they were getting some kernel panics after a macOS update.

I've updated and merged this change. Thanks again!

PR #1718, opened 2018-09-11
Merge branch 'master' into feature/check

jayconrod:
Also, update gcexportdata to latest version to support Go 1.11.


jayconrod:
cc @stjj89

PR #1721, opened 2018-09-17
Fix non-root package testing on Windows

### LINKED ISSUE #1720: go_test fails on windows if package is inside folder
When using Windows, If I have the exact same empty test inside and outside a folder inside my workspace, the test outside the folder will succeed and the one inside will fail.

Reproduction:
- `git clone https://github.com/filipesilva/bazel_go_windows`
- `cd bazel_go_windows`
- `bazel test ...`
```
$ bazel test ...
INFO: Analysed 4 targets (0 packages loaded).
INFO: Found 2 targets and 2 test targets...
FAIL: //foo:go_default_test (see C:/users/kamik/_bazel_kamik/5xrk7syg/execroot/__main__/bazel-out/x64_windows-fastbuild/testlogs/foo/go_default_test/test.log)
INFO: Elapsed time: 1.415s, Critical Path: 0.55s
INFO: 1 process: 1 local.
INFO: Build completed, 1 test FAILED, 2 total actions
//:go_default_test                                              (cached) PASSED in 0.7s
//foo:go_default_test                                                    FAILED in 0.5s
  C:/users/kamik/_bazel_kamik/5xrk7syg/execroot/__main__/bazel-out/x64_windows-fastbuild/testlogs/foo/go_default_test/test.log

Executed 1 out of 2 tests: 1 test passes and 1 fails locally.
INFO: Build completed, 1 test FAILED, 2 total actions
```
```
Executing tests from //foo:go_default_test
2018/09/17 15:19:58 could not change to test directory: chdir C:\users\kamik\_bazel_kamik\5xrk7syg\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\foo\windows_amd64_stripped\go_default_test.exe.runfiles\__main__\foo: The system cannot find the file specified.
```

If I had to best, it's related to looking up runfiles on Windows. `rules_go` has some helpers (https://github.com/bazelbuild/rules_go/tree/master/go/tools/bazel) that maybe can address this. Unsure why it doesn't happen on the root package though.

I'm using Windows 10, Bazel 0.17.1, and `rules_go` at https://github.com/bazelbuild/rules_go/commit/1bad20854a59785f4896d8b8847bdb57bfccf2ba.

filipesilva:
Use a temporary directory instead of piggybacking into symlinked runfiles directories, as those might not be there on Windows.

Fix #1720

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -68,6 +68,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"testing"
 	"testing/internal/testdeps"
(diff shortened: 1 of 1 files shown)


jayconrod:
Thanks for sending this. Unfortunately, this breaks the `rundir` of `go_test`, so I can't accept this as-is.

I would accept a (temporary) change that makes the error from `os.Chdir` non-fatal when `runtime.GOOS == "windows"`. I hope at some point Bazel will expose runfiles in a directory tree of symbolic links.

filipesilva:
@jayconrod I made the change you requested but am unsure of how it might affect testing on Windows. 

Is it problematic that `os.Setenv("PWD", abs)` is not set when running tests? Should it be set to something else?

PR #1723, opened 2018-09-18
More small nogo fixes

jayconrod:
* Rename vet.go to nogo_vet.go and remove from compile builder.
* Comment on importer and unsafe special case in nogo_main.go.
* Fix example in nogo.rst.


jayconrod:
@stjj89 Couple more small fixes. PTAL.

PR #1726, opened 2018-09-19
Merge branch 'master' into feature/check

jayconrod:


jayconrod:
cc @stjj89 This is merged into feature/check now, but it may still be good to wait a few days. Alan mentioned he's planning on renaming several of the types, so the API is not stable yet.

stjj89:
Ack. Thanks for the heads up.

PR #1730, opened 2018-09-24
Add sha256 attributes to http_archive deps in go_rules_dependencies()

gregmagolan:
This allows Bazel to use its repository cache when fetching these deps and prevents re-downloads on fresh builds downstream.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -35,6 +35,7 @@ def go_rules_dependencies():
         # master, as of 2018-09-18
         urls = ["https://codeload.github.com/golang/tools/zip/90fa682c2a6e6a37b3a1364ce2fe1d5e41af9d6d"],
         strip_prefix = "tools-90fa682c2a6e6a37b3a1364ce2fe1d5e41af9d6d",
+        sha256 = "d22945772a265fa5684b6a04e83fdfee458b7d4c1cefec423a6cd0866b3be719",
         type = "zip",
         overlay = manifest["org_golang_x_tools"],
         # importpath = "golang.org/x/tools",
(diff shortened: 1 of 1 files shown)


steeve:
I've read that sha256 against `codeload.github.com` is not safe since it's generated on the fly.

gregmagolan:
@steeve Does the `https://github.com/owner/repo/archive/commit_hash` url have the same issue? We use that in rules_nodejs:

```
    http_archive(
        name = "io_bazel_skydoc",
        url = "https://github.com/bazelbuild/skydoc/archive/0ef7695c9d70084946a3e99b89ad5a99ede79580.zip",
        strip_prefix = "skydoc-0ef7695c9d70084946a3e99b89ad5a99ede79580",
        sha256 = "491f9e142b870b18a0ec8eb3d66636eeceabe5f0c73025706c86f91a1a2acb4d",
    )
```

https://github.com/bazelbuild/rules_nodejs/blob/master/package.bzl#L100

steeve:
@gregmagolan i'm not sure, but that would be nice!

jayconrod:
Sorry I can't merge this one. As @steeve mentioned, GitHub generates zip files on demand, and the sha256 sums are not stable. As soon as GitHub upgrades their OS and the order of files or alignment changes in zip files, the sums will be different, and it will break *everyone*. This happened a couple times last year, and it was very painful.

I don't know whether the `https://github.com/owner/repo/archive/commit_hash` endpoint is stable. If you can point me to GitHub documentation that says it is, I'll be happy to migrate everything to use that. Otherwise, I assume it's not.

I would encourage everyone to include .zip or .tar.gz files as artifacts with their releases. rules_go and Gazelle do this.

gregmagolan:
@jayconrod @steeve Thanks for the info Jay. I'll do some investigation to see if what we are doing in rules_nodejs will end up with the same issue.

PR #1731, opened 2018-09-25
Do not depend on stdlib for CGo Codegen anymore

### LINKED ISSUE #1709: stdlib builds twice if CGo is enabled
Building the following example:
```python
go_library(
    name = "go_default_library",
    srcs = [
        "src.go",
    ],
    cgo = True,
    importpath = "github.com/myproject",
)

go_binary(
    name = "mybin",
    deps = [
        ":go_default_library",
    ],
    pure = "off",
)
```

If `cgo = True`, then `stdlib` is built twice:
- one with the correct mode
- the other with a default mode (`pure = True`)

I am failing to see where that info gets lost, as manually adding `pure` attributes to all cgo rules doesn't fix it somehow.

steeve:
Also fixes #1709.

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -210,7 +210,7 @@ def _cgo_codegen_impl(ctx):
 
     tool_args.add_all(["-objdir", out_dir])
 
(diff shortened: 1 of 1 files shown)


jayconrod:
Thanks!

PR #1750, opened 2018-10-05
nogo: use new golang.org/x/tools/analysis API

stjj89:
Remove the local fork of package analysis that we previously
relied on, and move to then official release instead.

While there:
- Support executing analyzer dependencies in parallel.
- Handle identical analyzer package names using unique import
  identifiers in the generated nogo source file.
- Correctly specify import paths in repositories.bzl.
- Change docs and code to refer to nogo "analyzers" instead of "checks".


go/nogo.rst
```diff
@@ -55,35 +57,32 @@ abstract syntax trees (ASTs) and type information for that package. For example:
...
         }
       }
-      return &analysis.Result{Findings: findings}, nil
+      return nil, nil
     }
 
-Any findings returned by the check will fail compilation. Do not emit findings
-unless they are severe enough to warrant interrupting the compiler.
+Any diagnostics reported by the analyzer will fail compilation. Do not emit
+diagnostics unless they are severe enough to warrant interrupting the compiler.
```
alandonovan: "stopping the build"

(the compiler runs to completion)

tests/core/nogo/custom/foo_func_analyzer.go
```diff
@@ -0,0 +1,37 @@
...
+
+var Analyzer = &analysis.Analyzer{
+	Name: "foo_func_analyzer",
+	Run:  run,
+	Doc:  doc,
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	for _, f := range pass.Files {
+		ast.Inspect(f, func(n ast.Node) bool {
```
alandonovan: In this example, consider using the new "inspect" package rather than ast.Inspect as it is much faster.  It will be checked in to x/tools today and google3 next week.

tests/core/nogo/custom/import_fmt_analyzer.go
```diff
@@ -0,0 +1,34 @@
...
+	Run:  run,
+	Doc:  doc,
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	for _, f := range pass.Files {
+		ast.Inspect(f, func(n ast.Node) bool {
+			switch n := n.(type) {
+			case *ast.ImportSpec:
+				if n.Path.Value == "\"fmt\"" {
```
alandonovan: if path, _ := strconv.Unquote(n.Path.Value); path == "fmt" {

There are many string literals that denote the same string (e.g. \`fmt\`).

tests/core/nogo/custom/return_bool_analyzer.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"golang.org/x/tools/go/analysis"
+)
+
+const doc = `report functions that return booleans
+
+The return_bool_analyzer analyzer reports functions that return booleans.`
+
+var Analyzer = &analysis.Analyzer{
+	Name: "return_bool_analyzer",
```
alandonovan: Don't use underscores here. Also, analyzer is redundant.

`Name: "boolreturn"`

tests/core/nogo/custom/return_bool_analyzer.go
```diff
@@ -0,0 +1,37 @@
...
+	Run:  run,
+	Doc:  doc,
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	for _, f := range pass.Files {
+		ast.Inspect(f, func(n ast.Node) bool {
+			switch n := n.(type) {
+			case *ast.FuncDecl:
+				if results := n.Type.Results; results != nil {
```
alandonovan: Not a big deal, but if you're matching types, use the types, not the syntax:

```
results := pass.InfoTypes.Defs[n.Name].Type().(*types.Signature).Results()
for i := 0; i < results.Len(); i++ {
    if results.At(i).Type() == types.Typ[types.Bool] { ... }
}
```

Do you want to match exactly bool, or named boolean types too?

tests/core/nogo/custom/return_bool_analyzer.go
```diff
@@ -0,0 +1,37 @@
+// returnboolanalyzer checks for functions that return bool.
```
alandonovan: There's nothing wrong with returning a bool. It would be better to use an example that at least hinted at an actual problem.

tests/core/nogo/deps/a.go
```diff
@@ -0,0 +1,21 @@
...
+var Analyzer = &analysis.Analyzer{
+	Name:     "a",
+	Doc:      "an analyzer that depends on c.Analyzer",
+	Run:      run,
+	Requires: []*analysis.Analyzer{c.Analyzer},
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	pass.Report(analysis.Diagnostic{Message: "ran a"})
+	pass.Report(analysis.Diagnostic{Message: fmt.Sprintf("a %s", pass.ResultOf[c.Analyzer].(string))})
```
alandonovan: pass.Reportf("a %s", pass.ResultOf[c.Analyzer]))

alandonovan: Combine these two diagnostics into one?

tests/core/nogo/deps/b.go
```diff
@@ -0,0 +1,21 @@
...
+
+var Analyzer = &analysis.Analyzer{
+	Name:     "b",
+	Doc:      "an analyzer that depends on c.Analyzer",
+	Run:      run,
+	Requires: []*analysis.Analyzer{c.Analyzer},
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	pass.Report(analysis.Diagnostic{Message: "ran b"})
```
alandonovan: ditto

tests/core/nogo/deps/c.go
```diff
@@ -0,0 +1,22 @@
...
+var Analyzer = &analysis.Analyzer{
+	Name:       "c",
+	Doc:        "an analyzer that depends on d.Analyzer",
+	Run:        run,
+	Requires:   []*analysis.Analyzer{d.Analyzer},
+	ResultType: reflect.TypeOf(""),
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	pass.Report(analysis.Diagnostic{Message: "ran c"})
```
alandonovan: ditto

tests/core/nogo/deps/d.go
```diff
@@ -0,0 +1,28 @@
+package d
+
+import (
+	"fmt"
+	"reflect"
+
+	"golang.org/x/tools/go/analysis"
+)
+
+// Since the test nogo rule does not explicitly depend on d, we package state
```
alandonovan: "we package state"?

Please don't put global state in an analyzer.

tests/core/nogo/deps/d.go
```diff
@@ -0,0 +1,28 @@
...
+
+var Analyzer = &analysis.Analyzer{
+	Name:       "d",
+	Doc:        "an analyzer that does not depend on other analyzers",
+	Run:        run,
+	ResultType: reflect.TypeOf(""),
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	numRuns++
```
alandonovan: This is a potential data race.

tests/core/nogo/custom/foo_func_analyzer.go
jayconrod: I've filed #1755. Since this is a test though, I'd suggest we proceed with this PR and fix it after we merge the `feature/check` branch into `master` and update `org_golang_x_tools` on master.

Unfortunately, this means we'll need another `go_tool_library` target, but many `Analyzer` implementations will want it.

tests/core/nogo/custom/return_bool_analyzer.go
alandonovan: Yes, it handles both.

tests/core/nogo/custom/import_fmt_analyzer.go
alandonovan: You should never do that, of course, but an analysis tool should expect the unexpected.

go/nogo.rst
```diff
@@ -55,35 +57,32 @@ abstract syntax trees (ASTs) and type information for that package. For example:
...
 
-Each check must be written as a `go_tool_library`_ rule. This rule
-is identical to `go_library`_ but avoids a bootstrapping problem, which
-will be explained later. For example:
+Each analyzer must be written as a `go_tool_library`_ rule and must import a
+`go_tool_library`_ version of the package `analysis`_ target. This rule is
```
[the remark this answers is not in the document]
jayconrod: Yeah, that message is good.

Users won't be able to use their own `golang.org/x/tools` unless they replace `@org_golang_x_tools` and use the same labels. For example, if they vendor `golang.org/x/tools`, they'll get build-time conflicts because the linker will see two different versions of `golang.org/x/tools/go/analysis.Analyzer` and other types.

PR #1751, opened 2018-10-05
Fix link to reproducible_binary BUILD file.

DarienRaymond:
The original link is dead. The new link is the closet one I can find.


jayconrod:
Thanks for pointing this out!

`//tests/reproducibility` is really complicated, so I'd hesitate to point to it as an example in the FAQ. I'd like to refactor it and probably move it some time soon, which would break the link again.

I've pushed #1753 instead, which points to `//tests/core/cross` instead. That is a much simpler test which demonstrates the same use of `$(location)`, and I don't expect it to change much.

PR #1754, opened 2018-10-05
Define the toolchain_type target used by go rules.

### LINKED ISSUE #1752: No actual toolchain_type for @io_bazel_rules_go//go:toolchain
With the fixes for https://github.com/bazelbuild/bazel/issues/6015, there will be an actual dependency from toolchains back to the respective toolchain_type target. rules_go currently does not actually define the toolchain_type, and so builds fail (see https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/480#5b3d386c-f1d5-412d-adf0-522dd5856b2c for a sample).

Error:
```
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ef9bf13ce4d6a32f18a051c5f9234a8d/external/go_sdk/BUILD.bazel:52:1: no such target '@io_bazel_rules_go//go:toolchain': target 'toolchain' not declared in package 'go'; however, a source directory of this name exists.  (Perhaps add 'exports_files(["toolchain"])' to go/BUILD, or define a filegroup?) defined by /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ef9bf13ce4d6a32f18a051c5f9234a8d/external/io_bazel_rules_go/go/BUILD.bazel and referenced by '@go_sdk//:go_windows_amd64'
```

katre:
Fixes #1752.

### DIFF
--- a/go/BUILD.bazel
+++ b/go/BUILD.bazel
@@ -3,3 +3,8 @@ filegroup(
     srcs = glob(["*.bzl"]) + ["//go/private:all_rules"],
     visibility = ["//visibility:public"],
 )
+
+toolchain_type(
+    name = "toolchain",
+    visibility = ["//visibility:public"],
+)


katre:
Previously, the "toolchain_type" attribute on the "toolchain" rule was a NODEP_LABEL, which simply checks that the value can be parsed as a label but doesn't create a dependency and so doesn't check whether the referenced target actually exists.

https://github.com/bazelbuild/bazel/commit/23488930607e4919a99e80de240d6d35f4037797 changes that to an actual dependency (part of some other cleanups I'm making), so this suddenly fails.

We're going to roll back https://github.com/bazelbuild/bazel/commit/23488930607e4919a99e80de240d6d35f4037797 and not roll it forward until either:
1. All projects are fixed to properly define a toolchain_type, or
2. I can figure out how to guard the change with an incompatible flag.

PR #1756, opened 2018-10-05
Add compiled archives to compilation_outputs group

schroederc:
This behavior matches the native C++ and Java rules[1] and allows users
to easily verify a target compiles without invoking the linker
unnecessarily.

[1]: https://github.com/bazelbuild/bazel/commit/1dcd8f0f12bb582cc449381f8245f943121031fb

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -85,6 +85,7 @@ def _go_binary_impl(ctx):
         archive,
         OutputGroupInfo(
             cgo_exports = archive.cgo_exports,
+            compilation_outputs = [archive.data.file],
         ),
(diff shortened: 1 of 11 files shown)


jayconrod:
Thanks, seems like a reasonable feature.

`go_proto_library` (in `//proto:def.bzl`) should support this for consistency.

Also, need tests. Creating a new directory `//tests/core/output_groups` is probably best. You can filter an output group out of a rule with `filegroup`, so you could just have a `go_test` that consumes `filegroups` via `data` and `args` and verifies those files exist and end with the `.a` suffix.

schroederc:
I've made the requested changes; PTAL.

PR #1758, opened 2018-10-08
Update feature/check to master and remove vendor directory

jayconrod:
1. Merge `master` into `feature/check`.
2. Point the dependency on `golang.org/x/tools/go/gcexportdata` to `@org_golang_x_tools` instead of vendor (requires updating the patch).
3. Delete `vendor`.


jayconrod:
cc @stjj89

PR #1763, opened 2018-10-11
nogo: rename apply_to and whitelist config fields

stjj89:
The new names, "only_files" and "exclude_files", make it clear that
this is an include/exclude system.


stjj89:
No problem. I'm looking forward to getting this out into the wild!

PR #1773, opened 2018-10-18
Update org_golang_x_tools dependency

ash2k:
I need a newer version of this dependency to pick up some fixes. I tried to declare it before calling `go_rules_dependencies()` and it worked with `0.15.4` but no longer does with `0.16.0`.
Here I'm trying to update the version to the latest but testing is failing with:
```console
ERROR: /private/var/tmp/_bazel_ash2k/976c6f6fb6cb58e5d210fd9e522a8491/external/org_golang_x_tools/go/types/typeutil/BUILD.bazel:3:1: GoCompile external/org_golang_x_tools/go/types/typeutil/darwin_amd64_stripped/go_default_library%/golang.org/x/tools/go/types/typeutil.a failed (Exit 1)
GoCompile: missing strict dependencies:
        /private/var/tmp/_bazel_ash2k/976c6f6fb6cb58e5d210fd9e522a8491/sandbox/darwin-sandbox/786/execroot/io_bazel_rules_go/external/org_golang_x_tools/go/types/typeutil/callee.go: import of "golang.org/x/tools/go/ast/astutil"
No dependencies were provided.
Check that imports in Go sources match importpath attributes in deps.
```

IIUC the patches that `http_archive` applies in `go_rules_dependencies()` need to be updated. But I'm not sure how to do that. Any help would be appreciated.


jayconrod:
If you want to update this in rules_go, you can find the procedure at [Updating dependencies](https://github.com/bazelbuild/rules_go/wiki/Updating-dependencies). The Gazelle patch needs to be updated, but the others probably don't.

It would likely be easier to just override `org_golang_x_tools` in your own WORKSPACE by following [Overriding dependencies](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#overriding-dependencies) though.

ash2k:
@jayconrod this is ready for review.

I've tried to override this dependency but one of the patches wouldn't apply. So I updated it in this branch, including the patches and the list of tests, as per instructions. All tests pass now, including my project.

jayconrod:
There were some issues with the extra patch. It's hard for me to review patches, so I just regenerated them both and updated the PR.

It appears GitHub Webhooks are down right now, which means CI is also down (I'll be looking forward to the post-mortem for this one). I've manually run the tests on Linux and macOS, and everything passed.

ash2k:
Thanks!

PR #1776, opened 2018-10-18
Exclude srcs and their runfiles from Go rule runfiles

### LINKED ISSUE #1761: rules_go added source files to runfiles if you specify `native.filegroup` in the srcs
Please take a look to repo
https://github.com/excavador/bazel_rules_go_default_info_provider_bug

jayconrod:
ctx.runfiles(collect_data = True) collects default runfiles from srcs,
deps, and data. When srcs comes from a filegroup or genrule, the
source files themselves are included in the runfiles, which is
surprising for users.

With this change, we gather data runfiles only from data and deps.

Fixes #1761


jayconrod:
The macOS builder seems to be going off into the weeds, but I've verified tests pass on my own Mac, so...

Issue #1777, opened 2018-10-19
How to activate nogo


Globegitter:
The documentation mentions the release of `nogo` which sounds very interesting but does not directly state on how to activate it. Does it just run by default? Or does it need to be put into a top level `BUILD.bazel` file similar to `gazelle`? Have not been able to find any mention of that in the docs.

justinsb:
I'd echo this - I tried to just get go vet working and couldn't get it to work.  It was complaining about a missing `/tmp/govet.cfg` file (IIRC) which I thought was a good sign, but then later it "just stopped" complaining but also wasn't picking up a go vet problem (of the `-printf` class), so I assume it stopped complaining because it stopped running.

My WIP branch is here if it's relevant: https://github.com/justinsb/etcd-manager/commit/169cbfad24b34a4095139247b214da6721f10293

jayconrod:
@Globegitter [go/nogo.rst](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst) is the best place to start. Basically, you need to declare a `nogo` target somewhere, then point to it from `go_register_toolchains`. It will run automatically for each Go package that gets built.

@stjj89 Do you have bandwidth today to look into @justinsb's example?

justinsb:
Ah - thanks @jayconrod  - the "point to it from go_register_toolchains" makes it obvious to me what I'm doing when I declare it in go_register_toolchains, and explains why I only have to put it in the root.  (That would be a great phrasing for the docs, IMO)

I have a theory that I'm running in parallel and all the tasks are writing to the same `/tmp/vet.cfg` file (and deleting it).  i.e. we should use a genuine `ioutil.TempFile` or a vet.cfg file in a tempdir if the name matters.  I think this could explain sometimes seeing file-not-found, and maybe I'm not getting vet errors where I think I should be because vet.cfg keeps being rewritten with different packages.  That said, this machine is only 2 cores, so it's not a great theory...

But `bazel build --jobs=1 //pkg/...` does indeed complete without the file-not-found errors, though I still have yet to have a `go vet` failure...

justinsb:
I added an egregious `if true || true` (borrowing the idea from the tests for nogo) and it does correctly no-go me :-)  https://github.com/kopeio/etcd-manager/pull/134/files  

And I'm pretty sure the single file is an issue, because with jobs=1 I didn't see problems, but without jobs=1 I saw: `vet: open /tmp/vet.cfg: no such file or directory`

[...]

PR #1778, opened 2018-10-19
nogo: write vet.cfg to random temporary configuration file

jayconrod:
Related #1777

### DIFF
--- a/go/tools/builders/nogo_vet.go
+++ b/go/tools/builders/nogo_vet.go
@@ -118,9 +118,11 @@ func buildVetcfgFile(packageFile, importMap map[string]string, stdImports, files
 	}
 	vcfgPath := vcfgFile.Name() // vcfgPath may be "" when returning
 	defer func() {
-		os.Remove(vcfgPath)
 		if cerr := vcfgFile.Close(); err == nil && cerr != nil {
 			err = cerr
+		}
+		if err != nil {
+			os.Remove(vcfgPath)
 		}
 	}()
 


go/tools/builders/nogo_vet.go
```diff
@@ -111,16 +110,27 @@ func buildVetcfgFile(packageFile, importMap map[string]string, stdImports, files
...
-	js = append(js, '\n')
-	vcfgFile := filepath.Join(os.TempDir(), "vet.cfg")
-	if err := ioutil.WriteFile(vcfgFile, js, 0666); err != nil {
 		return "", err
 	}
-	return vcfgFile, nil
+	vcfgPath := vcfgFile.Name() // vcfgPath may be "" when returning
+	defer func() {
+		if cerr := vcfgFile.Close(); err == nil && cerr != nil {
+			os.Remove(vcfgPath)
```
justinsb: Do you want to `os.Remove` on any error, not just an error closing the file, so you always clean up the temp file?

So

```
defer func() {
  if cerr := vcfgFile.Close(); err == nil && cerr != nil {
    err = cerr
  }
  if err != nil {
    os.Remove(vcfgPath)
  }
```

(I think)

jayconrod: Good catch, it should be on any error.

go/tools/builders/nogo_vet.go
```diff
@@ -111,16 +110,27 @@ func buildVetcfgFile(packageFile, importMap map[string]string, stdImports, files
...
+			err = cerr
+		}
+	}()
+
+	enc := json.NewEncoder(vcfgFile)
+	enc.SetIndent("", "\t")
+	if err := enc.Encode(vcfg); err != nil {
```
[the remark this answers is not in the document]
jayconrod: It shouldn't matter. The named return value `err` will be set when the `return` statement executes, and the deferred closure will see that one.

go/tools/builders/nogo_vet.go
```diff
@@ -91,7 +90,7 @@ func splitOutput(out string) []string {
 
 // buildVetcfgFile creates a vet.cfg file and returns its file path. It is the
 // caller's responsibility to remove this file when it is no longer needed.
-func buildVetcfgFile(packageFile, importMap map[string]string, stdImports, files []string) (string, error) {
+func buildVetcfgFile(packageFile, importMap map[string]string, stdImports, files []string) (vcfgPath_ string, err error) {
```
stjj89: Is the underscore in `vcfgPath_` a typo?

go/tools/builders/nogo_vet.go
```diff
@@ -111,16 +110,27 @@ func buildVetcfgFile(packageFile, importMap map[string]string, stdImports, files
...
-	}
-	js = append(js, '\n')
-	vcfgFile := filepath.Join(os.TempDir(), "vet.cfg")
-	if err := ioutil.WriteFile(vcfgFile, js, 0666); err != nil {
 		return "", err
 	}
-	return vcfgFile, nil
+	vcfgPath := vcfgFile.Name() // vcfgPath may be "" when returning
+	defer func() {
+		os.Remove(vcfgPath)
```
stjj89: Potentially stupid question: is this deferred Remove necessary or problematic when the calling function (nogo_main.run) also defers the removal of this file?

go/tools/builders/nogo_vet.go
jayconrod: That's actually intentional. There is a return statement below which sets `vcfgPath_` to `""`. If the deferred closure referenced it, it would not be able to remove the file. Instead, it references `vcfgPath` which is only set in one place. I thought about giving them unique names earlier, but that seemed less readable.

go/tools/builders/nogo_vet.go
jayconrod: The caller will only be responsible for removing the file if the call succeeds. The call could fail in a number of ways, including not being able to create the file, and the caller shouldn't have to worry about that. If the file is successfully created though, then it's the caller's responsibility.

go/tools/builders/nogo_vet.go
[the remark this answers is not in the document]
justinsb: Ah - quite right - sorry for noise!

stjj89:
The issue this PR fixes will block users trying to use ``nogo`` in 0.16.0. It would be unfortunate if we lose some potential users of this feature because they got turned away by this blocking bug. Can we emergency-drop a 0.16.1 release once this is PR is merged?

stjj89:
@jayconrod great, thanks!

PR #1779, opened 2018-10-19
Improve nogo documentation

stjj89:
Create a "Setup" section similar to the core rules_go README
that gives direct instructions on how to get run nogo. Also,
make miscellaneous improvements to other section of nogo.rst.

Related #1777

### DIFF
--- a/go/nogo.rst
+++ b/go/nogo.rst
@@ -50,11 +50,20 @@ want to run.
     nogo(
         name = "my_nogo",
         deps = [
+            # analyzer from the local repository
             ":importunsafe",
-            ":unsafedom",
-            "@analyzers//:loopclosure", # analyzers can be imported from a remote repo
+            # analyzer from a remote repository
+            "@org_golang_x_tools//go/analysis/passes/printf:go_tool_library",
         ],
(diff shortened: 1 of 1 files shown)


go/nogo.rst
```diff
@@ -31,15 +31,53 @@ compiler-like errors.
...
+want to run.
+
+.. code:: bzl
+
+    load("@io_bazel_rules_go//go:def.bzl", "nogo")
+
+    nogo(
+        name = "my_nogo",
+        deps = [
+            ":importunsafe",
```
jayconrod: Include a `go_tool_library` target in this example, along with a `load`. It's explained below, so I don't think you need to explain its purpose here, but it would be good to preclude people from trying `go_library` first and encountering a confusing error.

To keep this short, consider reducing `deps` to two entries, one defined here and one elsewhere.

Use a real example for the one defined elsewhere. I think we are going to want to write `go_tool_library` targets for everything in `golang.org/x/tools/go/analysis` so those can easily be referenced. I'ved filed #1780 for that. No need to block on the fix, but it would be good to have a runnable example soon.

stjj89: Done.

PR #1781, opened 2018-10-22
Add go_tool_library targets for analysis passes

### LINKED ISSUE #1780: Add go_tool_library targets for golang.org/x/tools/go/analysis/passes/...

jayconrod:
Fixes #1780

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -16,9 +16,50 @@ builders(
     visibility = ["//visibility:public"],
 )
 
+# default_nogo is the nogo target that nogo references by default. It
+# does not analyze anything, which means no binary is built or run
+# at compile time.
 nogo(
     name = "default_nogo",
     visibility = ["//visibility:public"],
+)
+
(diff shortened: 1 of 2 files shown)


stjj89:
Nice, good to see that the vet checks have been abstracted away into Analyzers.

Perhaps you should also include a `nogo` documentation change that instructs users how to run the checks-previously-know-as-vet. But it might also make sense to change the docs in another PR that also deletes the nogo vet hack altogether.

jayconrod:
> But it might also make sense to change the docs in another PR that also deletes the nogo vet hack altogether.

I think this is the best approach. In the next major release, we can probably remove documentation for `vet = True` (but maybe keep it around for a little while as a compatibility option).

Before actually switching vet functionality to use these analyzers, I'd like to confirm they're as fast as vet. Alan said earlier the whole-package-at-a-time approach was slower than vet's node-at-a-time approach due to cache locality, but we'd make some of that back in parallelism. I'm not sure where we ended up.

PR #1783, opened 2018-10-22
Fix execute arguments to use list see github.com/bazelbuild/rules_go/issues/1782

mauriciogg:
https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#execute


mauriciogg:
Thanks!

PR #1802, opened 2018-11-02
Update org_golang_x_tools

jayconrod:


jayconrod:
macOS failure in CI appears to be a failure in the CI framework (Mac builders are still flaky).

PR #1810, opened 2018-11-05
nogo: support go/analysis fact serialization

jayconrod:
nogo binaries can now save facts about definitions in analyzed
packages to a .x file for the package they are invoked on. nogo will
load facts from the .x files of every imported package before starting
an analysis.

- go.archive now declares an .x file when nogo is enabled. This is
  passed to go.compile as the out_export parameter.
- go.compile now passes -p in the builder arguments instead of the
  tool arguments. This is piped through to nogo, since go/types needs
  to know the package path in order to serialize facts correctly.
- nogo no longer accepts -src flags. These were redundant with
  positional arguments.
- facts are serialized and deserialized using
  golang.org/x/tools/go/analysis/internal/facts (which has a
  visibility exception for the builders).
- Replaced "boolreturn" test analyzer with "visibility" in
  tests/core/nogo/custom. This is a more realistic test that relies on
  serialized facts.


jayconrod:
@stjj89 PTAL

As it turns out, fact serialization is a prerequisite for migrating over to the vet analyzers, so I went ahead and implemented this. I also wanted to try out the framework myself, which is why the `visibility` test case is so big.

Issue #1612, opened 2018-07-27
go_test on Windows should set up runfiles symlinks


jayconrod:
Bazel has limited support for test data files (runfiles) on Windows. See #1212. They want to avoid creating symlinks, since this requires administrative privilege on older versions of Windows. Instead, they provide a manifest file (mapping short paths that would normally be used to access run files to actual file paths).

`go_test` should just create the symlinks before running the tests. We create symlinks in other places, so this does not reduce our compatibility.

laszlocsomor:
@meteorcloudy : you recently added symlink support to Bazel on Windows. Does that resolve this bug?

meteorcloudy:
I believe yes, we can enable symlink runfiles tree by `--experimental_enable_runfiles` flag. I think it's fine to just use it for tests.

jayconrod:
`--experimental_enable_runfiles` seems really promising. Thanks for adding it!

I tried this out on Windows 10 Enterprise N 2016 LTSB, version 1607. It looks like that's not new enough to have symbolic links available in developer mode. I did get it to work from an administrator shell though (I was initially confused that it didn't work, but I needed to `bazel shutdown` to kill the server started from my non-administrator shell).

I'll re-image my laptop with a newer Windows and verify this as soon as I get a chance. I'm going on vacation next week though, so it may be a while. I'd also like to wait until this feature is no longer experimental before closing this issue.

meteorcloudy:
@jayconrod That sounds good. The runfiles support should work well if you have the right to create symlink. I'll make this feature non-experimental soon.

PR #1816, opened 2018-11-07
Add runfiles support for Windows.

### LINKED ISSUE #1729: Runfile in go/tools/bazel does not work in Windows
Runfiles on windows are not present as symlinks inside the runfiles directory, but rather listed as entries in the `MANIFEST` file inside of it.

While using Windows 10 and Bazel 0.17.1, to reproduce follow:
```
git clone https://github.com/filipesilva/rules_go_runfiles_binary
cd rules_go_runfiles_binary
bazel test :bazel test :go_default_test
```

The test should fail with the following log
```
--- FAIL: TestLookupRunfile (0.00s)
    main_test.go:10: unable to find file "file.txt"
FAIL
```

The content of the `MANIFEST` file for this test run is:
```
__main__/file.txt D:/sandbox/rules_go_runfiles_binary/file.txt
__main__/windows_amd64_stripped/go_default_test.exe C:/users/kamik/_bazel_kamik/frdn7u4y/execroot/__main__/bazel-out/x64_windows-fastbuild/bin/windows_amd64_stripped/go_default_test.exe
```

### LINKED ISSUE #1815: runfiles library in Go
Transferred from https://github.com/bazelbuild/bazel/issues/6333.

jchv:
Hoi,

It seemed nobody was currently working on this so I decided to take a quick stab at it. This implements seemingly-working runfiles support for Windows, including mostly backwards compatible behavior for the existing `bazel.Runfile` routine.

Notably, this supports the Bazel runfiles manifest file. I don't know if there's a standard for the format of this file, so I just parse it similar to how other Bazel libraries are doing it. I also don't know if it's intentional that it looks like spaces in filenames would break this format 😕 Maybe that ought to be revisited some day.

I wrote this because I noticed a test failing in `bazel-watcher` that needed runfiles. Since I did this without really asking anyone first, I anticipate that parts of this may need to be rewritten, and that's OK. I'll await instruction on how to proceed.

Fixes #1729 
Fixes #1815 
Related #1612

### DIFF
--- a/go/tools/bazel/runfiles.go
+++ b/go/tools/bazel/runfiles.go
@@ -41,6 +41,7 @@ func newRunfilesResolver() (runfilesResolver, error) {
 		if err != nil {
 			return nil, err
 		}
+		defer f.Close()
 		return newManifestRunfilesResolver(f)
 	}
 


go/tools/bazel/runfiles/directory.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2018 Google Inc.
```
jayconrod: Please change to `Copyright 2018 The Bazel Authors.`. Same for other new .go files in this PR.

go/tools/bazel/bazel.go
```diff
@@ -20,43 +20,72 @@ import (
...
 
 const TEST_SRCDIR = "TEST_SRCDIR"
 const TEST_TMPDIR = "TEST_TMPDIR"
 const TEST_WORKSPACE = "TEST_WORKSPACE"
 
-var defaultTestWorkspace = ""
+var (
+	defaultTestWorkspace = ""
+
+	// Runfiles resolver state. We can't use sync.Once because of error handling.
```
jayconrod: I think you can actually use `sync.Once` here. Declare separate variables for `runfileResolver` and `runfileResolverErr error` and set both inside the initialization function.

go/tools/bazel/runfiles/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+// These are the possible Runfiles-related errors.
+var (
+	ErrNoRunfilesEnv   = errors.New("runfiles environment missing")
+	ErrManifestInvalid = errors.New("runfiles manifest syntax error")
+)
+
+// Resolver is an interface for a resolver that can take a runfiles path and
+// resolve it to a path on disk.
+type Resolver interface {
+	Resolve(string) (string, error)
```
jayconrod: I think it's better to return `(string, bool)` instead of `(string, error)`. In the case where a file is not found in a manifest, it's not really an error, it's just not found (and the caller will try a different variant of the path).

go/tools/bazel/runfiles/resolver.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// NewResolver creates a new runfiles resolver. The type of resolver and its
+// parameters are derived from the environment.
+func NewResolver() (Resolver, error) {
+	return newResolverWithGetenv(os.Getenv)
+}
+
+func newResolverWithGetenv(getenv func(string) string) (Resolver, error) {
```
jayconrod: I'd rather not inject a test replacement for `getenv`. It'd be way too easy to accidentally call `os.Getenv` explicitly in a future change.

I think it's better for the tests to explicitly set the environment variables. To make sure they get cleaned up at the end of each test case, define a helper function like this:

```
defer setenvForTest(key, value string) (cleanup func()) {
  if old, ok := os.LookupEnv(key); ok {
    cleanup = func() { os.Setenv(key, old) }
  } else {
    cleanup = func() { os.Unsetenv(key) }
  }
  os.Setenv(key, value)
  return cleanup
}
```

Then call like this:

```
cleanupEnv := setenvForTest(RUNFILES_MANIFEST_FILE, manifestfn)
defer cleanupEnv()
```

go/tools/bazel/runfiles/resolver_test.go
```diff
@@ -0,0 +1,111 @@
...
+func TestManifestRunfiles(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer os.RemoveAll(dir)
+
+	testStr := "This is a test"
+	mappedfn := filepath.Join(dir, "mapped_file.txt")
```
jayconrod: Nit: rename to `mappedFilename` and `manifestFilename`. `fn` is a more common abbreviation for "function".

go/tools/bazel/runfiles/resolver_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"testing"
+)
+
+func fakeGetenv(env map[string]string) func(string) string {
+	return func(key string) string {
+		return env[key]
+	}
+}
+
+func TestManifestRunfiles(t *testing.T) {
```
jayconrod: I think it would be better to test against `bazel.Runfile` directly instead of the `Resolver` interface. I'd prefer `Resolver` not be a public API, and it's better to test against public APIs whenever possible.

go/tools/bazel/runfiles/BUILD.bazel
```diff
@@ -0,0 +1,19 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
```
jayconrod: I think these sources should be folded into `//go/tools/bazel:go_default_library`.

`bazel.Runfile` should be the only function that test authors need to call directly. I don't think there's any reason to instantiate a `runfile.Resolver` outside of the `bazel` package. I'd prefer to avoid introducing public APIs if at all possible, so I think it makes sense for the `Resolver` interface to be private, folded into the `bazel` package.

jayconrod:
> Notably, this supports the Bazel runfiles manifest file. I don't know if there's a standard for the format of this file, so I just parse it similar to how other Bazel libraries are doing it. I also don't know if it's intentional that it looks like spaces in filenames would break this format confused Maybe that ought to be revisited some day.

Sadly, that's the state of things.

I think Bazel considered the manifest file more or less an internal format. There are libraries (in `@bazel_tools`) for Java, C++, and bash to locate runfiles, but there's no spec for Starlark rules to implement similar libraries. And yes, spaces in filenames will break everything.

go/tools/bazel/runfiles/directory.go
jchv: Done.

go/tools/bazel/bazel.go
jchv: You can, but of course that means you won't retry again in case of error. At first I thought that would be undesirable for tests, but then I realized it's also kind of undesirable for tests if your success is cached indefinitely, so may as well go both ways. Done.

go/tools/bazel/runfiles/resolver.go
jchv: SGTM, Done. For context, originally I had the manifest impl checking actual file existence and the directory impl resolving symlinks. I rebuked on this but never changed the interface back to booleans.

go/tools/bazel/runfiles/resolver.go
jchv: Works for me, Done. I am paranoid relying on serial execution of tests, but it seems tests do not run in parallel by default, so it's a non-issue.

go/tools/bazel/runfiles/resolver_test.go
jchv: Done.

go/tools/bazel/runfiles/BUILD.bazel
jchv: Done.

go/tools/bazel/runfiles/resolver_test.go
jchv: Alright, I've tried to make the tests use the public API. The memoization of the initialization adds a complication but otherwise it's a straightforward change. Let me know if I should do more to make this agnostic.

jchv:
Alright, I've tried to integrate the feedback. (I force pushed once to reset author to my google address, sorry.)

go/tools/bazel/runfiles.go
```diff
@@ -0,0 +1,53 @@
...
+type runfilesResolver interface {
+	Resolve(string) (string, bool)
+}
+
+// newRunfilesResolver creates a new runfiles resolver. The type of resolver and its parameters are derived
+// from the environment.
+func newRunfilesResolver() (runfilesResolver, error) {
+	manifest := os.Getenv(RUNFILES_MANIFEST_FILE)
+	if manifest != "" {
+		f, err := os.Open(manifest)
```
ash2k: If I'm reading this correctly, the file is not closed properly

jchv: Whoops, thanks.

filipesilva:
Great work @jchv!

jchv:
@filipesilva Thanks! Appreciated.

PR #1818, opened 2018-11-07
Add @com_github_golang_protobuf//protoc-gen-go/generator:go_default_library_gen

### LINKED ISSUE #1817: Create _gen variant for //protoc-gen-go/generator
The target `@com_github_golang_protobuf//protoc-gen-go/generator:go_default_library` depends on `@com_github_golang_protobuf//protoc-gen-go/descriptor:go_default_library`. This can lead to a conflict:

```
GoLink: warning: package "github.com/golang/protobuf/protoc-gen-go/descriptor" is provided by more than one rule:
    @com_github_golang_protobuf//protoc-gen-go/descriptor:go_default_library
    @io_bazel_rules_go//proto/wkt:descriptor_go_proto
Set "importmap" to different paths in each library.
This will be an error in the future.
```

This warning occurs when compiling grpc-gateway with the latest rules_go/gazelle, as [types.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/7d429ea4af6c17ae773bafb6dffae8e851a9b35f/protoc-gen-grpc-gateway/descriptor/types.go#L7) depends on both `protoc-gen-go/descriptor` and `protoc-gen-go/generator`.

I would expect the target `@com_github_golang_protobuf//protoc-gen-go/generator:go_default_library_gen` to exist to avoid this problem. Could this be added?

jayconrod:
It's important for go_default_library not to depend on any
go_proto_library because protoc-gen-go depends on it. However, other
binaries may need to depend on generator and go_proto_library targets,
and this helps them avoid conflicts.

Fixes #1817

### DIFF
--- a/third_party/com_github_golang_protobuf-extras.patch
+++ b/third_party/com_github_golang_protobuf-extras.patch
@@ -1,6 +1,6 @@
 diff -urN b/conformance/internal/conformance_proto/BUILD.bazel c/conformance/internal/conformance_proto/BUILD.bazel
---- b/conformance/internal/conformance_proto/BUILD.bazel	2018-09-28 12:25:59.940105588 -0400
-+++ c/conformance/internal/conformance_proto/BUILD.bazel	2018-09-28 12:28:40.234183077 -0400
+--- b/conformance/internal/conformance_proto/BUILD.bazel	2018-11-07 14:11:49.292167270 -0500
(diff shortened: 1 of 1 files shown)


jayconrod:
@drigz Could you verify this works for you before I merge it?

drigz:
Thanks, it works! Should I file an issue against bazel-gazelle so it's used in generated BUILD files?

jayconrod:
@drigz Thanks for verifying. Please do file an issue on Gazelle. I'll add a new case tomorrow.

PR #1828, opened 2018-11-22
Update protobuf version to 7b28271a61a3da0a37f6fda399b0c4c86464e5b3

meteorcloudy:
See https://github.com/bazelbuild/bazel/issues/6675#issuecomment-438741737

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -60,9 +60,9 @@ def go_rules_dependencies():
     _maybe(
         http_archive,
         name = "com_google_protobuf",
-        # v3.6.1, latest as of 2018-09-28
-        urls = ["https://codeload.github.com/google/protobuf/zip/48cb18e5c419ddd23d9badcfe4e9df7bde1979b2"],
-        strip_prefix = "protobuf-48cb18e5c419ddd23d9badcfe4e9df7bde1979b2",
+        # v3.6.1.1, latest as of 2018-11-22
+        urls = ["https://github.com/protocolbuffers/protobuf/archive/v3.6.1.1.zip"],
+        strip_prefix = "protobuf-3.6.1.1",
         type = "zip",
     )
     _maybe(
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -60,9 +60,12 @@ def go_rules_dependencies():
     _maybe(
         http_archive,
         name = "com_google_protobuf",
-        # v3.6.1.1, latest as of 2018-11-22
-        urls = ["https://github.com/protocolbuffers/protobuf/archive/v3.6.1.1.zip"],
-        strip_prefix = "protobuf-3.6.1.1",
+        # Latest tagged version at time of writing is v3.6.1, which doesn't
+        # include fixes for --incompatible_package_name_is_a_function,
+        # --incompatible_new_actions_api, and possibly others.
+        # TODO: Update to a newer tagged version when available.
+        urls = ["https://codeload.github.com/google/protobuf/zip/7b28271a61a3da0a37f6fda399b0c4c86464e5b3"],
+        strip_prefix = "protobuf-7b28271a61a3da0a37f6fda399b0c4c86464e5b3",
         type = "zip",
     )
     _maybe(


meteorcloudy:
Can we have a new release including this fix so that we can upgrade rules_typescript and finally enable it in Bazel's downstream test?

laurentlb:
Does it work with `--incompatible_package_name_is_a_function=false`?
My understanding is that the recent Proto releases didn't include the fix to the bzl files.

jayconrod:
@meteorcloudy Could you confirm this is a stable release? I see a tag for it in the protobuf repository, but there's no release notes and no artifacts.

Once this is in, I'll probably update all the other dependencies and do a release.

@laurentlb I haven't confirmed this, but is protocolbuffers/protobuf#5382 what you're thinking of?

laurentlb:
I'm thinking of all changes to protobuf.bzl: https://github.com/protocolbuffers/protobuf/commits/master/protobuf.bzl

When I checked, https://github.com/protocolbuffers/protobuf/pull/4650 was not in the Protobuf release (but I haven't checked the release from last week).

laurentlb:
I've just tried, bzl changes are not included here.

I suggest using this instead (I've used it in other rules as well):

```python
    _maybe(
        http_archive,
        name = "com_google_protobuf",
        strip_prefix = "protobuf-7b28271a61a3da0a37f6fda399b0c4c86464e5b3",
        sha256 = "d625beb4a43304409429a0466bb4fb44c89f7e7d90aeced972b8a61dbe92c80b",
        urls = ["https://github.com/google/protobuf/archive/7b28271a61a3da0a37f6fda399b0c4c86464e5b3.zip"],  # 2018-11-16
        type = "zip",
    )
```

jayconrod:
I'm cautious about using untagged versions of something as important as protobuf, but if other rules are using that version without problems, that should be fine.

One small nit: are we confident the zip files produced by `github.com/google/protobuf/archive` have stable SHA-256 sums? The zip files served by `codeload.github.com` have changed SHA-256 sums in the past after GitHub server software updates. I've mostly stopped including `sha256` attributes on `http_archive` unless I can point to a release artifact prepared by the maintainer.

meteorcloudy:
@jayconrod Yes, I can confirm 3.6.1.1 is a stable release. It's just one cherry pick on top of 3.6.1 which won't affect the built binaries. The protobuf release process is complicated, so we agree it's fine to just have a tag for 3.6.1.1. See https://github.com/bazelbuild/bazel/issues/6675#issuecomment-439172800

laurentlb:
@meteorcloudy see my comments. 3.6.1.1 is still broken and doesn't contain the fixes.

My suggestion is consistent with other repos, e.g.
https://github.com/bazelbuild/skydoc/blob/master/WORKSPACE#L24
https://github.com/tweag/rules_haskell/blob/master/WORKSPACE#L38

meteorcloudy:
@laurentlb I've asked a 3.6.1.2 for protobuf, https://github.com/protocolbuffers/protobuf/pull/4650#issuecomment-442000923, if they agree, we don't have to use protobuf at random commit. 
But I'm totally fine to follow your suggestion to fix things sooner! ;)

laurentlb:
I'd really prefer to get it submitted now (it turns out every Bazel repository uses rules_go, so it's blocking everyone). We can update when we have a proper release of Proto.

meteorcloudy:
It's done, please take a look~

laurentlb:
Tests are failing, because protobuf now has a dependency on skylib.
After discussion with @dslomov, I suggest adding the dependency on skylib in repositories.bzl (even if rules_go already vendors skylib):

```python
    _maybe(
        http_archive,
        name = "bazel_skylib",
        sha256 = "54ee22e5b9f0dd2b42eb8a6c1878dee592cfe8eb33223a7dbbc583a383f6ee1a",
        strip_prefix = "bazel-skylib-0.6.0",
        urls = ["https://github.com/bazelbuild/bazel-skylib/archive/0.6.0.zip"],
        type = "zip",
    )
```

I've verified tests pass after this change.

jayconrod:
Adding `bazel_skylib` to `go_rules_dependencies` should be fine, but I'd like there to be a comment something to this effect:

```
# bazel_skylib is a dependency of com_google_protobuf.
# Nothing in rules_go may depend on bazel_skylib, since it won't be declared
# when go/def.bzl is loaded. The vendored copy of skylib in go/private/skylib
# may be used instead.
```

jayconrod:
Closing, as this was superceded by #1834.

Issue #1829, opened 2018-11-22
Fix Bazel incompatible changes


laurentlb:
I'd like to be able to build rules_go with `--all_incompatible_changes`:
  `cd rules_go && bazel build --all_incompatible_changes //...`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

PR #1830, opened 2018-11-22
Explicitly reexport some symbols, reformat the .bzl files with Buildifier

laurentlb:
In the future, Bazel will not automatically re-export the symbols that come from a load. This PR is fixing that.

This diff is large, because I also reformatted all files with Buildifier. This will allow you to use Buildifier automated fixes in the future.

Progress towards #1829

### DIFF
--- a/go/private/common.bzl
+++ b/go/private/common.bzl
@@ -12,12 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//go/private:skylib/lib/dicts.bzl", "dicts")
-load("//go/private:skylib/lib/paths.bzl", "paths")
-load("//go/private:skylib/lib/sets.bzl", "sets")
-load("//go/private:skylib/lib/shell.bzl", "shell")
-load("//go/private:skylib/lib/structs.bzl", "structs")
-load("@io_bazel_rules_go//go/private:mode.bzl", "mode_string")
+load("//go/private:skylib/lib/dicts.bzl", _dicts = "dicts")
+load("//go/private:skylib/lib/paths.bzl", _paths = "paths")
+load("//go/private:skylib/lib/sets.bzl", _sets = "sets")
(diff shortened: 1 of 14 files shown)


go/private/common.bzl
```diff
@@ -12,12 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//go/private:skylib/lib/dicts.bzl", "dicts")
-load("//go/private:skylib/lib/paths.bzl", "paths")
-load("//go/private:skylib/lib/sets.bzl", "sets")
-load("//go/private:skylib/lib/shell.bzl", "shell")
-load("//go/private:skylib/lib/structs.bzl", "structs")
-load("@io_bazel_rules_go//go/private:mode.bzl", "mode_string")
+load("//go/private:skylib/lib/dicts.bzl", _dicts = "dicts")
```
jayconrod: There shouldn't be any need to re-export Skylib symbols from common.bzl. .bzl files that need Skylib symbols should load them directly.

Same for `mode_string`, actually.

go/private/skylib/lib/collections.bzl
```diff
@@ -14,60 +14,56 @@
...
 """Skylib module containing functions that operate on collections."""
 
-
 def _after_each(separator, iterable):
-  """Inserts `separator` after each item in `iterable`.
-
-  Args:
-    separator: The value to insert after each item in `iterable`.
-    iterable: The list into which to intersperse the separator.
+    """Inserts `separator` after each item in `iterable`.
```
jayconrod: I'd prefer not to make any local changes to our copy of Skylib (everything under go/private/skylib). These changes will be erased next time we update.

It looks like the latest version of Skylib has been formatted with Buildifier. I'll update that in a separate PR.

laurentlb:
Comments should be resolved

jayconrod:
In the future, Bazel will not automatically re-export the symbols that come from a load. This PR is fixing that.

Progress towards #1829

PR #1834, opened 2018-11-27
Fix for remaining Bazel incompatible changes

### LINKED ISSUE #1829: Fix Bazel incompatible changes

laurentlb:
- Update proto version
- Update Gazelle version
- Fix a remaining issue

Tested:
  bazel test //... --all_incompatible_changes

Fixes #1829

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -12,7 +12,7 @@ go_register_toolchains()
 # Needed for tests
 git_repository(
     name = "bazel_gazelle",
-    commit = "682c0e396e708a146e28f0796ae08b7913e2d9a9",  # master as of 2018-10-02
+    commit = "455e69320ee92c6f3bfb267aa211a4fa6ebc4e5d",  # master as of 2018-11-27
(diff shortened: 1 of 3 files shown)


laurentlb:
This replaces https://github.com/bazelbuild/rules_go/pull/1828 (because @meteorcloudy left the office and timezone differences make it hard to iterate), with additional fixes.

prattmic:
FWIW, it would be helpful if the release notes mentioned fixes like this. When my build broke this morning due to an old version of the protobuf repo, it took a long time to figure out that the latest release of rules_go already fixed this. The release notes make it sound like the only changes were new versions of Go.

jayconrod:
@prattmic Thanks for the feedback. I'm not really sure what level of detail is useful to people in release notes, but I'll include more information in the future.

PR #1840, opened 2018-12-03
Fix borken handling of 32-bit ARM platform

likan999:
The platform is represented inconsistently across the repository, some are armv6l, and in go/platform/list.bzl it is arm. This commit replaces all occurences of armv6l to arm because in the official package it is using arm rather than armv6l.


jayconrod:
Could you explain exactly how the current implementation is broken? I'm not really clear on the issue that this PR fixes. The host platform detection shouldn't work any differently right? But something like the example below would start working:

```
go_download_sdk(
    name = "go_sdk",
    goos = "linux",
    goarch "arm", # instead of armv6l
    sdks = SDK_TOOLCHAINS["1.11.2"],
)
```

However, this change would break anyone with `goarch = "armv6l"`.

likan999:
Are you sure goarch=“armv6l” works at the moment? If you look at https://github.com/bazelbuild/rules_go/blob/master/go/private/BUILD.sdk.bazel

You will find it expects a directory {goos}_{goarch} in the downloaded package. In the official package there is a directory linux_arm but not a linux_armv6l.

jayconrod:
Ok, I see what you're saying. `armv6l` is not a standard `GOARCH` value, but it looks like the SDK is named that way to indicate which `GOARM` it was prepared with.

Since you can confirm this works for `arm`, I'll merge this. However, be aware that `arm` is not officially supported, and we don't have an `arm` CI platform or `arm`-specific tests.

PR #1845, opened 2018-12-12
Make gofmt available via sdk.tools

vam-google:
`gofmt` is needed to format GAPIC generator output and make it compliant with Go code style best practices.


vam-google:
@jayconrod I removed .srcjar support as Bazel 0.19.0+ with `--incompatible_expand_directories` enabled plus a small hack (creating directories which end with `.go` "extension" so it can full the file extensions validation in rules_go) allow to build and test the set of .go files.

jayconrod:
@vam-google Great, I'm glad that worked without needing changes. I'll look at further support for directories in the new year, though I expect it will take some time.

To recap the earlier email thread:

* For the `gofmt` change, I'd prefer adding `gofmt` to the `tools` `filegroup` instead of adding a new property for it.
* Renaming `go_googleapis` can't be done atomically. We need a multi-step migration to avoid breaking people.

FYI, I have fixes in the pipeline for some of the `go_proto_library` issues we discussed earlier.

vam-google:
@jayconrod 

About `gofmt`. 

`gofmt` is implemented this way because it is a separate binary, which resides in same directory as `go` binary, that is why in `rules_go` it has a new property. The idea is to make rules_go sdk to correspond to actual go folder structure. Note, currently this is the case (as far as I can tell): sdk representation in `rules_go` and file structure of Golang sdk correspond to each other, if I move `gofmt` under tools, I will break this consistency between the two. 

Also, in `providers.bzl` on `Line 66` the description (for "tools" property) explicitly says: `"List of executable files from pkg/tool"`. Moving `gofmt` under tools will make that description wrong (`gofmt` binary is **not** under `pkg/tool`).

I strongly believe that `gofmt` should have its own label, but if you want me to move it under tools I can do it, it is not a problem. Do you still want me to move it under tools?

About `go_googleapis` vs `com_google_googleapis`.

I have two important questions here:

1) What are we going to do with one `proto_library` rule per go package priciple in go_googleapis vs one `proto_library` rule  per proto file in `com_goole_googleapis`? In my ideal word, we just move to `com_google_googleapis` version without any transitioning, assuming that there is a very low chance that anybody among current users of go_googleapis directly uses proto_library targets declared there. If no, what is your proposal? 

2) Not all "packages" (sets of proto files) have corresponding Bazel rules in com_google_googleapis (because the rules are not autogenerated). I really do not want to add redundant targets (and there will be quite a lot) to everything in googleapis. At least because the protos which do not have rules are not supported ones, and they exist in googleapis only for technical/historical reasons, providing build rules for them is misleading and increases maintenance cost.

vam-google:
@jayconrod A clarification about the most problematic part of go_googleapis vs com_google_googleapis conflict.

Unfortunatelly gooogleapis declare a bunch of common proto packages (like rpc, api etc), which are used in other repositories. Specifically `go_googleapis` in used by imported [org_golang_google_grpc](https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl#L134) repository from the corresponding [org_golang_google_grpc-gazelle.patch#L603](https://github.com/bazelbuild/rules_go/blob/master/third_party/org_golang_google_grpc-gazelle.patch) file. 

So the `googleapis` change affects the `grpc` repository (which looks wrong, because `grpc` is more fundamental than `googleapis` and it must be a unidirectional dependency: `googelapis` depends on `grpc`, not vice versa, but...).

jayconrod:
@vam-google Sorry for flow response -- I was out last week and am still catching up on email.

About gofmt, I'd prefer not to add a new field to `GoSDK`. I don't think it's important for `GoSDK` to match the directory structure; I think bin/go is separate from pkg/tool just because you wouldn't want the stuff in pkg/tool to be in your `PATH`, but that doesn't matter to Bazel. It makes more sense to me to change the `tools` description to "list of executable files in the SDK built for the execution platform, not including the Go command".

About migrating `proto_library` and `go_proto_library` targets:

1. You've mentioned earlier that `com_google_googleapis` will have one `proto_library` per .proto file. Will it also have a `proto_library` per package grouping those files? If yes, the `go_googleapis` `proto_library` targets can be changed to `alias` rules. If no, the `go_googleapis` `proto_library` targets can just have `deps` (no `srcs`) that point to the `com_google_googleapis` `proto_library` targets.

Either way, Gazelle will get updated dependency resolution rules and will help migrate users to the new targets before we turn down `go_googleapis`. 

We can't do this migration in a single step. Anyone who has a `proto_library` that depends on `go_googleapis` would be broken. I expect a lot of people are using things like `google.rpc.Status` and `google.longrunning`.

2. Could you provide some examples, just so I understand what packages should not be built? From the Go perspective, I think `com_google_googleapis` should be compatible with `google.golang.org/genproto`, a.k.a, `github.com/golang/genproto`, a.k.a. `@org_golang_google_genproto`. That repository is maintained by the Go team, but not by me. As I understand, it's generated from everything in googleapis, so it may have the same problem.

About circular dependencies, we may be able to avoid some of these. The gRPC plugin itself should be built with dependencies on `org_golang_google_genproto` instead of `go_googleapis`. Bazel allows circular dependencies at the repository level, just not the target level, so this shouldn't block anything.

vam-google:
@jayconrod `com_google_googleapis` won't have proto_library per Go package. This is technically possible to include it, but readability and simplicity of the `BUILD.bazel` files in googleapis is one of the main priorities for us. Also the configurations for different languages should look as similar as possible (ideally be the same, with only prefixes for rules being language-dependent). If https://github.com/bazelbuild/rules_go/issues/1856 is done relatively quickly, it is possible (othewise we would need the prosy library defined anyways, just for technical reasons).

I'm fine with having "meta/proxy" `proto_library` rules in `go_googleapis` which group different `proto_libraries` by Go package (this is under assumption that https://github.com/bazelbuild/rules_go/pull/1870 is pushed first), in this case there is no need to have aliases. Also once https://github.com/bazelbuild/rules_go/issues/1856 is implemented, it would be possible to do everything a user wants without having the package-level `proto_library` so I believe leaving package-level proto libraries in `go_googleapis` only (and not introducing them in `com_google_googleapis`) is the right decision.

As for `genproto` repository you are right. Id id not realize that they also include pretty much everything from googleapis. A good example of stuff in googleapis which should not be used is most of the stuff under [devtools](https://github.com/googleapis/googleapis/tree/master/google/devtools) (except few, like cloudtrace). Apparently genproto regenerates go stubs for everything in googleapis which has `go_package` option which starts with `"google.golang.org/genproto"`, which is pretty much every proto file there (see [regen.sh](https://github.com/google/go-genproto/blob/master/regen.sh) and [regen.go](https://github.com/google/go-genproto/blob/master/regen.go) for details)

Well, this is only Go so far... Most probably other languages (not java, but at least one more language in addition to Go) are also doing something like that =(. It looks like it would be easier to just include the `proto_library` rules for every .proto file than to convince everytbody that it should not be done.

So a very specific question about how the temporary step of having both `go_googleapis` and `com_google_googleapis` should look like? 

1) How can it be implemented technically? 
Assuming we have `com_google_googleapis` ready, by ready I mean the following:
- it has `proto_library` for each for each .proto
- it has corresponding go_proto_library defined properly (the way it is done depends on if https://github.com/bazelbuild/rules_go/issues/1856 is implemented first or no;

2) What is next? How is aliasing done? Do you mean that googleapis repository will be imported twice under two different names (if it is even allowed by bazel, if not allowed there can be "fake" repository for go_googleapis), and `go_googleapis` patch files will be referrencing com_google_googleapis stuff in it (so basically proxying new `com_google_googleapis`? 

3) If the assumption above is true, how do you envision the creation of the "proxying" patch files for `go_googleapis`? Can it be done manually as one-time job, or are you planning to make corresponding changes in gazelle?

jayconrod:
I'll try to get #1856 done within the next week. It should be a simple enough change that it doesn't make sense to add temporary measures to work around it. Once that's part of a tagged release (ideally next week, more conservatively, early the week of Jan 14), you can start adding rules.

So then the migration would look like this:

1. #1856 is included in a tagged release of rules_go, along with any other changes needed to make step 2 possible. This may include changes to `org_golang_google_grpc` to reduce dependencies on `go_googleapis`.
2. You can point `com_google_googleapis` to the new version of rules_go and add `go_proto_library` rules as appropriate. During this time, `go_googleapis` is effectively a fork of `com_google_googleapis` with different build files.
3. When `go_proto_library` rules are in place, `com_google_googleapis` can be added to `go_rules_dependencies`. `go_googleapis` will still be a separate workspace, but its build files will contain aliases and "proxy" `proto_library` rules to the real rules in `com_google_googleapis`. This will be part of a new rules_go release.
4. Gazelle's dependency resolution rules will be updated to point to `com_google_googleapis` instead of `go_googleapis`. Users should be able to run Gazelle to fix all their dependencies without needing to touch their build files.
5. At some point, `go_googleapis` can be removed from `go_rules_dependencies`. Users that haven't migrated will need to use an older version of rules_go.

vam-google:
@jayconrod Ok, but I'm still not sure how the step 3 should look like. How the go_googleapis the proxying rules will be created? Manually or no? Update to gazelle (step 4), as I understand it means that it basically uses "com_google_googleapis" instead of "go_googleapis" string literal in its generator logic. 

If the proxy rules/aliases created as a one-time job and manually it means that stuff must be kept in sync. Also the longer we try to keep them in sync (i.e. the longer go_googleapis exists in general) the more it "costs" for us (in terms of maintenance effort).

At the same time, trying to do it automatically (by some update to gazelle, I assume) feels like an overkill and simply not worth it (as for me at least) and will also add the cleanup work once the go_googleapis go away.

(Btw moved gofmt under tools).

I'm happy to do the #1856 myself if you are ok with it. I'm also ready to help with the gazelle change and the "proxying" work (but it depends on what exactly it means).

jayconrod:
> How the the proxying go_googleapis rules will be created? Manually or no?

Somewhat manual, but mostly automatic. We can build a list of `proto_library` targets, `go_proto_library` targets, and .proto files in the current repository using `bazel query` and `find`, then use a throwaway Python script or something to write build files to a clean checkout of the googleapis repository. Then just diff to create a patch. For reference, [Updating dependencies](https://github.com/bazelbuild/rules_go/wiki/Updating-dependencies) describes the process for producing patches for other dependencies. This will be different, since Gazelle won't be used for this step, but it's a useful starting point.

> If the proxy rules/aliases created manually as a one-time job, it means that the stuff must be kept in sync. 

I was assuming they didn't need to be kept in sync, but I'm not sure that's true. How often are breaking changes made in googleapis, and how significant are the breaks? If it's not often (hopefully the case), we wouldn't have to keep it in sync. People would need to migrate to `com_google_googleapis` to get the new stuff. If it's very often, then the above process would need to be more automatic to keep the cost reasonable. But keep in mind, most people won't be at tip at either of these repositories; they'll be at the latest version of googleapis on the day rules_go was tagged.

> I'm happy to do the #1856 myself if you are ok with it. I'm also ready to help with the gazelle change and the "proxying" work (but it depends on what exactly it means).

Sounds good, thanks! Couple notes to save you some time:

* The `go_proto_library` rule is in `proto/def.bzl`. We'll need a new `protos` attribute. (I thought we could define a macro wrapper for this, but it would involve renaming the rule, which would change query results. There's a workaround for that, but let's not mess with it for now.)
* `go_proto_compile` in `proto/compiler.bzl` will need to be changed to accept a list of `protos` instead of a single `proto`. Technically, this is a public interface, but I don't believe anyone actually uses it. It should be okay to change.
* The builder in `go/tools/builders/protoc.go` will need to be updated. This is the complicated part.
* A test in tests/core/go_proto_library should cover multiple protos.

Let me know if you have questions or run into any issues.

vam-google:
@jayconrod I've converted this PR to the small piece which should be ok to push as is now. Please take a look, so I can push at least a small part of required changes. The other parts will be handled separately.

jayconrod:
Looks good. Will submit when CI is green.

jayconrod:
Failures are due to something I think was fixed in #1870. Rebased and running again...

PR #1864, opened 2018-12-18
Pure mode when using the goos attribute

siddharthab:
Currently, the `goos` attribute on a go_binary rule is ignored when determining whether to force pure mode. The only check being performed right now is using the toolchain target.

This change forces pure mode for a binary when the toolchain target is the same as the host, but the specific binary is cross-compiled using the goos attribute. All other cases remain the same.

### DIFF
--- a/tests/core/go_binary/BUILD.bazel
+++ b/tests/core/go_binary/BUILD.bazel
@@ -19,5 +19,14 @@ go_binary(
     srcs = ["custom_bin.go"],
     out = "alt_bin",
 )
-    
+
+go_binary(
+    name = "goos_pure_bin",
+    srcs = [
+        "broken_cgo.go",
+        "hello.go",
+    ],
+    goos = "plan9",
+)
+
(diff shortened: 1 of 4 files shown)


tests/core/go_binary/BUILD.bazel
```diff
@@ -19,5 +19,14 @@ go_binary(
     srcs = ["custom_bin.go"],
     out = "alt_bin",
 )
-    
+
+go_binary(
+    name = "goos_pure_bin",
+    srcs = [
+        "broken_cgo.go",
```
jayconrod: It looks like this file is still being included in the build. I don't believe anything will filter out the `_cgo.go` suffix; that only works for goos / goarch suffixes. Instead, could you add `// +build cgo` to the top of the file?

siddharthab: Thanks for the suggestion. Looks like the tests are passing now.

PR #1868, opened 2018-12-20
Add documentation on configuring C/C++ toolchains [skip ci]

jayconrod:


jayconrod:
@ixdy Could you read this and let me know what you think?

This is a minimal example. Ideally, it would have instructions for building a custom, hermetic, cross-compiling toolchain, but I haven't had time to figure that out (and probably won't for a while to be honest). But at least it shows how to install a toolchain into a repository that can be referenced from other workspaces and how to get Bazel to use it.

go/crosstool.rst
```diff
@@ -0,0 +1,340 @@
...
+          "lib/**",
+          "libexec/**",
+          "share/**",
+      ]),
+  )
+
+Next, we'll create a ``cc_toolchain`` target that tells Bazel where to find some
+important files. This API is undocumented and will very likely change in the
+future. We need to create one of these for each ``toolchain`` in ``CROSSTOOL``.
+The ``toolchain_identifier`` and ``cpu`` fields should match, and the
```
ixdy: maybe clarify that "should match" here means "should match the values in `CROSSTOOL`. At first I was confused becfause I thought that `toolchain_identifier` should be the same as `cpu`, which it's not.

ixdy:
Thanks, this has been helpful for understanding the crosstool setup (along with reading lots of other docs).

I've been working on extending this to support cross-compiling (currently from linux/amd64 -> linux/arm).

I got something kinda working; I added the following to my `CROSSTOOL`:
```proto
toolchain {
  toolchain_identifier: "clang-arm"
  host_system_name: "linux"
  target_system_name: "linux"
  target_cpu: "arm"
  target_libc: "arm"
  compiler: "clang"
  abi_version: "unknown"
  abi_libc_version: "unknown"

  tool_path { name: "ar" path: "bin/llvm-ar" }
  tool_path { name: "cpp" path: "bin/clang-cpp" }
  tool_path { name: "dwp" path: "bin/llvm-dwp" }
  tool_path { name: "gcc" path: "bin/clang" }
  tool_path { name: "gcov" path: "bin/llvm-profdata" }
  tool_path { name: "ld" path: "bin/ld.lld" }
  tool_path { name: "nm" path: "bin/llvm-nm" }
  tool_path { name: "objcopy" path: "bin/llvm-objcopy" }
  tool_path { name: "objdump" path: "bin/llvm-objdump" }
  tool_path { name: "strip" path: "bin/llvm-strip" }

  compiler_flag: "-no-canonical-prefixes"
  linker_flag: "-no-canonical-prefixes"

  cxx_builtin_include_directory: "/usr/arm-linux-gnueabihf/include"

  linker_flag: "-fuse-ld=lld"

  compiler_flag: "-target"
  compiler_flag: "arm-linux-gnueabihf"
  compiler_flag: "-mfloat-abi=hard"
  linker_flag: "-target"
  linker_flag: "arm-linux-gnueabihf"

  compiler_flag: "-I/usr/arm-linux-gnueabihf/include"
  linker_flag: "-L/usr/arm-linux-gnueabihf/lib"

  compiler_flag: "-Wno-builtin-requires-header"
}
```

This requires running on a machine (or in a docker container) with the `libc-dev-armhf-cross` and `libgcc-6-dev-armhf-cross` debian packages installed, which isn't great (and won't work with RBE). (I also needed to explicitly set `pure=off` on the `go_binary` target).

I'd like to find some way to not depend on those packages, though it seems tricky. Debian has prebuilt packages for all relevant architectures, but I'm not sure we want to depend on them. Linaro has prebuilt sysroots for arm, but not ppc/s390x, which Kubernetes needs. The Chromium project also has prebuilt sysroots, but again not for ppc/s390x.

Another option to consider is using musl, though we still have the issue of needing to build everything, we'd probably want to statically link rather than dynamically link (maybe better), and there may be subtle differences from glibc, which we might not want.

We could drop the dependency on `libgcc` if we use clang's compiler-rt, but the clang binary distribution doesn't ship libraries for cross-targets, so we'd need to build those somehow too.

Along the way, I've also found a few other interesting projects:
https://github.com/vsco/bazel-toolchains - maybe not maintained anymore, but uses clang/llvm + chromium's sysroots (but not cross-compiling)
https://github.com/grailbio/bazel-toolchain - more recent, automatically downloads the relevant clang/llvm release tarball, not hermetic (uses system headers/libraries)

PR #1870, opened 2018-12-21
go_proto_library: support proxy proto_library

### LINKED ISSUE #1857: go_proto_library: support proxy proto_library
`go_proto_library` should be able to compile `proto_library` targets that only have `deps`. These libraries may be used to aggregate multiple protos in the same package.

For example, this does not build:

```

proto_library(
    name = "foo_proto",
    srcs = ["foo.proto"],
)

proto_library(
    name = "bar_proto",
    srcs = ["bar.proto"],
    deps = [":foo_proto"],
)

proto_library(
    name = "baz_proto",
    srcs = ["baz.proto"],
)

proto_library(
    name = "combined_proto",
    deps = [
        ":bar_proto",
        ":baz_proto",
        ":foo_proto",
    ],
)

go_proto_library(
    name = "combined_go_proto",
    importpath = "example.com/hello",
    proto = ":combined_proto",
)
```

Fixing this may be as simple as using `proto.check_deps_sources` instead of `proto.direct_sources`.

jayconrod:
In some repositories, a proto_library target will be declared for each
.proto file, even for .proto files in the same package. A
proto_library target with deps and no srcs may be used to combine
several proto_library rules.

With this change, go_proto_library treats a "proxy" proto_library as
if it contained the direct srcs of its deps.

Fixes #1857

### DIFF
--- a/tests/core/go_proto_library/proxy_test.go
+++ b/tests/core/go_proto_library/proxy_test.go
@@ -1,0 +1,30 @@
+/* Copyright 2018 The Bazel Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proxy_test
+
+import (
+	"testing"
+
+	"github.com/bazelbuild/rules_go/tests/core/go_proto_library/proxy"
+)
+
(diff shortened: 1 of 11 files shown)


vam-google:
@ianthehat Can you please take a look at this PR? This blocks GAPIC (google cloud autogenerated clients) Bazel extensions work.

tests/core/go_proto_library/proxy_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/bazelbuild/rules_go/tests/core/go_proto_library/proxy"
+)
+
+func use(interface{}) {}
+
+func TestProxy(t *testing.T) {
+	// just make sure both types exist
+	use(proxy.A{})
+  use(proxy.B{})
```
vam-google: Looks like bad formatting (tab vs spaces)

jayconrod: Done. CI needs to run again though. Will submit when it's green.

FYI Ian doesn't work on rules_go anymore, and he's also on vacation. So no need to @ him.

Issue #1486, opened 2018-05-04
cgo: C++ static initialization not working


cstrahan:
Here's the scenario:

* You have a `go_binary` which depends on a `cc_library`, let's call it `cxxlib`
* `cxxlib` in turn has a dependency on a second `cc_library`, let's call it `cxxdep`
* `cxxdep` does one thing: it uses static initialization for some sort of side effect (say, registering some type in a plugin system)
* Both uses of `cc_library` use `linkalways = 1`, as we want to ensure that our symbols remain, despite the fact that some may not appear to the linker as used.

What we expect:

* The initializer(s) should run.

What happens:

* The initializers are not run (presumably because `cxxdep` wasn't actually linked.

----

I have a minimal example that reproduces this here:  https://github.com/cstrahan/rules_go_repro

Instructions:

1. `$ bazel build :demogo`
2. `$ bazel-bin/linux_amd64_stripped/demogo`

In theory, you should see the following two lines, but you'll find that only the second line is printed:

```
Hello from cxxdep!
Hello from cxxlib!
```

You can compare that to what happens with `cc_binary` like so:

1. `$ bazel build :democc`
2. `$ bazel-bin/democc`

You'll see:

```
Hello from cxxdep!
Hello from cxxlib!
```

cstrahan:
Note that this also applies to code like:

```c
__attribute__((constructor))
static void init() {
    puts("Hello from init!");
}
```

cstrahan:
Here's the build log, along with some observations paced inline:

    $ bazel build -s ":demogo" --verbose_explanations
    WARNING: --verbose_explanations has no effect when --explain=<file> is not enabled
    INFO: Analysed target //:demogo (14 packages loaded).
    INFO: Found 1 target...
    SUBCOMMAND: # //:cxxlib [action 'Compiling lib.cc']
    (cd /home/cstrahan/.cache/bazel/_bazel_cstrahan/f02058a447d81a94bd1b377d77d4c78d/execroot/demo && \
      exec env - \
        PWD=/proc/self/cwd \
[...]

PR #1879, opened 2019-01-02
Add support for C dependencies with alwayslink=1

steeve:
This is done via an aspect that walks the C deps graph and builds a
dict between files and alwayslink properties, which is then used when
building the link command line.

Also, the command line is OS dependant, so add support for that too.

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -57,6 +57,40 @@ load(
 
 _CgoCodegen = provider()
 _CgoInfo = provider()
+_CcAlwaysLinkInfo = provider()
+
+_WHOLE_ARCHIVE_FLAGS_BY_GOOS = {
+    goos: ("-Wl,--whole-archive", "-Wl,--no-whole-archive")
+    for goos in GOOS.keys()
+} + {
+    "darwin": ("-Wl,-all_load", ""),
+    "js": ("", ""),
+    "windows": ("", ""),
+}
+
+def _always_link_aspect_impl(target, ctx):
+    # Make sure the rule has a srcs attribute.
+    transitive_always_links = {}
+    alwayslink = getattr(ctx.rule.attr, "alwayslink") in [1, True]
+    for dep in getattr(ctx.rule.attr, "deps", []):
+        if dep[_CcAlwaysLinkInfo]:
+            info = dep[_CcAlwaysLinkInfo]
+            transitive_always_links.update(info.transitive_always_links)
+            for f in dep.files:
+                # Only set the ones that are true
+                if info.alwayslink == True:
+                    transitive_always_links[f] = info.alwayslink
+    return [
+        _CcAlwaysLinkInfo(
+            alwayslink = alwayslink,
+            transitive_always_links = transitive_always_links,
+        ),
+    ]
+
+always_link_aspect = aspect(
+    implementation = _always_link_aspect_impl,
+    attr_aspects = ["cdeps", "deps"],
+)
 
 # Maximum number of characters in stem of base name for mangled cgo files.
 # Some file systems have fairly short limits (eCryptFS has a limit of 143),
@@ -219,6 +253,9 @@ def _cgo_codegen_impl(ctx):
     runfiles = ctx.runfiles(collect_data = True)
     for d in ctx.attr.deps:
         runfiles = runfiles.merge(d.data_runfiles)
+        always_links = {}
+        if d[_CcAlwaysLinkInfo]:
+            always_links = d[_CcAlwaysLinkInfo].transitive_always_links
         if hasattr(d, "cc"):
             inputs = sets.union(inputs, d.cc.transitive_headers)
             deps = sets.union(deps, d.cc.libs)
@@ -230,6 +267,9 @@ def _cgo_codegen_impl(ctx):
             for inc in d.cc.system_include_directories:
                 _include_unique(cppopts, "-isystem", inc, seen_system_includes)
             for lib in as_iterable(d.cc.libs):
+                is_whole_archive = (always_links.get(lib) == True)
+                if is_whole_archive:
+                    linkopts.append(_WHOLE_ARCHIVE_FLAGS_BY_GOOS[go.mode.goos][0])
                 # If both static and dynamic variants are available, Bazel will only give
                 # us the static variant. We'll get one file for each transitive dependency,
                 # so the same file may appear more than once.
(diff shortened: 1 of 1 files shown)


go/private/rules/cgo.bzl
```diff
@@ -57,6 +57,40 @@ load(
 
 _CgoCodegen = provider()
 _CgoInfo = provider()
+_CcAlwaysLinkInfo = provider()
+
+_WHOLE_ARCHIVE_FLAGS_BY_GOOS = {
```
jayconrod: I don't think the `{} + {}` operator is supported anymore in Starlark. If it is, it's deprecated.

I think it would be simpler to just define a function that returns the args though.

go/private/rules/cgo.bzl
```diff
@@ -57,6 +57,40 @@ load(
...
+    goos: ("-Wl,--whole-archive", "-Wl,--no-whole-archive")
+    for goos in GOOS.keys()
+} + {
+    "darwin": ("-Wl,-all_load", ""),
+    "js": ("", ""),
+    "windows": ("", ""),
```
jayconrod: Out of curiosity, why aren't the flags needed on Windows? Might be good to have a comment explaining it.

go/private/rules/cgo.bzl
```diff
@@ -57,6 +57,40 @@ load(
...
+}
+
+def _always_link_aspect_impl(target, ctx):
```
jayconrod: An aspect will be too expensive for this. It gets applied not only to the `cc_library` we depend on through `cdeps`, but to everything it depends on, which could be a lot of things. Aspects consume a lot of time and memory, and I'd love to get rid of the ones we already have.

It's unfortunate that `CcSkylarkApiProvider` doesn't expose this information. Do you know if there's an open issue to add it? If not, maybe we can file one. It seems like a reasonable thing to add.

Until then, I think it would be reasonable to make `-Wl,--whole-archive` the default for all libraries. This would probably fix #1486 (looks like you've already seen that one).

go/private/rules/cgo.bzl
steeve: I just tested on Bazel 0.21 and it worked without warning. Where did you see that?

go/private/rules/cgo.bzl
steeve: They are needed, but I'm not sure what they are

go/private/rules/cgo.bzl
steeve: Is there a way to measure this? We build a pretty complex codebase and we didn't feel it was slow?

go/private/rules/cgo.bzl
steeve: I meant depending on MSVC or MSYS

go/private/rules/cgo.bzl
jayconrod: I don't think it will work with --all_incompatible_changes.

See https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#dictionary-concatenation

go/private/rules/cgo.bzl
jayconrod: cgo doesn't support MSVC, so don't worry about that. I think MSYS should accept gcc flags, but I don't know if there's a slightly different variant on Windows. Maybe just take out the case for Windows, and I can fix it later if it's a problem.

go/private/rules/cgo.bzl
steeve: I've opened https://github.com/bazelbuild/bazel/issues/7033

jayconrod: You can build with `--profile=file` and then `bazel analyze-profile file`. It will tell you how much time Bazel spends in loading, analysis, execution, etc, but it won't say which lines of code cost the most. I don't think there's a way to do that yet.

cgo currently adds a huge amount of loading / analysis overhead because it declares `cc_library` rules for each configuration that could affect build files. That's separate from the `cdeps` libraries, but in any case, it's quite a bit slower than `go build`. An aspect would slow it further.

steeve: I'm not sure using `-Wl,--whole-archive` would be the best way really, it may bloat the binaries significantly.

That said, if we could find a way to get the `go.o` file, and use a normal `cc_library` to link it against the rest (bypassing `go link`), we might get all the niceties for free.

jayconrod: That may be possible, but it would interfere with tooling. `bazel query` would show your target as a `cc_binary` with a `go_binary` dependency. We have some hacks like this already (cgo), but Bazel API changes have made this less necessary.

Thanks for filing bazelbuild/bazel#7033. I think that's the right path forward. I'll comment there.

jayconrod:
Just to recap, Bazel 0.23.0 will introduce a new Starlark provider that will contain an `alwayslink` field. It was submitted in 94e5016. The new provider is not documented yet, so I don't know exactly how it will work. When rules_go migrates to that API (which we'll need to do because the old one will be deprecated), we'll observe that field.

I'll leave this PR open for now, but I may close it after there's a more concrete plan.

steeve:
Going back to this. Would you be open to bump the minimum bazel version to 0.23 ?

steeve:
It would remove this whole hack too: https://github.com/bazelbuild/rules_go/blob/master/go/private/rules/cgo.bzl#L239-L252

jayconrod:
I'd rather not drop support for old Bazel versions, but I hope we won't have to anymore. rules_go now has a compatibility layer that can provide a consistent interface across Bazel API breaks. It's in go/private/compat. Maybe we can add something in there to retrieve this field on versions that support it.

At the moment, I'm working on moving a lot of logic into the execution layer. Which means almost a complete rewrite of cgo. So let's not change anything here just yet. I'll likely ask you to take a look at that branch when it's passing CI here. You have a more complicated use case than most folks, and I want to make sure you're not broken by that change.

steeve:
So I take it you'll emit c artefacts from starlark and not via macro? That's great to hear!

steeve:
I'll try to rebase that PR and make use of the new field so that we can remove the aspect.

That said, it won't work on earlier versions of bazel.

jayconrod:
> So I take it you'll emit c artefacts from starlark and not via macro? That's great to hear!

That's right, we'll build an entire package in one action, including cgo code generation, C compilation, assembly, packing, coverage, and Go compilation.

Big caveat though: it won't work for Objective C because Bazel doesn't expose the Objective C toolchain to Starlark rules. So we'll still follow the old path when `objc = True`.

steeve:
That's amazing.

For objc I think we can away pretty far with using -x objc to a cc_binary and aggregating deps as their are both compatible with each other.

steeve:
@jayconrod i just remembered there is https://docs.bazel.build/versions/master/skylark/lib/apple_common.html#new_objc_provider which could perhaps do the trick?

jayconrod:
I think `new_objc_provider` would be used if an `objc_library` needs to depend on a `go_library` or `go_binary`. Is there a case where we need to do that?

steeve:
I'm not sure, it sure looks to me it could be used to emit an `objc_library` from a rule?

steeve:
Would it be an option to use the aspect on older Bazel versions but the real provider on newer ones ?

steeve:
Also, obviously this PR will change massively related to the changes in `GoCompilePkg`, not sure how much yet.

steeve:
Closing this, will reopen when the need arises.

Issue #1888, opened 2019-01-08
Fix incompatibility errors for Bazel 0.22


jayconrod:
The following flags indicate problems:

* `--incompatible_disable_legacy_cc_provider` - breaks cgo. Not sure what the replacement is. bazelbuild/bazel#7036
* `--incompatible_disable_legacy_proto_provider` - breaks `go_proto_library`. `ProtoInfo` is the replacement, but it's undocumented. bazelbuild/bazel#6901
* `--incompatible_disable_genrule_cc_toolchain_dependency` - some tests build .so files with `genrules` that reference `$(CC)`. We should be able to use `cc_binary` for this now.
* `--incompatible_depset_is_not_iterable` - recent unrelated proto change regressed here. Easy fix though.

PR #1889, opened 2019-01-08
Fix some incompatibilities in Bazel 0.22.0

jayconrod:
* go_proto_library now relies on the undocumented ProtoInfo instead of
  the .proto field.
* go_proto_library now uses to_list() when iterating ProtoInfo
  depsets.
* A test in tests/core/cgo now uses cc_binary to generate shared
  libraries.

Updates #1888

### DIFF
--- a/tests/core/cgo/BUILD.bazel
+++ b/tests/core/cgo/BUILD.bazel
@@ -46,25 +46,23 @@ cc_import(
     tags = ["manual"],
 )
 
-genrule(
-    name = "darwin_gen_imported_dylib",
+cc_binary(
+    name = "libimported.dylib",
     srcs = ["imported.c"],
-    outs = ["libimported.dylib"],
-    cmd = "$(CC) -shared -install_name @rpath/libimported.dylib $< -o $@",
+    linkshared = True,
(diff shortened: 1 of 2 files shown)


jayconrod:
TODO:

- [ ] Verify this works and other things aren't broken on macOS and Windows.
- [ ] This probably requires bumping the minimum version of Bazel.

jayconrod:
We can't start using `CcInfo` and `ProtoInfo` until our minimum version is Bazel 0.22.0 (which is not released yet). Those symbols are resolved eagerly by the Starlark interpreter, so the only way not to use them on older versions is to load .bzl files through a repository rule that dynamically selects which .bzl file to load. rules_go currently loads everything from def.bzl, so this would require splitting workspace definitions and rule definitions into two files, which would be a breaking change for everyone. I think it's better to just keep release branches that work with old / new versions of Bazel.

This PR will be cherry-picked onto older branches, which need to continue supporting older versions of Bazel. Another PR will migrate to `CcInfo` and `ProtoInfo`, probably in rules_go 0.18.0 (not the coming release but the one after).

Issue #1856, opened 2018-12-14
go_proto_library: support multiple proto dependencies


jayconrod:
`go_proto_library` currently generates .pb.go files from a single `proto_library` target, referenced with the `proto` attribute. This does not work well when there are multiple `proto_library` targets for the same package (e.g., one `proto_library` per .proto file).

The `proto` attribute should be replaced with a `protos` attribute that accepts a list of labels.

A wrapper macro may be used to migrate users of the old `proto` attribute until it is deprecated and removed.

vam-google:
@jayconrod can you please assign this to me?

jayconrod:
GitHub won't let me assign to anyone outside the bazelbuild org. But let's proceed as if you're assigned. :)

PR #1890, opened 2019-01-08
go_proto_library: support multiple proto dependencies

vam-google:
Add `protos` argument to `go_proto_library` (same as the existing `proto` argument, but supports multiple protos).

`go_proto_compile` changes one of the arguments from `proto` to `protos`, which is considered Ok (even though this is a technically public API there is no detected usages of the method anywhere else on github).

Implements https://github.com/bazelbuild/rules_go/issues/1856

### DIFF
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -42,8 +42,14 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
         for src in proto.check_deps_sources.to_list():
             path = proto_path(src, proto)
             if path in proto_paths:
+                if proto_paths[path] != src:
+                    fail("proto files {} and {} have the same import path, {}".format(
+                        src.path,
+                        proto_paths[path].path,
+                        path,
+                    ))
                 continue
-            proto_paths[path] = True
+            proto_paths[path] = src
             out = go.declare_file(
                 go,
                 path = importpath + "/" + src.basename[:-len(".proto")],
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -32,10 +32,16 @@ load(
 
 GoProtoCompiler = provider()
 
-def go_proto_compile(go, compiler, proto, imports, importpath):
+def go_proto_compile(go, compiler, protos, imports, importpath):
     go_srcs = []
     outpath = None
-    for src in proto.check_deps_sources:
+    dep_sources = depset(direct = [], transitive = [proto.check_deps_sources for proto in protos])
+    transitive_descriptor_sets = depset(
+        direct = [],
+        transitive = [proto.transitive_descriptor_sets for proto in protos],
+    )
+
+    for src in dep_sources:
         out = go.declare_file(go, path = importpath + "/" + src.basename[:-len(".proto")], ext = compiler.suffix)
         go_srcs.append(out)
         if outpath == None:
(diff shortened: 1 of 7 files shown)


proto/compiler.bzl
```diff
@@ -32,10 +32,16 @@ load(
 
 GoProtoCompiler = provider()
 
-def go_proto_compile(go, compiler, proto, imports, importpath):
+def go_proto_compile(go, compiler, protos, imports, importpath):
     go_srcs = []
     outpath = None
-    for src in proto.check_deps_sources:
+    dep_sources = depset(direct = [], transitive = [proto.check_deps_sources for proto in protos])
```
jayconrod: Since we need to iterate the depset anyway, I think it would be clearer to just have a doubly nested loop, and build `proto_paths` at the same time we're declaring output files.

Constructing a `depset` has a surprising amount of overhead (just in the Java implementation, not anything fundamental), and we should try to avoid it unless either a) we're passing it along a dependency graph and we can avoid iterating it at every step or b) we can pass it to `args.add_all` and avoid iterating it unless the action is executed.

proto/compiler.bzl
```diff
@@ -32,10 +32,16 @@ load(
...
+    transitive_descriptor_sets = depset(
+        direct = [],
+        transitive = [proto.transitive_descriptor_sets for proto in protos],
+    )
+
+    for src in dep_sources:
```
jayconrod: `dep_sources.to_list()`

Bazel will eventually make it an error to iterate a `depset`. There is a flag `--incompatible_depset_is_not_iterable`, which detects this. (My previous code here is wrong; I didn't realize `proto.check_deps_sources` was a `depset`).

proto/def.bzl
```diff
@@ -83,14 +83,23 @@ def _go_proto_library_impl(ctx):
         compilers = ctx.attr.compilers
     go_srcs = []
     valid_archive = False
+
+    if ctx.attr.proto:
```
jayconrod: I think it should be an error to specify both `proto` and `protos`: one or the other should be required. This should have been the case for `compilers` as well, but let's leave that aside for now.

tests/core/go_proto_library/BUILD.bazel
```diff
@@ -104,6 +104,34 @@ proto_library(
     srcs = ["proxy_b.proto"],
 )
 
+
```
jayconrod: There are a couple cosmetic errors here. Could you run `buildifier` on this file? It should clean them up. You can install it with `go get -u github.com/bazelbuild/buildtools/buildifier`.

proto/compiler.bzl
jayconrod: We should definitely avoid duplicates, but we're not able to take advantage of `depset` performance here because we're iterating through `check_deps_sources` below, so the cost isn't worth it.

I was thinking a change like this:

```
go_srcs = []
outpath = None
proto_paths = {}
for proto in protos:
    for src in proto.check_deps_sources.to_list():
        path = proto_path(src, proto)
        if path in proto_paths:
            continue
        proto_paths[path] = True
        out = go.declare_file(go, path = importpath + "/" + src.basename[:-len(".proto")], ext = compiler.suffix)
        go_srcs.append(out)
        if outpath == None:
            outpath = out.dirname[:-len(importpath)]
```

WDYT?

The construction of `transitive_descriptor_sets` looks good though. The `depset` makes sense there.

proto/def.bzl
vam-google: Done

vam-google:
@jayconrod PTAL. Also please pay attention to the change in `get_imports()`, which was not included in initial commit.

proto/compiler.bzl
jayconrod: This is a valid concern. Let's make it an error.

Ultimately, the `proto_path` is what gets passed to `protoc`, and that's what we use to find the right descriptor to generate code for. We have the full set of descriptors for transitive dependencies, and we don't have another way to identify direct dependencies. It's okay to have the same .proto file repeated (possibly seen through more than one `proto_library`), but it should be an error to have different files with the same effective import path.

So let's do this: `proto_paths` should be a map from `proto_path` values to Bazel `File`s. The code code be changed to something like this:

```
path = proto_path(src, proto)
if path in proto_paths:
    if proto_paths[path] != src:
        fail("proto files {} and {} have the same import path, {}".format(src.path, proto_paths[path].path, path))
    continue
proto_paths[path] = src
```

proto/def.bzl
```diff
@@ -38,20 +38,27 @@ load(
...
-        ]
+        protos = [attr.proto.proto]
+    elif hasattr(attr, "protos"):
+        for proto in attr.protos:
+            if hasattr(proto, "proto"):
+                protos.append(proto.proto)
+
+    direct = dict()
+    for proto in protos:
+        for src in proto.check_deps_sources:
```
jayconrod: I think `.to_list()` is needed here (previous code was incorrect, too). Maybe run tests with `--incompatible_depset_is_not_iterable` to double-check there aren't any other issues.

jayconrod:
Added two small comments, but this is almost ready to go.

proto/def.bzl
vam-google: changed

jayconrod:
Test failure is due to a CI configuration which will be fixed in #1902, so submitting.

Issue #1896, opened 2019-01-10
Build changes needed to support Go 1.12


jayconrod:
This is a catch-all issue for 1.12 related changes.

* [ ] Deprecate `go_vet_test` and direct usage of `vet` in `nogo` because `go tool vet` no longer works. `nogo(vet = True)` should use the analysis framework instead of invoking vet.
* [x] Generate symabis files for packages with assembly. The assembler now requires this.
* [x] Ensure cgo .o files in archives don't have their extensions truncated. The linker will no longer blindly link files.
* [x] Create a cache directory and set `GOCACHE` when building the standard library.
* [ ] Update to a newer Gazelle that can build `fetch_repo` and `gazelle` correctly (needed for tests).
* [x] Fix or disable tests that rely on removed features.

robfig:
@jayconrod good to close this out?

jayconrod:
Yep this can be closed.

PR #1897, opened 2019-01-10
nogo: translate 'vet = True' into appropriate deps

### LINKED ISSUE #1800: nogo with vet does not respect excludes_files
If you register `nogo()` with vet enabled, it doesn't look like you can exclude directories -- for example `vendor/*` -- from being processed like you can with analyzer configurations.

For example:
```bazel
nogo(
    name = "nogo",
    config = "nogo.json",
    vet = True,
)
```
Would run vet on all compiled libraries, including those from vendor.

Ideally vet would respect a configuration entry for itself. Eg.
```json
{
    "vet": {
      "exclude_files": {
        "vendor/*": "third party code"
      }
    }
}
```
At least it didn't look like this was possible with a quick glance at the nogo code.

jayconrod:
nogo no longer invokes vet at all. Setting 'vet = True' is handled by
a wrapper, which adds the equivalent libraries from
@org_golang_x_tools//go/analysis/passes.

'vet = True' was originally needed before the vet analyses had been
migrated to the analysis framework. There is no need for it anymore,
and the command line interface for vet is changing in Go 1.12.

Updates #1896
Fixes #1800

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -80,7 +80,7 @@ load(
 )
 load(
     "@io_bazel_rules_go//go/private:rules/nogo.bzl",
-    _nogo = "nogo",
+    _nogo = "nogo_wrapper",
 )
 
 # Current version or next version to be tagged. Gazelle and other tools may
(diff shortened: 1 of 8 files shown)


jayconrod:
cc @stjj89

jayconrod:

PR #1898, opened 2019-01-10
go/tool/compile: generate symabis for assembly

jayconrod:
Go 1.12 requires the assembler to generate a symabis file, which is
passed to the compiler. The compile action will now do this when there
are assembly sources.

In the future, we'll combine GoCompile, GoAsm, and other actions into
a single GoCompile action per package. Because of this plan,
generating the symabis file is not a separate action.

Updates #1896


jayconrod:

PR #1899, opened 2019-01-10
go/tools/builders/pack: preserve file extensions

jayconrod:
The .a format limits file names to 16 characters. There are various
extensions to this format which support longer names, but Go doesn't
recognize them and generally doesn't need them. Before Go 1.12, the
linker passed all extra files (other than the compiled Go code and the
export data) to the external link.

In Go 1.12, the linker only passes on .o files. The GoPack action
simply truncated long file names, which meant .o files were sometimes
ignored. GoPack will now preserve file extensions.

Updates #1896


jayconrod:

PR #1900, opened 2019-01-10
go/tools/builders/stdlib: set temporary GOCACHE

jayconrod:
A GOCACHE directory is required by 'go build', starting in Go
1.12. Normally the location is inferred from HOME or other environment
variables, but there's aren't set in the Bazel sandbox. To build the
standard library, we create a temporary cache that is deleted at the
end of the GoStdLib action.

Updates #1896


jayconrod:
Spurious test failures will be fixed by #1902.

PR #1902, opened 2019-01-11
Skip part of //tests/reproducibility:go_default_test in BuildKite

jayconrod:
It appears that Bazel CI is now using the username "bazel" when
running tests. This test verifies the current username does not appear
in its binary. "bazel" should be allowed though.

### DIFF
--- a/tests/reproducibility/reproducible_binary_test.go
+++ b/tests/reproducibility/reproducible_binary_test.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/user"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/bazelbuild/rules_go/go/tools/bazel"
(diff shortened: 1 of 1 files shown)


serathius:
Hey @jayconrod,
Since upgrading to rules_go 0.18.6, tests started to be cached.
In golang it's easy to disable with "GOCACHE=off go test", but I was unable to do that for "go_test" rule
I tried "bazel test --test_env=GOCACHE=off", but it doesn't work (change in environment will hit other cache bucket, but second run will still be cached).

Do you think this could be related to this PR?

jayconrod:
@serathius Do you mean tests in general are cached or `//tests/reproducibility:go_default_test` specifically is cached when it shouldn't be? This PR only affects the latter, but I think you mean the former.

Like the Go command, Bazel caches positive test results. That's true for all tests, not just `go_test`. It only re-runs tests when inputs change.

You can temporarily disable caching for a command with `--nocache_test_results`.

You can permanently disable caching for a specific test by adding `tags = ["no-cache"]` to the test rule.

serathius:
Thanks, I will try that.

Issue #1911, opened 2019-01-18
security releases 1.11.5 and 1.10.8 on Weds Jan 23rd


jmhodges:
The Go team has announced a security release coming out on Weds January 23rd around 8 PM UTC/12 PM PT/3 PM ET. See https://groups.google.com/forum/#!topic/golang-announce/yzoslEVBvIs

New versions will be 1.11.5 and 1.10.8

jayconrod:
I'll put out a new rules_go point release with support for these soon after they're published.

jmhodges:
Done long ago!

PR #1919, opened 2019-01-23
Support Go 1.11.5 and 1.10.8

jayconrod:


jayconrod:
Holding on golang/go#29906

Related #1911

PR #1923, opened 2019-01-26
change import_path_option default value to false

### LINKED ISSUE #1596: go_proto_library generates different package name from protoc
(Originally bazelbuild/bazel-gazelle#271)

In [`github.com/googleapis/gnostic/extensions`](https://github.com/googleapis/gnostic/tree/master/extensions), when we build with `go_proto_library`, we generate a .pb.go file with the package name `extensions`. When we manually generate .pb.go files with protoc, we generate a .pb.go file with the package name `openapiextension_v1` (which matches the other .go file in the directory).

There is no `option go_package` in this proto. Gazelle sets the `importpath` to `github.com/googleapis/gnostic/extensions`, which is correct, but I think we're passing that path to `protoc-gen-go` in a way that sets the package name to the last element, `extensions`.

blico:
Fixes #1596

`import_path_option` is a `go_proto_compiler` attribute with the following definition:
When true, the `importpath` attribute from `go_proto_library` rules using this compiler will be passed to the compiler on the command line as `--option import_path={}`.

When a proto file does not contain `option go_package`, go_protobuf and gogoprotobuf use the `import_path` parameter to determine the pb.go file's package name. When the `import_path` parameter is not passed in, the proto package name is used to determine the pb.go file's package name.

With the lack of `option go_package`, it appears to be the [expected behaviour](https://developers.google.com/protocol-buffers/docs/reference/go-generated) to use the proto package name for determining the generated go file's package name. Additionally, the `import_path` parameter is [deprecated](https://github.com/golang/protobuf/blob/master/README.md#parameters), and other go proto compilers do not have an `import_path` option. This is leading to package name mismatch compiler errors when a `go_proto_library` rule contains multiple compilers and some are not gogoprotobuf/go_protobuf compilers.

Therefore, I am proposing that the default behavior of the `go_proto_compiler` rule should be to not pass in the `import_path` option. Another solution could be to add a new set of rules for gogoprotobuf and go_protobuf that have the `import_path_option` set to false.

### DIFF
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -159,7 +159,7 @@ go_proto_compiler = go_rule(
         "options": attr.string_list(),
         "suffix": attr.string(default = ".pb.go"),
         "valid_archive": attr.bool(default = True),
-        "import_path_option": attr.bool(default = True),
+        "import_path_option": attr.bool(default = False),
         "plugin": attr.label(
             allow_single_file = True,
             executable = True,
--- a/tests/core/go_proto_library/BUILD.bazel
+++ b/tests/core/go_proto_library/BUILD.bazel
@@ -144,8 +144,8 @@ go_test(
 
 go_proto_library(
     name = "gofast_proto",
-    importpath = "github.com/bazelbuild/rules_go/tests/core/go_proto_library/foo",
     compilers = ["@io_bazel_rules_go//proto:gofast_proto"],
+    importpath = "github.com/bazelbuild/rules_go/tests/core/go_proto_library/foo",
     protos = [":foo_proto"],
 )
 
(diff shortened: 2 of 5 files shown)


jayconrod:
Thanks for sending this PR. I think this is the right thing to do.

Would you mind adding a test in `tests/core/go_proto_library` to verify the new behavior? A `go_proto_library` with a `go_test` that imports it using a package name based on the proto package name would be good enough. Make sure to add an entry in the README.rst file there.

cc @vam-google Do you think this would cause any issues in googleapis? I think most (all?) of the protos there set `option go_package`, so I'm guessing not.

vam-google:
@jayconrod I think googleapis should be fine. I think the procedure of pushing anything to googleapis ensures that `go_package` parameter is added (the tool, which makes the change list adds the parameter to protos if it does not exist).

jayconrod:
@blico Actually, nevermind, I'll go ahead and add that test now. I want to cut a release branch today, and I'd like to include this PR.

Issue #1943, opened 2019-02-08
go tools now resolve the real path of the input file


steeve:
It seems `go install` and `go tool compile` now resolve the real path of the input file, causing sandbox and real paths to leak into the binary.

See this `dwarfdump` sample:
```
                 AT_decl_file( "/Users/steeve/go/src/github.com/steeve/test/main.go" )
                 AT_decl_file( "/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/execroot/__main__/bazel-out/x86-dbg/bin/external/io_bazel_rules_go/android_386_debug_c-shared/stdlib%/src/runtime/cgo/gcc_android.c" )
                 AT_decl_file( "/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/execroot/__main__/bazel-out/x86-dbg/bin/external/io_bazel_rules_go/android_386_debug_c-shared/stdlib%/src/runtime/cgo/libcgo.h" )
```

Notice the first file is not in the sandbox, while the other is in the external bazel folder, not in the sandbox, which was:
```
/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/sandbox/darwin-sandbox/3/execroot/__main__
```

I'm not sure how to fix that properly....

steeve:
Looking at whole `dwarfdump`, it seems only `.s` and `.c` files are impacted for `stdlib`.

Which is weird since we pass trimpath to `go tool asm`

steeve:
After doing some bash-fu, I could narrow down the following packages for stdlib that seem impacted:
```
internal/bytealg
internal/cpu
reflect
runtime
runtime/cgo
runtime/internal/atomic
runtime/internal/sys
sync/atomic
syscall
```

steeve:
Ok so the reason is that the trimpath is:
```
-asmflags=all=-trimpath=/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/sandbox/darwin-sandbox/3/execroot/__main__
```

Which is not the path that ends up in the dwarf.

steeve:
Alright so, for `stdlib` `asmflags`, it seems we need to strip the output path instead of the sandbox path.
Now only `.c` and `.h` files remain.

steeve:
I managed to fix `runtime/cgo` by adding the output path to the `-fdebug-prefix-map` in `CGO_CFLAGS`.

[...]

PR #1945, opened 2019-02-08
Ensure bazel paths don't leak in stdlib builds

steeve:
It seems `go tool asm` and `runtime/cgo` uses the output path, so add them.

Fixes the `stdlib` part of #1943.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -153,6 +153,8 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, sSrcs, hSrcs []string, pa
 
 	// Compile C, C++, and assembly code.
 	defaultCFlags := defaultCFlags()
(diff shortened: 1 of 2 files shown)


steeve:
Also, I've changed the prefix map to output `./` as the base, as it is custom (see `--features=debug_prefix_map_pwd_is_dot` on the xcode crosstool, for instance).

It's easier to then add a source-map to gdb or lldb from `.=<PATH TO MY SOURCES>`

steeve:
Continuing discussion on #1943, actually there is more to it, as `runtime/cgo/gcc_*.S` leaks in the `.a` file.

steeve:
Please don't merge this PR just now, as per findings in #1943.
Also, I have a cleaner way of doing this easily.

steeve:
I've just rebased this PR.
Some things are still leaking but we're much better.

steeve:
For stdlib, the only remaining string leaking is related to https://bugs.llvm.org/show_bug.cgi?id=38050.

PR #1949, opened 2019-02-12
go_download_sdk: work around bazel bug extracting .tar.gz files

jayconrod:
This PR tests rules_go with go1.12rc1. It's main purpose is to run CI. It will eventually be submitted with test changes needed to support Go 1.12. Support for go1.12rc1 will not actually be merged.


jayconrod:
TODO: in `_remote_sdk`, support unzipping files and verify it works on Windows. No other platform has .zip downloads.

PR #1954, opened 2019-02-19
Make googleapis work on FreeBSD.

EdSchouten:
The patches that are shipped with googleapis have the intent of removing
files from the input. Though GNU patch does this properly, the copy of
patch shipped with FreeBSD does not do this explicitly. It needs the -E
flag to do so.

This makes it possible to build https://github.com/buildbarn on FreeBSD
11.2-STABLE.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -169,7 +169,7 @@ def go_rules_dependencies():
             "@io_bazel_rules_go//third_party:go_googleapis-gazelle.patch",
             "@io_bazel_rules_go//third_party:go_googleapis-fix.patch",
         ],
-        patch_args = ["-p1"],
+        patch_args = ["-E", "-p1"],
     )
 
     # Needed for examples


alexeagle:
Wow, after some archeology, I found that this PR causes Bazel to be unable to use its native `patch` function
https://cs.opensource.google/bazel/bazel/+/refs/tags/6.2.0:tools/build_defs/repo/utils.bzl;l=69
and so now all builds involving rules_go require a `patch` command to be on the users `$PATH` where it shouldn't.

I think we should guard this `-E` flag to be added to `patch-args` only on a BSD operating system, so that the bug @EdSchouten ran into is fixed in a way that doesn't cause a new system dependency for everyone else.

fmeum:
Great catch, @alexeagle!

Since the decision whether to use `-E` needs to be made in a `WORKSPACE` macro, I don't think that there is a way to get the "BSD or not" bit into it automatically.

However, since Bazel now has a native implementation, BSD should be supported just as well without the flag - we should just remove it.

fmeum:
As this is only used for googleapis, it will go away with https://github.com/bazelbuild/rules_go/pull/3603 anyway in the next release.

alexeagle:
Ah, then I'll just wait on #3603 and locally patch out this `-E` argument, thanks!

PR #1960, opened 2019-02-26
Add shallow_since to git_repository dependencies

keith:
This stops bazel from printing about this dependency:

```
DEBUG: Rule 'com_github_golang_protobuf' modified arguments {"shallow_since": "1534281267 -0700"}
```


jayconrod:
I think it's reasonable to add this for the `git_repository` dependencies. Please it add appropriate arguments for *all* `git_repository` dependencies though, not just that one.

keith:
I've updated this PR to add this to all `git_repository` rules

keith:
Thanks!


--
Keith Smiley

PR #1961, opened 2019-02-26
Add sha256 for org_golang_x_tools

keith:
This stops bazel from warning about this:

```
DEBUG: Rule 'org_golang_x_tools' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "11629171a39a1cb4d426760005be6f7cb9b4182e4cb2756b7f1c5c2b6ae869fe"
```

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -41,6 +41,7 @@ def go_rules_dependencies():
         # master, as of 2019-01-15
         urls = ["https://codeload.github.com/golang/tools/zip/bf090417da8b6150dcfe96795325f5aa78fff718"],
         strip_prefix = "tools-bf090417da8b6150dcfe96795325f5aa78fff718",
+        sha256 = "11629171a39a1cb4d426760005be6f7cb9b4182e4cb2756b7f1c5c2b6ae869fe",
         type = "zip",
         patches = [
             "@io_bazel_rules_go//third_party:org_golang_x_tools-gazelle.patch",


jayconrod:
Sorry, I can't accept this one. `codeload.github.com` URLs don't have stable SHA-256 sums. If GitHub changes anything about the archive format (e.g., the order of files, alignment, compression), the SHA-256 sums will change, and it will break everyone. This has happened multiple times in the past.

`git_repository` is considerably slower than `http_archive`. It's not clear to me if it provides better security guarantees (less need to trust GitHub).

At some point, `go mod download` may provide better speed and trust for Go repositories. At the moment, it's not strictly better than `git_repository` though, so we won't switch yet.

keith:
> `git_repository` is considerably slower than `http_archive`. It's not clear to me if it provides better security guarantees (less need to trust GitHub).

This makes sense. The core reason I wanted this is just to reduce the logging that happens every time you hit this codepath if you don't lock this somehow. I can change this to a `git_repository` if you'd like. The problem is because of the custom patches doing this in our own WORKSPACE wouldn't be feasible

jayconrod:
FYI, you can override repositories declared in `go_rules_dependencies` with your own versions. You can apply the patches in `third_party`, although you won't need the Gazelle patches if you're using `go_repository`. See [Overriding dependencies](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#overriding-dependencies).

keith:
Ah knowing I won't need the patches is much nicer, thanks!

PR #1967, opened 2019-02-27
Upgrade github.com/golang/protobuf from v1.2.0 to v1.3.0

peter-edge:
This is necessary because https://github.com/golang/protobuf/releases/tag/v1.3.0 resulted in a new `ProtoPackageIsVersion3` https://github.com/golang/protobuf/blob/v1.3.0/proto/lib.go#L946 meaning that code generated with v1.3.0 will not build against a dependency of v1.2.0. There's a bit of a chicken-and-egg problem that `go_rules_dependencies` depends on v1.2.0, but some use `go_repository` from Gazelle to depend on `com_github_golang_protobuf`, but `go_rules_dependencies` needs to be loaded before `go_repository` can be loaded from Gazelle.

I've verified this is a drop-in, I am using this in https://github.com/uber/prototool/blob/3c3a6dbdc3f490a385928221a8b6606f14ac1aa5/WORKSPACE#L15

It would probably be worthwhile to upgrade gogo/protobuf to v1.2.1 as well, but this isn't required.

`bazel test //...` passes locally using Bazel v0.22.0.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -55,7 +55,7 @@ def go_rules_dependencies():
         git_repository,
         name = "com_github_golang_protobuf",
         remote = "https://github.com/golang/protobuf",
-        commit = "aa810b61a9c79d51363740d207bb46cf8e620ed5",  # v1.2.0, as of 2018-09-28
+        commit = "c823c79ea1570fb5ff454033735a8e68575d1d0f",  # v1.3.0, as of 2019-02-26
         patches = [
             "@io_bazel_rules_go//third_party:com_github_golang_protobuf-gazelle.patch",
             "@io_bazel_rules_go//third_party:com_github_golang_protobuf-extras.patch",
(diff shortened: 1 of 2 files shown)


jayconrod:
Sorry, I'm not going to update any dependencies in `go_rules_dependencies` until we're closer to a new major release, which will be in a month or so, at which point, I'll do them all at once. Updating protobuf in particular means verifying the third_party patch that provides extra rules. I'd prefer not to repeat that work later on.

You can override the protobuf version in your own workspace though. There are instructions at [Overriding dependencies](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#overriding-dependencies). It should be safe to call `gazelle_dependencies` and `go_repository` before `go_rules_dependencies`. You can apply the patch in `@io_bazel_rules_go//third_party` (but you won't need the Gazelle patch with `go_repository`).

peter-edge:
Ok - FYI you’re going to have a lot of people broken on this, was trying to do a favor here to get ahead of it before others come along, but your call. This is likely going to break a lot of people if not fixed soon.

PR #1973, opened 2019-03-01
Add env attribute to go_test rule

blico:
This change adds the `env` attribute to the `go_test` rule.

`env` is a map of environment variables that will be set before the test is run.

### DIFF
--- a/go/core.rst
+++ b/go/core.rst
@@ -3,6 +3,7 @@ Core go rules
 
 .. _test_filter: https://docs.bazel.build/versions/master/user-manual.html#flag--test_filter
 .. _test_arg: https://docs.bazel.build/versions/master/user-manual.html#flag--test_arg
+.. _test_env: https://docs.bazel.build/versions/master/user-manual.html#flag--test_env
 .. _Gazelle: https://github.com/bazelbuild/bazel-gazelle
 .. _GoLibrary: providers.rst#GoLibrary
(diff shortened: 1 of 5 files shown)


jayconrod:
Why set environment variables in Bazel rules and not in the tests themselves?

If there's a reason environment variables must be set in rules, does it make sense for it to be a common attribute for all test rules in Bazel itself?

rules_go seems like the wrong place for this.

Globegitter:
We have also run into cases where it is useful to set env vars before a target runs (not so much in go but especially js). I opened https://github.com/bazelbuild/bazel/issues/7364 for that, @blico might be useful +1 that issue as well :)

Globegitter:
Btw currently as a workaround we are using the `command` rule from here: https://github.com/atlassian/bazel-tools/blob/master/multirun/README.md

jayconrod:
This PR won't actually set environment variables before the target runs. Instead, they'll be set in `main`. `init` functions would not see the new environment variables. So I'd encourage people to write their own `TestMain` functions to set variable instead so it's clear what's happening. This would let the same tests work with `go test`, too.

bazelbuild/bazel#7364 would be able to set variables before the test executes. I think that's a better solution.

blico:
Thank you for the feedback.

I will close this PR and look into [bazelbuild/bazel#7364](https://github.com/bazelbuild/bazel/issues/7364)

PR #1985, opened 2019-03-08
Pass ProtoInfo instead of Target to proto_path

robbertvanginkel:
During a refactor in 8670623e6e3a18123129173ba667c33631aca006, a `.proto` attribute was missed and now a `Target` gets passed into the `proto_path` function rather than a `ProtoInfo` provider.

When used with `{strip_}import_prefix`, this would result in the `import` flags passed to the `protoc-go` builder to contain incorrect paths:

```
tests/core/go_proto_library/_virtual_imports/adjusted_c_proto/adjusted/adjusted_c.proto=adjusted/c
```
instead of
```
adjusted/adjusted_c.proto=adjusted/c
```

When `go_package` isn't present in a proto file, protoc-gen-go relies in `Mfile.proto=go/importpath` to resolve proto files to import paths. With these passed incorrect packages for dependencies fail to compile.

I made a few modifications to the tests under `tests/core/go_proto_library/` to catch the issue, but have left the test targets uncommented due to the minimum bazel requirement.

### DIFF
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -57,7 +57,7 @@ def get_imports(attr):
     direct = dict()
     for dep in proto_deps:
         for src in proto_check_deps_sources(dep).to_list():
-            direct["{}={}".format(proto_path(src, dep), attr.importpath)] = True
+            direct["{}={}".format(proto_path(src, get_proto(dep)), attr.importpath)] = True
 
     deps = getattr(attr, "deps", []) + getattr(attr, "embed", [])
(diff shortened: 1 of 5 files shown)


robbertvanginkel:
I'm tracking down how to get on the list, will let you know when it should be fixed.

robbertvanginkel:
@googlebot CLA should be signed.

jayconrod:
Great, thanks!

Issue #1959, opened 2019-02-25
Migrate bazel_test rules to go_bazel_test


jayconrod:
`bazel_go_integration_test` is defined in https://github.com/bazelbuild/bazel-integration-testing, among other languages.

The framework there creates a scratch workspace (with a WORKSPACE that has `local_repository` rules already set up), then allows tests to run Bazel commands and check the results.

This is basically what `bazel_test` does. `bazel_test` is pretty ugly and difficult to modify, and I've bee thinking about a rewrite for a while. It would be better to leverage shared test infrastructure with other languages though.

jayconrod:
Instead of migrating to `bazel_go_integration_test`, I plan to migrate these tests to `go_bazel_test`.

gonzojive:
@jayconrod, may you please say more about the decision not to use [bazel-integration-testing](https://github.com/bazelbuild/bazel-integration-testing)?

One feature of `bazel-integration-testing` that I like is the approach to working with external repositories. Repository archives can be cached so that no network access is needed within the test of the scratch workspace. This is discussed in [import.bzl](https://github.com/bazelbuild/bazel-integration-testing/blob/165440b2dbda885f8d1ccb8d0f417e6cf8c54f17/tools/import.bzl) and [RepositoryCache.java](https://github.com/bazelbuild/bazel-integration-testing/blob/165440b2dbda885f8d1ccb8d0f417e6cf8c54f17/java/build/bazel/tests/integration/RepositoryCache.java).

I also noticed that go_bazel_test adds the "local" tag to rules, which [prevents running the test in a sandbox](https://bazel.build/reference/be/common-definitions). Presumably that means these tests don't get run as part of tools like buildkite, but I'm not sure about that.

PR #1989, opened 2019-03-12
Update travis.yml to use caches across CI jobs

rahul-malik:
You can specify directories for caching that will be preserved across multiple CI jobs which will reduce time needed to fetch dependencies while executing CI build steps.
More info here: https://docs.travis-ci.com/user/caching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,3 +42,8 @@ script:
 
 notifications:
   email: false
+  
+cache:
+  directories:
+    - $HOME/.bazel_repository_cache
+    - $HOME/.cache


rahul-malik:
I'm not sure if it's because I used the GitHub web UI to make this PR but I have a corporate CLA through Pinterest and have contributed to other Bazel / Google repositories before.

jayconrod:
Thanks for working on this!

I took a look at Travis CI caching about a year ago, and at the time, retrieving the cache took significantly longer than just doing a clean build. Do you know if that's improved? I think it was mostly problematic for Mac builds, since their Mac cluster is not in the same data center as the cache.

About the CLA, I've verified Pinterest has signed the corp CLA. Could you verify that the email used to create the commit is a member of the authorized contributor Google group and is attached to your GitHub?

rahul-malik:
Hey sorry for not getting back earlier. I don’t know if performance has been improved but feel free to test out the patch if you decide to go this route in the future

jayconrod:
Thanks for working on this. Can't proceed with this patch without CLA approval though.

Our primary pain point with Travis is long macOS queueing times. Downloads used to be a bigger issue, but they haven't been bad lately.

The reason we still use Travis CI is so we can test old versions of Bazel (BuildKite doesn't support that automatically). However, https://github.com/bazelbuild/bazel-integration-testing should let us do this within tests, so we should migrate to that (#1959) and drop Travis.

rahul-malik:
Given the longer term approach is going to be on BuildKite it probably makes sense to abandon this anyways. Thanks for the review!

PR #1991, opened 2019-03-13
Fix gRPC cross-compilation for Android

steeve:
This was an internal patch that I should have upstreamed a long time ago.

### DIFF
--- a/third_party/org_golang_google_grpc-crosscompile.patch
+++ b/third_party/org_golang_google_grpc-crosscompile.patch
@@ -27,3 +27,14 @@ diff -urN b/internal/channelz/BUILD.bazel c/internal/channelz/BUILD.bazel
          "@io_bazel_rules_go//go/platform:linux": [
              "@org_golang_x_sys//unix:go_default_library",
          ],
+diff -urN a/internal/syscall/BUILD.bazel b/internal/syscall/BUILD.bazel
+--- a/internal/syscall/BUILD.bazel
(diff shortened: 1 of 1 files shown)


third_party/org_golang_google_grpc-crosscompile.patch
```diff
@@ -27,3 +27,55 @@ diff -urN b/internal/channelz/BUILD.bazel c/internal/channelz/BUILD.bazel
          "@io_bazel_rules_go//go/platform:linux": [
              "@org_golang_x_sys//unix:go_default_library",
          ],
+diff --git a/internal/syscall/BUILD.bazel b/internal/syscall/BUILD.bazel
```
jayconrod: I think I understand why this patch is needed for Android, but not for other platforms. This packages has two files, `syscall_linux.go` and `syscall_nonlinux.go`. Android should build with the first file, which imports `@org_golang_x_sys//unix:go_default_library`. `syscall_nonlinux.go` doesn't import that library though, so it shouldn't be necessary.

steeve: i think it should be needed on linux as well right ?

steeve: an it's there already. yes `android` matches `linux` as well (kind of a weird situation)

steeve:
Updated the PR :)

PR #1997, opened 2019-03-19
Fix missing package name

codesuki:
Does not stamp if package name is missing.

I tried it locally with 0.18.1.


jayconrod:
The old example works for me, and that style is used in several tests. We automatically add the package path to the -X flag that gets passed to the linker if it's not specified in the `x_defs` attribute.

Do you want to open an issue instead? We might be able to figure out what's going wrong with more information.

PR #2007, opened 2019-03-26
Fix gRPC cross-compilation for Linux from Darwin

matt-e:
In the same vein as #1991 this appears to fix gRPC cross-compilation for Linux from Darwin.

I suspect the other Unix platforms need the same change, but don't have a way to test them.

### DIFF
--- a/third_party/org_golang_google_grpc-gazelle.patch
+++ b/third_party/org_golang_google_grpc-gazelle.patch
@@ -2264,6 +2264,7 @@ diff -urN a/internal/syscall/BUILD.bazel b/internal/syscall/BUILD.bazel
 +        ],
 +        "@io_bazel_rules_go//go/platform:darwin": [
 +            "//grpclog:go_default_library",
++            "@org_golang_x_sys//unix:go_default_library",
 +        ],
 +        "@io_bazel_rules_go//go/platform:dragonfly": [
 +            "//grpclog:go_default_library",
--- a/third_party/org_golang_google_grpc-gazelle.patch
+++ b/third_party/org_golang_google_grpc-gazelle.patch
@@ -2247,7 +2247,7 @@ diff -urN a/internal/leakcheck/BUILD.bazel b/internal/leakcheck/BUILD.bazel
 diff -urN a/internal/syscall/BUILD.bazel b/internal/syscall/BUILD.bazel
 --- a/internal/syscall/BUILD.bazel	1969-12-31 19:00:00.000000000 -0500
 +++ b/internal/syscall/BUILD.bazel	2000-01-01 00:00:00.000000000 -0000
-@@ -0,0 +1,48 @@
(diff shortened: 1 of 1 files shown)


matt-e:
I signed it!

jayconrod:
I'd like to understand the issue you're facing before proceeding this. What's the error you're seeing? Is there an open issue? I suspect you might be cross-compiling by setting `goos = "linux"` on a `go_binary`? That's a known issue, and we might be able to use something like this as a workaround until the underlying issue is resolved. But I'd strongly recommend using `--platforms` for cross-compilation instead.

This patch doesn't seem right to me. `golang.org/x/sys/unix` is only imported in the Linux version `google.golang.org/grpc/internal/syscall`. It shouldn't be needed for `darwin`. #1991 was needed because Gazelle doesn't know that `GOOS` being set to `android` implies `linux`.

matt-e:
> I suspect you might be cross-compiling by setting `goos = "linux"` on a `go_binary`? That's a known issue, and we might be able to use something like this as a workaround until the underlying issue is resolved. But I'd strongly recommend using --platforms for cross-compilation instead.

Close enough - we're setting `goos="linux"` on a `go_image` from `rules_docker`. I'm aware this is a workaround. That said, it's a massive quality of life improvement. We have a multi-platform repo (dev happens on darwin and linux, and deploy to darwin and linux), and without this change, `bazel build //...` fails for obvious reasons. So we end up with a makefile combining increasingly complex search queries and `--platforms` flags, and have basically had to drop use of the bazel cli.

jayconrod:
Fair enough. I wish this weren't necessary, but it's going to be a while before the root cause is fixed, and as you said, it's a massive quality of life improvement until then.

Could you make this following changes?

* Move the diff to a new patch. The Gazelle patch is generated by Gazelle, and local changes won't survive the next update.
* Update not just `darwin` but all operating systems on which `golang.org/x/tools/unix` is safe to build. I think the full list is `aix darwin dragonfly freebsd linux netbsd openbsd solaris`
* Add the patch to `org_golang_google_grpc` in `go_repositories`.

matt-e:
Thanks Jay - I also wish this wasn't necessary. 

Just to clarify:
> Move the diff to a new patch. The Gazelle patch is generated by Gazelle, and local changes won't survive the next update.

I see there is already a crosscompile patch (https://github.com/bazelbuild/rules_go/blob/master/third_party/org_golang_google_grpc-crosscompile.patch) - should I rather just update that, or add another?

jayconrod:
Updating the crosscompile patch should also be fine.

matt-e:
I updated the patch in #2009, and will close this one out. 

I noticed that `aix` doesn't seem to be a supported platform? At least from a cursory look at https://github.com/bazelbuild/rules_go/blob/master/go/platform/list.bzl

PR #2019, opened 2019-04-04
Add shallow_since to com_google_protobuf

Globegitter:
As provided by bazel: DEBUG: Rule 'com_google_protobuf' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1551387314 -0800"


fejta:
Any chance we can release an 0.18.4 with this tag so we stop seeing these messages when we build?

jayconrod:
I'll include this in the next point release. That will likely be in early May with the next Go release.

PR #2022, opened 2019-04-05
Support Go 1.12.2 and 1.11.7

jayconrod:


jayconrod:
Link errors are due to golang/go#13114.

It appears that the C compiler used when preparing the binary Go distribution emitted relocs that the linker in ubuntu14.04 doesn't understand.

PR #2023, opened 2019-04-06
Document How To Run Benchmark Test

### LINKED ISSUE #75: Run benchmarks
Does a `go_test` support running benchmarks? I tried running the resulting binary with `-test.bench=.`, but it doesn't seem to do anything, and I don't see any reference to benchmark handling in `geneate_test_main.go`

dayfine:
This fixes #75

### DIFF
--- a/go/core.rst
+++ b/go/core.rst
@@ -626,6 +626,12 @@ equivalent of ``go test ./...``), run
 ::
 
   bazel test --test_output=errors //...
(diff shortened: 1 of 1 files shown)


ns-ggeorgiev:
Hm, I think this still requires a rule. Only then such tests can be cached and distributed etc what the bazel is known for.

PR #2024, opened 2019-04-06
Add -linkmode=pie mode

### LINKED ISSUE #2020: Add support for linkmode = "pie"

mauriciogg:
This commit resolves #2020


go/private/mode.bzl
```diff
@@ -210,6 +224,9 @@ def link_mode_args(mode):
     elif mode.link == LINKMODE_PLUGIN:
         if platform in _LINK_PLUGIN_PLATFORMS:
             args.append("-dynlink")
+    elif mode.link == LINKMODE_PIE:
+        if platform in _LINK_PIE_PLATFORMS:
+          args.append("-shared")
```
jayconrod: I don't think this is enough. See src/cmd/go/internal/work/init.go for how `-buildmode="pie"` is interpreted. It passes `-shared` to the compiler. I think it also passes `-buildmode=pie` to the linker. And `-pie` is passed to `cgo -dynimport`.

Also, this indentation doesn't look correct. Make sure to run buildifier to format this file.

mauriciogg: Done.
Code now passes pie to cgo action. -buildmode=pie is already being passed in link.bzl whenever linkmode != LINKMODE_NORMAL.

mauriciogg: Added test. Basing check for PIE on https://android.googlesource.com/platform/bionic/+/master/linker/linker_main.cpp#388. PIE executables are implemented as shared libraries (type ET_DYN). Other than that, I think any kind of relocation in the text section is a no-no.

go/private/mode.bzl
```diff
@@ -210,6 +224,9 @@ def link_mode_args(mode):
     elif mode.link == LINKMODE_PLUGIN:
         if platform in _LINK_PLUGIN_PLATFORMS:
             args.append("-dynlink")
+    elif mode.link == LINKMODE_PIE:
+        if platform in _LINK_PIE_PLATFORMS:
+            args.append("-shared")
```
jayconrod: Based on the output of `go build -x -work -buildmode=pie` with a cgo binary, we need to:

* Build the standard library with `-shared` (done, I think).
* Pass `-shared` to the compiler. (done).
* Pass `-pie` to C compiler when building cgo binary (done).
* Pass `-shared` to assembler (done).
* Pass `-buildmode=pie` to linker (done).

So this should be good.

go/core.rst
```diff
@@ -597,6 +597,8 @@ Attributes
 |                                                                                                  |
 | :value:`normal`                                                                                  |
 |     Builds a normal executable with position-dependent code.                                     |
+| :value:`pie`                                                                               |
+|     Builds a position-independent executable.                                       |
```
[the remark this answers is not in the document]
mauriciogg: Missed that one. Should be aligned now.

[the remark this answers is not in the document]
mauriciogg: Tests is here [https://github.com/bazelbuild/rules_go/pull/2024/commits/19095550d34bfcd04f40b3860f799a46598e4ce0]. Its failing for mac right now, although test should be similar. There is a debug/macho-o package as well. How do you handle cross-platform tests? Do you check different targets or should I just add the mac src to the test with a compiler directive.

tests/core/go_binary/pie_test_darwin.go
```diff
@@ -0,0 +1,48 @@
+// +build darwin
```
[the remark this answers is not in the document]
mauriciogg: Done

tests/core/go_binary/BUILD.bazel
```diff
@@ -67,3 +67,30 @@ go_library(
         "DepSelf": "DepSelf",
     },
 )
+
+go_binary(
+    name = "hello_pie_bin",
+    srcs = ["hello.go"],
+    cgo = "on",
```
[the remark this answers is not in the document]
mauriciogg: Done

tests/core/go_binary/BUILD.bazel
```diff
@@ -67,3 +67,30 @@ go_library(
...
+    cgo = True,
+)
+
+go_test(
+    name = "pie_test",
+    srcs = [
+        "pie_darwin_test.go",
+        "pie_linux_test.go",
+    ],
+    data = [
```
jayconrod: It looks like there is an error in Windows CI because the linker complains when `-buildmode=pie` is passed to it. This is not a great error message, but it's consistent with what we do for other modes, so no need to change it in this PR.

I'd suggest adding `tags = ["manual"]` to both `go_binary` rules above, then changing this `data` attribute to not pull anything in on Windows.

Something like this should work:

```
data = select({
        ":hello_nopie_bin",
        ":hello_pie_bin",
    ],
        ":hello_nopie_bin",
        ":hello_pie_bin",
    ],
    "//conditions:default": [],
}),
```

mauriciogg: Yep, That works. Thanks for the suggestion.

mauriciogg:
Great. Thanks for the review!

PR #2027, opened 2019-04-07
GoCompilePkg: unified action for building a Go package

### LINKED ISSUE #1957: Consolidate actions needed to build a package
Currently, we define separate actions for:

* Compiling Go code
* Running cgo (actually a few different actions)
* Compiling C/C++ code (via `cc_library`, `cc_binary`).
* Assembling Go assembly
* Packing .o files produced by the assembler into .a files
* Coverage

We should consolidate all this functionality into a single action. This should reduce set up / tear down overhead in both local configurations (sandboxing) and remote configurations (file transfer).

Perhaps more importantly, cgo will be much easier to maintain if we move it out of loading / analysis phases into execution. Previously, we needed cgo logic to be in loading / analysis because we needed to compile generated C code with `cc_library` / `cc_binary` rules. Bazel now gives us enough information to run the C compiler on our own.

jayconrod:
GoCompilePkg is a new action implemented in the builder binary that builds a compile Go package. It supports:

* Go compilation
* Build constraints
* cgo code generation, imports, and C, C++, assembly
* Go assembly
* Nogo
* Coverage

This is all done in the execution layer in a single action. We no longer need to create a bunch of extra rules as we did with cgo. We also don't need to create extra actions for things like assembly and coverage. In general, this approach should be friendlier to remote execution.

GoCompilePkg does not support Objective C. The legacy cgo pathway is still present for this reason, but it will only be used if `objc = True` is set.

Fixes #1957


jayconrod:
TODO

- [ ] Confirm this works on supported platforms.
- [ ] Measure performance difference, if any, especially with remote execution.
- [ ] Investigate whether Objective C can be built with the C/C++ toolchain.

steeve:
Congratulations @jayconrod, this is amazing work!

go/tools/builders/cgo2.go
```diff
@@ -0,0 +1,353 @@
...
+		return "", nil, nil, err
+	}
+
+	// If we only have C/C++ sources without cgo, just compile and pack them
+	// without generating code. The Go command forbids this, but we've
+	// historically allowed it.
+	// TODO(jayconrod): this doesn't write CGO_LDFLAGS into the archive. We
+	// might miss dependencies like -lstdc++ if they aren't referenced in
+	// some other way.
+	if len(cgoSrcs) == 0 {
```
steeve: why not run them through `cc_library` ?

steeve:
It seems you're not leveraging `cc_common` ?

go/tools/builders/cgo2.go
jayconrod: I'm not sure I follow?

With this change, we won't need `cc_library` at all anymore. `go_library` will generate one action. It won't need to be a macro that generates a constellation of `filegroup`, `cc_library`, `cgo_codegen` and other rules anymore.

I think right thing to do is synthesize a .go file with `//cgo:linkflag` (or whatever the special pragma is). The Go compiler and linker will know what to do with those.

jayconrod:
`cc_common` is already used. We pull the compiler path and flags out of that in `context.bzl`. The information is stored in `go.cgo_tools`.

go/tools/builders/cgo2.go
steeve: sorry, I meant `cc_common.create_library_to_link`. It seems you are sending the cflags and then invoking `CC` manually ?

jayconrod: I think we need to do it this way. This case is needed when there are .c files but no .go files with `import "C"`. There may have been .go files with `import "C"` that were filtered out, and there must be other pure .go files (we will synthesize one if none pass the filter). So the output is still a Go package, not a C library, and we need to read the files in order to decide how to build it.

jayconrod:
This is finally, finally passing CI.

Notes from TODOs listed above:

* Confirmed this works on Darwin, Linux, and Windows.
* No significant performance difference. This should make future performance work easier though.
* While it may be possible to compile Objective C, we can't support the full range of `objc_library` attributes we currently do.
* `linkmode = "pie"` is supported.

achew22:
Woohoo! :clap: great work.

Do you know how this impacts Kythe?

jayconrod:
@achew22 I don't know enough about Kythe to say, but it can only make things better. This should eliminate the cloud of `cc_library`, `filegroup`, and random internal rules around each `go_library` with `cgo = True`. The output of `bazel query` should be clearer.

PR #2033, opened 2019-04-17
Make nogo work on Windows (path handling)

Xjs:
Recently, we tried out using nogo in our Go projects, and it failed on Windows because of too long relative path names. (I'm not surprised about this thing anymore ;))

I don't know if you think this one goes too far, feel free to adapt. This one works for me on Windows and Linux. (Haven't tried macOS)

### DIFF
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -34,6 +34,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


go/tools/builders/nogo_main.go
```diff
@@ -90,6 +91,9 @@ func run(args []string) error {
 		return fmt.Errorf("errors found by nogo during build-time code analysis:\n%s\n", diagnostics)
 	}
 	if *xPath != "" {
+		if runtime.GOOS != "darwin" {
```
jayconrod: Why not `abs` on darwin?

We avoid changing paths for executables and shared libraries on darwin because they get baked into the binary and affect loading. We're just writing files here though, so it should be fine.

Xjs: Honestly, I wasn't sure what exactly these files did, and I thought seeing some .a files on the command line (I may have been mistaken), that's why. If you think abs() is fine on any platform, so be it. (Maybe I'll even find time to test it on a darwin machine later today)

Xjs:
Confirmed that abs() works on Darwin.

Issue #1376, opened 2018-03-09
Cannot compile on NixOS


crawford:
### Environment ###

```
$ bazel version
Build label: 0.10.1- (@non-git)
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Sat Nov 2 00:00:00 +11968 (315532800000)
Build timestamp: 315532800000
Build timestamp as int: 315532800000
```

```
$ nixos-version
17.09.3129.1dcd022f01b (Hummingbird)
```

### Failure ###

```
$ bazel build tarball
DEBUG: /home/alex/.cache/bazel/_bazel_alex/9a62ea23e6672e5bdeba02f2fd03b905/external/io_bazel_rules_go/go/private/sdk.bzl:103:5: env: ‘bin/go’: No such file or directory
ERROR: /home/alex/.cache/bazel/_bazel_alex/9a62ea23e6672e5bdeba02f2fd03b905/external/io_bazel_rules_go/BUILD.bazel:7:1: every rule of type go_context_data implicitly depends upon the target '
@go_sdk//:packages.txt', but this target could not be found because of: no such package '@go_sdk//': Traceback (most recent call last):                                                        
        File "/home/alex/.cache/bazel/_bazel_alex/9a62ea23e6672e5bdeba02f2fd03b905/external/io_bazel_rules_go/go/private/sdk.bzl", line 50
                _prepare(ctx)                                                                                                             
        File "/home/alex/.cache/bazel/_bazel_alex/9a62ea23e6672e5bdeba02f2fd03b905/external/io_bazel_rules_go/go/private/sdk.bzl", line 104, in _prepare
                fail("failed to list standard package...")                                                                                              
failed to list standard packages                          
ERROR: Analysis of target '//:tarball' failed; build aborted: no such package '@go_sdk//': Traceback (most recent call last):
        File "/home/alex/.cache/bazel/_bazel_alex/9a62ea23e6672e5bdeba02f2fd03b905/external/io_bazel_rules_go/go/private/sdk.bzl", line 50
                _prepare(ctx)                                                                                                             
        File "/home/alex/.cache/bazel/_bazel_alex/9a62ea23e6672e5bdeba02f2fd03b905/external/io_bazel_rules_go/go/private/sdk.bzl", line 104, in _prepare
                fail("failed to list standard package...")                                                                                              
failed to list standard packages                          
[...]

PR #2041, opened 2019-04-24
Generate dyld wrappers scripts on NixOS

steeve:
On Nix/NixOS, the Go distribution can't run because of a dynamic linker
issue being in the wrong path. Create wrapper after the download to
manually invoke the dynamic linker.

We can't use patchelf yet because patchelf 0.10 is not yet released in
nixpkgs [1].

1. https://github.com/NixOS/patchelf/issues/66

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -132,6 +132,41 @@ def _remote_sdk(ctx, urls, strip_prefix, sha256):
             stripPrefix = strip_prefix,
             sha256 = sha256,
         )
+    _patch_nix(ctx)
+
+_NIX_DYLD_TPL = """\
+#!/bin/sh
+exec {dyld} {binary} ${{@}}
+"""
+
+def _patch_nix(ctx):
+    """
+    Generates wrappers for the Go SDK binaries to use the correct interpreter
+    on Nix/NixOS.
+    """
+    # Are we on Nix/NixOS? Check via $NIX_CC
+    nix_cc = ctx.os.environ.get("NIX_CC", None)
+    if not nix_cc:
+        return
+    # Get the dyld
+    res = ctx.execute(["cat", "%s/nix-support/dynamic-linker" % nix_cc])
+    if res.return_code:
+        fail("error reading the dynamic linker:\n" + res.stdout + res.stderr)
+    dyld = res.stdout.rstrip()
+    # Generate the wrappers. Unfortunately we can't use patchelf to do that
+    # until patchelf 0.10 is in NixOS. So manually invoke the dyld via wrappers
+    # instead.
+    # See https://github.com/NixOS/patchelf/issues/66
+    for binary in ctx.path("bin").readdir():
+        # Does the binary needs a wrapper?
+        if ctx.execute(["patchelf", "--print-interpreter", binary]).return_code:
+            continue
+        orig = "%s.1" % binary
+        ctx.execute(["mv", binary, orig])
+        ctx.file(binary, _NIX_DYLD_TPL.format(
+            dyld = dyld,
+            binary = orig,
+        ), executable = True)
 
 def _local_sdk(ctx, path):
     for entry in ["src", "pkg", "bin"]:
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -134,9 +134,9 @@ def _remote_sdk(ctx, urls, strip_prefix, sha256):
         )
     _patch_nix(ctx)
 
-_NIX_DYLD_TPL = """\
+_NIX_DYLD = """\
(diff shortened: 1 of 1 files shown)


steeve:
Should also fix https://github.com/bazelbuild/rules_go/issues/1376.

steeve:
Waiting on @QuentinPerez or @teh-cmc to confirm it works out of the box.

steeve:
You have a fair point about maintainability.

`go_wrap_sdk` doesn't fill the same use case not work for our case because it is used to wrap a Go distribution from nixpkgs, which another use case (and a cool one at that).

We'd like to run the vanilla Go distribution as fetched by `rules_go` on NixOS. Not everybody in our team is running NixOS (some are ubuntu, some are macos), so changing the `WORKSPACE` doesn't seem like an option either. Also, `go_download_sdk` for forked distributions would work properly (NixOS is `linux_amd64`).

Also, note that we are fetching the dyld using the method described by NixOS themselves: https://nixos.wiki/wiki/Packaging/Binaries#The_Dynamic_Loader.

steeve:
Thinking about it, I think that it's fine as long as it doesn't bleed on non-NixOS users somehow (which I don't think it will). To be honest I think it's fine if someday you break it. You can't expect to test the whole OS/Version/Use case matrix. If it does break, either nobody cares and after a while it can be removed, or people do care and their raise the flag.
Kind of like what we are already doing with objc and iOS (`apple.bzl`): yes, `master` is broken for our use case at the moment but we help fixing it as we understand we are a pretty special use case (well, we hope it'll be not so special later on!).

jayconrod:
Could you write a repository rule that downloads and extracts the Go SDK, and patches the binaries if running on NixOS with the logic here? Then you could use `go_wrap_sdk` over that to do all the Bazel-specific setup.

steeve:
It's worth noting that the problem is related to Nix, not NixOS specifically (Nix runs on Linux and macOS) so a test could be written for it I think. Would that put your worries at rest?

I absolutely could do a repo rule but I feel it would be better for everybody (including people running bazel in nix, which doesn't really work out of the box today) to have a turn-key solution upstreamed?

One other solution would be to provide patch hooks in the `_go_remote_sdk` rule so that only the patch would live inside the project. But then again people would need to share their snippet via side-channels.

jayconrod:
I'm mainly worried that I don't understand the Nix dynamic loader. Even if there is a test for this, if it ever breaks, it will take me a very long time to reproduce the problem and fix it. I don't think I can afford the time and effort to maintain this.

Another thought: this problem will affect any language that downloads a dynamically-linked binary toolchain. It's not just a Go issue.

Is there another repository of Nix related Bazel rules where this could live? I know the Tweag I/O folks are active Nix and Bazel users. Maybe they'll have some thoughts?

steeve:
You are absolutely right about anything dynamic downloaded.

The issue is that in the case of Nix, the glibc/dyld itself is sandboxed. This means that the ELF interpreter Go uses is hardcoded in the binary (`/lib/ld64.so`). On Nix the linker lives in another location.

Now Nix people made `patchelf` to change it (it seems this routine to them). However, due to the issue in the OP, we can't use it.  The solution is to call dyld manually in a wrapper script.

This PR isn't complete as I'm not properly sandboxing the `.1` files too. It would be indeed much easier if `patchelf` was used. From what I understand, it's just that it hasn't been relased in `nixpkgs` yet.

Perhaps @mboes could chime in ?

mboes:
We're users of rules_go and users of Nix and NixOS. In rules_haskell (which depends on rules_go), we do this:

```
go_register_toolchains(go_version = "host")
```

This works well for everyone: non Nix users download their own Go toolchain and make sure it's available in the `PATH`, and Nix users likewise make the toolchain available in the `PATH` but using a nix-shell. This is not very hermetic for non-Nix users, but for Nix users it is. TBH within a team I would recommend sticking to everyone using Nix or no one using Nix (that's not an onerous requirement, since Nix is just a package manager and runs everywhere except windows).

I feel a lot more comfortable with this scheme than anything involving `patchelf`. Which is a brittle hack patching binaries.

mboes:
Also, you should be able to use rules_nixpkgs to feed rules_go with a Go toolchain from Nix, even without using a nix-shell. And if you want that to happen only on NixOS, do a tiny bit of autodetection in a custom repository rule. (We use platform constraints in rules_haskell but I'm not sure whether those are available in WORKSPACE.)

jayconrod:
Closing this PR. If any changes are needed to `go_wrap_sdk` or other rules to enable external modifications for Nix, I'm open to that. I'd rather not have Nix-specific logic in SDK rules though.

PR #2049, opened 2019-05-08
Support go1.12.5 and go1.11.10

cblecker:


jayconrod:
Sorry, I don't accept PRs for Go versions. The review process is basically generating an equivalent PR, then confirming the two are identical.

#2050 is most of the way through CI. I'll cherry-pick it and tag releases today.

cblecker:
Sounds good, @jayconrod! Thought I would try to make it easier, but just happy to see this change getting in. Thank you as always!

PR #2052, opened 2019-05-08
remove -lstdc++ from CGO_LDFLAGS if -static-libstdc++ is present

gengyl08:
This pull request allows libstdc++ to be statically linked if -static-libstdc++ is specified in the link options.

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -317,10 +317,11 @@ def _cgo_codegen_impl(ctx):
         objc_outs.append(gen_file)
         builder_args.add("-src", gen_file.path + "=" + src.path)
 
-    # Filter out -lstdc++ in CGO_LDFLAGS if we don't have any C++ code. This
-    # also gets filtered out in link.bzl.
+    # Filter out -lstdc++ in CGO_LDFLAGS if we don't have any C++ code or libstdc++
+    # is statically linked. This also gets filtered out in link.bzl.
     have_cc = len(source.cxx) + len(source.objc) + len(ctx.attr.deps) > 0
-    if not have_cc:
+    static_cc = "-static-libstdc++" in linkopts
+    if not have_cc or static_cc:
         linkopts = [o for o in linkopts if o not in ("-lstdc++", "-lc++")]
 
     tool_args.add("-objdir", out_dir)
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -56,21 +56,28 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, sSrcs, hSrcs []string, pa
 	}
 	defer cleanup()
 
-	// Filter out -lstdc++ and -lc++ from ldflags if we don't have C++ sources,
-	// and set CGO_LDFLAGS. These flags get written as special comments into cgo
-	// generated sources. The compiler encodes those flags in the compiled .a
-	// file, and the linker passes them on to the external linker.
+	// Filter out -lstdc++ and -lc++ from ldflags if we don't have C++ sources
+	// or -static-libstdc++ is explicitly specified, and set CGO_LDFLAGS. These
+	// flags get written as special comments into cgo generated sources. The
+	// compiler encodes those flags in the compiled .a file, and the linker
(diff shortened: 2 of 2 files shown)


jayconrod:
I'm not sure rules_go is the right place to filter out this flag. It doesn't seem like a Go-specific issue.

rules_go does not add `-lstdc++` or `-lc++` automatically. These come from the C/C++ toolchain configuration in CROSSTOOL. How are you setting `-static-libstdc++`?

gengyl08:
@jayconrod We tried to remove -lstdc++ from the c/c++ toolchain however that prevents other dependencies from building (specifically protobuf and grpc in our case). We are adding -static-libstdc++ in clinkopts in the go_rule. Without the changes in this pull request -lstdc++ will take precedence over -static-libstdc++ and libstdc++ will still be dynamically linked. The changes in this pull request allows libstdc++ to be statically linked if the developer explicitly specifies -static-libstdc++ in the go_rule.

jayconrod:
> We tried to remove -lstdc++ from the c/c++ toolchain however that prevents other dependencies from building (specifically protobuf and grpc in our case).

Do these dependencies need to be built in the same configuration as your Go targets, i.e., do your Go targets link against them? I'd expect the protobuf compiler to be built in a host configuration, so it could be built with a different toolchain from the target configuration.

What happens when a `go_binary` is built with a `cc_library` in `cdeps`? It seems like the `cc_library` will be built with `-lstdc++` (which would be the default) may not be compatible with any Go library with these flags.

In general, I really want to avoid special cases like this. I'm not convinced this is correct, and if something goes wrong, it will be difficult to diagnose. This also still seems like a C/C++ toolchain configuration issue to me. I'd expect you'd have the same problem with `rust_binary` or anything else that links against C++. rules_go is probably not the right place to fix it.

gengyl08:
@jayconrod Thank you for your quick responses!

Specifically, what we are doing is:
1. we have a cc_library rule A
2. We imported com_github_golang_protobuf and org_golang_google_grpc by calling go_rules_dependencies() from WORKSPACE
3. we have a go_library rule B which has A in its cdeps and com_github_golang_protobuf and org_golang_google_grpc in its deps

My understanding is that in this case the cc_library, com_github_golang_protobuf and org_golang_google_grpc has to use the same cc_toolchain. It'd be great if you know a way to use different cc_toolchains for them.

In our case, we actually have a customized CROSSTOOL that specifies -lstdc++ as a feature (following ttps://github.com/bazelbuild/bazel/issues/2954). We enable this feature by default in .bazelrc and remove this feature from cc_library when we want to link statically.

The above works fine for cc_library but not for go_library because it doesn't have a "features" attribute. The resulting dilemma is:
1. com_github_golang_protobuf and org_golang_google_grpc requires -lstdc++ so that we need to have the -lstdc++ feature enabled in .bazelrc.
2. By removing this feature in cc_library, we can link A statically
3. However cdeps also has this feature from .bazelrc and adds -lstdc++ back, resulting in dynamically linked libstdc++ in B

Hope this explains our problem. And hope you agree the problem cannot be fixed with CROSSTOOL.

If you think this pull request may introduce potential problems, do you think adding a, say, "staticlibstdcxx" flat in go_library is a good idea? This way people can explicitly remove -lstdc++ by specifying this flag, which should be safe and backward compatible.

jayconrod:
Thanks for explaining. I'm not sure I fully understand the issue yet though.

* Why do targets in `com_github_golang_protobuf` and `org_golang_google_grpc` need `-lstdc++`? I know there is some cgo code in there, but there aren't any `cdeps`, and I don't think there are any cgo C++ sources. Even if there is C++, why doesn't `-static-libc++` work?

* Controlling whether `-lstdc++` is passed to the linker using features seems strange to me. @hlopko is this still the recommended way? The advice in bazelbuild/bazel#2954 is a couple years old, and I know there have been a lot of changes since then.

Earlier, I thought this was about deciding to pass `-lstdc++` or `-static-libstdc++`. My concern that at most of these flags should be passed to the entire build. I'm fine with removing both of these flags from Go targets if we know there's no C++ code, but I still think it's dangerous to remove one flag if the other is present. CROSSTOOL should be configured so that both `-lstdc++` and `-static-libstdc++` are not passed.

PR #2054, opened 2019-05-10
Migrate for Bazel's --incompatible_require_ctx_in_configure_features

hlopko:
This PR migrates rules_go for Bazel's incompatible change https://github.com/bazelbuild/bazel/issues/7793, which is present in 0.25.

### DIFF
--- a/go/private/compat/compat_repo.bzl
+++ b/go/private/compat/compat_repo.bzl
@@ -38,16 +38,18 @@ _go_rules_compat = repository_rule(
 )
 
 def go_rules_compat(**kwargs):
-    v = native.bazel_version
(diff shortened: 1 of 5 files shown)


hlopko:
@jayconrod I see your TODO a line above my change regarding toolchain files. [`cc_toolchain.all_files`](https://docs.bazel.build/versions/master/skylark/lib/CcToolchainInfo.html#all_files) method is now available and it is forward compatible with toolchains.

jayconrod:
I changed this a little to go through the compatibility layer. Old versions of Bazel (rules_go supports back to 0.18.0) don't know about the `ctx` parameter.

PR #2064, opened 2019-05-16
Windows: future-proof single_output_test.bzl

laszlocsomor:
Fix single_output_test() on Windows with the new
native test wrapper.

Bazel 0.25 introduced the
--incompatible_windows_native_test_wrapper flag.

This flag enables a new test wrapper on Windows
that doesn't require Bash. On other platforms this
flag has no effect.

The new test wrapper is a C++ binary and doesn't
use Bash, but therefore test rules whose
executable is a naked shell script also don't
work with it.

The solution here is to write a special script
that does nothing and is valid both as `.bat` and
as `.sh`.

See https://github.com/bazelbuild/bazel/issues/6622

### DIFF
--- a/go/private/tools/single_output_test.bzl
+++ b/go/private/tools/single_output_test.bzl
@@ -13,25 +13,45 @@
 # limitations under the License.
 
 def _impl(ctx):
+    exe = ctx.outputs.out
     ctx.actions.write(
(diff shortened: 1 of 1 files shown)


laszlocsomor:
Thanks for the rapid review!

Issue #2037, opened 2019-04-22
Restrictive shared library extensions causes incorrect link arguments


clintharrison:
### What version of rules_go are you using?

master

### What version of Bazel are you using?

0.24.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Linux x64

### Any other potentially useful information about your toolchain?

Using the output of a `configure_make` target from `rules_foreign_cc` is how I triggered this issue.
The package in question (MIT Kerberos) produces a handful of versioned `.so` files. This is allowed by the standard C++ rules: https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CppFileTypes.java;l=163

It turns out `rules_go` will only look for `.so`, which means that when we have a list of `.so.##` files, without a `.so` present, the `-rpath` flag doesn't get generated. See https://github.com/bazelbuild/rules_go/blob/da0ac6f8d552baef52abf8c36e669df306f416d0/go/private/actions/link.bzl#L115.

PR #2066, opened 2019-05-16
Allow dependencies on versioned shared libraries

clintharrison:
This is an attempt at fixing #2037, based on the logic used for the native C++ rules -- see https://github.com/bazelbuild/bazel/blob/e95234d4e06cd5c4442434f5867224aebcb31ab7/src/main/java/com/google/devtools/build/lib/rules/cpp/CppFileTypes.java#L169-L171.

It's unfortunately pretty ugly in Starlark, since the native logic uses a simple regex, and instead we must do that string matching manually.

There might be a better way to do this? I am pretty out of my depth with this 🙂


go/private/rules/cgo.bzl
```diff
@@ -345,13 +346,20 @@ def _cgo_codegen_impl(ctx):
                 # us the static variant. We'll get one file for each transitive dependency,
                 # so the same file may appear more than once.
                 if (lib.basename.startswith("lib") and
-                    any([lib.basename.endswith(ext) for ext in SHARED_LIB_EXTENSIONS])):
+                    has_shared_lib_extension(lib.basename)):
```
jayconrod: Please also update the corresponding logic in `cgo_configure`. That's the new path. I hope to delete `_cgo_codegen_impl` and other rules soon, but both paths need to be supported.

tests/core/starlark/common_tests.bzl
```diff
@@ -0,0 +1,32 @@
+load("@io_bazel_rules_go//go/private:common.bzl", "has_shared_lib_extension")
+load("@io_bazel_rules_go//go/private:skylib/lib/unittest.bzl", "asserts", "unittest")
+
+def _versioned_shared_libraries_test(ctx):
```
jayconrod: I'm glad to have a unit test for this, but could you also add something in `tests/core/cgo` that confirms a library with a name like this can be linked?

Please make sure to update README.rst (and add one in this directory) and run `tests/update.py`, too. Test build files aren't readable on their own, so we keep test docs separate.

tests/core/starlark/common_tests.bzl
```diff
@@ -0,0 +1,32 @@
...
+    asserts.true(env, has_shared_lib_extension("somelibrary.so.20.2"))
+    asserts.true(env, has_shared_lib_extension("a/somelibrary.so.2"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.so.e"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.so.2e"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.so.e2"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.so.20.e2"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.a.2"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.a..2"))
+    asserts.false(env, has_shared_lib_extension("somelibrary.so.2."))
+    asserts.false(env, has_shared_lib_extension("somelibrary.so."))
```
jayconrod: Please add some non-ASCII characters in here before and after `.so`. Some emoji or anything outside the BMP would be good.

I think the string operations in `has_shared_lib_extension` are correct, but let's add coverage.

tests/core/starlark/common_tests.bzl
clintharrison: Created (and updated) the tests, thanks!

clintharrison:
Thanks for the quick review!

I'm a little more confident now in the behavior. I...don't think it was actually working before, but now it matches the behavior suggested by https://source.bazel.build/bazel/+/69095be0988711b42dad047892cd20581e9e9e10:src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java;l=319;bpv= in using `-lfoo` and `-l:libfoo.so.1` accordingly. (In the case of the tests, this is `-limported` or `-l:libimported.so.2` accordingly.)

Struggling a little bit with the Windows tests, but hopefully excluded it correctly this time.

PR #2068, opened 2019-05-19
go_library: add experimental importpath_aliases attribute

### LINKED ISSUE #2058: Minimal module compatibility
A change is needed to rules_go in order to support Go repositories that rely on [minimal module compatibility](https://github.com/golang/go/wiki/Modules#how-are-v2-modules-treated-in-a-build-if-modules-support-is-not-enabled-how-does-minimal-module-compatibility-work-in-197-1103-and-111).

Suppose we depend on repository `example.com/foo` which has not migrated to modules. `example.com/foo` depends on repository `example.com/bar`, which has migrated to modules. The module path in the latter repository is `example.com/bar/v2`. There is no major version subdirectory in this repository (the go.mod is at the repository root directory), so `example.com/foo` must import `example.com/bar`, not `example.com/bar/v2`.

The Go command has a special case for this when building in GOPATH-mode. rules_go needs an equivalent special case. In each `GoCompilePkg` or `GoCompile` action, we map import paths to archive files when generating importcfg files. If we have an import path like `example.com/bar/v2` that matches no known dependencies, we should try to match it without the `vN` path component (`example.com/bar`) before reporting a strict dependency error.

jayconrod:
importpath_aliases lets a go_library specify a list of secondary names
by which it may be imported.

This is necessary to support minimal module compatibility. For
example, suppose we have module example.com/foo/v2 which doesn't use
major version subdirectories. example.com/bar imports a library in
this module as example.com/foo (not v2). In GOPATH mode, 'go build'
internally translates the import example.com/foo/v2 to
example.com/foo. To support this in bazel, we'll require
importpath_aliases = ["example.com/foo"] so the library may be
imported with either name. Gazelle may generate these attributes,
given a command line option.

Go 1.14 will likely have more powerful replace directives, which will
also let users import packages within a module by multiple names.
importpath_aliases will be useful for supporting that as well.

Fixes #2058


jayconrod:
Holding off on submitting until the corresponding change is in Gazelle so the whole thing can be verified.

PR #2072, opened 2019-05-23
Making toolchain work without turning off SIP

rosun82:
Significantly simplify the set up process. Now the toolchain will

- Automatically detect local go installation

- Warn about incompatible Bazel versions.

This is a temporary solution though.


rosun82:
Mistake

PR #2076, opened 2019-05-25
go/tools/bazel: refactor runfiles functionality

### LINKED ISSUE #1728: Runfile in go/tools/bazel does not work with binaries
Binaries can have their runfiles, but calling `bazel.Runfile(file)` will return an error about a missing TEST_SRCDIR variable.

To reproduce follow:
```
git clone https://github.com/filipesilva/rules_go_runfiles_binary
cd rules_go_runfiles_binary
bazel run :binary
```

You should see
```
environment variable "TEST_SRCDIR" is not defined, are you running with bazel test
```

### LINKED ISSUE #2031: go/tools/bazel: FindBinary does not work on Windows

0.18.3

0.24.1

yes

windows, amd64

### What did you do?

Tried to use `bazel.FindBinary("tests/core/cgo/tag_pure_bin")` in a test.

`FindBinary` should translate slashes in the path and append a `.exe` extension on Windows. It should search directories or the runfiles manifest depending on which is being used.

### What did you see instead?

An error. Currently, the runfiles manifest is not used at all, and there is no special handling for Windows paths.

jayconrod:
* Refactored Runfile, FindBinary, and other functions. Everything now
  calls a single initialization function through sync.Once.
* Added ListRunfiles. This will be used to re-create trees of
  runfiles, which will be needed for the new version of bazel_test.
* Moved and rewrote tests in tests/core/runfiles.

Fixes #2031
Fixes #1728


alexeagle:
hey there, I think this change broke rules_typescript
https://github.com/bazelbuild/rules_typescript/blob/master/devserver/runfiles/runfiles.go
it seems that `Runfile("build_bazel_rules_nodejs/packages/protractor/test/protractor-2/app.js")` used to work (starts with workspace name) but now it must be `Runfile("packages/protractor/test/protractor-2/app.js")` (no workspace name)?
Does that sound right?

jayconrod:
Are you able to switch to not using the workspace name? `Runfile` should only accept relative paths from the workspace root, but it was (and is still) way too permissive on the input path. 

I'm sorry for the break though. `Runfile` was originally written by a few external contributors, and it wasn't adequately documented or tested. I needed to refactor it and make it work on Windows to make progress, but it was hard to make any change without subtly breaking something.

alexeagle:
Is it really the right interface to accept a path with no workspace name? If I have a file `index.js` in the local workspace runfiles and also in an external one, I would have no way to locate the external index.js in the runfiles.

jayconrod:
It depends on the situation, but no, `Runfile` is a poor interface.

`ListRunfiles` provides the most complete information. I'd recommend using that if files may be in multiple workspaces or if you're using a mix of short paths and regular paths.

If you can pass the location directly through a build file with something `args = ["$(location :label)"]`, that may let you avoid this library altogether.

alexeagle:
I'm not ready to figure out that refactoring right now so I'll just patch out this commit from our rules_go http_archive fetch. One day the merge conflict will bring someone back to this thread. Maybe me! :)

PR #2080, opened 2019-05-29
Update protobuf

keith:
This updates protobuf, which allows us to remove the explicit zlib
dependency. The downside of this is users have to call the
`protobuf_deps` function from their WORKSPACEs.


jayconrod:
Sorry, I don't accept PRs updating dependencies. I'll update all dependencies together before the next release in a couple weeks.

Please follow the instructions at [Overriding dependencies](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#overriding-dependencies) to update `com_google_protobuf` in your own projects.

keith:
Ok, thanks. Please do update protobuf when you do though!

PR #2081, opened 2019-05-29
WIP: Update vendored Skylib to 0.8.0 with some edits

laszlocsomor:
Motivation: lib/unittest.bzl now works with Bazel's
native test wrapper on Windows, which will replace
the current Bash-script-based test wrapper.

The "some edits" are described in
"go/private/skylib/README.rst".

See https://github.com/bazelbuild/bazel/issues/6622


laszlocsomor:
I believe the tests failed because the Bazel version is too old for Skylib 0.8.0
I try a different approach: https://github.com/bazelbuild/rules_go/pull/2082

PR #2082, opened 2019-05-29
Fix vendored Skylib w/ Windows-native test wrapper

laszlocsomor:
Partially back-port a Skylib commit [1] to the
vendored Skylib code, to fix unittest.bzl for
Bazel's native test wrapper on Windows.

The native test wrapper will replace the
Bash-script test wrapper on Windows [2].

I tried updating the current Skylib 0.5.0 sources
to 0.8.0, but tests failed on Travis CI. I suspect
that an older Bazel version is used.

[1] https://github.com/bazelbuild/bazel-skylib/commit/daf513702286fe211f291675443235e35e79f34f
[2] https://github.com/bazelbuild/bazel/issues/6622

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -16,6 +16,7 @@
 
 load("@io_bazel_rules_go//go/private:common.bzl", "MINIMUM_BAZEL_VERSION")
 load("@io_bazel_rules_go//go/private:compat/compat_repo.bzl", "go_rules_compat")
+load("@io_bazel_rules_go//go/private:skylib/lib/unittest.bzl", "register_unittest_toolchains")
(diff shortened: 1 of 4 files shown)


laszlocsomor:
Tests look good \o/

jayconrod:
Thanks for fixing this, but I ended up removing the vendored copy of Skylib in #2083 instead.

There is a set of files loaded from `//go:deps.bzl` which can't depend on any external repositories, since they may not have been loaded yet. Those files only depend on `versions.bzl`, which remains.

Everything not loaded from `//go:deps.bzl` can depend on `@bazel_skylib` directly, which is declared at `0.7.0`. I think that's new enough that we shouldn't have problems.

laszlocsomor:
Cool, thanks!

Issue #2079, opened 2019-05-28
Bazel 0.26 break iOS and perhaps Android cross compilation


steeve:
### What version of rules_go are you using?

```
ddea1ade19b0f7cab64d859d95fa6494714019d7
```

### What version of gazelle are you using?

```
325c4cf53b65148f5392f5f65462970a7e5cb6fa
```

### What version of Bazel are you using?

```
Build label: 0.26.0
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue May 28 08:38:24 2019 (1559032704)
Build timestamp: 1559032704
Build timestamp as int: 1559032704
```

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

```
Darwin XXXX 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64
```

### The problem

Bazel 0.26 ships with new platform enabled toolchains, which are defined like this:
```python
OSX_TOOLS_CONSTRAINTS = {
    "darwin_x86_64": ["@bazel_tools//platforms:osx", "@bazel_tools//platforms:x86_64"],
    "ios_i386": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:x86_32"],
    "ios_x86_64": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:x86_64"],
    "watchos_i386": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:x86_32"],
    "watchos_x86_64": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:x86_64"],
    "tvos_x86_64": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:x86_64"],
    "armeabi-v7a": ["@bazel_tools//platforms:arm"],
    "ios_armv7": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:arm"],
    "ios_arm64": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:arm"],
    "ios_arm64e": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:arm"],
    "watchos_armv7k": ["@bazel_tools//platforms:ios", "@bazel_tools//platforms:arm"],
[...]

PR #2086, opened 2019-06-03
Add iOS as a platform

steeve:
Opening in draft for now.

iOS is a special variant of GOOS=darwin. In order to support the new
CC toolchain constraints that come with bazel 0.26, we need to propagate
the same OS that what bazel expects, or else CGo won't work because it
can't find a toolchain.

This needs Bazel 0.26.1 as 0.26.0 has a bug regarding this [1].

1. https://github.com/bazelbuild/bazel/pull/8522

Also, this might require gazelle to handle the ios case, too.

### DIFF
--- a/go/platform/list.bzl
+++ b/go/platform/list.bzl
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 GOOS = {
-    "android": None,
+    "android": "@bazel_tools//platforms:android",
     "darwin": "@bazel_tools//platforms:osx",
+    "ios": "@bazel_tools//platforms:ios",
     "dragonfly": None,
     "freebsd": "@bazel_tools//platforms:freebsd",
     "linux": "@bazel_tools//platforms:linux",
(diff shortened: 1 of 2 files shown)


steeve:
I'm wondering is it would be better that `GOOS=ios` be propagated down to right before Go invocation, maintaining the illusion everywhere.

steeve:
For gazelle, which may be required as now this select doesn't work:
```python
    deps = select({
            "//vendor/golang.org/x/sys/unix:go_default_library",
        ],
```

It should be:
```python
    deps = select({
            "//vendor/golang.org/x/sys/unix:go_default_library",
        ],
        "@io_bazel_rules_go//go/platform:ios": [
            "//vendor/golang.org/x/sys/unix:go_default_library",
        ],
```

Unless we trick it somehow ?

jayconrod:
I've started to work on a fix for this. Currently, we have a `GOOS_GOARCH` table of valid platforms, and everything is generated from that. I'm working on expanding this to be a wider table with `struct` elements. The `GOOS` and `GOARCH` may be independent from the `constraint_values`.

I'd rather not introduce `ios` as anything other than a prefix for a few `platform` and `toolchain` targets. It's not a real `GOOS` value. I haven't figured out how `"@io_bazel_rules_go//go/platform:darwin"` is going to work yet in `select`. Need to experiment more.

jayconrod:
Let's discuss on #2079 instead of here.

steeve:
Close in favor of #2090

steeve:
Close in favor of #2090

Issue #2089, opened 2019-06-04
Bazel 0.26.0 fixed the autoconfigured C++ toolchains and now break cross-compilation of Go


hlopko:
Hello :)

With the fix for https://github.com/bazelbuild/bazel/issues/8330 we discovered one flaw in our design of Go cross compilation. Currently rules_go depend on C++ toolchain in //go/private:context.bzl (`toolchains = ["@bazel_tools//tools/cpp:toolchain_type"]`). This dependency exists no matter if we actually have C++ dependencies or not.

The problem is that with https://github.com/bazelbuild/bazel/issues/8330 C++ autoconfigured toolchains properly report their constraints, and now Bazel detects that we don't actually have a C++ cross-compilation toolchain available. As a result the cross-compilation build of go fails: 

```
ERROR: While resolving toolchains for target @io_bazel_rules_go//:go_context_data: no matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type
ERROR: Analysis of target '//:main' failed; build aborted: no matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type
```

To reproduce, create a hello world go_binary, and build it with `bazel build //:hello_world --platforms=@io_bazel_rules_go//go/toolchain:darwin_amd64` on linux, (or if you're on mac, use a different platform).

hlopko:
CC @katre, @philwo

jayconrod:
@hlopko @katre Do you have a solution for this in mind already? Is there any way to make a toolchain dependency optional?

I think we could define a rule that requires the C++ toolchain. Go rules could depend on that target or not via a `select`. WDYT?

hlopko:
I didn't have time to think about it too much.

I guess the simple workaround for the meantime is to revert https://github.com/bazelbuild/rules_go/commit/dd527c7d01af29fa7aabba8860c1ad3ded785153 (I didn't try it yet but it should help).

The separate rule approach is worth thinking about. What will you select on? Ideally you want to take C++ toolchain if its present...

@lberki had a nice evil idea - have an aspect traverse go dependencies, and if it finds C++ target, it steals its C++ toolchain from it. There's a lot of handwaving there :))

hlopko:
Another not great alternative is to register Null-object-like C++ toolchains that can only fail the build when actually used.

jayconrod:
Would reverting dd527c7 break cgo compilation in general? Is that required in Bazel 0.26.0, or still optional?

[...]

PR #2090, opened 2019-06-04
Add iOS toolchains, platforms, and constraint_values

### LINKED ISSUE #2079: Bazel 0.26 break iOS and perhaps Android cross compilation

jayconrod:
Refactored the description of target platforms. The source of truth is
now the PLATFORMS table go/private/platforms.bzl. Declarations for
config_settings in //go/platform; constraint_values and aliases in
//go/toolchain; and toolchains in @go_sdk are generated from this
table.

In addition to the normal GOOS_GOARCH list, there are some additional
entries for iOS. iOS platforms still have GOOS=darwin, but they are
compatible with a different constraint_value
(@bazel_tools//platforms:ios instead of @bazel_tools//platforms:osx).

Fixes #2079


jayconrod:
@steeve Please take a look. This should make it possible to `bazel build --platforms=@io_bazel_rules_go//go/toolchain:ios_arm`.

This is incomplete. I don't have a solution for `select` yet. Although it seems to compile correctly, it won't link (but my toolchain is probably just messed up).

jayconrod:
I added a commit that should fix `@io_bazel_rules_go//go/platform:darwin`. It should be true if target platform is any Go `darwin` or `ios` platform or if the target platform is the default (not cross-compiling) and the host OS is macOS.

steeve:
@jayconrod it seems I can't use the `@io_bazel_rules_go//go/toolchain:ios` with your patch. I suspect only the `os_arch` combination exist ?

steeve:
I tried constraining on `@bazel_tools//platforms:ios`, and I'm happy to report that it works!

steeve:
Now I'm having issues on Android:
With `android_arm64`:
```
ERROR: While resolving toolchains for target //android/java/main:ZenlyCore: no matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type
ERROR: Analysis of target '//android:publish_snapshot' failed; build aborted: no matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type
```

I suspect this is a problem in the NDK rules of Bazel.

And on `android_arm`:
```
ERROR: /private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/external/org_golang_google_grpc/internal/syscall/BUILD.bazel:3:1: Illegal ambiguous match on configurable attribute "deps" in @org_golang_google_grpc//internal/syscall:go_default_library:
@io_bazel_rules_go//go/platform:android
@io_bazel_rules_go//go/platform:darwin
Multiple matches are not allowed unless one is unambiguously more specialized.
INFO: SHA256 (http://jcenter.bintray.com/org/ow2/asm/asm/6.2/asm-6.2.jar) = 917bda888bc543187325d5fbc1034207eed152574ef78df1734ca0aee40b7fc8
ERROR: Analysis of target '//android:publish_snapshot' failed; build aborted:

/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/external/org_golang_google_grpc/internal/syscall/BUILD.bazel:3:1: Illegal ambiguous match on configurable attribute "deps" in @org_golang_google_grpc//internal/syscall:go_default_library:
@io_bazel_rules_go//go/platform:android
@io_bazel_rules_go//go/platform:darwin
Multiple matches are not allowed unless one is unambiguously more specialized.
INFO: Elapsed time: 6.105s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (23 packages loaded, 8805 targets configured)
```

steeve:
Note that this happens also on bazel 0.25.3
`android_arm`
```
/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/external/org_golang_google_grpc/internal/channelz/BUILD.bazel:3:1: Illegal ambiguous match on configurable attribute "deps" in @org_golang_google_grpc//internal/channelz:go_default_library:
@io_bazel_rules_go//go/platform:darwin
@io_bazel_rules_go//go/platform:android
Multiple matches are not allowed unless one is unambiguously more specialized.
```

`android_arm64`:
```
/private/var/tmp/_bazel_steeve/309ac22c92b055a3e3f78825fe63406b/external/org_golang_google_grpc/internal/channelz/BUILD.bazel:3:1: Illegal ambiguous match on configurable attribute "deps" in @org_golang_google_grpc//internal/channelz:go_default_library:
@io_bazel_rules_go//go/platform:darwin
@io_bazel_rules_go//go/platform:android
Multiple matches are not allowed unless one is unambiguously more specialized.
```

steeve:
cc @jin since he made the NDK rules
@jin could it be that the NDK rules suffered from the same bug from https://github.com/bazelbuild/bazel/pull/8522 ?

steeve:
Hum this is weird, when I constrain on `@bazel_tools//platforms:aarch64` instead of `@io_bazel_rules_go//go/toolchain:arm64`, I get the same issue as `android_arm`.

jayconrod:
> it seems I can't use the @io_bazel_rules_go//go/toolchain:ios constraint with your patch. I suspect only the os_arch combination exist ?
> The reason is we have our own platform that used to apply rules_go constraints

Some explanation about the naming here: `//go/platform` contains `config_setting` targets for each `GOOS`, `GOARCH`, and `GOOS_GOARCH` pair. `//go/toolchain` contains `constraint_setting`, `constraint_value`, and `platform` targets. Ideally `platform` targets would be in `//go/platform`, but that package predates Bazel support for platforms and toolchains by quite a bit, and displacing the old `config_settings` would break builds.

Anyway, you should be able to define a `platform` like this:

```
platform(
    name = "custom_ios_arm",
    constraint_values = [
        "@bazel_tools//platforms:ios",
        "@bazel_tools//platforms:arm',
        "@io_bazel_rules_go//go/toolchain:is_darwin",  # needed for @io_bazel_rules_go//platform:darwin to work
        # any other values relevant to your toolchains
    ],
)
```

I think the Android problems you're experiencing are #2089, which is affecting cross-compilation in general. I'm planning to fix that in another PR before the next release.

PR #2097, opened 2019-06-17
Fix cgo build failures with specific GCC flags

AustinSchuh:
cgo runs the compiler over a test file as part of the build process.  It then parses the compiler errors.  This requires the compiler to be configured well enough to work for cgo.

-fdiagnostics-color is another way to enable color support on GCC.  This adds colors unconditionally to the output (bazel doesn't present a TTY, so to get colors with bazel, you need to force them on).

-fmax-errors=... is a way to prevent GCC from completely flooding your terminal when something goes wrong.  A missed semicolon or misplaced {} can cause thousands of lines of errors.  This can stop showing errors before cgo sees all the errors it is looking for.  We want to strip it out too.

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -69,9 +69,15 @@ GoContext = provider()
 _GoContextData = provider()
 
 _COMPILER_OPTIONS_BLACKLIST = {
+    # cgo parses the error messages from the compiler.  It can't handle colors.
+    # Ignore both variants of the diagnostics color flag.
     "-fcolor-diagnostics": None,
     "-fdiagnostics-color": None,
-    "-fmax-errors=20": None,
+
+    # cgo also wants to see all the errors when it is testing the compiler.
+    # fmax-errors limits that and causes build failures.
+    "-fmax-errors=": None,
+
     "-Wall": None,
 
     # Symbols are needed by Go, so keep them
@@ -88,8 +94,18 @@ _LINKER_OPTIONS_BLACKLIST = {
     "-Wl,--gc-sections": None,
 }
 
+def _match_option(option, pattern):
+    if pattern.endswith("="):
+        return option.startswith(pattern)
+    else:
+        return option == pattern
+
 def _filter_options(options, blacklist):
-    return [option for option in options if option not in blacklist]
+    return [
+        option
+        for option in options
+        if not any([_match_option(option, pattern) for pattern in blacklist])
+    ]
 
 def _child_name(go, path, ext, name):
     childname = mode_string(go.mode) + "/"
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -70,6 +70,8 @@ _GoContextData = provider()
 
 _COMPILER_OPTIONS_BLACKLIST = {
     "-fcolor-diagnostics": None,
+    "-fdiagnostics-color": None,
+    "-fmax-errors=20": None,
     "-Wall": None,
 
     # Symbols are needed by Go, so keep them
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -795,7 +795,10 @@ def setup_cgo_library_for_mode(name, srcs, cdeps, copts, cxxopts, cppopts, clink
         name = cgo_o_name,
         srcs = [select_main_c],
         deps = cdeps + cgo_o_deps,
-        copts = copts + cppopts,
+        copts = copts + cppopts + [
+            # The generated thunks often contain unused variables.
+            "-Wno-unused-variable",
+        ],
         linkopts = clinkopts,
         visibility = ["//visibility:private"],
         **common_attrs


go/private/rules/cgo.bzl
```diff
@@ -795,7 +795,10 @@ def setup_cgo_library_for_mode(name, srcs, cdeps, copts, cxxopts, cppopts, clink
         name = cgo_o_name,
         srcs = [select_main_c],
         deps = cdeps + cgo_o_deps,
-        copts = copts + cppopts,
+        copts = copts + cppopts + [
```
jayconrod: This code is deprecated and is only used to build cgo code for a very narrow legacy case.

`_cgo_context_data_impl` is probably the right place to add this flag (to `c_compile_options`).

go/private/context.bzl
```diff
@@ -70,6 +70,8 @@ _GoContextData = provider()
 
 _COMPILER_OPTIONS_BLACKLIST = {
     "-fcolor-diagnostics": None,
+    "-fdiagnostics-color": None,
+    "-fmax-errors=20": None,
```
jayconrod: Why exclude this flag? And why 20 specifically?

jayconrod:
Test failures are caused by #2100, unrelated this this PR.

AustinSchuh:
Process question: should I use an amend/rebase workflow, or just add more commits?  I normally use gerrit for reviews.

jayconrod:
I prefer more commits because it preserves comment threads, but either works: the PR will be squashed and merged, so it ends up as a single commit in a linear history.

(I also use Gerrit, and I wish GitHub code review would copy some things from there)

go/private/context.bzl
AustinSchuh: In our code base, we add -fmax-errors=20.  Generally only the first couple of errors are actually helpful when something fails to compile.  cgo tests the compiler with a file which has more than 20 errors, so the presence of this flag drops errors it is expecting and fails the build.

Currently your filter function doesn't support a startswith list.  I can add one, though it'll be slower to parse.  I could also find the limit and add -fmax-errors={0..N} to filter those out.  What would you prefer?

jayconrod: I'd suggest adding a `_match_option` function:

```
def _match_option(opt, pattern):
    if pattern.endswith("="):
        return opt.startswith(pattern)
    else:
        return opt == pattern
```

Then using that as a predicate in `_filter_options`:

```
def _filter_options(options, blacklist):
    return [
        option
        for option in options
        if not any([_match_option(option, pattern) for pattern in blacklist])
    ]
```

A quick comment that `-fmax-errors=N` breaks cgo would be helpful.

go/private/rules/cgo.bzl
AustinSchuh: I've confirmed by A/B building with and without this line that this is the exact target that is causing the problem.

I'm building github.com/mattn/go-sqlite3

The build rule generated is:
go_library(
    name = "go_default_library",
    srcs = [ 
        "backup.go",
        "callback.go",
        "doc.go",
        "error.go",
        "sqlite3.go",
        "sqlite3-binding.c",
        "sqlite3-binding.h",
        "sqlite3_context.go",
        "sqlite3_go18.go",
        "sqlite3_load_extension.go",
        "sqlite3_other.go",
        "sqlite3_solaris.go",
        "sqlite3_type.go",
        "sqlite3_windows.go",
        "sqlite3ext.h",
        "static_mock.go",
    ],  
    cgo = True,
    clinkopts = select({
            "-ldl",
        ],
        "@io_bazel_rules_go//go/platform:solaris": [
            "-lc",
        ],
        "@io_bazel_rules_go//go/platform:windows": [
            "-lmingwex -lmingw32",
        ],
        "//conditions:default": [], 
    }), 
    copts = [ 
        "-std=gnu99",
        "-DSQLITE_ENABLE_RTREE -DSQLITE_THREADSAFE=1 -DHAVE_USLEEP=1",
        "-DSQLITE_ENABLE_FTS3 -DSQLITE_ENABLE_FTS3_PARENTHESIS -DSQLITE_ENABLE_FTS4_UNICODE61",
        "-DSQLITE_TRACE_SIZE_LIMIT=15",
        "-DSQLITE_DISABLE_INTRINSIC",
        "-Wno-deprecated-declarations",
    ] + select({
        "@io_bazel_rules_go//go/platform:android": [
            "-I.",
        ],
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:dragonfly": [
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:freebsd": [
            "-I.",
        ],
            "-DHAVE_PREAD64=1 -DHAVE_PWRITE64=1",
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:nacl": [
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:netbsd": [
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:openbsd": [
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:plan9": [
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:solaris": [
            "-D__EXTENSIONS__=1",
            "-I.",
        ],
        "@io_bazel_rules_go//go/platform:windows": [
            "-I. -fno-stack-check -fno-stack-protector -mno-stack-arg-probe",
        ],
        "//conditions:default": [], 
    }) + select({
        "@io_bazel_rules_go//go/platform:windows_386": [
            "-D_USE_32BIT_TIME_T",
        ],
        "//conditions:default": [], 
    }), 
    importpath = "github.com/mattn/go-sqlite3",
)

I put it here since restricting it to the smallest scope seemed like a good idea rather than being more broad.  I'm happy to do whatever you want though.

go/private/context.bzl
AustinSchuh: Sweet, done.

go/private/rules/cgo.bzl
jayconrod: Hmm, that's strange. Is that on `master` or an older version of rules_go? I tried `bazel build @com_github_mattn_go_sqlite3//:go_default_library` with rules_go and go-sqlite3 at master, but `setup_cgo_library_for_mode` wasn't called (checked by adding a `fail` in there).

By any chance, do you have `objc = True` set in a rule somewhere?

In any case please do add this to `_cgo_context_data_impl`. The legacy path should disappear at some point soon.

AustinSchuh: Just for you, I upgraded to the latest released rules_go (0.18.6), and I'm still seeing it.  I'm using Gazelle 0.17.0.  The only ```objc = True``` in our code base is in a test in rules_go.

I removed the offending rule and gazelle added it back without changes.  So I think it's not some legacy flag on the rule.  I suspect Bazel is building the binary even if it is unused for completeness.

I used ```-s``` with bazel and changed some cgo code to add an unused variable warning and already found ```-Wno-unused-variable``` in the command line used to build the cgo generated code, and in the command line used to build standalone c code.  I'm reluctant to add it somewhere else since that would duplicate the flag.

Let me know what you want to do.

jayconrod: Ah, the `release-0.18` branch was cut before the new cgo stuff was merged, so this path will still be active there. It shouldn't be active at `master` though (unless `objc = True`).

So I think this option should in *both*  `_cgo_context_data_impl` and `setup_cgo_library_for_mode`. I'll backport the latter part to `release-0.18` after this is merged (the next minor release will likely be early July).

Some time after `release-0.19` is cut, I'll remove the legacy path, which will un-duplicate this.

AustinSchuh: Your request got hard...  I want to test master with the failing code before sending it out for review, but that's a fair amount of work.  I'm tempted to just roll this part back and revisit when you cut a new release.

jayconrod: Up to you. I'm fine with this being left here, just be aware that the old path will disappear in a few weeks.

AustinSchuh: Let's just do that then.  When we upgrade again, I'll take stock of how things look and send you any reviews needed to finish this.

Thanks Jay!

PR #2099, opened 2019-06-18
Fix iOS cross compilation and enable test

jayconrod:
Updates #2079


jayconrod:
Test failures are due to #2100, which is unrelated to this.

Issue #2100, opened 2019-06-18
Coverage tests are broken in Bazel 0.27.0


jayconrod:
Example: https://buildkite.com/bazel/rules-go-golang/builds/1197#7fba7969-7a88-4201-b0a4-8982d9ebf8d0

* //tests/core/coverage:coverage_test_test
* //tests/core/nogo/coverage:coverage_test
* //tests/core/race:race_feature_test

robfig:
The latest build looks like all these tests are passing, e.g.
https://buildkite.com/bazel/rules-go-golang/builds/3263#ad864e99-1e32-4dfd-a83b-951fa4fad57c

```
//tests/core/coverage:coverage_test                                      PASSED in 16.8s
//tests/core/nogo/coverage:coverage_test                                 PASSED in 5.5s
//tests/core/race:race_test                                              PASSED in 57.3s
```

Closing, but please reopen if I've missed something

PR #2101, opened 2019-06-18
go_proto_compiler: remove dependence on C/C++ toolchain

### LINKED ISSUE #2089: Bazel 0.26.0 fixed the autoconfigured C++ toolchains and now break cross-compilation of Go

jayconrod:
The C compiler was used to set PATH. PATH is needed on Windows to
locate DLL files. Instead of requiring the C compiler, just use the
same environment as the rest of the Go toolchain. PATH should be set
there.

Updates #2089

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -172,7 +172,6 @@ platforms:
     - "-//tests/core/coverage:cross_cover_test_test"
     - "-//tests/core/cross:cross_test"
     - "-//tests/core/cross:ios_select_test"
-    - "-//tests/core/cross:proto_test"
     - "-//tests/core/go_binary:go_default_test"
     - "-//tests/core/go_path:go_path"
     - "-//tests/core/go_path:go_path_test"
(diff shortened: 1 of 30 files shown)


jayconrod:
Test failures due to #2100, which is unrelated.

Issue #2091, opened 2019-06-05
Having GRPC as a dependency after calling `go_rules_dependencies()` on windows causes failure


mason-bially:
### What version of rules_go are you using?

0.18.5

### What version of gazelle are you using?

N/A

### What version of Bazel are you using?

0.26.0

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Windows / x86-64

### Any other potentially useful information about your toolchain?

Windows C++ 10.0.17763.0

### What did you do?

Build a project using GRPC after calling `go_rules_dependencies()` in my WORKSPACE file.

### What did you expect to see?

A successful build.

### What did you see instead?

```
/_bazel_user/hwwnc6kc/external/net_zlib/BUILD.bazel:32:1: C++ compilation of rule '@net_zlib//:zlib' failed (Exit 2)
cl : Command line error D8021 : invalid numeric argument '/Wno-unused-variable'
```

mason-bially:
I believe it's this section of code: https://github.com/bazelbuild/rules_go/blame/master/go/private/repositories.bzl#L86

Which has apparently been corrected:
https://github.com/protocolbuffers/protobuf/issues/5472

jayconrod:
MSVC is not supported with rules_go. The Go toolchain requires GCC or clang in order to build cgo code.

You may find [these instructions for setting up msys and MinGW](https://github.com/bazelbuild/rules_go/blob/master/windows.rst) useful. We're working on improving Windows C/C++ toolchain selection, but that probably won't help until Bazel 0.27.0 at least.

It is surprising that `net_zlib` doesn't compile since it doesn't have anything to do with Go, but this isn't something I can support.

mason-bially:
I guess I was more asking for including rules_go in my workspace file to not break other dependencies. I'm not expecting it to work, I'm expecting it to *not break GRPC* by being included.

mason-bially:
@jayconrod 

To reiterate my issue is not with `go_rules` not working on Windows. Read my issue carefully. My problem is that using `go_rules` in a workspace file *breaks GRPC on Windows*.

jayconrod:
@mason-bially Could you give a more complete example? It would be helpful to have a WORKSPACE file and a `bazel build` command with a specific target that reproduces the problem.

Simply calling `go_rules_dependencies` and building Go targets without proto dependencies seems to work fine. I'm also able to build `go_proto_libraries` with services using MinGW as described above.

[...]

PR #2105, opened 2019-06-21
Update repositories.bzl

mauriciogg:
3.7.1 adds fixes missing errorprone dependency in protobuf_java_util. Java libraries that depend on com_google_protobuf break  with external/com_google_protobuf/java/util/src/main/java/com/google/protobuf/util/Timestamps.java:39: error: package com.google.errorprone.annotations does not exist

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -81,7 +81,7 @@ def go_rules_dependencies():
         git_repository,
         name = "com_google_protobuf",
         remote = "https://github.com/protocolbuffers/protobuf",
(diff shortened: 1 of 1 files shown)


jayconrod:
Sorry, I don't accept PRs to update dependencies. Please follow instructions at [Overriding dependencies](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#overriding-dependencies) to use a specific version in your own project. There's a known issue with `com_google_protobuf` in particular due to conflicts with its own dependencies, so see #2091 as well.

All dependencies will be updated together before the next major release.

PR #2106, opened 2019-06-21
Add a config with Bazel 0.26.1 to find out what broke the build

philwo:


philwo:
Please ignore this PR - I'm just using this to run a test on CI with an additional Bazel version.

jayconrod:
I opened bazelbuild/bazel#8670 with details about the break.

#2100 is the tracking issue here.

philwo:
It is supported :) If you give the task a „name“ attribute, too, it will easily be distinguishable in the Buildkite results. Example: https://github.com/bazelbuild/bazel/blob/168ee7c589955372843115f67d9df2b4cc42e6ba/.bazelci/postsubmit.yml#L188 results in “OpenJDK 10 Javabase on Ubuntu 18.04”.

jayconrod:
CI is green now that 0.27.1 is released, so I'll close this.

Issue #1508, opened 2018-05-22
bazel_test: allow multiple test cases


jayconrod:
It should be possible to have multiple test cases (i.e., multiple invocations of Bazel) within a single `bazel_test` rule. This would allow us to consolidate some tests and speed up the presubmit.

jayconrod:
Closing as obsolete. `go_bazel_test`, the successor to `bazel_test`, supports this.

PR #2109, opened 2019-06-23
go_bazel_test: wrapper and test library for running Bazel from go_test

jayconrod:
go_bazel_test is a macro that wraps go_test. It will be used for
integration tests in rules_go as a replacement for bazel_test. The
macro adds dependencies on bazel_testing and runfiles from rules_go.

bazel_testing is a new test framework. Its TestMain function may be
called to construct a Bazel workspace (specified as a txtar
archive). Tests are run within this workspace. Bazel may be invoked
with the RunBazel function.

Updates #1508
Updates #1959


jayconrod:
Remaining test failure are due to #2100.

Issue #2102, opened 2019-06-19
Cache misses when using volatile keys in x_defs 


robbertvanginkel:
I'm running into unexpected caching behavior when using rules_go with stamping trough `x_defs`. Mainly based on experience with c binaries I'd expect to be able to get 100% cache hit, but the go binaries keep re-linking when using stamping with variables from the workspace status.

### What did you do?
```
$ cat WORKSPACE
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "io_bazel_rules_go",
    urls = ["https://storage.googleapis.com/bazel-mirror/github.com/bazelbuild/rules_go/releases/download/0.18.6/rules_go-0.18.6.tar.gz"],
    sha256 = "f04d2373bcaf8aa09bccb08a98a57e721306c8f6043a2a0ee610fd6853dcde3d",
)
load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
go_rules_dependencies()
go_register_toolchains()
$ cat BUILD.bazel 
load("@io_bazel_rules_go//go:def.bzl", "go_binary")
go_binary(
    name = "go",
    srcs = ["main.go"],
    x_defs = {"main.buildtime" : "{BUILD_TIMESTAMP}"}
)
$ cat main.go 
package main

import "fmt"

var buildtime = ""

func main() {
	fmt.Printf("Hello from Go at %s\n", buildtime)
}
```

Also available at https://gist.github.com/robbertvanginkel/925b10198e98010d0ca7ad722b5acbef

### What did you expect to see?

Expected 100% cache hit when doing:
```
$ bazel build --disk_cache ... //go_binary
$ bazel clean
$ bazel build --disk_cache ... //go_binary
```

### What did you see instead?

Only a partial cache hit, with recompilation of 

```
$ bazel build --disk_cache /tmp/examplecache //:go
INFO: Invocation ID: 8a7942c6-2371-4a47-8132-b23a05e5d37d
INFO: Analyzed target //:go (24 packages loaded, 6583 targets configured).
INFO: Found 1 target...
Target //:go up-to-date:
  bazel-bin/darwin_amd64_stripped/go
INFO: Elapsed time: 1.798s, Critical Path: 0.30s
INFO: 3 processes: 2 remote cache hit, 1 darwin-sandbox.
INFO: Build completed successfully, 7 total actions
```

[...]

PR #2110, opened 2019-06-25
Add configurable stamp support with --[no]stamp and stamp attrs

robbertvanginkel:
This adds support for rules_go to make stamping of workspace variables conditional to Bazel's [`--nostamp`](https://docs.bazel.build/versions/master/user-manual.html#flag--stamp) flag. 

Fixes https://github.com/bazelbuild/rules_go/issues/2102.


robbertvanginkel:
@jayconrod I can't seem to reproduce the build failure locally, or find the raw non summarized logs. Can you let me know if we can retry or help me out with the logs?

robbertvanginkel:
I originally went without the `stamp` attribute, but added it for the tests later. I couldn't figure out a way to test the stamping behavior as the `go_test` targets could be build with `--stamp` and `--nostamp`. I went looking for the other rules did this, and for C the docs on `stamp = 1` mention

> `stamp = 1`: ... Use this if there are tests that depend on the build information.

Happy to remove the stamp attribute, but need some guidance on what to do with the existing tests under `tests/legacy/examples/stamped_bin`.

jayconrod:
Ah, I just wrote a new testing thing last weekend that would work for this: [`//go/tools/bazel_testing`](https://github.com/bazelbuild/rules_go/tree/master/go/tools/bazel_testing). You can write a `go_bazel_test` that runs in a synthetic workspace. The test itself is written in Go and uses the `bazel_testing` library. It can run `bazel` commands and verify the result. I haven't migrated all the integration tests yet, but [`//tests/legacy/no_prefix:no_prefix_test`](https://github.com/bazelbuild/rules_go/blob/master/tests/legacy/no_prefix/BUILD.bazel#L4) is an example.

So in this case, the test would define a test workspace with a binary to be stamped. It could `bazel run` the binary with and without the `--stamp` option. The binary can print something that would be stamped, and the test can look for that in stdout.

munnerz:
I've just pulled in a revision of `rules_go` that includes this change (upgrading from 0.18.6) - I've found that with this rule definition:

```
go_library(
    name = "go_default_library",
    srcs = [
        "context.go",
        "ingress.go",
        "useragent.go",
        "util.go",
        "version.go",
    ],
    importpath = "github.com/jetstack/cert-manager/pkg/util",
    x_defs = {
        "AppVersion": "{STABLE_APP_VERSION}",
        "AppGitCommit": "{STABLE_APP_GIT_COMMIT}",
        "AppGitState": "{STABLE_APP_GIT_STATE}",
    },
)
```

My `AppVersion` var is being set to `{STABLE_APP_VERSION}`. Presumably if I were to set `--stamp` on my `bazel build` command, this would be set appropriately.

> With go_binary, we already explicitly ask for specific keys to be stamped using x_defs. Setting stamp = 0 and x_defs = {"Timestamp": "{BUILD_TIMESTAMP}"} wouldn't make sense. Likewise, setting stamp = 1 without setting x_defs wouldn't make sense.

This makes sense, however if len(x_defs) > 0, should `stamp` not default to True to maintain backwards compatibility? Esp. given I cannot set `stamp = True` on my go_library or go_binary, meaning this becomes a user facing change to how they build the project? (developers now have to specify `--stamp`, or it has to be added to our `.bazelrc`).

munnerz:
update: even when setting `--stamp=true`, I'm still seeing the same behaviour: https://github.com/jetstack/cert-manager/pull/1855/commits/cd113898ab1a79142eb3c82edbef6e6687e75e69

Any insight as to what I'm doing wrong here? This seems at least like a regression or otherwise a breaking change 😅

jayconrod:
@munnerz This was intentionally a user-facing change. We were not being consistent with other languages.

I can see an argument for adding a `stamp` attribute on `go_binary` similar to `cc_binary`, but I think the default would have to be to observe the `--stamp` flag. Setting `--stamp` in `.bazelrc` seems like a better option though.

PR #2118, opened 2019-07-02
toolchains: remove default value from cgo and darwin constraints

### LINKED ISSUE #2115: specific SHA has interoperability issue with rules_docker

```python
http_archive(
    name = "io_bazel_rules_go",
    urls = ["https://github.com/bazelbuild/rules_go/archive/8ea79bbd5e6ea09dc611c245d1dc09ef7ab7118a.tar.gz"],
    sha256 = "039c159e8f16a365a5f0ce31c8e6ca8acba5708962e904c1f164ba79fb8efe4d",
    strip_prefix = "rules_go-8ea79bbd5e6ea09dc611c245d1dc09ef7ab7118a",
)
```

```python
http_archive(
    name = "bazel_gazelle",
    urls = ["https://github.com/bazelbuild/bazel-gazelle/releases/download/0.17.0/bazel-gazelle-0.17.0.tar.gz"],
    sha256 = "3c681998538231a2d24d0c07ed5a7658cb72bfb5fd4bf9911157c0e9ac6a2687",
)
```

tested 0.27.0 -> 0.23.1

Using the latest

darwin, amd64

Trying to get cross compilation working.

### What did you do?

See https://github.com/chrislovecnm/external-dns/tree/work-on-bazel

Cross compiled go binary running in a container.

### What did you see instead?

```console
ERROR: While resolving toolchains for target //:external_dns_image: no matching toolchains found for types @io_bazel_rules_docker//toolchains/docker:toolchain_type
ERROR: Analysis of target '//:external_dns_image' failed; build aborted: no matching toolchains found for types @io_bazel_rules_docker//toolchains/docker:toolchain_type
```

I have tried rules_docker 0.7.x -> 0.8.1 branches and I am still experiencing the same issues. I have tried lower versions of bazel and have had the exact same issues.  So the cross compile for go, works like a champ!  But add rules_docker, and we get the no matching toolchain issue.

See https://github.com/bazelbuild/rules_docker/issues/943

tl;dr

Using latest bazel and a SHA from go_rules, cross build `build --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64` works like a champ.  Add rules_docker to the mix, and I am getting a toolchain error.

jayconrod:
Previously, //go/toolchains:cgo_constraint and darwin_constraint had
default values cgo_on and is_darwin (on macOS only). This enabled cgo
and the darwin config_setting (again, macOS only) when the default
target platform (@bazel_tools//platforms:target_platform) is used,
which is normally true.

I believed these defaults would only apply to platforms that didn't
specify a value otherwise. It appears that these defaults also add
constraints to toolchains that don't specify a value otherwise
(bazelbuild/bazel#8778). These means non-Go toolchains (e.g.,
docker_toolchain) are incompatible with non-default target platforms,
since they don't specify values for these constraints.

With this change, cgo_constraint no longer has a default
value. darwin_constraint and its values are deleted
entirely. Platforms defined in //go/toolchain specify cgo_on or
cgo_off. Toolchains defined in @go_sdk (or any similar workspace)
require cgo_on or no value (i.e., cgo is off by default), except for
the toolchains for the host platform, which require cgo_off or no
value (i.e., cgo is on by default).

darwin_constraint and its values are deleted entirely. I can't think
of an automatic way to make a single case in a select expression
work. To slightly mitigate this, ios and ios_$goarch are added to
//go/platform for use in separate select cases. The darwin
config_setting will only be true on macOS, but the darwin build
constraint will continue to work on both operating systems.

Fixes #2115
Updates #2089


jayconrod:
@chrislovecnm PTAL. I think this should fix the issue.

@steeve Unfortunately, this means `@io_bazel_rules_go//go/platform:darwin` will only be true on macOS, not iOS. I've added `@io_bazel_rules_go//go/platform:ios`, but you'll need extra cases in each `select` expression where it's relevant. Without default values, I can't think of any way to make a `config_setting` true on two operating systems.

chrislovecnm:
Will test drive 😁

steeve:
@jayconrod I'm not worried about the new constraint, in fact I think it's nicer (like android vs linux). But I think gazelle might need updating too.

jayconrod:
@steeve I've opened bazelbuild/bazel-gazelle#566 for this.

steeve:
great job jay thank you

steeve:
For the record I can't update Bazel nor rules_go because of issues related to this. I'll try this PR !

steeve:
Would it make sense to leverage https://github.com/bazelbuild/platforms too ?

jayconrod:
> Would it make sense to leverage https://github.com/bazelbuild/platforms too ?

Eventually yes, but not until that's canonical for the minimum supported version of Bazel. I think there's currently an incompatible flag that turns the targets in `@bazel_tools//platforms` into aliases into that repo.

chrislovecnm:
@jayconrod just tested external dns and I can run the container.  If you want to double check my WORKSPACE file as I pulled in the PR directly from the github archive.  Just want to make sure I tested it correctly.

https://github.com/chrislovecnm/external-dns/blob/work-on-bazel/WORKSPACE

chrislovecnm:
@nlopezgi if you want to take a look.  This seems to be working :)

chrislovecnm:
@jayconrod / @nlopezgi we talked about testing this, probably in rules_docker, how do you recommend proceeding??

jayconrod:
@chrislovecnm Thanks for testing!

I think this change and the underlying change is too complicated to backport to the release-0.18 branch. I'm planning to cut release-0.19 soon, once a few more cgo issues are resolved and dependencies are updated.

PR #2126, opened 2019-07-09
Update dependencies

### LINKED ISSUE #2113: org_golang_x_tools is almost 4 months old
Is it time to upgrade? Since `go_rules_dependencies()` is called very early, it forces all repos to use the old version of `org_golang_x_tools`

jayconrod:
* com_google_protobuf is removed because we can't call its macro from
  go_rules_dependencies.
* org_golang_google_grpc and its dependencies are removed because
  there are too many of them.
* Rewrote go/workspace.rst documentation.

Fixes #2113


linzhp:
Thanks for doing this. We need the newer version of x/tools from different tasks

PR #2137, opened 2019-07-13
Add a dynamic load test for c-shared link mode

siddharthab:
This test has a non-trivial dependency (golang.org/x/crypto) to check for more complex modes.

On master, this test currently fails. But it passes on 0.18.7.


jayconrod:
Thanks for this test case! This is really helpful for reproducing the problem. I've pushed a fix.

PR #2165, opened 2019-08-01
Specify sha256 sum of org_golang_x_tools archive

SilverRainZ:
Both `--repository_cache` and `--distdir` require archive has sha256sum specified.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -41,6 +41,7 @@ def go_rules_dependencies():
         # master, as of 2019-01-15
         urls = ["https://codeload.github.com/golang/tools/zip/bf090417da8b6150dcfe96795325f5aa78fff718"],
         strip_prefix = "tools-bf090417da8b6150dcfe96795325f5aa78fff718",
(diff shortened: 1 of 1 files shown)


SilverRainZ:
This PR is just for `release-0.17` because I use it.

steeve:
zip files from `codeload.github.com` are not stable (they are generated on the fly) and therefore can't have `sha256` unfortunately. there are known cases of that happening.

SilverRainZ:
Is `git_repository` acceptable? This issue totally breaks my offline CI.

jayconrod:
Sorry, I can't accept this for the reason @steeve gave. If GitHub changes the way they build zip files, it will retroactively break everyone on this branch. This has happened multiple times in the past. We no longer use `http_archive` on `codeload.github.com` for this reason. For rules that did use that end point previously, this is why `sha256` is not set. You can override this repo by declaring it in your WORKSPACE before calling `go_rules_dependencies`, but make sure to apply the same patches.

PR #2169, opened 2019-08-06
Fix "local variable 'importmap' is referenced before assignment"

prestonvanloon:
There were code paths where `importmap` is undefined on line 157 and 174.

These errors only occur when cgo_archives is set on the source. 
I ran into this with some custom rules around GoSource.

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -70,6 +70,8 @@ def emit_archive(go, source = None):
         if a.source.mode != go.mode:
             fail("Archive mode does not match {} is {} expected {}".format(a.data.label, mode_string(a.source.mode), mode_string(go.mode)))
 
+    importmap = "main" if source.library.is_main else source.library.importmap
+            
     if not source.cgo_archives:
         # TODO(jayconrod): We still need to support the legacy cgo path when
         # Objective C sources are present, since the Objective C toolchain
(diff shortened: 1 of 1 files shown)


jayconrod:
I'm curious what your use case is for `cgo_archives`. I think the only thing that produces this now is libraries with `objc = True`. That functionality is deprecated and will be removed in the next release. I'm intending to remove handling for `cgo_archives` as well, but if you're using that in custom rules, please let me know and we can figure something out.

prestonvanloon:
@jayconrod, my use case here is that I want to link archive files to a go library without using cgo. The idea I am exploring is to compile a rust library to expose methods to go code via FFI. This requires packing the rust archives into the go library.

This is the code that I am most interested in:
https://github.com/bazelbuild/rules_go/blob/9b614efc502964e60de3f8e5d0f0eb740a2a6930/go/private/actions/archive.bzl#L166-L197

My custom rule looks something like

```python
def _go_rust_library_to_source(go, attr, source, merge):
    merge(source, attr.src[GoSource])
    source["cgo_archives"] = attr.rust_lib[DefaultInfo].files.to_list()

def _go_rust_library_impl(ctx):
    go = go_context(ctx)

    library = go.new_library(go, resolver = _go_rust_library_to_source, importable = True)
    source = go.library_to_source(go, ctx.attr, library, ctx.coverage_instrumented())

    return [
        library,
        source,
        archive,
        DefaultInfo(
            files = depset([archive.data.file]),
        ),
        OutputGroupInfo(
            cgo_exports = archive.cgo_exports,
            compilation_outputs = [archive.data.file],
        ),
    ]

go_rust_library = go_rule(
    _go_rust_library_impl,
    attrs = {
        "src": attr.label(
            providers = [GoLibrary],
            mandatory = True,
        ),
        "importpath": attr.string(
            mandatory = True,
        ),
        "rust_lib": attr.label(
            # TODO: add provider to restrict to rust_library.
            mandatory = True,
        ),
    },
)
```

jayconrod:
Very cool. I've filed #2170 to migrate the `cgo_archives` functionality to the new path.

PR #2174, opened 2019-08-12
go_toolchain: move cgo_context_data dep to go_context_data

### LINKED ISSUE #2166: cgo_context_data is not resolved for the target platform
# What version of rules_go are you using?

`master`

`master`

`0.28.0`

Yes.

`darwin_amd64`

Android NDK crosstool

### What did you do?

It seems that because `cgo_context_data` is resolved at toolchain creation time, the flags are the one from the host architecture, rather than the target.

I did a custom rule that depends on it and things are fine:
```python
load(
    "@bazel_tools//tools/cpp:toolchain_utils.bzl",
    "find_cpp_toolchain",
)
load(
    "@io_bazel_rules_go//go/private:providers.bzl",
    "CgoContextData",
)


def _test_cpp_impl(ctx):
    # cc_toolchain = find_cpp_toolchain(ctx)
    # print(">>>>", cc_toolchain)
    cgodata = ctx.attr._cgo_data[CgoContextData]
    print(">>>>>", cgodata)
    return []

test_cpp = rule(
    implementation = _test_cpp_impl,
    attrs = {
        "_cc_toolchain": attr.label(default = "@bazel_tools//tools/cpp:current_cc_toolchain"),
        "_cgo_data": attr.label(default = "@io_bazel_rules_go//:cgo_context_data"),
        "_xcode_config": attr.label(
            default = "@bazel_tools//tools/osx:current_xcode_config",
        ),
    },
    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
    fragments = ["apple", "cpp"],
    provides = [CcInfo],
)
```

If I compile for android, everything in `cgo_context_data` is indeed for android. But if I build a `go_binary`, I see the content is from the host somehow.

My take on that is that the target is pulled before the platform split, and thus may still be referencing the host.

jayconrod:
Previously, we defined separate cgo and non-cgo variants of
go_toolchain. The cgo variant depended on cgo_context_data, which
requires a C/C++ toolchain. However, go_toolchain is built in host
mode, so we always got a toolchain for the execution platform, not the
target platform. This broke cross-compilation with cgo.

With this change, we only declare a go_toolchain per platform,
regardless of cgo. go_context_data depends on cgo_context_data
if the cgo_off constraint is not set.

This means a C/C++ toolchain (for the correct platform) will be
required for the default target platform and platforms that have not
specifically opted out of cgo. Platforms declared in //go/toolchain
are already fine. However, users defining their own platforms may be
broken by this change.

Fixes #2166

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -63,14 +63,18 @@ nogo(
 # It may depend on cgo_context_data via go_toolchain.
 go_context_data(
     name = "go_context_data",
-    strip = select({
-        "@io_bazel_rules_go//go/private:strip-always": "always",
-        "@io_bazel_rules_go//go/private:strip-sometimes": "sometimes",
-        "@io_bazel_rules_go//go/private:strip-never": "never",
+    cgo_context_data = select({
+        "@io_bazel_rules_go//go/platform:internal_cgo_off": None,
+        "//conditions:default": ":cgo_context_data",
     }),
     stamp = select({
         "@io_bazel_rules_go//go/private:stamp": True,
         "//conditions:default": False,
+    }),
+    strip = select({
+        "@io_bazel_rules_go//go/private:strip-always": "always",
+        "@io_bazel_rules_go//go/private:strip-sometimes": "sometimes",
+        "@io_bazel_rules_go//go/private:strip-never": "never",
     }),
     visibility = ["//visibility:public"],
 )
--- a/go/platform/list.bzl
+++ b/go/platform/list.bzl
@@ -68,5 +68,14 @@ def declare_config_settings():
             constraint_values = [
                 "@bazel_tools//platforms:ios",
                 "@io_bazel_rules_go//go/toolchain:" + goarch,
-            ]
+            ],
         )
+
+    # Setting that determines whether cgo is enabled.
+    # This is experimental and may be changed or removed when we migrate
+    # to build settings.
+    native.config_setting(
+        name = "internal_cgo_off",
+        constraint_values = ["@io_bazel_rules_go//go/toolchain:cgo_off"],
+        visibility = ["@io_bazel_rules_go//:__pkg__"],
+    )
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -23,8 +23,8 @@ load(
     "CPP_LINK_EXECUTABLE_ACTION_NAME",
     "CPP_LINK_STATIC_LIBRARY_ACTION_NAME",
     "C_COMPILE_ACTION_NAME",
-    "OBJC_COMPILE_ACTION_NAME",
     "OBJCPP_COMPILE_ACTION_NAME",
+    "OBJC_COMPILE_ACTION_NAME",
 )
 load(
     "@io_bazel_rules_go_compat//:compat.bzl",
@@ -79,7 +79,6 @@ _COMPILER_OPTIONS_BLACKLIST = {
     # cgo also wants to see all the errors when it is testing the compiler.
     # fmax-errors limits that and causes build failures.
     "-fmax-errors=": None,
-
     "-Wall": None,
 
     # Symbols are needed by Go, so keep them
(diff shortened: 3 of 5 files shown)


jayconrod:
@steeve, please take a look and see if this will work for you.

I tested this manually with the default target, a non-cgo target in `//go/toolchain`, and a custom Android target with cgo enabled and a locally installed NDK. Other platforms seem to get the correct toolchain (or not). Android fails when cgo is disabled, but that seems to be expected.

At the moment, I'm not adventurous enough to set up a cross-compiling C/C++ toolchain in CI.

steeve:
That looks about right. A few things are missing I think: the sandboxing of the `cc_toolchain`, and the `ldflags`, here is my diff:
```diff
diff --git a/go/private/compat/v23.bzl b/go/private/compat/v23.bzl
index 325250f2..37098880 100644
--- a/go/private/compat/v23.bzl
+++ b/go/private/compat/v23.bzl
@@ -80,6 +80,9 @@ def cc_compile_flags(target):
 
     return options
 
+def cc_toolchain_all_files(ctx):
+    return ctx.files._cc_toolchain
+
 # Compatibility for --incompatible_disable_legacy_proto_provider
 
 PROTO_PROVIDER_NAME = ProtoInfo
diff --git a/go/private/compat/v25.bzl b/go/private/compat/v25.bzl
index 0826d3aa..ff157722 100644
--- a/go/private/compat/v25.bzl
+++ b/go/private/compat/v25.bzl
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 load("@io_bazel_rules_go//go/private:common.bzl", "as_iterable")
+load(
+    "@bazel_tools//tools/cpp:toolchain_utils.bzl",
+    "find_cpp_toolchain",
+)
 
 # Compatibility for --incompatible_disable_legacy_cc_provider
 
@@ -80,6 +84,9 @@ def cc_compile_flags(target):
 
     return options
 
+def cc_toolchain_all_files(ctx):
+    return find_cpp_toolchain(ctx).all_files.to_list()
+
 # Compatibility for --incompatible_disable_legacy_proto_provider
 
 PROTO_PROVIDER_NAME = ProtoInfo
diff --git a/go/private/context.bzl b/go/private/context.bzl
index ff0f9c8e..f54509eb 100644
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -29,6 +29,7 @@ load(
 load(
     "@io_bazel_rules_go_compat//:compat.bzl",
     "cc_configure_features",
+    "cc_toolchain_all_files",
 )
 load(
     "@io_bazel_rules_go//go/private:providers.bzl",
@@ -683,7 +685,7 @@ def _cgo_context_data_impl(ctx):
     )
 
     return [CgoContextData(
-        crosstool = ctx.files._cc_toolchain,
+        crosstool = cc_toolchain_all_files(ctx),
         tags = tags,
         env = env,
         cgo_tools = struct(
diff --git a/go/private/go_toolchain.bzl b/go/private/go_toolchain.bzl
index 50b76926..bb3bde48 100644
--- a/go/private/go_toolchain.bzl
+++ b/go/private/go_toolchain.bzl
@@ -54,7 +54,6 @@ def _go_toolchain_impl(ctx):
 
         # Internal fields -- may be read by emit functions.
         _builder = ctx.executable.builder,
-        _cgo_context_data = ctx.attr.cgo_context_data[CgoContextData] if ctx.attr.cgo_context_data else None,
     )]
 
 go_toolchain = rule(
@@ -81,10 +80,6 @@ go_toolchain = rule(
             doc = "The SDK this toolchain is based on",
         ),
         # Optional extras to a toolchain
-        "cgo_context_data": attr.label(
-            providers = [CgoContextData],
-            doc = "A target that collects information about the C/C++ toolchain.",
-        ),
         "link_flags": attr.string_list(
             doc = "Flags passed to the Go internal linker",
         ),
 
diff --git a/go/tools/builders/stdlib.go b/go/tools/builders/stdlib.go
index af161392..b9622563 100644
--- a/go/tools/builders/stdlib.go
+++ b/go/tools/builders/stdlib.go
@@ -82,7 +82,8 @@ func stdlib(args []string) error {
    sandboxPath := abs(".")
 
    // Strip path prefix from source files in debug information.
-   os.Setenv("CGO_CFLAGS", os.Getenv("CGO_CFLAGS")+" -fdebug-prefix-map="+sandboxPath+"=. -fdebug-prefix-map="+output+"=.")
+   os.Setenv("CGO_CFLAGS", os.Getenv("CGO_CFLAGS")+" "+strings.Join(defaultCFlags(output), " "))
+   os.Setenv("CGO_LDFLAGS", os.Getenv("CGO_LDFLAGS")+" "+strings.Join(defaultLdFlags(), " "))
 
    // Build the commands needed to build the std library in the right mode
    // NOTE: the go command stamps compiled .a files with build ids, which are
```

Other than that, that looks about right!

steeve:
Also this cosmetic one:
```
 # go_context_data collects build options and is depended on by all Go targets.
-# It may depend on cgo_context_data via go_toolchain.
+# It may depend on cgo_context_data if CGo isn't disabled.
 go_context_data(
     name = "go_context_data",
     strip = select({
@@ -72,13 +79,10 @@ go_context_data(
         "@io_bazel_rules_go//go/private:stamp": True,
         "//conditions:default": False,
     }),
```

steeve:
Just tested, with my diff above, everything is ok

jayconrod:
@steeve Thanks for testing. Could you submit that diff as a separate PR? I'll merge it at the same time as this one.

steeve:
@jayconrod I've opened #2184, thanks

PR #2181, opened 2019-08-14
go_path: add include_transitive

steeve:
This PR improves `go_path` in ~two~ one ways:
- ~add `go_archive_aspect` properties so that `include_pkg` are compiled with the right mode (since it consumes archives)~
- add `include_transitive` so that only the packages in `deps` are included in the GoPath. When used with `include_pkg`, this allows to only sandbox the needed packages, and thus invalidate potentially less.

When using gomobile's `gobind`, this reduces hot rebuild times by 2-3x, because less is invalidated.

### DIFF
--- a/go/core.rst
+++ b/go/core.rst
@@ -863,6 +863,11 @@ Attributes
 | included in the output directory. Files listed in the :param:`data` attribute                    |
 | for this rule will be included regardless of this attribute.                                     |
 +----------------------------+-----------------------------+---------------------------------------+
+| :param:`include_pkg`       | :type:`bool`                | :value:`False`                        |
++----------------------------+-----------------------------+---------------------------------------+
+| When true, a `pkg` subdirectory containing the compiled libraries will be created in the         |
+| generated `GOPATH` containing compiled libraries.                                                |
++----------------------------+-----------------------------+---------------------------------------+
 
 Defines and stamping
 --------------------
--- a/go/core.rst
+++ b/go/core.rst
@@ -868,6 +868,12 @@ Attributes
 | When true, a `pkg` subdirectory containing the compiled libraries will be created in the         |
 | generated `GOPATH` containing compiled libraries.                                                |
 +----------------------------+-----------------------------+---------------------------------------+
+| :param:`include_transitive`| :type:`bool`                | :value:`True`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| When true, the transitive dependency graph will be included in the generated `GOPATH`. This is   |
+| the default behaviour. When false, only the direct dependencies will be included in the          |
+| generated `GOPATH`.                                                                              |
++----------------------------+-----------------------------+---------------------------------------+
 
 Defines and stamping
 --------------------
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -38,7 +38,9 @@ def _go_path_impl(ctx):
     mode_to_archive = {}
     for mode, archives in mode_to_deps.items():
         direct = [a.data for a in archives]
-        transitive = [a.transitive for a in archives]
+        transitive = []
+        if ctx.attr.include_transitive:
+            transitive = [a.transitive for a in archives]
         mode_to_archive[mode] = depset(direct = direct, transitive = transitive)
 
     # Collect sources and data files from archives. Merge archives into packages.
@@ -163,6 +165,7 @@ go_path = rule(
         ),
         "include_data": attr.bool(default = True),
         "include_pkg": attr.bool(default = False),
+        "include_transitive": attr.bool(default = True),
         "_go_path": attr.label(
             default = "@io_bazel_rules_go//go/tools/builders:go_path",
             executable = True,
--- a/tests/core/go_path/BUILD.bazel
+++ b/tests/core/go_path/BUILD.bazel
@@ -27,6 +27,14 @@ go_path(
     deps = ["//tests/core/go_path/pkg/lib:go_default_library"],
 )
 
+go_path(
+    name = "notransitive_path",
+    testonly = True,
+    include_transitive = False,
+    mode = "copy",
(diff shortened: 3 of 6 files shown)


steeve:
This made me think that the aspect attrs should be defined once.

jayconrod:
Could you open a new issue or describe here why these features are needed?

In general, I'm hesitant to expand the API surface of rules_go unless there's a problem that can't be solved any other way. I'm especially hesitant to expand the use of `go_archive_aspect`, since we need to remove it and replace it with build settings soon.

This seems like it might be a very specialized use case, and I wonder if you might be better served with your own custom rule or locally modified version of `go_path`.

steeve:
Sure thing. I completely agree about not expanding the public API. That said, I find those very useful.
I'll split it in two points, which are not dependant on one another, but I figure I would put them in the same PR since they were touching `go_path`.

The context is that some of the packages/tools I saw requiring a proper `GOPATH` was because they were using the `go/build` and `go/importer` packages (gomobile's gobind being the one I'm targeting here). They either need a proper gopath filled with sources (`importer.For("source")`) and/or archives (`importer.For("gc")`). The latter being _much_ (1000x) faster than the former.
This was the reason for `go_path.include_pkg`.

However, `include_pkg` will will trigger the creation/compilation of archives, and thus requires proper mode setting, hence the `go_archive_aspect` and its properties. Doing that, we can ensure that a `GOPATH`'s archives are created with a proper mode (`pure` etc...).

The second is kind of more specialised, and it's the reason why I put it in its own commit. I think it fixes a widespread problem and is a nice addition. I'm going to take gomobile's `gobind` as an example but it applies to anything requiring a `GOPATH`.

The idea here is to minimise cache invalidation when running tools/packages on the `GOPATH`. When filling the `GOPATH` using transitive dependencies, any dependency change will trigger a re-creation of the `GOPATH` and the downstream graph. In the case of `gobind`, this means when changing non bound packages, `go_path` + `gobind` are rerun anyway, where a simple `GoLink` would have done the job.

In my case, when running `gobind` after a transitive dependency change, only a compilation of the package and a `GoLink` are performed, which saves time (`gobind` can be 10+ seconds). Think of it `rules_go` equivalent to `ijar`, so to speak.

jayconrod:
Thanks for explaining. That makes sense.

In an ideal world, `gobind` would use `golang.org/x/tools/go/packages`, and we'd have a driver for that framework that worked inside a Bazel action. Then it could consume archives more or less directly. There's a lot of work to be done to get into that state. `go_path` is meant as a stopgap, and I agree that it makes sense to augment it here.

A few things to fix:

* [ ] It looks like `//tests/core/go_path:go_path_test` is failing because a binary that `go_path` depends on also has these attributes set. I think this is the first time something with `go_archive_aspect` could depend on another with `go_archive_aspect`, so I think the outer aspect is overriding the inner aspect. If possible, can you fix that? If none of the mode attributes are set in `go_path`, that shouldn't override mode attributes of targets it depends on.
* [ ] Need tests to verify the new behavior.
* [ ] Please update the `go_path` documentation in go/core.rst.

steeve:
Doc updated and test added.
That said, about the aspect, I see the problem, but I'm not sure where to look ? If you can point me in a direction, that'd be great. Still digging.

jayconrod:
go/private/rules/aspect.bzl is where `go_archive_aspect` is implemented.

The idea is that rules such as `go_library` do their normal analysis and return `GoSource` and `GoArchive`. The aspect runs over all the same targets and returns `GoArchiveAspect` (which wraps another `GoSource` and `GoArchive`) if the aspect mode is different than the default mode. The heavy lifting of actually creating those providers is done by `go.library_to_source` and `go.archive`, which are called in the rules and in the aspect. `get_source` and `get_archive` are called to get the correct `GoSource` or `GoArchive`, depending on whether `GoArchiveAspect`.

This is all very hacky, and we should replace this with build settings and transitions, now that those work. Just need to find the time.

I'm not sure what happens when an aspect runs over a subgraph that uses the same aspect. I don't know what the precedence should be. I think the right behavior is that if `go_path` depends on a `go_binary`, and both have mode attributes set, `go_binary` should win. If one or the other has mode attributes set, that one should win.

I don't know whether it's possible to express that in the current framework though. If it's very difficult (e.g., if it requires introducing new wrappers or providers or aspects), we should hold back that part of the PR until the build setting / transition migration is done.

A workaround on your side might be to define a rule that has dependencies and mode attributes and uses an aspect like `go_binary` but doesn't actually link an executable. Then `go_path` could depend on that without modification.

steeve:
Since transitions got merged, this PR is now considerably simpler, so I'm reviving it.

go/core.rst
```diff
@@ -863,6 +863,17 @@ Attributes
...
+| :param:`include_pkg`       | :type:`bool`                | :value:`False`                        |
++----------------------------+-----------------------------+---------------------------------------+
+| When true, a `pkg` subdirectory containing the compiled libraries will be created in the         |
+| generated `GOPATH` containing compiled libraries.                                                |
++----------------------------+-----------------------------+---------------------------------------+
+| :param:`include_transitive`| :type:`bool`                | :value:`True`                         |
++----------------------------+-----------------------------+---------------------------------------+
+| When true, the transitive dependency graph will be included in the generated `GOPATH`. This is   |
+| the default behaviour. When false, only the direct dependencies will be included in the          |
+| generated `GOPATH`. (the default).                                                               |
```
jayconrod, line 875: Delete "(the default)" since true is the default.

jayconrod (commented):
Going through old PRs. Sorry I didn't notice this was updated earlier.

One small documentation change needed. Also please update. Looks good otherwise.

steeve:
Will do, thank you

steeve:
Rebased and updated.

jayconrod (approved):
Looks good, thanks!

Issue #1803, opened 2018-11-02
Split compiled code and export data for each archive


jayconrod:
When we build a Go package, we currently produce a single .a archive file in the [ar format](https://en.wikipedia.org/wiki/Ar_(Unix)). The archive contains two files: one is compiled code, consumed by the linker. The other is export data (type information for exported declarations), consumed by the compiler for dependent packages. For packages that contain cgo or assembly code, there may be additional .o files that the linker consumes.

Instead of this, we should produce two separate files (both still in the ar format): a .a file that contains compiled code, and a .x file that contains export data.

The advantage of producing separate files is that when the implementation of a function changes in a .go file, only the .a file built from that source would change; the .x file would stay the same. Consequently, we wouldn't need to re-compile dependent packages. Another advantage is that we could pack serialized facts from the analysis framework into .x files without confusing the linker.

steeve:
From what I'm ready, `go tool compile` can output the two file separately via the `-linkobj` param. This then creates 2 archives.

steeve:
I just made a poc, what was surprisingly easy!

steeve:
I'm opening a draft PR to showcase this. I gotta say I'm impressed that the go tool handles it by default.

PR #2199, opened 2019-08-23
Split compiled code and export data for each archive

steeve:
This a PoC to showcase #1803. Tests are not passing, but builds are. I'll update the tests asap.

It's pretty nice.

The metadata archive and the linking archives are smaller than the full archive, this means:
- most actions are faster
- linking is faster since the linking archive is smaller than the full archive
- i would figure this leads to significant speedups in RBE

All in all, in my non trivial project, building an Android AAR, changing a deep package:

Don't look at the time too much, as least half is spent zipping the AAR.

Without this PR:
```
INFO: Elapsed time: 15.596s, Critical Path: 14.72s
INFO: 22 processes: 22 darwin-sandbox.
```

With this PR:
```
INFO: Elapsed time: 12.385s, Critical Path: 11.73s
INFO: 6 processes: 6 darwin-sandbox.
```

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -48,6 +48,11 @@ def emit_archive(go, source = None):
     split = split_srcs(source.srcs)
     lib_name = source.library.importmap + ".a"
     out_lib = go.declare_file(go, path = lib_name)
+    out_archives = [out_lib]
+    out_linklib = None
+    if go.mode.linkobj:
+        out_linklib = go.declare_file(go, path = source.library.importmap + ".link.a")
+        out_archives.append(out_linklib)
     if go.nogo:
         # TODO(#1847): write nogo data into a new section in the .a file instead
         # of writing a separate file.
@@ -96,6 +101,7 @@ def emit_archive(go, source = None):
             importmap = importmap,
             archives = direct,
             out_lib = out_lib,
+            out_linklib = out_linklib,
             out_export = out_export,
             out_cgo_export_h = out_cgo_export_h,
             gc_goopts = source.gc_goopts,
@@ -119,6 +125,7 @@ def emit_archive(go, source = None):
             importmap = importmap,
             archives = direct,
             out_lib = out_lib,
+            out_linklib = out_linklib,
             out_export = out_export,
             gc_goopts = source.gc_goopts,
             cgo = False,
@@ -133,6 +140,7 @@ def emit_archive(go, source = None):
         importpath_aliases = source.library.importpath_aliases,
         pathtype = source.library.pathtype,
         file = out_lib,
+        linkfile = out_linklib,
         export_file = out_export,
         srcs = as_tuple(source.srcs),
         orig_srcs = as_tuple(source.orig_srcs),
@@ -151,7 +159,7 @@ def emit_archive(go, source = None):
         data = data,
         direct = direct,
         searchpaths = depset(direct = [searchpath], transitive = [a.searchpaths for a in direct]),
-        libs = depset(direct = [out_lib], transitive = [a.libs for a in direct]),
+        libs = depset(direct = out_archives, transitive = [a.libs for a in direct]),
         transitive = depset([data], transitive = [a.transitive for a in direct]),
         x_defs = x_defs,
         cgo_deps = depset(transitive = [cgo_deps] + [a.cgo_deps for a in direct]),
--- a/go/private/mode.bzl
+++ b/go/private/mode.bzl
@@ -28,7 +28,7 @@ LINKMODE_C_ARCHIVE = "c-archive"
 
 LINKMODES = [LINKMODE_NORMAL, LINKMODE_PLUGIN, LINKMODE_C_SHARED, LINKMODE_C_ARCHIVE, LINKMODE_PIE]
 
-def new_mode(goos, goarch, static = False, race = False, msan = False, pure = False, link = LINKMODE_NORMAL, debug = False, strip = False):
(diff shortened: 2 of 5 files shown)


steeve:
More benchmarks, from a widely used package in our repo:
```
INFO: Elapsed time: 17.364s, Critical Path: 14.68s
INFO: 57 processes: 57 darwin-sandbox.
```

```
INFO: Elapsed time: 13.916s, Critical Path: 13.56s
INFO: 6 processes: 6 darwin-sandbox.
```

steeve:
Here is a quick view of the sizes for huge gRPC proto file we have internally:
```
-r-xr-xr-x  1 steeve  wheel   501K Aug 24 01:28 obfs.a
-r-xr-xr-x  1 steeve  wheel    14M Aug 24 01:28 obfs.link.a
```

steeve:
Everything seems to be working, save for plugin mode. Apparently it needs the full link file.

I'm looking into disabling linkobj in `plugin` mode.

steeve:
One thing we can do is, if the mode is `plugin`, at link, merge the `.a` and `.link.a` and pass that to the linker. That way the benefit is kept.

steeve:
Ok so if a binary imports `plugin`, then it can't link with `.link.a` archives, it needs the full archives. The problem is that it can't be detected in `go.mode`, hence why the test is failing

steeve:
Since the plugin use case is so niche, I'm wondering if having an attribute for binaries that use `import "plugin"` would be useful, and it would disable linkobj mode (which the vast majority of users would benefit from)

steeve:
Thinking out loud, @jayconrod, since you have access to the google rules, how do they deal with that case?

steeve:
So after discussing on https://github.com/golang/go/issues/33820, here are the options:
- make the feature opt-in (perhaps via `features`), and thus make it part of the `go.mode`
- make the feature opt-out, since I would figure not many people use `plugin` anyway

steeve:
I've been thinking perhaps the best way is to do what rules_apple does at https://github.com/bazelbuild/rules_apple/blob/master/doc/common_info.md

Something like:
```
--define=io.bazel.rules_go.linkobj=1
--define=go.linkobj=1
```

steeve:
So I went ahead and added opt-in via `--define=go.linkobj=1`. I'll update the README if you guys are fine with the solution.

go/private/actions/archive.bzl
```diff
@@ -52,6 +52,11 @@ def emit_archive(go, source = None):
     split = split_srcs(source.srcs)
     lib_name = source.library.importmap + ".a"
     out_lib = go.declare_file(go, path = lib_name)
+    out_archives = [out_lib]
+    out_linklib = None
+    if go.mode.linkobj:
+        out_linklib = go.declare_file(go, path = source.library.importmap + ".link.a")
```
jayconrod: Ideally, we should store export data in the .x file instead of in a new file. Currently, the .a file contains export data and compiled code, and the .x file contains facts. We should make it so both .a and .x files are archives, and .a contains compiled code, and .x contains export data and facts. 

With that change, we wouldn't need to change the toolchain or provider API at all. However, nogo would need to be able to read facts from a file within an archive instead of reading the whole file (we already have code for parsing archives, so probably not too hard).

go/private/mode.bzl
```diff
@@ -125,6 +128,7 @@ def get_mode(ctx, host_only, go_toolchain, go_context_data):
         # You are not allowed to compile in race mode with pure enabled
         race = False
     debug = ctx.var["COMPILATION_MODE"] == "dbg"
+    linkobj = ctx.var.get("go.linkobj", "") == "1"
```
jayconrod: I'm not sure why plugins don't work with `-linkobj`. To answer your question about what happens within Google, I don't believe plugins are supported at all; the Bazel rules have a greater scope and feature set than the internal rules.

If there's not a simple way to fix this, I'd rather not use `--define=go.linkobj=1` though. It would be better to just test `link == LINKMODE_PLUGIN` if that's the only affected mode.

steeve: The reason it's not working is because: https://github.com/golang/go/issues/33820#issuecomment-524898207

I was using `link == LINKMODE_PLUGIN`, but it's not when building plugins, it's also when importing the `plugin` package, in regular go binaries... If any package imports `plugin`, then we can't use linkobj on any package.

So either we make it opt-in, or opt-out. There is an argument to be made about opt-out since `plugin` usage is so low, but still.

go/private/actions/archive.bzl
steeve: I don't see that `.x` file being part of the provider? You'd rather add a field?
Because that `.x` file needs to be passed to `compilepkg` and only the compiled code to the linker.

jayconrod: The .x file is already exposed as `GoArchive.data.export_file`. It doesn't look like it's documented yet, but it probably should be.

go/private/mode.bzl
jayconrod: Ah, interesting and unfortunate.

In general, I'm opposed to adding flags and additional configurations. More flags means testing is difficult, regressions are likely, and I can't afford to spend more time on support than I already do.

Maybe a workaround is possible for golang/go#33820 though? We could invoke the compiler without `-linkobj`. After it runs, the builder could copy the export data from the compiled .a file into the .x file. So the export data would appear in both archives. It's not as efficient as it could be with better toolchain support, but the I/O should be pretty cheap since the newly written .a file should still be in RAM.

Incidentally, I checked what Blaze does if a `go_binary` imports `"plugin"`: the linker just fails with the same error.

steeve:
I'm just going back to this. First, rebasing.

linzhp:
Superseded by #2461

Issue #2157, opened 2019-07-24
Cannot build go_library with bazel_skylib 0.9.0


prestonvanloon:
### What version of rules_go are you using?

0.19.1

### What version of gazelle are you using?

0.17.0

### What version of Bazel are you using?

0.28.1

### Does this issue reproduce with the latest releases of all the above?

Yes. Updated gazelle to 0.18.1 and faced same issue.

### What operating system and processor architecture are you using?

ubuntu 18. linux_amd64

### Any other potentially useful information about your toolchain?

The only thing that changed was that I updated bazel-skylib from 0.8.0 to 0.9.0.

https://github.com/bazelbuild/bazel-skylib/releases/tag/0.9.0

### What did you do?

bazel build //:go_default_library

### What did you expect to see?

Successful build

### What did you see instead?

```
ERROR: /home/preston/.cache/bazel/_bazel_preston/2bf51e8690558a83a0a0514ab422fd4b/external/io_bazel_rules_go/go/tools/coverdata/BUILD.bazel:3:1: in go_tool_library rule @io_bazel_rules_go//go/tools/coverdata:coverdata:
Traceback (most recent call last):
        File "/home/preston/.cache/bazel/_bazel_preston/2bf51e8690558a83a0a0514ab422fd4b/external/io_bazel_rules_go/go/tools/coverdata/BUILD.bazel", line 3
                go_tool_library(name = 'coverdata')
        File "/home/preston/.cache/bazel/_bazel_preston/2bf51e8690558a83a0a0514ab422fd4b/external/io_bazel_rules_go/go/private/rules/library.bzl", line 42, in _go_library_impl
                go.archive(go, source)
        File "/home/preston/.cache/bazel/_bazel_preston/2bf51e8690558a83a0a0514ab422fd4b/external/io_bazel_rules_go/go/private/actions/archive.bzl", line 218, in go.archive
                sets.union(cgo_exports, *[a.cgo_exports for a...])
        File "/home/preston/.cache/bazel/_bazel_preston/2bf51e8690558a83a0a0514ab422fd4b/external/bazel_skylib/lib/new_sets.bzl", line 184, in sets.union
                struct(_values = dicts.add(*[s._values ...]))
        File "/home/preston/.cache/bazel/_bazel_preston/2bf51e8690558a83a0a0514ab422fd4b/external/bazel_skylib/lib/new_sets.bzl", line 184, in struct
                dicts.add(*[s._values for s in args])
[...]

PR #2207, opened 2019-08-29
use skylib's old_sets.bzl to build on skylib 0.9.0

jmhodges:
Bumps skylib to 0.9.0 and uses its old_sets.bzl to maintain compatibility.

This addresses #2157, but old_sets.bzl will be removed in the next version of
skylib, and a longer term solution would be good.

Updates #2157

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 load(
-    "@bazel_skylib//lib:sets.bzl",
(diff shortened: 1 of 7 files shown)


jmhodges:
Ping!

jayconrod:
Thanks for fixing this. #2208 was very similar, so I've merged that one. I'd prefer not to actually bump the Skylib version in the same PR, since that would prevent it from being backported to release branches.

(sorry for the delayed response; I was on vacation for the past two weeks, and I'm still catching up)

PR #2240, opened 2019-10-10
Set RULES_GO_VERSION to 0.20.0

jayconrod:


jayconrod:
Test failure is caused by bazelbuild/bazel#9995. Merging, since this change is not related.

PR #2241, opened 2019-10-10
Remove support for go1.10 and update deprecation page

jayconrod:


jayconrod:
Test failures due to bazelbuild/bazel#9995. Merging since this change is not related.

PR #2243, opened 2019-10-14
Update skylib dependency to 1.0.2

jayconrod:


jayconrod:
Test failure is caused by bazelbuild/bazel#9995. Merging since this change is not related.

PR #2247, opened 2019-10-15
Fix minor bugs in bazel_tests

asuffield:
Create a query:isolate entry in the config so that bazel_test rules
can be written that call bazel query.

Excitingly, --output_base is not the outermost directory, so it ends
up with some files left in $HOME. Change to --output_user_root so that
everything goes into the temp dir.

### DIFF
--- a/tests/bazel_tests.bzl
+++ b/tests/bazel_tests.bzl
@@ -44,6 +44,8 @@ test --nocache_test_results
 
 build:isolate --
 build:fetch --fetch=True
+
+query:isolate --
 """.format(strategy = strategy)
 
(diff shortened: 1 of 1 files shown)


jayconrod:
I'd rather not make any changes to `bazel_test`. It's basically deprecated for internal testing, and it was never really intended for external testing. `go_bazel_test` in go/tools/bazel_testing should be preferred instead.

Is there some test that doesn't pass without this PR in a configuration that's important to you? Can it be migrated to `go_bazel_test` instead?

asuffield:
This is just an old local fix I was trying to get rid of - I'll take a look at migrating to `go_bazel_test` instead.

PR #2248, opened 2019-10-15
go_bazel_test: add test to ensure data and args keep working

jmhodges:


jmhodges:
Totally okay if you don't want to include this, but it's a fairly small test that hit stuff we really ran into.

jmhodges:
Okay, feedback handled. Thank you!

jayconrod:
Looks good, but seems to fail on Windows.

```
Executing tests from //tests/core/go_bazel_test:dataargtest_test
--- FAIL: TestGoldenPath (0.00s)
    dataargtest_test.go:46: unable to open Go binary file: CreateFile D:\b\26kghsey\execroot\io_bazel_rules_go\bazel-out\x64_windows-fastbuild\bin\tests\core\go_bazel_test\windows_amd64_stripped\dataargtest_test.exe.runfiles\io_bazel_rules_go\tests\core\go_binary\windows_amd64_stripped\hello.exe: The system cannot find the path specified.
FAIL
cleanup error: remove D:\b\26kghsey\bazel_testing\bazel_go_test\main: The process cannot access the file because it is being used by another process.
```

The problem is that on Windows, Bazel doesn't create symlinks to runfiles because creating symlinks requires admin privilege by default. So Bazel just gives you a manifest file you have to figure it out for yourself. [`bazel.FindBinary`](https://godoc.org/github.com/bazelbuild/rules_go/go/tools/bazel#FindBinary) and [`bazel.ListRunfiles`](https://godoc.org/github.com/bazelbuild/rules_go/go/tools/bazel#ListRunfiles) are the main cross-platform functions for handling this.

That's not what this test covers though. I'd suggest skipping Windows for this test with either `// +build !windows` in the test source file or `select` on `srcs` and `data`.

jmhodges:
Rebased and with fixed comments.

If you'd like, I could make this be an exemplary test? I'm happy to change things to Do Things Right instead of adding a misleading test.

jmhodges:
Yeah, I should do that exemplary work, because this data and args trick doesn't work unless you use `bazel.Runfile` in the `go_test` case (whereas this is the go_bazel_test case). It seems like an accident that it works here.

jmhodges:
There, better

Issue #2219, opened 2019-09-13
Use build_settings and transitions instead of aspect for cross-compilation


jayconrod:
rules_go provides two ways to cross-compile code:

1. Users may set a target platform using `--platforms` on the command line. This selects a toolchain whose "default platform" is the target platform. We declare a separate `go_toolchain` and `toolchain` for each supported platform (even though the toolchains include the same files) to make this work. This mechanism generally works well.
2. Users may set `goos`, `goarch` and other attributes on `go_binary` and `go_test`. Both rules use an aspect that creates a second graph of actions if `goos`, `goarch` and other attributes are different than the default. This mechanism the the source of a lot of problems. In particular, `select` expressions (used for platform-specific dependencies) do not work because `select` is applied before analysis.

Recent versions of Bazel support [Starlark Build Configurations](https://docs.bazel.build/versions/master/skylark/config.html), which give users and Starlark rule authors much greater control over configuration transitions. We should use this functionality to remove the aspect. We may also be able to simplify toolchain declarations.

siddharthab:
#1600 can be marked as a duplicate of this issue.

jmillikin-stripe:
https://github.com/bazelbuild/rules_go/pull/2249 is related -- i want to use build settings to avoid the architecture-specific path component of `go_binary` outputs.

PR #2249, opened 2019-10-17
Add a build setting to disable architecture-specific binary output paths.

jmillikin-stripe:
This is another attempt to fix https://github.com/bazelbuild/rules_go/issues/1239 (`go_binary` output paths containing architecture-specific components). See https://github.com/bazelbuild/rules_go/pull/1257 for a previous approach that wasn't accepted.

This time I'm using the "Starlark build settings" feature of newer Bazels (https://docs.bazel.build/versions/1.0.0/skylark/config.html), which lets us cleanly plumb the setting from a command-line flag to `GoContext`. The end result:

```
$ bazel build //tests/core/go_binary:hello
[...]
Target //tests/core/go_binary:hello up-to-date:
  bazel-bin/tests/core/go_binary/darwin_amd64_stripped/hello

$ bazel build //tests/core/go_binary:hello --@io_bazel_rules_go//go/build_settings:per_mode_binaries=false
Target //tests/core/go_binary:hello up-to-date:
  bazel-bin/tests/core/go_binary/hello
```

In addition to being set by a flag, Bazel has an in-development mechanism ("configuration transitions") for overriding this on a per-target basis, and having it propagate through the dependency graph. So a target that built a deployable container image could use this to ensure its contained binaries have consistent paths regardless of whether they're written in C++ or Go.

Note: Bazel's option parser doesn't accept flags for build settings defined in external repositories (https://github.com/bazelbuild/bazel/issues/10039). The fix is https://github.com/bazelbuild/bazel/pull/10052 and it should hopefully land in v1.2.

### DIFF
--- a/go/build_settings/BUILD.bazel
+++ b/go/build_settings/BUILD.bazel
@@ -1,0 +1,27 @@
+# Copyright 2019 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load(
+    "@bazel_skylib//rules:common_settings.bzl",
+    "bool_flag",
+)
+
+# If true (the default), go_binary output paths will include the build mode
+# and target architecture. If false, binary output paths will match the
+# rule name.
+bool_flag(
+    name = "per_mode_binaries",
+    build_setting_default = True,
+    visibility = ["//visibility:public"],
+)
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -67,6 +67,10 @@ load(
     "@bazel_skylib//lib:paths.bzl",
     "paths",
 )
+load(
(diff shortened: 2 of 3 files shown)


jayconrod:
We're planning to eventually migrate all parts of the build configuration to build settings (#2219). I'd prefer to have a complete plan for that migration before adding any build settings. Could you comment on that issue to make sure this is investigated?

jmillikin-stripe:
> We're planning to eventually migrate all parts of the build configuration to build settings (#2219). I'd prefer to have a complete plan for that migration before adding any build settings. Could you comment on that issue to make sure this is investigated?

Refactoring the entire cross-compilation system to use build settings instead of platforms seems like a big project. Would it be possible to land this without blocking on it?

I'd be fine with naming it `experimental_`, then you can feel free to change/move/remove it in the future.

jayconrod:
> Refactoring the entire cross-compilation system to use build settings instead of platforms seems like a big project. Would it be possible to land this without blocking on it?

I'd prefer not to merge this. In general, I'd like to have as few settings as possible, since each one adds a dimension to the space of configurations that need to be tested. A large fraction of rules_go bugs are in under-tested combinations of configurations, so I don't want to expand that space further unless a new configuration provides a lot of value.

As part of the design for build setting migration, I want to understand how they affect output paths and caching. I remember at some point a hash of the configuration was part of the output path. If that's still true, we won't need to encode the Go configuration in the output path, and we won't need a setting like this at all.

> I'd be fine with naming it experimental_, then you can feel free to change/move/remove it in the future.

I think the best way to maintain this might be as a patch applied on the `http_archive` rule for `io_bazel_rules_go`.

Experimental features still have a maintenance cost in testing and avoiding unintentional changes when the surrounding code.

Issue #512, opened 2017-06-06
Expose generated Go files to editors and IDEs


jayconrod:
When a library is created with `go_proto_library` or some other mechanism that generates Go code (.pb.go files in this case), the generated sources files are not exposed to editors and IDEs. Consequently, code completion and refactoring tools don't work, lowering developer productivity.

The Go team is developing a workspace abstraction mechanism that will allow editors and Bazel (and other build systems) to communicate with each other without the need for direct integration. This is the long-term solution to this problem.

This issue will track progress on the workspace abstraction and will track editor awareness of code generated with Bazel.

jayconrod:
This issue is in response to [Bazel build, protobuf and code completion](https://stackoverflow.com/q/44375665/1891).

TODO: update answer when this issue is resolved.

zamotivator:
What the status of the task?

zamotivator:
So, I have proposal how to close this issue.

Suppose you generated code:
bazel-genfiles/somedir/somefile.go
bazel-genfiles/somedir/anotherfile.go

Let's add symlink
src/somedir => ../bazel-genfiles/somedir

Use in your files
`import "somedir"`

As result you will get workable code-completition AND workable build.

jayconrod:
@excavador We're still developing a more complete solution to this (the workspace abstraction I mentioned above). Ideally, tools won't need to be aware of the project layout in use. I don't have any ETA for that yet, but we are making progress.

As you showed, it's possible to work around this with symlinks. You can also copy or symlink generated files into your source directory. Bazel will prefer a generated file if a source file of the same name is present. I'd recommend against checking generated files into VCS though.

excavador:
Thank you for notify! Will wait. You can consider me as closed beta-tester :)

raliste:
Any updates?

excavador:
@raliste it works for me in IDEA Ultimate + Bazel plugin with following limitation: autocomplete works only if all files inside golang package are auto-generated. If mix together auto-generated and manually written file, autogenerated file would not be inspected

jayconrod:
The workspace abstraction mentioned above is being actively worked on. We've gone through a lot of internal prototypes, and we're in the process of reviewing, testing, and merging one that we're pretty happy with. It's a huge body of code though, and it will take time. Following that, a large number of Go tools will need to be updated to use the workspace abstraction so they can function in "go build", vgo, Bazel, and other kinds of workspaces.

[...]

PR #2261, opened 2019-10-25
Adding GoAspectProviders to go_test

linzhp:
This is an alternative implementation to #2260 to provide `srcs` and `importpath` to aspects of `go_test` targets.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -37,6 +37,7 @@ load(
     "GoLibrary",
     "INFERRED_PATH",
     "get_archive",
+    "GoAspectProviders",
 )
 load(
     "@io_bazel_rules_go//go/private:rules/aspect.bzl",
@@ -173,6 +174,9 @@ def _go_test_impl(ctx):
                 files = depset([executable]),
                 runfiles = runfiles,
                 executable = executable,
+            ),
+            GoAspectProviders(
+                archive = internal_archive,
             ),
             OutputGroupInfo(
                 compilation_outputs = [internal_archive.data.file],


jayconrod:
I don't understand the problem this solves. Could you open an issue and explain?

FYI, `GoAspectProviders` is not part of the public API of rules_go, and I intend to remove it (and the cross-compiling aspect) in the future.

linzhp:
Before #512 is solved, we have an aspect to identify generated sources in `srcs` of go_library, go_binary, go_test etc, and copy them to `vendor` so IDEs can find them. In the aspect, we extract `srcs`, `importpath` from available providers such as GoArchive and GoAspectProviders, so we know where the generated code is, and which directory under `vendor` it should go to.

If `GoAspectProviders` is going to be removed, then we can instead extract that information from GoLibrary and GoSource if you are OK with #2260

jayconrod:
This sounds similar to [`go_path`](https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#go-path). Could you use that?

If not, can you use the same mechanism that `go_path` uses? `go_test` already provides `GoArchive`. For any given [`GoArchive`](https://github.com/bazelbuild/rules_go/blob/master/go/providers.rst#goarchive), you can get the corresponding `GoSource` via the `source` field, and the `GoLibrary` via the `library` field of `GoSource`.

linzhp:
The GoArchive provider go_test returns is the testmain binary. The orig_srcs is `testmain.go` and the import path is `testmain`. This PR tries to expose `internal_archive` to the aspect, beside the `test_archive`.

jayconrod:
How about the `transitive` field of `GoArchive`? That is a `depset` of `GoArchiveData` objects. It includes data for the archive itself and everything it transitively depends on. For a binary (or a test), that corresponds to every archive the linker would care about. Each [`GoArchiveData`](https://github.com/bazelbuild/rules_go/blob/master/go/providers.rst#goarchivedata) object has a `srcs` field, which is what `go_path` uses.

linzhp:
Since `transitive` is a set of `GoArchiveData` objects, how do I find which one is from the `internal_archive`?

I had a similar implementation previous, where I did `test_target[GoArchive].source.deps[-1].source.deps[-1]`. It's relying on the assumption that the `external_archive` is the last element in `deps` of the testmain binary, and the `internal_archive` is the last element in `deps` of the `external_archive`. Future changes to the `go_test` implementation could accidentally break this assumption.

jayconrod:
Agree that depset iteration order is probably not stable.

To match internal archives specifically, you could look for archives that contain source files ending with `_test.go` with import paths that do *not* end with `_test`. That assumes that people haven't intentionally mixed `_test.go` into libraries.

There's also a `GoLibrary` field, `testfilter` that is set to `"exclude"` for internal archives, but that's not public and could change.

Why are internal archives needed specifically though? Since you wanted to use an aspect, I assumed you needed all source from all transitively imported archives.

linzhp:
> Why are internal archives needed specifically though? Since you wanted to use an aspect, I assumed you needed all source from all transitively imported archives.

Yes, we need all transitive imported archives, but this is done recursively, which the aspect of each target writing to its own manifest file according to the information of that target. So for a `go_test` target, we need to know the generated file in the `srcs` and its `importpath`. The internal archive has the correct `importpath`.

> To match internal archives specifically, you could look for archives that contain source files ending with _test.go with import paths that do not end with _test. That assumes that people haven't intentionally mixed _test.go into libraries.

This would match both the external and the internal archives, but the external archive doesn't have the correct import path. So we have to do some hacking to distinguish between the two. Maybe checking if the import path ends with `_test` to determine if it's an external archive?

jayconrod:
> Maybe checking if the import path ends with _test to determine if it's an external archive?

That should work.

However, if you're running an aspect that creates a manifest file for each target, in the action that creates that file, you can read the source files and check whether the package name ends with `_test` or not. That's what `GoCompilePkg` and `GoTestGenTest` do to split the two packages.

I'm curious by the way what kind of generated files you're filtering. Is this the generated test main? Or generated cgo / coverage files? Or something else?

linzhp:
> I'm curious by the way what kind of generated files you're filtering. Is this the generated test main? Or generated cgo / coverage files? Or something else?

I am only interested in generated files in `srcs` of go_library, go_binary and go_test. It is easy to get the `srcs` of go_library and go_binary from their aspects, but it turns out difficult to get `srcs` of go_test. The orig_srcs of go_test archive gives me the testmain.go, instead of the list appears in the BUILD.bazel file.

I am a bit concerned about the performance of iterating through all `srcs` of all transitive dependency of the `test_archive` just to find the `internal_archive`. Would #2260 be a better alternative?

jayconrod:
So theoretically at least, the testmain archive should only directly depend on the internal and external test archives. So you could take the `GoArchive` returned from `go_test` and look at the `direct` field for those two `GoArchives`.

I say theoretically because looking at the code, that may not be true: in go/private/rules/test.bzl, it looks like the external archive is the only test dependency. I'd be happy to take a PR to fix that.

I don't think #2260 or this PR should be merged though. `GoLibrary` and `GoSource` are primarily used for embedding, and a `go_test` shouldn't be embedded (people have tried!). And I'd rather not extend use of `GoAspectProviders` because I'd like to remove it soon.

PR #2264, opened 2019-10-28
delete unused cgoarc flag from compilepkg builder

jmhodges:
I came across this while poking around compilepkg to learn how to pull out the cgo-generated Go files from it for #512 and I might not have context on what its for. It's unused, but perhaps there was some expectation of future use?


jmhodges:
Similarly, I'm not seeing a place where the `-cgoexport` flag is set and so _cgo_exports.h isn't ever created. Is that also a future thing or can I post it for removal?

jmhodges:
Ah, thank you for the info!

PR #2271, opened 2019-11-02
add cgo-generated source files to GoArchiveData

jmhodges:
This adds the Go source file we create with cgo to the `srcs` field of
GoArchiveData. This is useful in AST and IDE tooling as discussed in #512.

Updates #512.


jmhodges:
There's one lame hack in there: the code creates empty files to satisfy the outputs requirement when compilepkg exits early successfully even though cgo was expected to be used.

An alternative would be to use an output directory, maybe? Let me know which way you want this to go

jmhodges:
I was off handling other things, but I'll take a look at getting that feedback done!

jmhodges:
Okay, I've run into something confusing and I feel I'm missing something.

The cgo-generated `*.cgo1.go` files are intended to be replacements of the specific source file they were each generated from. But in `cgo2`, those files (in `genGoSrcs) are appended to the list of go source files and then later in `compilepkg.go` we pass both the original and the cgo-generated files to things like `runNogo` and `compileGo` and they seem to work. 

But that would include duplicate definitions for any code in a file with an `import "C"`. I must have missed the filtering step for these. Do you know where that filtering occurs?

(This is relevant because the docs for `go list` and `go/packages.Package`, the CompiledFiles field contains the exact files that were passed to `go compile`.  Looking at the output of `go list -compiled` for packages that use cgo, `go list` clearly keeps the cgo-generated versions of the files plus those sources files that don't import "C". So I'm trying to keep that behavior up for #512 and ran into this)

jmhodges:
Oh, right, right, cgoSrcs contains the import "C" files.

jmhodges:
Writing down a fun lil extra wrinkle so I don't forget:

To work in #512, we need the cgo-generated `//line` directives to have working paths to the original cgo source files (because of uses like in the [cgocall analysis](https://github.com/golang/tools/blob/76a3b8da50ef7c9da5c87b2fdff51392aad781a0/go/analysis/passes/cgocall/cgocall.go#L180-L182))

jayconrod:
> To work in #512, we'd like the cgo-generated //line directives to have working paths to the original cgo source files (because of uses like in the cgocall analysis)

That doesn't seem like it would be possible, since we'll only see sandbox paths in execution. Are workspace-relative paths good enough? That might not work for generated files though.

Maybe the driver can fix these paths up when it's post-processing package JSON? But then it would need to copy the files somewhere which would cause even more problems.

jmhodges:
Well, the driver returns absolute file paths -- like the `go list` default one in `go/packages` -- while the line directives created by cgo are relative. So that relative to absolute mismatch should, in some way, be handled somewhere or else that code seems like it wouldn't ever work.  I'm not sure from context how this code handles it, of course, and it might do so with by changing directory to the PkgPath (which doesn't work in bazel) which would wreck our ideas. However, that would effect every line directive (not just cgo ones) and seems like a bug they'd have to address.

I'll have to look into what the behavior is.

In any case, "we'll fix it in post" might have to be the answer here. At the very least, getting some forward motion and re-adjusting as we close in on full #512 completion is probably the better idea.

jmhodges:
(Oh, and the cgo-generated files made by `go` are in different directories than the files the line directives point to. That's the other half of why the `go/packages` analysis code has to be handling this somewhere)

jmhodges:
Oh, right, declare_directory/Args.add_all doesn't work like I thought it did. We want each cgo-generated file listed out explicitly in the providers.

In order to get that, the options I've come up with all add another step to the compile: 1) emit a zip file and manifest and add a call to bazel's `zipper` as a separate step in `emit_archive`, or 2) add a step before `builders compilepkg` to get a manifest of all the files `compilepkg` would generate (or just generate them, but that idea was discarded already). 

We've been trying to avoid more steps in the compile path, so do you have an alternate idea?

jmhodges:
Ugh, no, the manifest idea won't work, of course, because there's no way to read a file in skylark.

I'm sure I'm missing some simple API call (that's maybe poorly documented).

jmhodges:
Okay, okay, what if we declare a directory for all compilepkg generated files (not the coverage ones, just cgo for now), and add that directory to the GoArchiveData as a separate field like `generated_src_dir`?

That would mean the gopackagesdriver and other tools would know where to look for generated files even though they don't specifically know what's in there. To get the cgo-generated files, specifically, the tools could do some file name pattern matching to "get it right". Those tools would have to know how cgo names files it generates, but that might be worth the tradeoff.

Maybe there's a better idea kicking around.

jmhodges:
I implemented a prototype of that new field on GoArchiveData. My version fails the //tests/reproducibility test, though (that thing rules, btw). That's because in c-archive mode, one of the generated C files contains a full path to a place in the work directory that's deleted when the rule finishes.

In the test, it's the adder.cgo2.c file, and the line that breaks thing is:

```
#line 3 "/private/var/tmp/_bazel_jmhodges/3d68d33e850dbf6fe19f5c927ef82602/sandbox/darwin-sandbox/4698/execroot/io_bazel_rules_go/tests/reproducibility/cgo/adder.go"
```

I looked but didn't see a flag that manages that in `go tool cgo` (I could be wrong!). 

What's the precedent for handling this kind of "unfixable" reproducibility problem? (sed?)

jayconrod:
A generated directory might be the way to go. When output directories were first introduced in Bazel, there were a lot of problems, but I think those must have been ironed out by now. Still, not being able to explicitly list the generated files in `GoArchiveData` is unfortunate. I wonder if it makes more sense to declare a `.cgo1.go` file for each source file, regardless of whether it contains cgo code? If the `GoCompilePkg` action finds that it does not contain `import "C"`, that file could simply be a copy of the original.

About the reproducibility test, I think it's fine to trim the path in the output file. `sed` may not be available, but it's easy to do the same thing in Go.

Globegitter:
@jmhodges have you been able to make any progress on this?

jayconrod:
Closing inactive PRs.

This will still be needed for #512 if someone tackles that in the future though.

PR #2275, opened 2019-11-06
go_compat_repo: alias constraint values and settings to bazel_tools or platforms

jayconrod:
os and cpu constraint settings and values are being removed from
@bazel_tools//platforms. They are currently available in the external
@platforms repo (now declared in go_rules_dependencies). So we should
reference those instead of @bazel_tools//platforms.

In order to support old Bazel versions (the current minimum is still
0.23.0), @io_bazel_rules_go_compat//platforms now declares aliases on
either @platforms or @bazel_tools//platforms.

Note that bazel-skylib 1.0.2 (the latest released version) is not
compatible with --incompatible_use_platforms_repo_for_constraints yet,
so not all rules_go tests pass with that flag yet.

Updates bazelbuild/bazel#8622

### DIFF
--- a/go/platform/list.bzl
+++ b/go/platform/list.bzl
@@ -59,14 +59,14 @@ def declare_config_settings():
     native.config_setting(
         name = "ios",
         constraint_values = [
-            "@bazel_tools//platforms:ios",
+            "@io_bazel_rules_go_compat//platforms:ios",
         ],
     )
     for goarch in ("arm", "arm64", "386", "amd64"):
         native.config_setting(
             name = "ios_" + goarch,
             constraint_values = [
-                "@bazel_tools//platforms:ios",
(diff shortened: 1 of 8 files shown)


jayconrod:
@hlopko I took a stab at this. Rather than declaring functions in the compat repo, I'm defining aliases in that repo, and anything that needs to refer to a standard `constraint_setting` or `constraint_value` will refer to an alias in there instead.

Seems to work with Bazel 0.23.0 and 1.1.0 with `--incompatible_use_platforms_repo_for_constraints`. I had to comment out some stuff that depended on the Skylib unittest toolchain, which is not compatible yet. Looks like you've already fixed that in bazelbuild/bazel-skylib#214, but it's not released yet.

hlopko:
Wow this is great, thanks Jay!

PR #2280, opened 2019-11-15
Add support for dirs in txtar expander

achew22:


go/tools/bazel_testing/bazel_testing.go
```diff
@@ -343,6 +343,11 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
 func extractTxtar(dir, txt string) error {
 	ar := txtar.Parse([]byte(txt))
 	for _, f := range ar.Files {
+		if parentDir := filepath.Dir(f.Name); parentDir != "." {
+			if err := os.MkdirAll(filepath.Join(dir, parentDir), 0755); err != nil {
```
jayconrod: Nit: change mode to `0777`. As a general rule, we should let the user's `umask` remove permission bits instead of doing it ourselves.

achew22:
PTAL

Issue #2187, opened 2019-08-20
go_path should preserve path of data files


duarten:
### What version of rules_go are you using?

0.19.1

### What version of gazelle are you using?

0.18.1

### What version of Bazel are you using?

0.28.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Reproduces in darwin and linux, both amd64.

### What did you do?

Consider the following rule in `BUILD.bazel`:

```python
load("@io_bazel_rules_go//go:def.bzl", "go_path")

go_path(
	name = "gqlgen_path",
	mode = "copy",
	deps = ["@com_github_99designs_gqlgen//:gqlgen"],
	data = ["@com_github_99designs_gqlgen//plugin/modelgen:models.gotpl"],
	visibility = ["//visibility:public"],
)
```

### What did you expect to see?

I would expect to see the `models.gotpl` file under the directory `bazel-bin/gqlgen_path/src/github.com/99designs/gqlgen/plugin/modelgen/`.

### What did you see instead?

I see that file in the root directory `bazel-bin/gqlgen_path/models.gotpl`. This file is needed during the execution of gqlgen, which expects to find it in that particular location.

This is the `WORKSPACE`:

```python
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

rules_go_version = "0.19.1"

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "8df59f11fb697743cbb3f26cfb8750395f30471e9eabde0d174c3aebc7a1cd39",
    urls = [
        "https://storage.googleapis.com/bazel-mirror/github.com/bazelbuild/rules_go/releases/download/{}/rules_go-{}.tar.gz".format(rules_go_version, rules_go_version),
        "https://github.com/bazelbuild/rules_go/releases/download/{}/rules_go-{}.tar.gz".format(rules_go_version, rules_go_version),
    ],
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")

go_rules_dependencies()

go_register_toolchains()

gazelle_version = "0.18.1"

http_archive(
    name = "bazel_gazelle",
    sha256 = "be9296bfd64882e3c08e3283c58fcb461fa6dd3c171764fcc4cf322f60615a9b",
    urls = [
[...]

Issue #1872, opened 2018-12-26
go_path rule fails due to cgo in go-libsass


robfig:
I have the https://github.com/wellington/go-libsass library working with both Bazel and the Go command line tool, but the `go_path` rule fails to result in a working GOPATH. Since go_path is used for code generation rules, any library that depends on go-libsass can not be used with e.g. gomock or wire.

I have a simple reproducer repo, which imports [my fork of go-libsass]:
https://github.com/robfig/bazel-libsass

bazel-libsass shows that go-libsass works both from Bazel and go:
```
$ go run github.com/robfig/bazel-libsass
$ bazel run :bazel-libsass
```

To see the go_path rule fail:
```go
bazel build :gopath
cd bazel-bin/gopath
GOPATH=`pwd` go run github.com/robfig/bazel-libsass
```

Results in many errors about undefined symbols:
```
Undefined symbols for architecture x86_64:
  "_libsass_version", referenced from:
      __cgo_a72fb0af35f5_Cfunc_libsass_version in _x006.o
     (maybe you meant: __cgo_a72fb0af35f5_Cfunc_libsass_version)
  "_sass2scss", referenced from:
      __cgo_a72fb0af35f5_Cfunc_sass2scss in _x008.o
     (maybe you meant: __cgo_a72fb0af35f5_Cfunc_sass2scss)
....
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The contents of the GOPATH look pretty reasonable, including .so files for the cgo:

```
$ find bazel-bin/gopath
bazel-bin/gopath
bazel-bin/gopath/src
bazel-bin/gopath/src/golang.org
bazel-bin/gopath/src/golang.org/x
bazel-bin/gopath/src/golang.org/x/net
bazel-bin/gopath/src/golang.org/x/net/context
bazel-bin/gopath/src/golang.org/x/net/context/pre_go17.go
bazel-bin/gopath/src/golang.org/x/net/context/go19.go
bazel-bin/gopath/src/golang.org/x/net/context/pre_go19.go
bazel-bin/gopath/src/golang.org/x/net/context/go17.go
bazel-bin/gopath/src/golang.org/x/net/context/context.go
bazel-bin/gopath/src/github.com
bazel-bin/gopath/src/github.com/robfig
bazel-bin/gopath/src/github.com/robfig/bazel-libsass
[...]

Issue #393, opened 2017-04-19
Integration with Go editor tooling


nelhage-stripe:
I'm not sure if this is a feature request or just a request for thoughts, but are there any recommended techniques for making `rules_go`-built repositories play nice with Go editor tooling like `gocode` and `godefs` and `goimports`? I check out my repository into the place it *would* go on `GOPATH`, which works well for many purposes, but if I generate code using bazel (e.g. to compile protobufs), `goimports` and `godef` can't see it. `goimports` even sometimes gets confused and tries to insert a `.../bazel-bin/...` import.

Are there any thoughts or best practices about how to make this experience better? Do we need to write bazel-aware versions of all these tools? Should we perhaps add a rule to create a symlink-farm pseudo-GOPATH inside `bazel-bin` we can point editor tooling at?

alandonovan:
My team is actively working to solve the problem of how to build source code tools that work unmodified in any workspace---go build, bazel, or blaze.  We hope to share some prototype tools, an API, and a proposal document later this quarter.

cc: @matloob

kgreenek:
Major +1 for this.
In addition, IDE's don't know how to handle .go files generated from protobufs. Our current workaround is to run a script that copies the generated *.pb.go files from the gen dir to the proto directory in source.

kerinin:
:+1: for a "preferred" method of exposing generated code to the standard toolchain. It would be nice if similar code-generation tools like `gomock` could be used.

toffaletti:
A command or some manual steps to get a GOPATH would be good enough for now. Not having anything is really painful right now.

nelhage-stripe:
I briefly experimented with a proof-of-concept aspect that can generate a `GOPATH` symlink farm from a `rules_go` target, but I abandoned it because most of the tools I looked at didn't like following symlinks, so it didn't help that much. It might be an approach worth pursing, though. I can post the code somewhere if someone'd find it useful.

micanzhang:
is it possible write a command tool which generate go source files and copy these go files  into GOPATH? or copy to ./vendor directory?

jayconrod:
There is an experimental rule that does this in [path.bzl](https://github.com/bazelbuild/rules_go/blob/34e7b12c2a76c80d65b4cd152cc2b0e5a69a0217/go/private/tools/path.bzl). You can use this as:

```
load("@io_bazel_rules_go//go:def.bzl", "go_path")

go_path(
    name = "path",
    deps = [":go_default_library"],
[...]

PR #2283, opened 2019-11-17
go_path: maintain subdirectory structure for data and cgo srcs when contained in a package.

### LINKED ISSUE #1916: Unexpected file collision when generating go_path
While writing some rules for myself on top of rules_go I came across, what I suspect is, a bug in the implementation of `_go_path_impl` in [`path.bzl`](https://github.com/bazelbuild/rules_go/blob/72dff4e33d24aac4953ee794815dc569af0d8093/go/private/tools/path.bzl#L32).

It appears that further down in the function we have [a code-block](https://github.com/bazelbuild/rules_go/blob/72dff4e33d24aac4953ee794815dc569af0d8093/go/private/tools/path.bzl#L67-L104) that creates the map of all files to link and which at the same time checks for the uniqueness of the providers for each individual file in the resulting directory tree. More specifically for data sections [we use](https://github.com/bazelbuild/rules_go/blob/72dff4e33d24aac4953ee794815dc569af0d8093/go/private/tools/path.bzl#L82-L91):
```python
# [...]
    if ctx.attr.include_data:
        for pkg in pkg_map.values():
            for f in pkg.data:
                parts = f.path.split("/")
                if "testdata" in parts:
                    i = parts.index("testdata")
                    dst = pkg.dir + "/" + "/".join(parts[i:])
                else:
# [...]
```
As far as my reading goes this works in all usual cases where people use `testdata`. However if one were to depend on a (non-`testdata`) `filegroup` rule that uses a glob of type `glob(['foo/**']` then this will break if the underlying globbed sub-directories contain two files with the same name.

Will be creating a PR with a proposed fix but am not entirely sure what style of fix would be preferred in-line with the rest of the rules' implementation style.

gnattishness:
I understand (as in #1449) that there is no general solution for translating data deps and cgo src paths in directories other than the package directory. However, it seems safe to maintain a subdirectory structure when the paths are children of the package path.

As discussed in #1872. Fixes #1916 for data contained in the pkg.

This maintains relative cgo `#include`s and reduces go_path failures due to clashing dst paths.

Maintains previous "flattening" behavior for other cases (espectially for `.go` files where they need to be in the main directory)

----

This is quite useful when using go_path with dependencies that wrap c libraries.

Somewhat based on @Helcaraxan's Helcaraxan/rules_go@74301f5 - thanks!

Other related issues I'd found:
#2187, #393

### DIFF
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -13,6 +13,14 @@
 # limitations under the License.
 
 load(
+    "@bazel_skylib//lib:paths.bzl",
+    "paths",
+)
+load(
+    "@bazel_skylib//lib:new_sets.bzl",
+    "sets",
+)
+load(
     "@io_bazel_rules_go//go/private:context.bzl",
     "go_context",
 )
@@ -58,6 +66,7 @@ def _go_path_impl(ctx):
                 srcs = as_list(archive.orig_srcs),
                 data = as_list(archive.data_files),
                 pkgs = {mode: archive.file},
+                labels = [archive.label],
             )
             if pkgpath in pkg_map:
                 _merge_pkg(pkg_map[pkgpath], pkg)
@@ -70,7 +79,14 @@ def _go_path_impl(ctx):
     manifest_entry_map = {}
     for pkg in pkg_map.values():
         for f in pkg.srcs:
-            dst = pkg.dir + "/" + f.basename
+            if f.extension in ("c", "h", "s", "S") and _is_in_pkg_subdir(pkg, f):
+                # NOTE assumes cgo=true for the pkg
+                # Maintain position of C files relative to package,
+                # if contained in a subdirectory of the pkg
+                dst = paths.normalize(paths.join(pkg.dir, _path_rel_to_pkg(pkg, f)))
+            else:
+                # Flatten into the package
+                dst = paths.join(pkg.dir, f.basename)
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
     if ctx.attr.include_pkg:
         for pkg in pkg_map.values():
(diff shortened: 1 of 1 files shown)


gnattishness:
Let me know if this seems reasonable, and I'll flesh out the tests.

I'm in the process of sorting out the CLA for my company.

gnattishness:
Hi @jayconrod, thanks for the quick response and sorry for taking a little while to get back to you.

I agree with you regarding .c, .cpp etc file dependencies, that cgo would  not compile them unless in the same directory. However, relative path `#include`s can and do work. From my understanding, the recommendation for header files to be in the same directory is to ensure recompilation upon modification of the .h files, but external and relative includes do work.

> [Note that changes to files in other directories do not cause the package to be recompiled, so all non-Go source code for the package should be stored in the package directory, not in subdirectories.](https://golang.org/cmd/cgo/#hdr-Using_cgo_with_the_go_command)

A cc_library with cdeps could be viable, but then we'd need to solve the problem of integrating those into a `go_path` rule and GOPATH-style directory, which doesn't appear to have a simple, general solution.

**Possible adjustments:**

- Adding a boolean flag to to enable/disable behavior for source files would be reasonable?
- Enable the behavior for only header files?

**Re the use case:**
Exporting an existing bazel-based project (and its dependencies) into a GOPATH structure for integration with other tooling.
One of the project's external dependencies is a go wrapper around a C library. For that particular case, the dependency is already in a GOPATH/cgo-suitable structure, so copying it verbatim would work! (but is not a general solution)

The library in question:
https://github.com/ethereum/go-ethereum/blob/master/crypto/secp256k1/secp256.go

For this, the current `go_path` flattening breaks #include references, which the PR resolves.

jayconrod:
Including headers outside of the package may work, but it's not supported. The C compiler may be able to find those headers, but `go list` and tools based on it won't. The Go command also won't attempt to compile any C files in subdirectories that aren't Go packages, so this ends up being very brittle. The advice in the cmd/cgo documentation is understated; headers should really be in the package directory. Structure and reproducibility are very important.

I'd rather not merge this functionality into `go_path`. The objective of `go_path` is to map a set of Go packages into a set of the directories that the `go` command will understand. The `go` command doesn't understand C headers and sources in subdirectories, so `go_path` shouldn't place them there.

I don't mean to say this doesn't work in your project of course. But you may be better off forking `go_path` and adding this.

PR #2286, opened 2019-11-19
Update go_toolchain to use execution transitons.

katre:


jayconrod:
Is `cfg = "host"` being renamed to `cfg = "exec"`? What is the minimum Bazel version that supports that?

The current minimum Bazel version for rules_go is `0.23.0`, but we'll make `1.1.0` the minimum somewhat soon.

katre:
Host and exec behave differently. Host is the current host transition: there's only one, it depends on the default flag values, and is set for building tools to be invoked on Host. The [execution transition](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-execution-transitions.md), on the other hand, allows for multiple different execution platforms and is more flexible.

Eventually I want to move everything to use the execution transition and get rid of the host, but that's not feasible anytime soon.

The execution transition was added before 1.0, but I would need to search to see what version exactly.

This draft PR was just created to run the tests and see if there are obvious failures. Do you have a way to run downstream projects with a release candidate?

jayconrod:
Ah, interesting. My understanding was that `cfg = "host"` actually meant the execution platform (a misnomer from way back). So anywhere that appears in rules_go, `cfg = "exec"` is what's actually intended.

> This draft PR was just created to run the tests and see if there are obvious failures. Do you have a way to run downstream projects with a release candidate?

I'm afraid not. Kubernetes is the biggest open source client but they're not using the most recent version. Most other clients are closed-source.

katre:
Thanks for the additional commit. If you're fine with this change let's go ahead and commit (it should be a no-op until the actual toolchain transition changes land).

katre:
Is there a way to kick off tests again?

jayconrod:
Ran the tests again through BuildKite. Looks good. Thanks for sending this.

(I resurrected this since the rules_go minimum version of Bazel is now 1.2.0).

steeve:
For the record, I ran into problems with the `exec` transition on my custom rules (it would fail to detect Android toolchain).
Not sure how the reason, I'll try again and report back here if I run into issues.

jayconrod:
@steeve Interesting, please do open an issue if you're able to reproduce that.

I'm hoping to get a rules_go release out today with this, but we can do a point release if it causes problems.

katre:
If you see an exec transition failure, please open a bug and let me know.

The change here is a no-op: toolchains are always in the host transition currently (until https://github.com/bazelbuild/bazel/issues/10523 is done), and the host transition is terminal, so these cfg changes don't change anything.

steeve:
Reviving this issue.

Having a `go_binary -> cc_import -> android_binary` breaks with `cfg = "exec"` with the following:
```
ERROR: /private/var/tmp/_bazel_steeve/8c5731f383ec5885e1a8afb0f4a77ff6/external/androidndk/BUILD.bazel:41:1: in cc_toolchain_suite rule @androidndk//:toolchain-libcpp: cc_toolchain_suite '@androidndk//:toolchain-libcpp' does not contain a toolchain for cpu 'darwin'
```

@katre do you want me to open an issue on bazel itself ?

katre:
@steeve Can you please file an issue on Bazel with detailed reproduction steps? I'll look at is as soon as I can.

steeve:
Sure thing!

PR #2291, opened 2019-11-22
Clear GOPATH environment variable in builder

andrew-d:
I'm still not 100% clear on the exact interaction, but this was precipitated by trying to do a Bazel build of something with a `host` Go toolchain that was on a read-only mount. When doing so, builds would fail with an error trying to `go install vendor/golang.org/x/net/nettest`, since it couldn't open the `nettest.a` file in the host Go toolchain for writing. My strong suspicion here is that the `stdlib` builder isn't setting a `GOPATH`, so it's trying to write to the file in `GOROOT`, which it can't actually do since that's read-only.

Either way, this fixes things and feels pretty low-impact.

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -383,6 +383,14 @@ def go_context(ctx, attr = None):
         "GOROOT": goroot,
         "GOROOT_FINAL": "GOROOT",
         "CGO_ENABLED": "0" if mode.pure else "1",
+
+        # If we use --action_env=GOPATH, or in other cases where environment
+        # variables are passed through to this builder, the SDK build will try
+        # to write to that GOPATH (e.g. for x/net/nettest). This will fail if
+        # the GOPATH is on a read-only mount, and is generally a bad idea.
+        # Explicitly clear this environment variable to ensure that doesn't
+        # happen. See #2291 for more information.
+        "GOPATH": "",
     })
 
     # TODO(jayconrod): remove this. It's way too broad. Everything should


jayconrod:
Could you open an issue instead with enough information to reproduce the problem? This doesn't seem like the right solution, but I'm not sure why the problem is happening. `GOPATH` should not be needed to rebuild the standard library. Nothing should be installed there.

andrew-d:
Okay, this turned out to be _way_ more complicated than I'd expected. Here's a summary, and you can decide whether or not this warrants fixing or not :sweat_smile: 

* NixOS patches `bazel` to avoid clearing process environments ([here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/build-managers/bazel/nix-hacks.patch#L40)) when running a build (this is to ensure that compiler wrappers work; it doesn't cause reproducibility problems since Nix packages are built with a deterministic environment).
* When building `gvisor` for NixOS, I patched the `WORKSPACE` file to use the host's Go toolchain, instead of downloading one, since the downloaded SDKs won't run on NixOS due to libc being in a different location.
* When building with the Go package as an input, the Nix tooling sets `GOPATH` to a value in the read-only Nix store [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/go/setup-hook.sh#L2).
* As a result, the `GOPATH` environment variable is passed all the way to this script here. In the case where the `x/net/nettest` package is being built, the Go tooling then tries to install that into the "existing `GOPATH`", which is mounted read-only, and thus fails.

This behavior would theoretically also be reproducible if using `--action_env=GOPATH=$GOPATH` and where that `GOPATH` value is mounted read-only, though I can't think of a reason that I'd want to do that.

Either way, I solved this in my Nix build by just clearing `GOPATH` prior to calling Bazel. For `rules_go`, my suggestion is that perhaps, instead of setting up a `GOPATH`, it'd be worth explicitly clearing it instead, to ensure that any provided value (through Nix, `action_env`, or otherwise) doesn't affect the build.

Thoughts on this?

jayconrod:
Wow, that's pretty far off the beaten path. I can't really support this, but I'd accept a simple fix. No promises about it being stable in the future.

It might be simpler to explicitly set `GOPATH` to the empty string where `env` is defined in [context.bzl](https://github.com/bazelbuild/rules_go/blob/1d299a6bc714fedde8f6b461cd7f97732c7eb1a2/go/private/context.bzl#L380). That should clear out the environment for all Go actions, not just `GoStdlib`. Can you check whether that works? If so, I'd rather go with that. Make sure to add a quick comment pointing to this issue.

 I still think it's troubling that it's trying to write anything into `GOPATH`. That might be a bug, either in `GoStdlib` or the `go` command. I'll look into that.

andrew-d:
@jayconrod - Confirmed that the change I just pushed solves the issue I mentioned above.  Thanks for the suggestion!

jayconrod:
I couldn't reproduce this issue by pointing `GOPATH` to a read-only directory. I tried `bazel build --action_env=GOPATH=/read/only/dir --platforms=...` (which should cross-compile the standard library) and `go install std`.

I don't think there's any reason for `GOPATH` to be set in any Bazel action. It certainly shouldn't come from the user environment.

If you're able to reproduce this issue outside of NixOS, either in rules_go or in `cmd/go`, please open a new issue with instructions.

Issue #2224, opened 2019-09-21
rules_go upgrade from v0.18.6 => 0.19.4 results in stamping not expanding variables


achew22:
### What version of rules_go are you using?

Upgrading from `v0.18.6` to `v0.19.4`

n.b. that commit https://github.com/bazelbuild/rules_go/commit/37896e31341fbdcc9d8ca790431f4336110b0145 is not included in the `v0.18.6` release, but is in the `v0.19.4` release so I think this is possibly user error from #2110, but I'm still confused.

### What version of Bazel are you using?

```
$ bazelisk version
Bazelisk version: development
Starting local Bazel server and connecting to it...
WARNING: ignoring LD_PRELOAD in environment.
Build label: 0.29.1
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Sep 10 13:44:39 2019 (1568123079)
Build timestamp: 1568123079
Build timestamp as int: 1568123079
```

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Linux amd64

### Any other potentially useful information about your toolchain?

No additional toolchains.

### What did you do?

Here is the difference https://github.com/bazelbuild/bazel-watcher/compare/master...achew22:v0.10.4?expand=1

To run this on your local box:

> THIS WILL INVOKE RANDOM CODE THAT YOU DOWNLOADED FROM THE INTERNET 

```
git clone https://github.com/achew22/bazel-watcher && cd bazel-watcher && git fetch origin v0.10.4 && git reset --hard FETCH_HEAD && bazelisk build -s --stamp --workspace_status_command="python tools/workplace_status.py" --config=release //release/npm:package.json
```

I'm explicitly passing `--stamp` and `--workspace_status_command` which, when invoked with `python tools/workplace_status.py` prints out a `STABLE_GIT_VERSION`.

### What did you expect to see?

I expect to have the build succeed

### What did you see instead?

I catch an [assert statement](https://github.com/bazelbuild/bazel-watcher/compare/master...achew22:v0.10.4?expand=1#diff-ae25fa204120f7bf5cf26e1487f9a341R28) in my code

```
panic: The version string was not expanded. Got: {STABLE_GIT_VERSION}
```

jayconrod:
I took a quick look at this. I was initially able to reproduce the error, but as soon as I made any change at all, it would stop reproducing.

[...]

PR #2305, opened 2019-12-04
Create test exhibiting issue #2224

achew22:
Demonstrate that stamps are not compiled into objects when referred to
through genrules.


jayconrod:
Closing inactive PRs.

I think this was an issue with Bazel not stamping host binaries.

PR #2306, opened 2019-12-04
benchmark: fix mktemp args for MacOS

lenage:
Make bazel_benchmark.sh script works on MacOS


jayconrod:
Why change this?

This was an internal test script that isn't used anymore. It's not part of rules or other infrastructure.

lenage:
i see, PR closed

PR #2327, opened 2019-12-29
Update gogo/protobuf=1.3.1

steeve:
Contains a critical fix for using `proto.Buffer`.

Without this patch, the wire format is invalid.


jayconrod:
I'd rather not merge this PR by itself. I'm planning to release `v0.21.0` within the next week or so, and that will include an update to all dependencies.

steeve:
Good, thanks!

PR #2331, opened 2020-01-05
Import all pkg which tests depend on

### LINKED ISSUE #1999: Bazel wrongly returns cached test results with ginkgo under some circumstances

0.18.0

0.16.0

0.22.0

Yes.

Linux, Fedora

### What did you do?

Create a test suite:

```go
package virt_api_test

import (
	"testing"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
)

func TestVirtApi(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Virt_Api Suite")
}
```

which is in the _test package, so `virt_api_test`. 

Then create actual tests in a separate file in the same directory

```go
package virt_api

import (
	. "github.com/onsi/ginkgo"
)

var _ = Describe("Filesystem", func() {

	It("should succeed", func() {

	})
})
```
Note that here package `virt_api` is used. Gazelle sets both files as sources for `go_test`:

```
go_test(
    name = "go_default_test",
    srcs = [
        "api_test.go",
        "virt-api_suite_test.go",
    ],
```

The first test run executs the tests. When you then add more tests into the the test file bazel still returns the cached results.

When I set `package` in both files to the same name, does not matter if `virt_api` or `virt_api_test`, bazel detects all changes properly.

It is probably fair to say that the package setup is not right in my example, but `go test` just works like expected, so it is hard to track that your tests are actually not run by bazel.

I would have expected that bazel checks if _test files change and runs all tests again, no matter in which package they are.

### What did you see instead?

Bazel thought that there are no changes in the tests and only returns cached results.

rmohr:
All imports which gazelle identified as test dependencies should be
imported in the test main package.

This way "hidden" global registrations are properly added to the binary.

One use-case is Ginkgo. It has its own wrappers for tests and rules_go
would miss them in some cases, which would lead to tests being not added to the binary.

Fixes #1999

### DIFF
--- a/tests/core/go_test/BUILD.bazel
+++ b/tests/core/go_test/BUILD.bazel
@@ -143,3 +143,12 @@ go_bazel_test(
     name = "test_filter_test",
     srcs = ["test_filter_test.go"],
 )
+
+go_test(
+    name = "testmain_import_test",
+    srcs = [
+        "testmain_import_indirect_test.go",
+        "testmain_import_main_test.go",
+    ],
+    importpath = "example.com/imports/test_main",
+)
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -313,9 +313,13 @@ func genTestMain(args []string) error {
 			}
 		}
 	}
-	// Add only the imports we found tests for
-	for pkg := range pkgs {
-		cases.Imports = append(cases.Imports, importMap[pkg])
+
+	for name := range importMap {
+		// Set the names for all unused imports to "_"
+		if !pkgs[name] {
+			importMap[name].Name = "_"
+		}
+		cases.Imports = append(cases.Imports, importMap[name])
 	}
 	sort.Slice(cases.Imports, func(i, j int) bool {
 		return cases.Imports[i].Name < cases.Imports[j].Name
--- a/tests/core/go_test/README.rst
+++ b/tests/core/go_test/README.rst
@@ -76,3 +76,10 @@ test_filter_test
 ----------------
 
 Checks that ``--test_filter`` actually filters out test cases.
+
+testmain_import_test
+----------------
+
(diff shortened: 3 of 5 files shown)


rmohr:
@jayconrod need to add a test, but I think that this should be the fix.

rmohr:
Test added. Thanks for the help in #1999, the PR should be ready for a review.

tests/core/go_test/BUILD.bazel
```diff
@@ -143,3 +143,8 @@ go_bazel_test(
     name = "test_filter_test",
     srcs = ["test_filter_test.go"],
 )
+
+go_bazel_test(
```
jayconrod: This can be a `go_test` instead of a `go_bazel_test`. Bazel doesn't need to be invoked in any special way in a separate workspace. `go_test` is much faster.

tests/core/go_test/testmain_import_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func TestMain(m *testing.M) {
+	bazel_testing.TestMain(m, bazel_testing.Args{
+		Main: `
+-- BUILD.bazel --
+load("@io_bazel_rules_go//go:def.bzl", "go_test")
+
+go_test(
+    name = "import_test",
+    srcs = ["indirect_test.go", "main_test.go"],
```
jayconrod: Make sure to run buildifier when bringing this back to `tests/core/go_test/BUILD.bazel`. These files should be on separate lines.

rmohr: Done

tests/core/go_test/BUILD.bazel
rmohr: Done. Indeed much faster. :+1:

Issue #236, opened 2016-12-22
Expose test cases in test XML


ixdy:
Over on the Kubernetes project, we're currently running `go test -v` piped into [go-junit-report](https://github.com/jstemmer/go-junit-report), which gives nice JUnit XML down to test case level, e.g. for the test added in https://github.com/kubernetes/kubernetes/pull/39161:
```xml
<testsuite tests="104" failures="1" time="7.000" name="k8s.io/kubernetes/pkg/kubelet">
  <properties>
    <property name="go.version" value="go1.7.4"/>
  </properties>
  ...
  <testcase classname="kubelet" name="TestAlwaysFail" time="0.000">
    <failure message="Failed" type="">kubelet_pods_test.go:1491: this should always fail</failure>
  </testcase>
</testsuite>
```

When I run this same test under bazel, the `test.xml` only notes that the overall test failed, with no details about the failing test case.
```xml
<testsuite name="pkg/kubelet/go_default_test" tests="1" failures="0" errors="1">
  <testcase name="pkg/kubelet/go_default_test" status="run"><error message="exited with error code 1"></error></testcase>
</testsuite>                                                                                                                             
```

If look at `test.log`, I can eventually find the failing test case, though it's not as straightforward as the `go-junit-report`-produced XML. (It looks like this log is just `go test -v` output.)

I'm guessing the Bazel `go_test` output can't be improved until https://github.com/golang/go/issues/2981 is resolved, but if you have any suggestions/workarounds, those would be welcome, too.

nelhage-stripe:
@ixdy fwiw I'm currently working around in CI by running `bazel test` with `--test_arg=-test.v` and then generating my own XML using `go-junit-report` on the log files in `bazel-out/local-fastbuild/testlogs/`. It's  a hack, but it gets me full reports in my CI.

Would also love to see this upstreamed somehow.

ashi009:
@ianthehat any pointer on this?  I'd love to improve test report feature as well.  I'll have some time to work on this next week (probably).

brendandouglas:
This is also important for IDE integration -- the Bazel IntelliJ plugin relies on test xml to show passing/failing test cases.

tristanz:
Any public workarounds for this issue?   I'm assuming other folks have already solved this in various ways.

robbertvanginkel:
[...]

PR #2348, opened 2020-01-23
Add XML testreport generation using testwrapper

robbertvanginkel:
Based on the discussion in https://github.com/bazelbuild/rules_go/issues/236#issuecomment-577425507, add a testwrapper library that can be imported from the generated testmain.

The testlibrary has a Wrap function, that will check if the test was executed from Bazel. If it was, the wrapper function executes the binary itself, wraps some functionality like capturing stdout and exits with the exitcode from the child process. If the wrapper function decides to wrap, it does not return.

### DIFF
--- a/go/tools/testwrapper/test2json.go
+++ b/go/tools/testwrapper/test2json.go
@@ -1,0 +1,450 @@
+// Copyright 2017 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package test2json implements conversion of test binary output to JSON.
+// It is used by cmd/test2json and cmd/go.
+//
+// See the cmd/test2json documentation for details of the JSON encoding.
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
+	"unicode/utf8"
+)
+
+// Mode controls details of the conversion.
+type Mode int
+
+const (
+	Timestamp Mode = 1 << iota // include Time in events
+)
+
+// event is the JSON struct we emit.
+type event struct {
+	Time    *time.Time `json:",omitempty"`
+	Action  string
+	Package string     `json:",omitempty"`
+	Test    string     `json:",omitempty"`
+	Elapsed *float64   `json:",omitempty"`
+	Output  *textBytes `json:",omitempty"`
+}
+
+// textBytes is a hack to get JSON to emit a []byte as a string
+// without actually copying it to a string.
+// It implements encoding.TextMarshaler, which returns its text form as a []byte,
+// and then json encodes that text form as a string (which was our goal).
+type textBytes []byte
+
+func (b textBytes) MarshalText() ([]byte, error) { return b, nil }
+
+// A converter holds the state of a test-to-JSON conversion.
+// It implements io.WriteCloser; the caller writes test output in,
+// and the converter writes JSON output to w.
+type converter struct {
+	w        io.Writer  // JSON output stream
+	pkg      string     // package to name in events
+	mode     Mode       // mode bits
+	start    time.Time  // time converter started
+	testName string     // name of current test, for output attribution
+	report   []*event   // pending test result reports (nested for subtests)
+	result   string     // overall test result if seen
+	input    lineBuffer // input buffer
+	output   lineBuffer // output buffer
+}
+
+// inBuffer and outBuffer are the input and output buffer sizes.
+// They're variables so that they can be reduced during testing.
+//
+// The input buffer needs to be able to hold any single test
+// directive line we want to recognize, like:
+//
+//     <many spaces> --- PASS: very/nested/s/u/b/t/e/s/t
+//
+// If anyone reports a test directive line > 4k not working, it will
+// be defensible to suggest they restructure their test or test names.
+//
+// The output buffer must be >= utf8.UTFMax, so that it can
+// accumulate any single UTF8 sequence. Lines that fit entirely
+// within the output buffer are emitted in single output events.
+// Otherwise they are split into multiple events.
+// The output buffer size therefore limits the size of the encoding
+// of a single JSON output event. 1k seems like a reasonable balance
+// between wanting to avoid splitting an output line and not wanting to
+// generate enormous output events.
+var (
+	inBuffer  = 4096
+	outBuffer = 1024
+)
+
+// NewConverter returns a "test to json" converter.
+// Writes on the returned writer are written as JSON to w,
+// with minimal delay.
+//
+// The writes to w are whole JSON events ending in \n,
+// so that it is safe to run multiple tests writing to multiple converters
+// writing to a single underlying output stream w.
+// As long as the underlying output w can handle concurrent writes
+// from multiple goroutines, the result will be a JSON stream
+// describing the relative ordering of execution in all the concurrent tests.
+//
+// The mode flag adjusts the behavior of the converter.
+// Passing ModeTime includes event timestamps and elapsed times.
+//
+// The pkg string, if present, specifies the import path to
+// report in the JSON stream.
+func NewConverter(w io.Writer, pkg string, mode Mode) io.WriteCloser {
+	c := new(converter)
+	*c = converter{
+		w:     w,
+		pkg:   pkg,
+		mode:  mode,
+		start: time.Now(),
+		input: lineBuffer{
+			b:    make([]byte, 0, inBuffer),
+			line: c.handleInputLine,
+			part: c.output.write,
+		},
+		output: lineBuffer{
+			b:    make([]byte, 0, outBuffer),
+			line: c.writeOutputEvent,
+			part: c.writeOutputEvent,
+		},
+	}
+	return c
+}
+
+// Write writes the test input to the converter.
+func (c *converter) Write(b []byte) (int, error) {
+	c.input.write(b)
+	return len(b), nil
+}
+
+var (
+	bigPass = []byte("PASS\n")
+	bigFail = []byte("FAIL\n")
+
+	updates = [][]byte{
+		[]byte("=== RUN   "),
+		[]byte("=== PAUSE "),
+		[]byte("=== CONT  "),
+	}
+
+	reports = [][]byte{
+		[]byte("--- PASS: "),
+		[]byte("--- FAIL: "),
+		[]byte("--- SKIP: "),
+		[]byte("--- BENCH: "),
+	}
+
+	fourSpace = []byte("    ")
+
+	skipLinePrefix = []byte("?   \t")
+	skipLineSuffix = []byte("\t[no test files]\n")
+)
+
+// handleInputLine handles a single whole test output line.
+// It must write the line to c.output but may choose to do so
+// before or after emitting other events.
+func (c *converter) handleInputLine(line []byte) {
+	// Final PASS or FAIL.
+	if bytes.Equal(line, bigPass) || bytes.Equal(line, bigFail) {
+		c.flushReport(0)
+		c.output.write(line)
+		if bytes.Equal(line, bigPass) {
+			c.result = "pass"
+		} else {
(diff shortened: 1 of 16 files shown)


robbertvanginkel:
@jayconrod I'm not sure exactly how I should go about adding a library that's importable by the testmain. I've followed how the code is layed out for the `tools/coverdata` library, but I'm not sure if this is the right approach. Please let me know if the approach is right and I'll add some tests to the PR.

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,31 @@
...
+	"os"
+
+	"os/exec"
+)
+
+// WrapIfNecessary wraps the execution of the test
+func WrapIfNecessary() {
+	if os.Getenv("RAN_FROM_BAZEL") != "1" {
+		return
+	}
```
robbertvanginkel: Maybe we should add an override to turn it off, in case the wrapper doesn't work for them for any reason (incompatible  code in `init()`, tests being dependent  on `-test.v`/`t.Verbose`). Maybe a global env variable `RULES_GO_NO_TESTWRAP` settable with `--test_env=`?

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,31 @@
...
+	os.Unsetenv("RAN_FROM_BAZEL")
+
+	exitCode := 0
+	var stdoutBuf bytes.Buffer
+
+	args := append([]string{"-test.v"}, os.Args[1:]...)
```
robbertvanginkel: I just found in the CI results modifying this causes issues, I'll look into different ways of instructing the child process to run in verbose mode when using the wrapper.

go/tools/testwrapper/wrap.go
jayconrod: There should be an environment variable that turns this on or off, so that if someone runs the test binary directly, it doesn't trigger the wrapping. If this should be on default (probably yes?), let's pick something related set by Bazel's test wrapper like `XML_OUTPUT_FILE`.

There should be another environment variable that overrides this. The wrapper can set this in the child process to prevent recursion. Maybe `GO_TEST_WRAP=0`? This could also be used to explicitly turn wrapping on, even when the executable is not run by Bazel.

```
if wrapEnv, ok := os.LookupEnv("GO_TEST_WRAP"); ok {
  wrap, err := strconv.ParseBool(wrapEnv)
  if err != nil {
    log.Fatalf("invalid value for GO_TEST_WRAP: %q", wrapEnv)
  }
  // explicitly wrap or not
}
```

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"bytes"
+	"io"
+	"os"
+
+	"os/exec"
+)
+
+// WrapIfNecessary wraps the execution of the test
```
jayconrod: Might be better to split this into two functions so that `main` has more control over how the process exits.

```
func shouldWrap() bool
func wrap() error
```

The call would be like:

```
if shouldWrap() {
  err := wrap()
  if xerr, ok := err.(*exec.ExitError); ok {
    os.Exit(xerr.ExitCode())
  } else if err != nil {
    log.Print(err)
    os.Exit(6) // something distinctive for abnormal termination
  } else {
    os.Exit(0)
  }
}
```

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,31 @@
...
+func WrapIfNecessary() {
+	if os.Getenv("RAN_FROM_BAZEL") != "1" {
+		return
+	}
+	os.Unsetenv("RAN_FROM_BAZEL")
+
+	exitCode := 0
```
jayconrod: Let's pick a distinctive value for an abnormal termination. Not sure if there's a standard code, but Bazel uses 6 to mean `RUN_FAILURE` in [ExitCodes.java](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/ExitCode.java), so I suggested that above.

go/tools/testwrapper/wrap.go
jayconrod: It looks like `//tests/legacy/transitive_data:go_default_test` is the only one affected, and that test is wrong. It would be broken with `go test -v`, too. The test should read `flag.Args()` instead of `os.Args`.

robbertvanginkel: #2348

robbertvanginkel:
I'll address the comments. Also made https://github.com/bazelbuild/rules_go/pull/2349 to resolve the other test.

For the high level approach of adding wrap as a single source file, do you think we should do the  same for embedding the code that will do the stdout -> xml conversion?

jayconrod:
> For the high level approach of adding wrap as a single source file, do you think we should do the same for embedding the code that will do the stdout -> xml conversion?

Yes. I don't expect that will be a large amount of code. I think it only makes sense to split it out if something else needs to import it or it causes a measurable increase in compile time across a large number of tests (more than the corresponding dependency wrangling and linking overhead).

go/tools/testwrapper/test2json.go
```diff
@@ -0,0 +1,450 @@
+// Copyright 2017 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package test2json implements conversion of test binary output to JSON.
+// It is used by cmd/test2json and cmd/go.
+//
+// See the cmd/test2json documentation for details of the JSON encoding.
+package main
```
jayconrod: Let's add a note here explaining where this file was copied from.

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,72 @@
...
+func shouldWrap() bool {
+	if wrapEnv, ok := os.LookupEnv("GO_TEST_WRAP"); ok {
+		wrap, err := strconv.ParseBool(wrapEnv)
+		if err != nil {
+			log.Fatalf("invalid value for GO_TEST_WRAP: %q", wrapEnv)
+		}
+		return wrap
+	}
+	if _, ok := os.LookupEnv("XML_OUTPUT_FILE"); ok {
+		return true
```
jayconrod: Nit: `return ok`

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,72 @@
...
+	if out, ok := os.LookupEnv("XML_OUTPUT_FILE"); ok {
+		writeReport(jsonBuffer, pkg, out)
+	}
+	return err
+}
+
+func writeReport(jsonBuffer bytes.Buffer, pkg string, path string) {
+	xml, cerr := json2xml(&jsonBuffer, pkg)
+	if cerr != nil {
+		log.Printf("error converting test output to xml: %s", cerr)
```
jayconrod: Maybe return these errors instead of logging? It seems better to fail loudly and exit non-zero if we can't write the XML for some reason.

go/tools/testwrapper/wrap_test.go
```diff
@@ -0,0 +1,49 @@
+package main
```
jayconrod: Nit: Need a copyright notice.

go/tools/testwrapper/wrap_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+func TestShouldWrap(t *testing.T) {
```
jayconrod: Separately, it would be good to have a `go_bazel_test` in `//tests/core/go_test` that invokes `bazel test` on a test target and verifies the output XML. That could be instead of or in addition to `xml_test.go`.

go/tools/testwrapper/xml_test.go
```diff
@@ -0,0 +1,41 @@
+package main
```
jayconrod: Need a copyright notice

go/tools/testwrapper/wrap.go
robbertvanginkel: Originally I had the idea that the XML generation shouldn't be able to change the exit code of the wrapper's execution. Returning an error here would make it possible for successful test runs to fail on xml generation.

Rethinking, I agree with you that it's probably better to error loudly instead of silently not writing a report. Will fix and adjust the comment for `testWrapperAbnormalExit` to indicate it could also be used for this.

go/tools/testwrapper/wrap_test.go
robbertvanginkel: Added it as an addition. The integration test under `//tests/core/go_test` is a little more tricky to completely verify as the test runtime will be different each execution, so I think this unittest is useful to test the explicit conversion from the json events to xml.

go/tools/testwrapper/wrap.go
```diff
@@ -0,0 +1,77 @@
...
+	err := cmd.Run()
+	jsonConverter.Close()
+	if out, ok := os.LookupEnv("XML_OUTPUT_FILE"); ok {
+		werr := writeReport(jsonBuffer, pkg, out)
+		if werr != nil {
+			if err != nil {
+				return fmt.Errorf("error while generating testreport: %s, (error wrapping test execution: %s)", werr, err)
+			}
+			return fmt.Errorf("error while generating testreport: %s", werr)
+		}
```
robbertvanginkel: @jayconrod the nesting here seems a little ugly, if you have any better suggestions for reporting back both errors please let me know.

jayconrod: Seems good enough. This shouldn't be a common user-facing error, and there's enough information to trace it back here. Also, `err` should be very short since stderr is piped to `os.Stderr`.

tests/core/go_test/xmlreport_test.go
```diff
@@ -0,0 +1,130 @@
...
+	err := bazel_testing.RunBazel("test", "//:xml_test")
+	if err == nil {
+		t.Fatal("expected bazel test to have failed")
+	}
+	if xerr, ok := err.(*bazel_testing.StderrExitError); !ok {
+		t.Fatalf("expected bazel tests to fail with exit code 3 (TESTS_FAILED), got: %s", err)
+	} else if xerr.Err.ExitCode() != 3 {
+		t.Fatalf("expected bazel tests to fail with exit code 3 (TESTS_FAILED), got: %s", xerr)
+	}
+	b, err := ioutil.ReadFile("bazel-out/darwin-fastbuild/testlogs/xml_test/test.xml")
```
jayconrod: This test is failing on non-darwin platforms because the xml gets written somewhere else. `bazel info bazel-testlogs` should give you the right directory.

tests/core/go_test/xmlreport_test.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func Test(t *testing.T) {
+	err := bazel_testing.RunBazel("test", "//:xml_test")
+	if err == nil {
+		t.Fatal("expected bazel test to have failed")
+	}
+	if xerr, ok := err.(*bazel_testing.StderrExitError); !ok {
+		t.Fatalf("expected bazel tests to fail with exit code 3 (TESTS_FAILED), got: %s", err)
+	} else if xerr.Err.ExitCode() != 3 {
```
jayconrod: Simplification: `!ok || xerr.Err.ExitCode() != 3`

tests/core/go_test/xmlreport_test.go
robbertvanginkel: Must have been half asleep while writing this :)

robbertvanginkel:
I'll retitle the PR to indicate its not just scaffolding a wrapper anymore but also includes adding the  testcode.

I was thinking maybe we should add documentation on `GO_TEST_WRAP` and the wrapper automatically calling all tests with `-test.v` set?

jayconrod:
Documentation sounds good. Maybe add a note in go/core.rst under `go_test`?

Looks good otherwise. Thanks for working on this!

robbertvanginkel:
No changes planned from my side, please merge when you have a chance :)

jayconrod:
Merged. Thanks again for working on this!

PR #2355, opened 2020-01-30
Remove patching for WORKSPACE file

meteorcloudy:
Bazel will always override the external repository's WORKSPACE file by `workspace_file`, `workspace_file_content` or a created file with "workspace(name =`<name in repo rule>`)"

### DIFF
--- a/third_party/go_googleapis-directives.patch
+++ b/third_party/go_googleapis-directives.patch
@@ -6,15 +6,6 @@ diff -urN b/BUILD.bazel c/BUILD.bazel
 +# gazelle:proto_group go_package
 +# gazelle:exclude gapic
 +# gazelle:exclude third_party
-diff -urN b/WORKSPACE c/WORKSPACE
---- b/WORKSPACE	2000-01-01 00:00:00.000000000 -0000
-+++ c/WORKSPACE	2000-01-01 00:00:00.000000000 -0000
-@@ -1,4 +1,4 @@
--workspace(name = "com_google_googleapis")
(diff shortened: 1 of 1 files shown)


meteorcloudy:
This will fix https://github.com/bazelbuild/bazel/issues/10681 for rules_go

philwo:
Thank you, Yun!

To clarify, this means that the problematic part of the patch was simply ignored until now (Bazel would patch the WORKSPACE file as requested and then replace it with a completely new one), so you can just remove it. Bazel@HEAD contains a bugfix for this behavior, so that patches now get applied *after* replacing the WORKSPACE file - which means that if the patch was not valid, it will now fail.

jayconrod:
Seems fine. I think this was necessary at some point, but not anymore.

What goes into the replacement WORKSPACE file?

philwo:
@jayconrod The default WORKSPACE (if neither `workspace_file` nor `workspace_file_content` are specified on the `http_repository`) simply contains one line:

```
workspace(name = "{name}")
```

where {name} is the name attribute of the `http_repository`.

This is implemented here: https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/utils.bzl#L55

nlopezgi:
Hi, could we have a patch release with this commit included. See https://github.com/bazelbuild/rules_k8s/pull/517#issuecomment-581451960

jayconrod:
@nlopezgi Just tagged releases `v0.20.6` and `v0.21.2` with this.

nlopezgi:
thanks!

PR #2361, opened 2020-02-05
go/tools/bazel: support bazel.Runfile inside actions

### LINKED ISSUE #2359: runfiles.go should look at os.Args[0] + ".runfiles" etc.

0.21.2

None

2.0.0

Yes

Debian GNU/Linux, x86-64

No

### What did you do?

I created the following workspace:

- BUILD:

```python
load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")

go_binary(
    name = "test",
    srcs = ["test.go"],
    data = ["test.txt"],
    deps = ["@io_bazel_rules_go//go/tools/bazel:go_default_library"],
)

genrule(
    name = "genrule",
    outs = ["copy.txt"],
    cmd = "$(location :test) > $@",
    tools = [":test"],
)
```

- test.go:

```go
package main

import (
	"io"
	"log"
	"os"

	"github.com/bazelbuild/rules_go/go/tools/bazel"
)

func main() {
	filename, err := bazel.Runfile("test.txt")
	if err != nil {
		log.Fatal(err)
	}
	fd, err := os.Open(filename)
	if err != nil {
		log.Fatal(err)
	}
	defer fd.Close()
	if _, err := io.Copy(os.Stdout, fd); err != nil {
		log.Fatal(err)
	}
}
```

- test.txt with arbitrary contents.

Then run `bazel build :copy.txt`.

Command succeeds

### What did you see instead?

```none
could not locate runfiles directory
Target //:copy.txt failed to build
```

* * * *

The issue is that Bazel actions don't pass runfiles environment variables, and the current directory is also incorrect. Rather, runfiles.go should look at `os.Args[0] + ".runfiles"`, `os.Args[0] + ".runfiles/MANIFEST"`, and `os.Args[0] + ".runfiles_manifest"`. See the C++ code for comparison: https://github.com/bazelbuild/bazel/blob/master/tools/cpp/runfiles/runfiles_src.cc#L271-L280.

jayconrod:
Fixes #2359


jayconrod:
Pausing work on this PR for now due to lack of time.

Windows has wildly different behavior of course. For a test, there's a .runfiles directory and no environment variables are set, but the directory contains a manifest file.

PR #2367, opened 2020-02-12
upgrade x/tools

linzhp:
Upgraded x/tools to the master as of 2020-02-11, because we need it...


jayconrod:
Normally I don't accept PRs updating dependencies out of cycle. A lot of things depend on x/tools though, and it's difficult to override.

Are you able to follow [Overriding dependencies](https://github.com/bazelbuild/rules_go/blob/master/go/workspace.rst#overriding-dependencies)? I'm not 100% sure it will work for x/tools because of the patches required, but that should be the first thing to try. Repositories overridden that way must be declared before calling `go_rules_dependencies`.

If not, please follow [Updating dependencies](https://github.com/bazelbuild/rules_go/wiki/Updating-dependencies) to get this PR working.

If that doesn't work, I can also update x/tools in another PR today or tomorrow.

linzhp:
We normally override dependencies of rules_go and use our go.mod as the source of true for the versions. However, the latest version of x/tools requires changes to `third_party/org_golang_x_tools-extras.patch`, which is the main focus of this PR (sorry, VS Code removed some trailing whitespaces).

Please update x/tools if you have time. I need to fix another issue of importing x/tools today.

Issue #2002, opened 2019-03-22
Argument list too long on macOS


jarreds:
- rules_go v2.30
- gazelle 0.17.0
- bazel 0.23.1

I put up a simple repro project that you anyone can clone:

```
~/src/github.com/jarreds $ git clone git@github.com:jarreds/toomanyfiles.git
Cloning into 'toomanyfiles'...
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 5 (delta 0), reused 5 (delta 0), pack-reused 0
Receiving objects: 100% (5/5), done.
~/src/github.com/jarreds $ cd toomanyfiles/
~/src/github.com/jarreds/toomanyfiles $ bazel build //:toomanyfiles
Loading: 
Loading: 0 packages loaded
Analyzing: target //:toomanyfiles (1 packages loaded, 0 targets configured)
INFO: Analysed target //:toomanyfiles (3 packages loaded, 30 targets configured).
INFO: Found 1 target...
bazel: Entering directory `/private/var/tmp/_bazel_jward/f47a410108a7b6bad68cd01a437e2533/execroot/__main__/'
[0 / 7] [-----] BazelWorkspaceStatusAction stable-status.txt
ERROR: /private/var/tmp/_bazel_jward/f47a410108a7b6bad68cd01a437e2533/external/com_github_awslabs_goformation/cloudformation/resources/BUILD.bazel:3:1: GoCompile external/com_github_awslabs_goformation/cloudformation/resources/darwin_amd64_stripped/go_default_library%/github.com/awslabs/goformation/cloudformation/resources.a failed (Exit 1) builder failed: error executing command bazel-out/host/bin/external/go_sdk/builder ... (remaining 4 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
compile: error starting compiler: fork/exec external/go_sdk/pkg/tool/darwin_amd64/compile: argument list too long
bazel: Leaving directory `/private/var/tmp/_bazel_jward/f47a410108a7b6bad68cd01a437e2533/execroot/__main__/'
Target //:toomanyfiles failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.137s, Critical Path: 0.40s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
FAILED: Build did NOT complete successfully
```

jayconrod:
[...]

PR #2375, opened 2020-02-19
Use paramfile for invoking nogo

robbertvanginkel:
When calling the builder binary from starlark, the `args.use_param_file`
option is used to support long argument lists. For nogo, the builder
calls the nogo binary with a list of arguments for which the size is
proportional to what the builder was called with. Currently nogo isn't
called with a parameter file but instead gets it's arguments from the
commandline.

This change writes the nogo arguments to file and calls nogo with an
argument to read the parameter file for arguments.

References #2002.

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -321,7 +321,7 @@ func compileArchive(
 		ctx, cancel := context.WithCancel(context.Background())
 		nogoChan = make(chan error)
 		go func() {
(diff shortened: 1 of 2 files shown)


robbertvanginkel:
@jayconrod I have verified with our own project that this resolves the issue that we were running into with https://github.com/bazelbuild/rules_go/issues/2002#issuecomment-585435383.

With nogo testing mostly relying on integration tests, adding a regression test that checks this specific detail of using/not using param files is a little tricky. Do you have a suggestion for if/how we should test that or are the tests that check if nogo works at all good enough for this change?

PR #2383, opened 2020-02-25
add go 1.14 to sdk

rickypai:


jayconrod:
Glad you're eager to use Go 1.14, but sorry, I don't accept PRs to add new Go versions or update dependencies. It's more work to review and verify than it is to generate them.

Superceded by #2384

rickypai:
np. thank you for the diligence

PR #2392, opened 2020-02-28
Separating patch for golang/protobuf/protoc-gen-go/testdata/grpc

### LINKED ISSUE #2391: Not able to override golang/protobuf version
After rules_go upgrades golang/protobuf to 1.13.3, we are no longer able to override the version, because `com_github_golang_protobuf-extras.patch` includes patches for `protoc-gen-go/testdata/grpc/BUILD.bazel` but `protoc-gen-go/testdata/grpc` has a go.mod file, making it a different Go module. When using `go_repository` rule to override `golang/protobuf`, `protoc-gen-go/testdata/grpc` is not included in the downloaded source, so the patch no longer apply

0.22.1

92bc7b03d768a3581bd9bdd30d5e3635bbf854da

2.0.0

Should reproduce

macOS 10.15.3, amd64

linzhp:
Fixed #2391

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -140,7 +140,7 @@ def go_rules_dependencies():
             "@io_bazel_rules_go//third_party:com_github_golang_protobuf-extras.patch",
             # protoc-gen-go/testdata/grpc is a different Go module.
             # Separating this so the above patch can be used by go_repository rules
-            "@io_bazel_rules_go//third_party:com_github_golang_protobuf-testdata-grpc.patch.patch",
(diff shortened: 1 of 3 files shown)


jayconrod:
I don't think these `filegroup` rules are actually necessary anymore. They were generated by an older version of Gazelle (before we started patching), and this patch just adds them back in case anyone was still depending on them.

No one should depend on targets in `testdata` though. I'll send a PR that just deletes this part of the patch.

jayconrod:
Superceded by #2394

PR #2409, opened 2020-03-21
[Fix] allow compilation with wasm js

grahamjenson:
To use protobuf builds with `goarch = "wasm"` and  `goos = "js"` the protobuf libraries need to include both versions of pointer, `unsafe` and `reflect`. The correct version will be selected with calling `go_binary` so both can be included safely. 

`gazelle` cannot be used to generate a BUILD file that contains both, so these need to be added to the `extras` patches.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -170,6 +170,8 @@ def go_rules_dependencies():
         patches = [
             # gazelle args: -repo_root . -go_prefix github.com/gogo/protobuf -proto legacy
             "@io_bazel_rules_go//third_party:com_github_gogo_protobuf-gazelle.patch",
+            # additional targets may depend on generated code for well known types
+            "@io_bazel_rules_go//third_party:com_github_gogo_protobuf-extras.patch",
         ],
         patch_args = ["-p1"],
     )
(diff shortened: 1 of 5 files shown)


jayconrod:
Please open an issue to discuss this.

If the issue is that Gazelle doesn't generate anything for the `wasm` / `js`, that should be fixed in Gazelle instead of adding a patch here.

grahamjenson:
@jayconrod This is not specific to WASM but I think maybe a fundamental floor in gazelle. In a situation where a project has two versions of code that are mutually exclusive  (e.g. in protobuf library with pointer_reflect.go and pointer_unsafe.go ), gazelle (because of build_tags) will only ever include one file. So, if a project has two go_binaries that require the different versions it cannot build both.

Usually, to fix this you could just pull in two go_repositories, and then use different `build_tags` option to get a version of the library for both environments, but because rules_go statically builds a patch, this means that it cannot do that. 

I created an issue here https://github.com/bazelbuild/bazel-gazelle/issues/740 on gazelle.

jayconrod:
Closing inactive PRs.

"wasm" and "js" are now among the platforms and tags recognized by rules_go and Gazelle, so I think this is resolved. If not, please open a new issue.

Issue #2302, opened 2019-12-02
go: deprecate the go_rule wrapper


jayconrod:
See bazelbuild/bazel-gazelle#676.

A Bazel incompatibility flag (possibly `--incompatible_remap_main_repo`, see bazelbuild/bazel#7130) causes default labels in attributes to be resolved in `@io_bazel_rules_go//go/private` (where `go_rule` is defined) instead of wherever the rule should actually have been defined.

We use `go_rule` as a wrapper around `rule` to add hidden attributes and a toolchain dependency. Instead, we should move hidden attributes to the toolchain *or* provide a set of common attributes. The toolchain dependency can be explicit.

jmhodges:
I see the deprecation commit made it in and would love to hear what the migration should look like! (bazel_gomock is still using `go_rule`, so it's relevant to my interests)

jayconrod:
@jmhodges I apologize for creating a classic Google "old thing is deprecated, new thing is not ready yet" dilemma. I thought it was important to mark this is deprecated, given that there are potential problems caused by Bazel.

The change I would like to make is for a skeleton Go rule to look something like this:

```
go_example = rule(
    implementation = _go_example_impl,
    attrs = {
        "_go_context": attr.label(default = "@io_bazel_rules_go//:go_context_data"),
    },
    toolchains = ["@io_bazel_rules_go//go:toolchain"],
)
```

The rule would get nearly all of the information it needs from the toolchain. There are a few things that must be configured for the target platform (C toolchain, flags), and toolchains are always configured for the host platform, so we still need `go_context_data` to provide those.

After this is ready and `rules_go` is migrated, `go_rule` would continue to be available for at least a full release cycle (about a quarter). It may be reduced to just adding the `_go_context` attribute and the toolchain.

PR #2414, opened 2020-03-26
go: use configuration transitions instead of aspect

### LINKED ISSUE #1351: go_binary: support build constraints via build_setting
I have a library which requires custom build tags to build it (because there are two modes of operation, and we switch between them using build tags).

Specifically, if building from make I need to do this:

` go build -o $@ -tags 'peer_name_alternative peer_name_hash' k8s.io/kops/protokube/cmd/protokube`

When I try what I believe to be the equivalent syntax with bazel the tags appear to be ignored - they aren't visible on the command line with `bazel build -s` for example, and I get a compilation error that suggests the code was ignored.

I'm adding this to go_library: `tags = [  "peer_name_alternative", "peer_name_hash"  ],  # keep`.  Actual code is here: https://github.com/kubernetes/kops/pull/4514/files 

Is there a different place / way I should be passing the tags?

### LINKED ISSUE #2219: Use build_settings and transitions instead of aspect for cross-compilation

jayconrod:
This changes how the goos, goarch, race, msan, static, and pure
attributes of go_binary and go_test are implemented.

Previously, the go_binary and go_test rules used an aspect on the deps
and embed attributes. If any of the mode attributes listed above were
set, the aspect declared additional output files and actions to
generate them for the specified configuration. This approach had two
significant problems. First, the aspect caused the analysis to be
performed twice, wasting time and memory even in the case
where no cross-compilation was needed. Second, the aspect was not
compatible with conditional dependencies specified with select
expressions: aspects run on the configured target graph, but the mode
attributes could not affect Bazel's configuration.

This change deletes the aspect and implements these attributes using a
different mechanism: build settings and configuration transitions.

A new package is added //go/config with a build setting for each
attribute. Most settings are flags that may be set on the
command-line. This is an alternative to --feature flags, which will be
deprecated and eventually removed.

The target //:go_config gathers build settings relevant to Go and
provides a GoContextInfo (private).

The target //:go_context_data depends on //:go_config, //:stdlib,
//:nogo, and a few other standard dependencies. go_binary and go_test
rules no longer need to depend on these targets directly. Go rules
should now only need to depend on //:go_context_data and the Go
toolchain. Unfortunately, neither //:go_context_data nor //:go_config
can be part of the toolchain because the toolchain is always analyzed
for the execution configuration, not the target configuration.

Because of the simplified dependencies, the go_rule function is no
longer needed by most rules. This change makes it possible to
deprecate it.

Fixes #1351
Fixes #2219
Updates #2302


achew22:
Woohoo! Congratulations on doing this @jayconrod. This is the number one PR of the week as far as I'm concerned!

steeve:
So happy to see this!

seh:
Given that #1351 is linked from here, does this change how Go build tags will work with Bazel? It looks like your patch retains the "gotags" attribute.

jayconrod:
@seh At the moment, it's possible to set build tags with `--define=gotags=foo,bar`. The `--define` flag was never really meant to be used like this though, so this is undocumented. It will eventually be deprecated and removed.

After this change, you'll be able to set build tags either by setting `gotags = ["foo", "bar"]` on a `go_binary` rule or by using the `--@io_bazel_rules_goo//go/config:tags` command line flag. Both would be equivalent.

I don't think command line flags like this are quite working yet in Bazel, but rules_go shouldn't need to do anything further to enable that in future versions.

jayconrod:
I decided to leave `gotags` out of this change, since I don't want to make it bigger than necessary. The infrastructure is there though, so it should be an easy follow-up. I've milestoned #1351 for the next release.

jayconrod:
I think this is ready to go, so merging now.

I'm hoping to ship this with 0.23.0 in early May. Please try it out and let me know if you have issues.

steeve:
Congratulations! I’ll test it asap!

jmillikin-stripe:
Now that rules_go has been migrated to build settings, does this mean the change in https://github.com/bazelbuild/rules_go/pull/2249 (a build setting to disable platform-specific output paths) would be acceptable?

pcj:
Thanks @jayconrod for landing this!  

Looking forward to testing this... Any notes to share regarding suggested/reqiured migration steps?

steeve:
> Now that rules_go has been migrated to build settings, does this mean the change in https://github.com/bazelbuild/rules_go/pull/2249 (a build setting to disable platform-specific output paths) would be acceptable?

I even feel like it should be the default? To be consistent with other bazel rules.

However when using a transition, paths are much much bigger.

steeve:
By the way tested this weekend, it works great and opens up so much possibilities!

jayconrod:
> Now that rules_go has been migrated to build settings, does this mean the change in #2249 (a build setting to disable platform-specific output paths) would be acceptable?

@jmillikin-stripe @steeve Rather than having a build setting control output path prefixes, I think now we can get rid of the path prefixes entirely. I just opened #2448 to track that.

jayconrod:
> Looking forward to testing this... Any notes to share regarding suggested/reqiured migration steps?

@pcj I'll include more information in the v0.23.0 release notes when it's ready.

There shouldn't be any incompatible change to the `go_binary` or `go_test` rule API. Just that `goos` and `goarch` will now work with `select` (they previously did not affect what `select` picked).

`--feature` flags should still work in this release, but I'll deprecate them before the release. Weird things may happen if they're mixed with the new flags.

Hopefully that's it.

pcj:
Thanks @jayconrod: for the docs (and/or this thread) I'd like to know the recommended strategy for a go_image/go_binary intended for linux container: should one use `goos = "linux", goarch = "amd64"` in the rule, or should one continue to specify `--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64`?

steeve:
They are equivalent since `goos`/`goarch` will trigger a transition on `platform`.

However, if you want your binary to be always built for a specific `goos`/`goarch` tuple, without fiddling with `.bazelrc` or `bazel` flags, I'd say put it in the target itself.

jayconrod:
@pcj Basically, what @steeve said. You should get the same output (not sure if bit for bit identical, but very close at least).

If the `go_binary` fundamentally should produce a Linux / amd64 executable, I'd set `goos` and `goarch`. That may be true for a Docker container.

It might make sense in the future for `go_image` itself to do a transition though.

PR #2416, opened 2020-03-27
Export `TOOLS_NOGO` deps list.

### LINKED ISSUE #2415: Export TOOLS_NOGO with all analyzers
rules_go exposes a "tools_nogo" rule that depends on all analyzers from "golang.org/x/tools/go/analysis/passes".  However, if you want to write a custom nogo vetter, you have to write your own "nogo" rule, but there's no way to also depend on all analyzers.  Basically, I'd like to be able to write:

```
load("@io_bazel_rules_go//go:def.bzl", "TOOLS_NOGO")

nogo(
    name = "mynogo",
    deps = TOOLS_NOGO + [":vetter_foobar"],
    config = "nogo.config",
)
```

bartle-stripe:
Fixes #2415

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -75,6 +75,48 @@ load(
     _nogo = "nogo_wrapper",
 )
 
+# TOOLS_NOGO is a list of all analysis passes in
+# golang.org/x/tools/go/analysis/passes.
+# This is not backward compatible, so use caution when depending on this --
+# new analyses may discover issues in existing builds.
+TOOLS_NOGO = [
+    "@org_golang_x_tools//go/analysis/passes/asmdecl:go_tool_library",
+    "@org_golang_x_tools//go/analysis/passes/assign:go_tool_library",
+    "@org_golang_x_tools//go/analysis/passes/atomic:go_tool_library",
+    "@org_golang_x_tools//go/analysis/passes/atomicalign:go_tool_library",
+    "@org_golang_x_tools//go/analysis/passes/bools:go_tool_library",
+    "@org_golang_x_tools//go/analysis/passes/buildssa:go_tool_library",
+    "@org_golang_x_tools//go/analysis/passes/buildtag:go_tool_library",
+    # TODO(#2396): pass raw cgo sources to cgocall and re-enable.
+    # "@org_golang_x_tools//go/analysis/passes/cgocall:go_tool_library",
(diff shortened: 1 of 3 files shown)


go/def.bzl
```diff
@@ -75,6 +75,44 @@ load(
     _nogo = "nogo_wrapper",
 )
 
+TOOLS_NOGO = [
```
jayconrod: Please add a comment here, similar to the one above `tools_nogo`.

```
# TOOLS_NOGO is a list of all analysis passes in
# golang.org/x/tools/go/analysis/passes.
# This is not backward compatible, so use caution when using on this --
# new analyses may discover issues in existing builds.
```

go/nogo.rst
```diff
@@ -86,6 +86,39 @@ even if the target is imported from an external repository. However, ``nogo``
 will not run when targets from the current repository are imported into other
 workspaces and built there.
 
+If you'd like to run all analyzers from ``golang.org/x/tools``, you can add a
+dependency on ``TOOLS_NOGO``.
```
jayconrod: "add a dependency on" doesn't seem quite right, since it's a list of dependencies.

Perhaps instead, switch the order of the two examples, since using `tools_nogo` alone will be more common.

To run all the `golang.org/x/tools` analyzers, use `@io_bazel_rules_go//:tools_nogo`.

...

To run the analyzers from `tools_nogo` together with your own analyzers, use the `TOOLS_NOGO` list of dependencies.

...

go/nogo.rst
```diff
@@ -86,6 +86,39 @@ even if the target is imported from an external repository. However, ``nogo``
...
+    )
+
+If you'd just like to depend on the ``golang.org/x/tools`` analyzers, you can depend
+on the ``@io_bazel_rules_go//:tools_nogo`` rule.
+
+.. code:: bzl
+
+    load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
+    go_rules_dependencies()
+    go_register_toolchains(nogo="@io_bazel_rules_go//:tools_nogo")
```
jayconrod: Nit: add space around the `=` operator.

bartle-stripe:
PTAL @jayconrod

bartle-stripe:
@jayconrod When would you expect to have a release that includes this?

jayconrod:
This would go in the next minor release. Hard to say exactly, but that's roughly a month away I'd guess.

PR #2424, opened 2020-04-07
Only include crosstool in stdlib inputs when using cgo

asuffield:
Noticed this while experimenting with C++ toolchains, it causes
unnecessary cache misses for pure builds.


jayconrod:
Closing since this will almost certainly be obsoleted by #2414.

Please file an issue if you'd like to track this or discuss further.

PR #2442, opened 2020-04-18
README: add missing strip_prefix on protobuf release

johnsonj:
**What type of PR is this?**
Documentation

**What does this PR do? Why is it needed?**
Fixes build error when copy/pasting protobuf dependency example

**Which issues(s) does this PR fix?**
Fixes #2441

PR #2444, opened 2020-04-18
Fixes com_google_protobuf reference

rodrigoaliste:
This fix enables direct copy-paste on the WORKSPACE file for bazel versions that don't handle strip_prefix automatically.

**What type of PR is this?**

Documentation

Fixes the README reference of `com_google_protobuf`. Adds the strip_prefix so users don't spend minutes of their lives searching for the issue when getting the following error:

    ERROR: error loading package '': Every .bzl file must have a corresponding package, but 
    '@com_google_protobuf//:protobuf_deps.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.


jayconrod:
Sorry, I can't review anything without a signed CLA.

Based on the description I'm guessing this is the same as #2442 and #2447 though.

PR #2450, opened 2020-04-20
Check that file is not nil before accessing the name

prestonvanloon:
**What type of PR is this?**

 Bug fix

In some rare cases, go/tools/builders/nogo_main.go:389 caused a nil pointer deference.

```
compilepkg: panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x6cd735]

goroutine 1 [running]:
go/token.(*File).Name(...)
        GOROOT/src/go/token/position.go:110
main.checkAnalysisResults(0xc00022e000, 0x1d, 0x1d, 0xc0002d8e60, 0x2, 0x2)
        external/io_bazel_rules_go/go/tools/builders/nogo_main.go:389 +0x475
main.checkPackage(0x9ff7c0, 0x1d, 0x1d, 0xc000221003, 0x2b, 0xc0002015f0, 0xc000201620, 0xc000201560, 0xc000113d20, 0x2, ...)
        external/io_bazel_rules_go/go/tools/builders/nogo_main.go:192 +0x3f0
main.run(0xc000113ba0, 0x1a, 0x1a, 0x0, 0x0)
        external/io_bazel_rules_go/go/tools/builders/nogo_main.go:84 +0x47c
main.main()
        external/io_bazel_rules_go/go/tools/builders/nogo_main.go:57 +0xa6
```

Nothing filed, I think

**Other notes for review**

### DIFF
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -386,7 +386,11 @@ func checkAnalysisResults(actions []*action, pkg *goPackage) string {
 		}
 		// Discard diagnostics based on the analyzer configuration.
 		for _, d := range act.diagnostics {
-			filename := pkg.fset.File(d.Pos).Name()
+			file := pkg.fset.File(d.Pos)
+			if file == nil {
+				continue
+			}
+			filename := file.Name()
 			include := true
 			if len(config.onlyFiles) > 0 {
 				// This analyzer emits diagnostics for only a set of files.


jayconrod:
Please open an issue to discuss. When does this happen, exactly?

I don't think this is the right fix. It looks like this discards errors if `fset` isn't set. Either `fset` should always be set, or we should print something sensible when it's not/

prestonvanloon:
I didn't open an issue because I thought this was a rather simple fix

What I noticed was happening is that d.Pos was 0. I am not sure if fset was set or not, but file was definitely nil and breaking build at the time. 

I'll open an issue, although I'm not sure how to replicate it.

jayconrod:
Closing inactive PRs.

Please open an issue if you find a way to reproduce this.

Issue #2486, opened 2020-05-10
Don't depend on msys on Windows


tomlu:
run_shell (just like genrules and the like) requires msys, which isn't available on Windows by default and isn't a desirable dependency. It's trivial to fix and I already have a pull request ready that does so.

The run_shell invocation is in go_tool_binary in rules_go/blob/master/go/private/rules/binary.bzl.

### What version of rules_go are you using?

HEAD ce51edf0e2980bf7548ce7314e7d9c849994b39e

### What version of gazelle are you using?

Irrelevant.

### What version of Bazel are you using?

bazel 3.1.0

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Windows 10

### What did you do?

Build any go_binary

### What did you expect to see?

Successful build without needing msys installed.

### What did you see instead?

Failure when the a shell command is attempted to be executed.

jayconrod:
Marking this as a future enhancement. However, this is not likely to be resolved in the foreseeable future. We could probably avoid depending on `sh` and `cat`, but we at least need `patch` to install build files in Go dependencies in `http_archive` and `git_repository`. It's difficult to find a working version of `patch` on Windows outside `msys`.

tomlu:
`patch` has a native Bazel implementation if you are only doing trivial patch commands. Bazel is in charge of the patching for BUILD files is it not?

We're going through issues one-by-one, fixing them as we come across them. This one I have a pull request for, like I may have mentioned in the original reported issue.

jayconrod:
Interesting, it looks like [`repository_ctx.patch`](https://docs.bazel.build/versions/0.29.0/skylark/lib/repository_ctx.html) was introduced in Bazel 0.29.0. I'm not sure if that will work directly with the patches we have, but it's promising.

In any case, I'd be happy to take a PR for the `go_tool_binary` issue.

dzbarsky:
There's no longer usage of run_shell. We can reopen if there are further issues.

PR #2488, opened 2020-05-11
Avoid use of run_shell in bazel rule.

tomlu:
**What type of PR is this?**
> Bug fix

run_shell requires bash which is not available on Windows
without a msys installation.

The shell was only used for "&&" anyway. We can just use
two separate actions.

Commands should be equivalent, except -trimpath=. is used
instead of -trimpath=$PWD. I tried building a binary and it
came out identical, so guessing it works either way.

Fixes # https://github.com/bazelbuild/rules_go/issues/2486

**Other notes for review**

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -118,9 +118,9 @@ def _go_tool_binary_impl(ctx):
     cout = ctx.actions.declare_file(name + ".a")
     if sdk.goos == "windows":
         cmd = "@echo off\n {go} tool compile -o {cout} -trimpath=%cd% {srcs}".format(
-            go = sdk.go.path.replace('/', '\\'),
+            go = sdk.go.path.replace("/", "\\"),
             cout = cout.path,
-            srcs = ' '.join([f.path for f in ctx.files.srcs]),
+            srcs = " ".join([f.path for f in ctx.files.srcs]),
         )
         bat = ctx.actions.declare_file(name + ".bat")
         ctx.actions.write(
@@ -138,7 +138,7 @@ def _go_tool_binary_impl(ctx):
         cmd = "{go} tool compile -o {cout} -trimpath=$PWD {srcs}".format(
             go = sdk.go.path,
             cout = cout.path,
-            srcs = ' '.join([f.path for f in ctx.files.srcs]),
+            srcs = " ".join([f.path for f in ctx.files.srcs]),
         )
         ctx.actions.run_shell(
             command = cmd,
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -36,10 +36,6 @@ load(
     "LINKMODE_PLUGIN",
     "LINKMODE_SHARED",
 )
-load(
-    "@bazel_skylib//lib:shell.bzl",
-    "shell",
-)
 
 def _go_binary_impl(ctx):
     """go_binary_impl emits actions for compiling and linking a go executable."""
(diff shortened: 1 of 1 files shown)


go/private/rules/binary.bzl
```diff
@@ -118,23 +114,34 @@ def _go_tool_binary_impl(ctx):
...
-                   "rm {out}.a")
-    command = command_tpl.format(
-        go = shell.quote(sdk.go.path),
-        goroot = shell.quote(sdk.root_file.dirname),
-        out = shell.quote(out.path),
+    cout = ctx.actions.declare_file(name + ".a")
+    cargs = ctx.actions.args()
+    cargs.add_all(["tool", "compile"])
+    cargs.add("-o", cout)
+    cargs.add("-I", sdk.root_file.dirname)
```
jayconrod, line 122: It doesn't actually seem like the `-I` flag is necessary, so let's take it out.

go/private/rules/binary.bzl
```diff
@@ -118,23 +114,34 @@ def _go_tool_binary_impl(ctx):
...
+        mnemonic = "GoToolchainBinaryCompile",
     )
 
-    ctx.actions.run_shell(
-        inputs = sdk.libs + sdk.headers + sdk.tools + ctx.files.srcs + [sdk.go],
+    out = ctx.actions.declare_file(name)
+    largs = ctx.actions.args()
+    largs.add_all(["tool", "link"])
+    largs.add("-o", out)
+    largs.add("-L", sdk.root_file.dirname)
```
jayconrod, line 138: It looks like the `-L` flag can be removed, too.

go/private/rules/binary.bzl
```diff
@@ -118,23 +114,34 @@ def _go_tool_binary_impl(ctx):
...
-    command = command_tpl.format(
-        go = shell.quote(sdk.go.path),
-        goroot = shell.quote(sdk.root_file.dirname),
-        out = shell.quote(out.path),
+    cout = ctx.actions.declare_file(name + ".a")
+    cargs = ctx.actions.args()
+    cargs.add_all(["tool", "compile"])
+    cargs.add("-o", cout)
+    cargs.add("-I", sdk.root_file.dirname)
+    cargs.add("-trimpath=.")
```
jayconrod, line 123: This won't work: the compiler treats these paths literally. The path has to be the absolute path to the directory containing the source files, but there's no way during analysis to know what that will be.

I think this means some kind of shell script is still needed. Perhaps on Windows it could be a .bat file instead?

go/private/rules/binary.bzl
tomlu: Are we sure about that? It means we're doing two things in one CL; A pure refactor and some cleanup.

go/private/rules/binary.bzl
tomlu: Could we fix this upstream instead? It seems pretty reasonable to allow relative paths.

tomlu: Are we sure trimpath doesn't work actually? How does this work internally? Are all Go builds surrounded by such a wrapper script? You can't put absolute paths into the command line or it won't be cacheable when distributed.

go/private/rules/binary.bzl
jayconrod: It's pretty low risk, but I'm fine with splitting it into two CLs.

The `-I` and `-L` flags are actually wrong: they should point to `$GOROOT/src` and `$GOROOT/pkg/${GOOS}_${GOARCH}`. But it doesn't seem like the compiler or linker need to be told where to find packages in `GOROOT`, so it never caused problems.

PR #2506, opened 2020-05-15
go: disable race and msan modes when pure mode is set

### LINKED ISSUE #2499: go_binary: setting pure = "on" should disable race mode by default
Interestingly, I didn't think I actually had any cgo code. I wonder if this is because of the --race flag?

Moving io_bazel_rules_go v0.22.4 -> v0.23.0

https://github.com/bazelbuild/bazel-watcher/pull/388

0.21.0

3.1.0

Yes

* Ubuntu 16.04
* Ubuntu 18.04
* OpenJDK Mac
* OpenJDK Windows

https://buildkite.com/bazel/bazel-watcher/builds/1169

They all fail with the same error

```
go install: -race requires cgo; enable cgo by setting CGO_ENABLED=1
```

### What did you do?

Any build

Build success

### What did you see instead?

```
go install: -race requires cgo; enable cgo by setting CGO_ENABLED=1
```

jayconrod:
In go_transition, race and msan settings will be disabled if pure mode
is explicitly enabled. If race or msan are explicitly enabled together
with pure mode, an error will be reported, but this isn't fatal.

In get_mode, --features=race and --features=msan will be ignored when
pure mode is enabled. If race or msan are explicitly enabled together
with pure mode, a fatal error will be reported.

Fixes #2499


q3k:
Any chance we could get a patch release for 0.23 that contains this? As far as I can see it's not in the 0.23.1 prerelease.

jayconrod:
@q3k There will be a patch release this week. I expect it will include this fix.

Issue #2518, opened 2020-05-28
Windows coverage tests are failing in CI


jayconrod:
Some tests are disabled because they are failing:

* `//tests/core/coverage:coverage_test`
* `//tests/core/nogo/coverage:coverage_test`
* Coverage part of `//tests/core/race:race_test`

Errors look like:

```
            FAIL: //:coverrace_test (see C:/b/4xosdfd5/execroot/__main__/bazel-out/x64_windows-fastbuild-ST-e550172346ce7259f92416ab36072baf4df0deb3e748a2c931fafa98361b25ba/testlogs/coverrace_test/test.log)
            ERROR: C:/b/gmkpsojl/bazel_testing/bazel_go_test/main/BUILD.bazel:59:8: output 'coverrace_test/coverage.dat' was not created
            ERROR: C:/b/gmkpsojl/bazel_testing/bazel_go_test/main/BUILD.bazel:59:8: not all outputs were created or valid
            Target //:coverrace_test up-to-date:
```

PR #2509, opened 2020-05-17
go/tools/bazel_testing: provide a way to configure C toolchain in tests

jayconrod:
go_bazel_tests that require cgo have been failing on Windows since the
test environment hasn't been passing additional flags needed to
configure the C toolchain on Windows.

This CL adds GO_BAZEL_TEST_BAZELFLAGS, an environment variable to be
set with --test_env, which provides a list of additional flags to
Bazel.

Fixes #2507

PR #2502, opened 2020-05-14
Setting coverage mode to atomic when race detection is enabled

linzhp:
**What type of PR is this?**
Bug fix

**What does this PR do? Why is it needed?**
`-covermode=set` cannot be run with race detection. This PR set `-covermode=atomic` when race detection is enabled. More information at https://hermanschaaf.com/running-the-go-race-detector-with-cover/

**Which issues(s) does this PR fix?**

Fixes #2498

**Other notes for review**

PR #2519, opened 2020-05-28
tests/core/race: disable cover_race on windows

jayconrod:
Test case was added in #2502. #2509 enabled tests and caused it to
start failing. Skipped for now.

Updates #2518


meteorcloudy:
Thanks for the fix!

PR #2542, opened 2020-06-17
Windows: Use actions.run and explicitly format cmd.exe .bat invocation

sunjayBhatia:
**What type of PR is this?**

Bug fix

- Correctly formats the cmd.exe batch file invocation on Windows
- This was noted as a new failure with commit 280a1ed on particular GCP
RBE environments given some combination of PATHEXT and other system
factors which didn't correctly kick off the given batch file and treated
that invocation as using forward slash notation, confusing the cmd.exe
command invocation.
- See https://dev.azure.com/cncf/4684fb3d-0389-4e0b-8251-221942316e06/_apis/build/builds/42282/logs/32 namely the below

```
ERROR: C:/_eb/external/go_sdk/BUILD.bazel:43:1: GoToolchainBinaryCompile external/go_sdk/builder.exe.a failed (Exit 1): builder.exe.bat failed: error executing command 
  cd C:/_eb/execroot/envoy
  SET GOROOT=external/go_sdk
  bazel-out/host/bin/external/go_sdk/builder.exe.bat
Execution platform: @rbe_windows_msvc_cl//config:platform
'bazel-out' is not recognized as an internal or external command, operable program or batch file.
```

N/A

**Other Notes**

If there is an alternate method of fixing this issue within Bazel, we would be open to it; there seems to be some discrepancy between how `ctx.actions.run` work on Windows when running locally (this change is not required for purely local builds) vs. via RBE (where this patch or reverting the relevant bits of https://github.com/bazelbuild/rules_go/pull/2488 is required).

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -128,8 +128,9 @@ def _go_tool_binary_impl(ctx):
             content = cmd,
         )
         ctx.actions.run(
-            executable = bat,
-            inputs = sdk.libs + sdk.headers + sdk.tools + ctx.files.srcs + [sdk.go],
+            executable = "cmd.exe",
+            arguments = ["/S", "/C", bat.path.replace("/", "\\")],
+            inputs = sdk.libs + sdk.headers + sdk.tools + ctx.files.srcs + [sdk.go, bat],
             outputs = [cout],
             env = {"GOROOT": sdk.root_file.dirname},  # NOTE(#2005): avoid realpath in sandbox
             mnemonic = "GoToolchainBinaryCompile",


sunjayBhatia:
@googlebot I signed it!

sunjayBhatia:
Still working out corporate CLA, we were covered by Pivotal but need to get moved over to VMware

jayconrod:
Sorry, I can't review this until the CLA is done.

In the mean time, could you open an issue describing the problem? From the PR description, it's not really clear what causes this problem.

wrowe:
@googlebot I fixed it

sunjayBhatia:
This actually looks like a general Bazel issue rather than something specific to `rules_go`, this is just the first place we've run into it, you can see https://github.com/greenhouse-org/bazel-issue-repro/tree/master/ctx_actions_run_rbe if youre interested @jayconrod

wrowe:
@googlebot I fixed it

jayconrod:
It looks like the CLA bot still is not happy. It looks like VMware is on the list of corporate signers. Some things to try:

* Make sure all commit and author email addresses are on the approved CLA for your org. You may need to amend the comment and force-push the PR.
* Make sure your GitHub accounts are associated with the same emails.

If that doesn't work, I can escalate to the folks that run the bot.

I'm still not really clear on what's going wrong here though. If this is a general Bazel issue (and from [your repro](https://github.com/greenhouse-org/bazel-issue-repro/tree/master/ctx_actions_run_rbe) it looks like this not Go-specific), it would be better to open an issue in bazelbuild/bazel and close this PR. If this is an issue with rules_go, please open an issue here with more information.

sunjayBhatia:
We've started some dialogue with RBE/Bazel folks and will continue digging into the exact cause of this issue (whether it is RBE or Bazel doing odd things with path conversions); we have a workaround for rules_go in Envoy so we can close this

sunjayBhatia:
https://github.com/bazelbuild/bazel/issues/11636 is our Bazel issue if youre interested

PR #2461, opened 2020-04-29
Copy PKGDEF and nogo facts to the same archive

linzhp:
Fixes #1803 based on the review comments of #2199:

>  We could invoke the compiler without -linkobj. After it runs, the builder could copy the export data from the compiled .a file into the .x file. So the export data would appear in both archives.

So this PR:

* makes the .x file an archive
* makes nogo be able to read the archive to get the facts
* copies `__.PKGDEF` from .a file to .x file
* stops passing .a file and instead always and only passing .x file to compiler

Tested on Uber's Go monorepo. All builds and tests passed.

PR #2585, opened 2020-07-28
Print out subcommand when it fails

linzhp:
**What type of PR is this?**

Feature

`--verbose_failures` only prints out the command that Bazel rules call directly. If a binary calls several subcommands (e.g., compilepkg), we don't know which subcommand fails until we run the parent command manually. For failures that cannot be reproduced locally (e.g., the Windows errors of #2461), printing out the subcommand gives us better idea what could have gone wrong.

### DIFF
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -165,7 +165,7 @@ func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
 	cleanup := passLongArgsInResponseFiles(cmd)
 	defer cleanup()
 	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("error running subcommand %q: %v", buffer.String(), err)
+		return fmt.Errorf("error running the following subcommand: %v\n%s", err, buffer.String())
 	}
 	return nil
 }
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -157,15 +157,14 @@ func absEnv(envNameList []string, argList []string) error {
 }
 
 func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
-	var buffer bytes.Buffer
-	formatCommand(&buffer, cmd)
+	formattedCmd := formatCommand(cmd)
 	if verbose {
-		io.Copy(os.Stderr, &buffer)
+		os.Stderr.WriteString(formattedCmd)
 	}
 	cleanup := passLongArgsInResponseFiles(cmd)
 	defer cleanup()
 	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("error running the following subcommand: %v\n%s", err, buffer.String())
+		return fmt.Errorf("error running the following subcommand: %v\n%s", err, formattedCmd)
 	}
 	return nil
 }
@@ -270,7 +269,7 @@ func absArgs(args []string, flags []string) {
 
 // formatCommand writes cmd to w in a format where it can be pasted into a
 // shell. Spaces in environment variables and arguments are escaped as needed.
-func formatCommand(w io.Writer, cmd *exec.Cmd) {
+func formatCommand(cmd *exec.Cmd) string {
 	quoteIfNeeded := func(s string) string {
 		if strings.IndexByte(s, ' ') < 0 {
 			return s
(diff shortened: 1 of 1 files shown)


go/tools/builders/env.go
```diff
@@ -157,13 +157,15 @@ func absEnv(envNameList []string, argList []string) error {
 }
 
 func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
+	var buffer bytes.Buffer
+	formatCommand(&buffer, cmd)
```
jayconrod, line 161: Nit: I think this would be more readable if `formatBuffer` just returned a `string`. There's probably no good reason for it to accept an `io.Writer` anyway.

robbertvanginkel:
My terminal output has been a bit more verbose on local compile failures recently and I think it might be due to this. Now if I have syntactical error and run a build locally with `--verbose_failures=false`
```
$ bazel test //:test --verbose_failures=false
INFO: Invocation ID: 6aa8fdc4-0a3e-4c64-8b01-7a1b355e495d
INFO: Analyzed 5 targets (0 packages loaded, 0 targets configured).
INFO: Found 4 targets and 1 test target...
ERROR: /Users/robbert/tmp/testing/BUILD.bazel:3:11: GoCompilePkg testing/go_default_library.a failed (Exit 1) builder failed: error executing command bazel-out/host/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix darwin_amd64 -src testing/env.go -importpath example.com/testing -p ... (remaining 13 argument(s) skipped)

/private/var/tmp/_bazel_robbert/b21ffbdb27bc569a3409d8e5058938fa/sandbox/darwin-sandbox/395/execroot/__main__/testing/env.go:7:9: undefined: os.GetEnv
compilepkg: error running the following subcommand: exit status 2
PATH=/usr/bin:external/local_config_cc:/bin \
CGO_ENABLED=1 \
GOARCH=amd64 \
GOPATH= \
GOOS=darwin \
GOROOT=external/go_sdk \
TMPDIR=/var/folders/p4/px0ds0qs4w58bk9fjp34q32h0000gn/T/ \
GOROOT_FINAL=GOROOT \
__CF_USER_TEXT_ENCODING=0x1F5:0x0:0x0 \
external/go_sdk/pkg/tool/darwin_amd64/compile -p example.com/testing -importcfg /private/var/tmp/_bazel_robbert/b21ffbdb27bc569a3409d8e5058938fa/sandbox/darwin-sandbox/395/execroot/__main__/bazel-out/darwin-fastbuild/bin/testing/importcfg581623091 -pack -trimpath=/private/var/tmp/_bazel_robbert/b21ffbdb27bc569a3409d8e5058938fa/sandbox/darwin-sandbox/395/execroot/__main__ -o /private/var/tmp/_bazel_robbert/b21ffbdb27bc569a3409d8e5058938fa/sandbox/darwin-sandbox/395/execroot/__main__/bazel-out/darwin-fastbuild/bin/testing/go_default_library.a -- /private/var/tmp/_bazel_robbert/b21ffbdb27bc569a3409d8e5058938fa/sandbox/darwin-sandbox/395/execroot/__main__/testing/env.go
```
Its pretty hard to spot the `testing/env.go:7:9: undefined: os.GetEnv` in my wrapped terminal output. Before, the error was easier to find and almost always at the bottom:
```
/private/var/tmp/_bazel_robbert/b21ffbdb27bc569a3409d8e5058938fa/sandbox/darwin-sandbox/395/execroot/__main__/testing/env.go:7:9: undefined: os.GetEnv
compilepkg: error running the following subcommand: exit status 2
```
Would it be possible to put this behind a flag or is it possible to make the behavior conditional to having used `--verbose_failures`?

jayconrod:
I tend to agree this is a bit too chatty. @linzhp WDYT?

I don't think Bazel communicates the `--verbose_failures` flag to Starlark rules or actions, so we can't really do anything with that. Since this output is mainly useful for debugging the rules and the builder, how about something like `--action_env=GO_VERBOSE_FAILURES=1`? The builder could just check for that environment variable.

robbertvanginkel:
Could a `config_setting` for verbose_failures be used, like how `--stamp` is passed into `go_config`?https://github.com/bazelbuild/rules_go/blob/c0696e47cc6de36a3a2e30c5bd394822fcd1453f/go/private/BUILD.bazel#L29-L32
https://github.com/bazelbuild/rules_go/blob/f17cccbc0cc35c03d5a62a857aefc71a7e6bec97/BUILD.bazel#L115-L118

jayconrod:
It's possible to define a build setting (like the flags in `//go/config`) that would do this, but build settings go into action cache keys, so using it would cause everything to be rebuilt.

linzhp:
`--action_env=GO_VERBOSE_FAILURES=1` sounds good to me.

PR #2586, opened 2020-07-31
Manifest single entry

dmitrig01:
**What type of PR is this?**

Bug fix

My Bazel seems to output lines in the runfiles manifest where the path and shortpath are the same as just one entry on the line. This causes issues with the runfile manifest parser.

### DIFF
--- a/go/tools/bazel/runfiles.go
+++ b/go/tools/bazel/runfiles.go
@@ -330,12 +330,7 @@ func initRunfiles() {
 				continue
 			}
 			e := bytes.SplitN(line, []byte(" "), 2)
-			if len(e) < 2 {
-				entry := RunfileEntry{ShortPath: string(e[0]), Path: string(e[0])}
-			} else {
-				entry := RunfileEntry{ShortPath: string(e[0]), Path: string(e[1])}
-			}
-
+			entry := RunfileEntry{ShortPath: string(e[0]), Path: string(e[len(e)-1])}
 			if i := strings.IndexByte(entry.ShortPath, '/'); i >= 0 {
 				entry.Workspace = entry.ShortPath[:i]
 				entry.ShortPath = entry.ShortPath[i+1:]
--- a/go/tools/bazel/runfiles.go
+++ b/go/tools/bazel/runfiles.go
@@ -331,11 +331,11 @@ func initRunfiles() {
 			}
 			e := bytes.SplitN(line, []byte(" "), 2)
 			if len(e) < 2 {
-				runfiles.err = fmt.Errorf("error parsing runfiles manifest: %s:%d: no space", manifest, lineno)
-				return
+				entry := RunfileEntry{ShortPath: string(e[0]), Path: string(e[0])}
+			} else {
+				entry := RunfileEntry{ShortPath: string(e[0]), Path: string(e[1])}
 			}
 
-			entry := RunfileEntry{ShortPath: string(e[0]), Path: string(e[1])}
 			if i := strings.IndexByte(entry.ShortPath, '/'); i >= 0 {
 				entry.Workspace = entry.ShortPath[:i]
 				entry.ShortPath = entry.ShortPath[i+1:]


jayconrod:
> My Bazel seems to output lines in the runfiles manifest where the path and shortpath are the same as just one entry on the line. This causes issues with the runfile manifest parser.

I'm not clear on what the problem here is. Could you please open an issue? Be sure to include 1) Bazel version, 2) operating system, 3) whether this happens in an action, a test, or with `bazel run`, 4) a copy of the manifest (at least a snippet showing the problem), 5) instructions to reproduce.

The change in this PR is not correct. There must be a space between the short path and the full file path in the manifest.

dmitrig01:
### Bazel version

``` 
% bazel version
Build label: 2.2.0
Build time: Tue Mar 3 09:28:15 2020 (1583227695)
Build timestamp: 1583227695
Build timestamp as int: 1583227695
```

### operating system

macOS 10.15

### whether this happens in an action, a test, or with bazel run

It happens in `bazel run` for me. I haven't tried in actions or tests.

### a copy of the manifest (at least a snippet showing the problem)

```
% bazel build //jupyter:deploy.apply
% head -n 10 bazel-bin/jupyter/deploy.apply.runfiles_manifest
__init__.py
branch_deploy/kubernetes/kube-config /private/var/tmp/_bazel_dmitri/667cfcf5f282ca2d9f6991a083f72fc0/external/branch_deploy/kubernetes/kube-config
data_platform/bazel/__init__.py
data_platform/bazel/k8s/__init__.py
data_platform/bazel/k8s/darwin_amd64_stripped/uploader /private/var/tmp/_bazel_dmitri/667cfcf5f282ca2d9f6991a083f72fc0/execroot/data_platform/bazel-out/darwin-fastbuild/bin/bazel/k8s/darwin_amd64_stripped/uploader
data_platform/bazel/k8s/templater /private/var/tmp/_bazel_dmitri/667cfcf5f282ca2d9f6991a083f72fc0/execroot/data_platform/bazel-out/darwin-fastbuild/bin/bazel/k8s/templater
data_platform/bazel/k8s/templater.py /Users/dmitri/Projects/data-platform/bazel/k8s/templater.py
data_platform/bazel/k8s/vars /private/var/tmp/_bazel_dmitri/667cfcf5f282ca2d9f6991a083f72fc0/execroot/data_platform/bazel-out/darwin-fastbuild/bin/bazel/k8s/vars
data_platform/bazel/k8s/vars.py /Users/dmitri/Projects/data-platform/bazel/k8s/vars.py
data_platform/deploy/env.bzl /Users/dmitri/Projects/data-platform/deploy/env.bzl
```

### instructions to reproduce

it happens on every target I've tried.

dmitrig01:
Scanning through the manifest, it seems like it's only `__init__.py` files. I bet it's because they are empty and not actually present in the source tree.

dmitrig01:
Maybe we should skip those lines?

jayconrod:
This seems like a bug in Bazel? Even if a file is empty, it should exist somewhere. Does this reproduce with the latest version of Bazel? If so, could you open an issue in bazelbuild/bazel and confirm with them whether this is the correct behavior?

The C++ runfiles library [reports an error](https://github.com/bazelbuild/bazel/blob/441586bec9048ba07d5960d5954c103783eed33f/tools/cpp/runfiles/runfiles_src.cc#L207) for this case. Go should do the same thing.

dmitrig01:
On the other hand, Java [uses the logic that I have in my PR](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java#L207-L209).

I have looked around and haven't actually found a spec for what the runfiles manifest should be, so I'm not sure how to call this a bug if there isn't a defined spec that Bazel is deviating from. However, if there is some spec I don't know about, I'd love to see it :-).

jayconrod:
Closing inactive PRs.

There is unfortunately no runfiles manifest spec. Given a choice of consistency with C++ and Java, I'd choose C++, since Go is more often used together there.

This should be clarified with Bazel folks before a change here though.

PR #2596, opened 2020-08-10
Export zoneinfo from go sdk.

andrewtar:
**What type of PR is this?**

Other

To run the CockroachDB node, you need [to provide an archive of time zones from your environment](https://www.cockroachlabs.com/docs/v20.1/known-limitations.html#location-based-time-zone-names). And if you\`re using Bazel, it would be very convenient to use `zoneinfo.zip` from the go sdk provided by the role_go.
With this export you can refer to this archive like this `"@go_sdk//:lib/time/zoneinfo.zip"`.

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -70,9 +70,7 @@ filegroup(
     ]),
 )
 
-filegroup(
-    name = "zoneinfo",
-    srcs = [
-        "lib/time/zoneinfo.zip",
-    ],
+exports_files(
+    ["lib/time/zoneinfo.zip"],
+    visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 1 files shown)


andrewtar:
> I think you meant @go_sdk//:zoneinfo ?

Sure. It will be even better. To avoid dependence on internal structure

go/private/BUILD.sdk.bazel
```diff
@@ -69,3 +69,10 @@ filegroup(
         "pkg/**",
     ]),
 )
+
+filegroup(
```
jayconrod, line 73: Left a comment here earlier, but GitHub seems to have swallowed it.

Instead of a `filegroup`, I think this should be:

```
exports_files(
    ["lib/time/zoneinfo.zip"],
)
```

jayconrod (commented):
Thanks. It looks like CI is failing on Windows. I think this is an old CI issue, not related to this PR. To confirm, please merge the latest `master`.

jayconrod (approved):
Looks good, thanks!

PR #2607, opened 2020-08-17
Bump validators dependency

Helcaraxan:
**What type of PR is this?**

Other

Updates the dependency on github.com/mwitkow/go-proto-validators to the latest patch version.

**Other notes for review**

The v0.3.1 release of `go-proto-validators` had a broken `go_proto_compiler` rule (due to missing dependencies).

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -185,13 +185,13 @@ def go_rules_dependencies(is_rules_go = False):
     _maybe(
         http_archive,
         name = "com_github_mwitkow_go_proto_validators",
-        # v0.3.1, latest as of 2020-08-11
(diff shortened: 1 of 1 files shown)


jayconrod:
Can this wait a couple days? I'm planning to bump all of the dependencies in `go_rules_dependencies` to their latest versions this week in preparation for `v0.24.0`.

Helcaraxan:
Sure. It can wait. But it will need to get in before the next release as the current version is broken as suggested by my description. :slightly_smiling_face: 

Will also fix the PR to make it pass CI.

jayconrod:
Right, will definitely be updated in the next release. I'll update this together with other dependencies in a later PR. I'd prefer to close this one.

PR #2620, opened 2020-08-24
Add bzl_library targets for Stardoc compatibility

### LINKED ISSUE #2619: rules_go provides no bzl_library targets

v0.23.8

none

3.3.1

Yes

Ubuntu 20.04 x86_64

n/a

### What did you do?

I've defined a [`stardoc`](https://github.com/bazelbuild/stardoc) target for a rule set that itself uses `rules_go`. The `stardoc` rule has a `deps` attribute that expects `bzl_library` targets. The transitive closure of the `deps` attribute must cover all `*.bzl` files that are loaded directly or indirectly by the rules to be documented. `rules_go` does not expose a `bzl_library` attribute, hence I cannot specify it as a dependency on my use-case and cannot generate Stardoc documentation for my use-case.

That rules_go defines a `bzl_library` target for `rules_go` that includes all relevant `.bzl` files.

### What did you see instead?

That rules_go defines no `bzl_library` target.

aherrmann:
**What type of PR is this?**

Bug fix

Adds `bzl_library` targets for `//go:go_lib` and `//proto:proto_lib`. The latter is a dependency of `//go:go_lib`.
These targets are required to generate Stardoc documentation for rule sets that themselves depend on `rules_go`.

Fixes #2619

### DIFF
--- a/go/BUILD.bazel
+++ b/go/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
 filegroup(
     name = "all_files",
     testonly = True,
@@ -19,6 +21,13 @@ filegroup(
         "//go/toolchain:all_rules",
     ],
     visibility = ["//visibility:public"],
+)
+
+bzl_library(
+    name = "go_lib",
+    srcs = [":all_rules"],
+    visibility = ["//visibility:public"],
+    deps = ["//proto:proto_lib"],
 )
 
 toolchain_type(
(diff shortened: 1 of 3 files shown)


jayconrod:
I'd prefer to go with #2621 instead of this PR. It's important that these targets follow conventions and can be maintained automatically.

aherrmann:
> I'd prefer to go with #2621 instead of this PR. It's important that these targets follow conventions and can be maintained automatically.

Yes, that makes sense. Thank you for looking into it so quickly!

PR #2621, opened 2020-08-24
Create bzl_library entries using gazelle-skylark

### LINKED ISSUE #2619: rules_go provides no bzl_library targets

achew22:
There are a few minor manual modifications that needed to be performed.
For example, some `.bzl` files depend on `.bzl` files that are generated
outside of this repo and do not presently exist. `@bazel_tools` is a
great example of this.

Additionally this effort has revealed that bazel gazelle has an explicit
dependency on the private parts of rules_go in
internal/gazelle_binary.bzl on line 21 where it imports from
`go/private:rules/transitions.bzl` which is no longer the path to import
from once I created the BUILD file in `go/private/rules`.

As a temporary fix for this, I'm included a patch that changes that to
point to the new path. I will provide a follow up patch once rules_go
has been released to fix it properly in the gazelle repo and then remove
the patch from rules_go.

Fixed: https://github.com/bazelbuild/rules_go/issues/2619
Bug: https://github.com/bazelbuild/bazel-skylib/issues/250
Bug: https://github.com/bazelbuild/bazel-gazelle/issues/803

**What type of PR is this?**

Feature

Adds a tree of `bzl_library` targets (generated by Gazelle with only small hand modifications for a few places where their deps don't actually exist).

Fixes #2619

**Other notes for review**

This was great feedback for a couple of features that need to be added to the gazelle generator for `bzl_library`

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -27,17 +27,6 @@ load("@io_bazel_rules_go//extras:embed_data_deps.bzl", "go_embed_data_dependenci
 
 go_embed_data_dependencies()
 
-http_archive(
-    name = "rules_proto",
-    sha256 = "4d421d51f9ecfe9bf96ab23b55c6f2b809cbaf0eea24952683e397decfbd0dd0",
-    strip_prefix = "rules_proto-f6b8d89b90a7956f6782a4a3609b2f0eee3ce965",
-    # master, as of 2020-01-06
-    urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/f6b8d89b90a7956f6782a4a3609b2f0eee3ce965.tar.gz",
-        "https://github.com/bazelbuild/rules_proto/archive/f6b8d89b90a7956f6782a4a3609b2f0eee3ce965.tar.gz",
-    ],
-)
-
 # Used by //tests:buildifier_test.
 # Latest release is not compatible with the incompatible bazel flags we use
 # in CI, in particular, --incompatible_load_proto_rules_from_bzl.
--- a/extras/BUILD.bazel
+++ b/extras/BUILD.bazel
@@ -31,5 +31,6 @@ bzl_library(
     name = "embed_data_deps",
     srcs = ["embed_data_deps.bzl"],
     visibility = ["//visibility:public"],
-    deps = ["@bazel_tools//tools/build_defs/repo:git.bzl"],
+    # Don't list dependency on @bazel_tools//tools/build_defs/repo.bzl
+    deps = [],  # keep
 )
--- a/extras/BUILD.bazel
+++ b/extras/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
 filegroup(
     name = "all_rules",
     srcs = glob(["*.bzl"]) + ["//go/private:all_rules"],
@@ -9,4 +11,25 @@ filegroup(
     testonly = True,
     srcs = glob(["**"]),
     visibility = ["//visibility:public"],
+)
+
+bzl_library(
+    name = "bindata",
+    srcs = ["bindata.bzl"],
+    visibility = ["//visibility:public"],
+    deps = ["@io_bazel_rules_go//go:def"],
+)
+
+bzl_library(
+    name = "embed_data",
+    srcs = ["embed_data.bzl"],
+    visibility = ["//visibility:public"],
+    deps = ["@io_bazel_rules_go//go/private:context"],
+)
+
+bzl_library(
+    name = "embed_data_deps",
+    srcs = ["embed_data_deps.bzl"],
+    visibility = ["//visibility:public"],
+    deps = ["@bazel_tools//tools/build_defs/repo:git.bzl"],
 )
--- a/go/private/BUILD.bazel
+++ b/go/private/BUILD.bazel
@@ -34,13 +34,12 @@ config_setting(
 )
 
 bzl_library(
-    name = "common",
-    srcs = ["common.bzl"],
-)
-
-bzl_library(
     name = "context",
     srcs = ["context.bzl"],
+    visibility = [
+        "//extras:__pkg__",  # Manually added
+        "//go:__subpackages__",
+    ],
     deps = [
         ":common",
         ":mode",
@@ -56,6 +55,7 @@ bzl_library(
 bzl_library(
     name = "go_toolchain",
     srcs = ["go_toolchain.bzl"],
+    visibility = ["//go:__subpackages__"],
     deps = [
         "@io_bazel_rules_go//go/private:platforms",
         "@io_bazel_rules_go//go/private:providers",
@@ -71,40 +71,22 @@ bzl_library(
 )
 
 bzl_library(
-    name = "mode",
-    srcs = ["mode.bzl"],
-)
-
-bzl_library(
-    name = "nogo",
-    srcs = ["nogo.bzl"],
-)
-
-bzl_library(
-    name = "platforms",
-    srcs = ["platforms.bzl"],
-)
-
-bzl_library(
-    name = "providers",
-    srcs = ["providers.bzl"],
-)
-
-bzl_library(
     name = "repositories",
     srcs = ["repositories.bzl"],
+    visibility = ["//go:__subpackages__"],
+    # Don't list dependency on @bazel_tools//tools/build_defs/repo:http.bzl
     deps = [
         ":common",
         ":nogo",
         "//go/private/skylib/lib:versions",
         "//proto:gogo",
-        "@bazel_tools//tools/build_defs/repo:http.bzl",
-    ],
+    ],  # keep
 )
 
 bzl_library(
     name = "sdk",
     srcs = ["sdk.bzl"],
+    visibility = ["//go:__subpackages__"],
     deps = [
         "@io_bazel_rules_go//go/private:common",
         "@io_bazel_rules_go//go/private:nogo",
@@ -115,6 +97,40 @@ bzl_library(
 )
 
 bzl_library(
+    name = "common",
+    srcs = ["common.bzl"],
+    visibility = ["//go:__subpackages__"],
+)
+
+bzl_library(
+    name = "mode",
+    srcs = ["mode.bzl"],
+    visibility = ["//go:__subpackages__"],
+)
+
+bzl_library(
+    name = "nogo",
+    srcs = ["nogo.bzl"],
+    visibility = ["//go:__subpackages__"],
+)
+
+bzl_library(
+    name = "platforms",
+    srcs = ["platforms.bzl"],
+    visibility = ["//go:__subpackages__"],
+)
+
+bzl_library(
+    name = "providers",
+    srcs = ["providers.bzl"],
+    visibility = [
+        "//go:__subpackages__",
+        "//proto:__pkg__",  # keep
+    ],
+)
+
+bzl_library(
     name = "sdk_list",
     srcs = ["sdk_list.bzl"],
+    visibility = ["//go:__subpackages__"],
 )
--- a/go/private/BUILD.bazel
+++ b/go/private/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
 filegroup(
     name = "all_rules",
     srcs = glob(["**/*.bzl"]),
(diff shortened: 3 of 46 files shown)


WORKSPACE
```diff
@@ -89,6 +89,9 @@ bazel_skylib_workspace()
 
 http_archive(
     name = "bazel_gazelle",
+    patches = [
+        "//third_party:com_google_github_bazel_gazelle-internal-in-their-new-location.patch",
```
jayconrod, line 93: I'd prefer to fix upstream issues first instead of maintaining a patch here. I'll add a note to bazelbuild/bazel-gazelle#803, which is the only reason we have that internal dependency.

achew22:
@aherrmann, I would like to add you as a `Co-authored-by` on the PR before merging, is that okay with you?

go/private/rules/BUILD.bazel
```diff
@@ -0,0 +1,147 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
```
jayconrod, line 1: Adding a build file here and in other subdirectories of `//go/private` will break old versions of Gazelle.

It's probably the right thing to do, but it would be better to fix Gazelle, then not release this change for a while so people have some time to upgrade. People are much more likely to upgrade rules_go than Gazelle.

achew22: This turned out to be a [relatively simple fix](https://github.com/bazelbuild/bazel-gazelle/pull/879) in gazelle. Hurray! Note that it won't pass until #2621 is merged and released.

WORKSPACE
achew22: I think this is a chicken and egg situation. One of these repos will have to go first. I tried creating an `alias` in the `BUILD` file associated with it, but it turns out `load` statements are interpreted outside of the context of `BUILD` files... so I think we're stuck unfortunately. I'm happy to merge these in rapid fire though

1) commit rules_go with the patch
2) update Gazelle to refer to the new version of rules_go
3) update rules_go to refer to the new version of Gazelle

achew22:
@jayconrod, I'm looking through the error output:

```
    --- FAIL: Test/test_feature (1.74s)
        race_test.go:269: running: bazel test --features=race //:racy_test --test_arg=-wantrace=true
        race_test.go:277: unexpected build failure: exit status 1
            stderr:
            Loading: 
            Loading: 0 packages loaded
            INFO: Build option --features has changed, discarding analysis cache.
            Analyzing: target //:racy_test (0 packages loaded, 0 targets configured)
            INFO: Analyzed target //:racy_test (0 packages loaded, 7299 targets configured).
            INFO: Found 1 test target...
            [0 / 3] [Prepa] BazelWorkspaceStatusAction stable-status.txt
            ERROR: C:/b/udu4fkru/bazel_testing/bazel_go_test/main/BUILD.bazel:27:8: GoLink racy_test_/racy_test.exe failed (Exit 1)
            c:/tools/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/lib/../lib/libmsvcrt.a(/2203): duplicate symbol reference: _unlock_file in both libgcc(.text) and libgcc(.data)
            libgcc(.text): relocation target ___chkstk_ms not defined
            libgcc(.text): relocation target atexit not defined
            link: error running the following subcommand: exit status 2
            CGO_ENABLED=1 \
            GOARCH=amd64 \
            GOOS=windows \
            GOPATH= \
            GOROOT=bazel-out/x64_windows-fastbuild/bin/external/io_bazel_rules_go/stdlib_ \
            GOROOT_FINAL=GOROOT \
            PATH=c:/tools/msys64/mingw64/bin \
            SYSTEMDRIVE=C: \
            SYSTEMROOT=C:\Windows \
            TEMP=C:\temp \
            TMP=C:\temp \
            external\go_sdk\pkg\tool\windows_amd64\link.exe -importcfg C:\b\gxuofdwa\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\racy_test_\importcfg627235515 -o C:\b\gxuofdwa\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\racy_test_\racy_test.exe -extld c:/tools/msys64/mingw64/bin/gcc -race -s -w -buildid=redacted C:\b\gxuofdwa\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\racy_test.a
            [13 / 13] 1 / 1 tests, 1 failed; checking cached actions
            Target //:racy_test failed to build
            INFO: Elapsed time: 1.711s, Critical Path: 1.13s
            INFO: 4 processes: 4 local.
```

Is that the output that you would expect to see if the race were back, or has this ticked something new? Also an interesting data point, I looked up [@aherrmann's run](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/8e8f5441-9868-4568-bb18-b7af728d0990/tests%5Ccore%5Crace%5Crace_test%5Cattempt_1.log) only to discover that we have the same set of failures (the [complete run](https://buildkite.com/bazel/rules-go-golang/builds/2341#8e8f5441-9868-4568-bb18-b7af728d0990)).

I'm happy to dig into this if it were anything other than a test with the word "race" in it. That usually means "this is something that the maintainer spent a lot of time thinking about... ask me before messing with it".

aherrmann:
@achew22 Thank you for implementing this! I've confirmed that this works for my use-case.

> @aherrmann, I would like to add you as a `Co-authored-by` on the PR before merging, is that okay with you?

Yes, absolutely. Thank you!

achew22:
@aherrmann, I pushed the change listing you as a co-author. Could I have you appease the CLA robot?

aherrmann:
@googlebot I consent.

achew22:
@jayconrod what would next steps be here? Do you want to wait more before we make this change or should I clean it up and get it passing on TAP?

jayconrod:
@achew22 Updating to the latest Gazelle broke some tests. I'll do that as soon as I have the time. I tried to do it while updating the dependencies before `v0.24.0`, but it looked like it would take a few hours, and since it's just a test dependency, I put it on the back burner.

I apologize for the delay. This is an extremely busy time. I've been trying to get a few features into the Go command before we can flip `GO111MODULE=on` by default in development (golang/go#41330). We need to do that as early within the development cycle as possible.

achew22:
No worries at all. That work is important ecosystems wide and I appreciate you prioritizing it above this. I'll ping you again about this after a small delay. Thanks for all that you do!

achew22:
Friendly ping

jayconrod:
I should be able to get to this in a couple days. I was sprinting to get [CL 255052](https://go-review.googlesource.com/c/go/+/255052) and its prerequisite features merged. Now that it's done, I should have time to work on more things.

achew22:
@jayconrod I won't do this on the gerrit because it has such a wide distribution and I don't wanna bother ppl, but I wanted to say, great job! You pour so much of yourself into this and have made it a delightful place to code. Thank you, Jay!

hugelgupf:
CLA bot doesn't seem to recognize the co-author email address in the commit as one of @aherrmann's, just FYI. I can't figure out why, though.

jayconrod:
This should be unblocked by #2658.

CLA verified manually.

achew22:
@jayconrod sorry but pushing broke the CLA again 😦 could I have you poke it one more time?

extras/BUILD.bazel
```diff
@@ -10,3 +12,32 @@ filegroup(
...
+        "@io_bazel_rules_go//go/private/rules:rule",
+    ],
+)
+
+bzl_library(
+    name = "embed_data_deps",
+    srcs = ["embed_data_deps.bzl"],
+    visibility = ["//visibility:public"],
+    deps = [
+        #"@bazel_tools//tools/build_defs/repo:git",
```
jayconrod, line 41: Add comment and `# keep` (assuming this is the right reason):

```
# Don't list dependency on @bazel_tools//tools/build_defs/repo.bzl
deps = [], # keep
```

go/private/BUILD.bazel
```diff
@@ -30,3 +37,103 @@ config_setting(
...
 )
+
+bzl_library(
+    name = "context",
+    srcs = ["context.bzl"],
+    visibility = [
+        "//extras:__pkg__",
+        "//go:__subpackages__",
+    ],
+    deps = [
```
jayconrod, line 48: Same deal with `# keep` here.

In general, if there's any manual change that will get removed by another run of Gazelle, let's add a `# keep` to prevent that from happening.

go/private/rules/BUILD.bazel
jayconrod: This is #2621; did you mean a different PR?

tests/BUILD.bazel
```diff
@@ -1,10 +1,20 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
```
jayconrod, line 1: Let's exclude the `//tests` directory and not make changes here. None of the `.bzl` files here should be imported by other repos, and almost nothing here follows Gazelle conventions.

achew22:
@jayconrod I took a look at updating this to the `main` branch but I have to admit I'm slightly confused.

There are a number of places where I think there is a dangling import. For example, in [`transition.bzl`](https://github.com/bazelbuild/rules_go/blob/312a801084712314ea7da6f1c226306af58eb097/go/private/rules/transition.bzl#L16) the `load` calls for `LINKMODES` from `:mode.bzl`. When I look in that same wd, [`go/private/rules`](https://github.com/bazelbuild/rules_go/tree/master/go/private/rules), I don't see a file with that name. I do, however, see that defined in a file at [`go/private/mode.bzl`](https://github.com/bazelbuild/rules_go/blob/312a801084712314ea7da6f1c226306af58eb097/go/private/mode.bzl#L29), but that doesn't exist in the correct directory.

Maybe I'm doing something really silly, but I don't understand how this code works. My rule of thumb is, when something does work but you can't explain how it does it's a good time to ask questions. So, can you help me understand how these imports are working? Is there some kind of `PATH` variable for imports? Do imports occur relative to the place that they are being imported from? Am I completely wrong here in my understanding of the dir structure?

Sorry to put the burdeon back onto you, but I'm mighty confused. I'll take another look in the AM, but I wanted to write out my thoughts.

Thanks again!

jayconrod:
Labels like `:mode.bzl` refer to files in the same Bazel package using a path relative to the directory containing the `BUILD` or `BUILD.bazel` file. So from `go/private/rules/transition.bzl`, the enclosing build file is `go/private/BUILD.bazel`, so the path is relative to `go/private`.

Since this PR adds build files in directories that didn't have them, some of those loads will need adjustment. That one will need to be changed to `//go/private:mode.bzl`.

achew22:
Interesting discovery, looks like `rules_proto` was already accidentally upgraded!

I was expecting 

https://github.com/bazelbuild/rules_go/blob/99e0cf0e35f21ae9ba601e3c42d133c23abf1c25/WORKSPACE#L30-L39

To be the thing that imported `rules_proto`, but it turns out that when `com_google_protobuf` was upgraded, the preexisting invocation of

https://github.com/bazelbuild/rules_go/blob/99e0cf0e35f21ae9ba601e3c42d133c23abf1c25/WORKSPACE#L22-L24

Now invokes

https://github.com/protocolbuffers/protobuf/blob/666b592e36fbebeaaccbee8714f50ed81e801fee/protobuf_deps.bzl#L51-L57

Which brings in the version of `rules_proto` that I thought we needed to upgrade to! So I removed it in this CL from the `WORKSPACE` to make things a little less confusing.

tests/BUILD.bazel
achew22: Done

go/private/rules/BUILD.bazel
achew22: Sorry, I meant https://github.com/bazelbuild/bazel-gazelle/pull/760

achew22: Turns out we were able to resolve it none the less.

jayconrod (approved):
Looks good I think.

CI seems to have gone off into the weeds, so I've canceled it and restarted.

jayconrod:
CI's remote cache seems to be down. Opened bazelbuild/continuous-integration#1055 for tracking.

This is a large enough change that I'd like to wait for CI to be green on all platforms before submitting.

achew22:
> This is a large enough change that I'd like to wait for CI to be green on all platforms before submitting.

I couldn't agree more. Thanks for not submitting it.

achew22:
Looks like CI is happier with this now. Could you take another look? @jayconrod

jayconrod:
Yep, all clear. Thanks for persevering on this!

PR #2642, opened 2020-09-04
Use absolute paths in GoPath builder (mainly for Windows compatibility)

Xjs:
**What type of PR is this?**

Bug fix

Windows has a limit of 256 characters for relative characters when using the syscalls the Go standard library uses. In a Bazel project I encountered recently I couldn't run the GoPath builder because of an `google.golang.org/grpc` dependency that included the file `bazel-out\x64_windows-fastbuild\bin\external\org_golang_google_grpc\reflection\grpc_reflection_v1alpha\grpc_reflection_v1alpha_go_proto_\google.golang.org\grpc\reflection\grpc_reflection_v1alpha\reflection.pb.go`, which even with my very short bazel root of `C:\tmp\bzl` was too long to open, and resulted in the error message `GoPath: open bazel-out\x64_windows-fastbuild\bin\external\org_golang_google_grpc\reflection\grpc_reflection_v1alpha\grpc_reflection_v1alpha_go_proto_\google.golang.org\grpc\reflection\grpc_reflection_v1alpha\reflection.pb.go: The system cannot find the path specified.` (pasting verbatim so people googling it can find it)

Using abs() and FromSlash() like in other builders fixes this issue.

### DIFF
--- a/go/tools/builders/BUILD.bazel
+++ b/go/tools/builders/BUILD.bazel
@@ -69,7 +69,11 @@ go_source(
 
 go_binary(
     name = "go_path",
-    srcs = ["go_path.go"],
+    srcs = [
+        "env.go",
+        "flags.go",
+        "go_path.go"
+    ],
     visibility = ["//visibility:public"],
 )
 
--- a/go/tools/builders/BUILD.bazel
+++ b/go/tools/builders/BUILD.bazel
@@ -72,7 +72,7 @@ go_binary(
     srcs = [
         "env.go",
         "flags.go",
-        "go_path.go"
+        "go_path.go",
     ],
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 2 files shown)


go/tools/builders/BUILD.bazel
```diff
@@ -69,7 +69,11 @@ go_source(
 
 go_binary(
     name = "go_path",
-    srcs = ["go_path.go"],
+    srcs = [
+        "env.go",
+        "flags.go",
+        "go_path.go"
```
jayconrod, line 75: Please run buildifier on this file. I think the comma is missing on the end of this line.

jayconrod (commented):
One comment to make buildifier happy. Looks good otherwise. Thanks for fixing this.

go/tools/builders/BUILD.bazel
Xjs: You are right. I ran buildifier and pushed again.

jayconrod (approved):
Looks good, thanks!

PR #2651, opened 2020-09-11
Fix positional argument in docs

hugelgupf:
**What type of PR is this?**

Documentation

When I copy and paste this line, I get

Error in label: label() got unexpected positional argument

from bazel.


jayconrod (approved):
Thanks for fixing this!

PR #2652, opened 2020-09-11
archive: use library's name for archive name

hugelgupf:
**What type of PR is this?**

Bug fix

If a custom Go rule generates more than one library, or otherwise
changes the name of the GoLibrary, the emitted archive will still have
the _ctx.label.name. To deconflict, use the library's name as the
emitted archive's name.

In the normal case, this is already the default, so it shouldn't hurt.

In the [very work in progress PR here](https://github.com/u-root/gobusybox/pull/23/files) I create an aspect that applies to go_binary rules. It takes either its srcs, or every embed'ed library's sources and rewrites them such that the go_binary is importable as a go_library. (The transformation is documented [here](https://github.com/u-root/gobusybox#command-transformation).)

This means every go_binary this is applied to generates 2 of GoLibrary, GoSource, and GoArchive. The aspect, of course, groups them into a new aspect (GoBusyboxLibrary) since the rule itself cannot return more than one aspect.

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -56,10 +56,10 @@ def emit_archive(go, source = None, _recompile_suffix = ""):
         pre_ext = ".external"
     if _recompile_suffix:
         pre_ext += _recompile_suffix
-    out_lib = go.declare_file(go, ext = pre_ext + ".a")
+    out_lib = go.declare_file(go, name = source.library.name, ext = pre_ext + ".a")
 
     # store __.PKGDEF and nogo facts in .x
-    out_export = go.declare_file(go, ext = pre_ext + ".x")
+    out_export = go.declare_file(go, name = source.library.name, ext = pre_ext + ".x")
     out_cgo_export_h = None  # set if cgo used in c-shared or c-archive mode
 
     direct = [get_archive(dep) for dep in source.deps]


go/private/actions/archive.bzl
```diff
@@ -56,10 +56,14 @@ def emit_archive(go, source = None, _recompile_suffix = ""):
         pre_ext = ".external"
     if _recompile_suffix:
         pre_ext += _recompile_suffix
-    out_lib = go.declare_file(go, ext = pre_ext + ".a")
+
+    name = ""
```
jayconrod, line 60: I think this can be simplified to `name = source.library.name`: no if statement needed. If `name` is falsy, then `_child_name` will use `go._ctx.label.name`, as it does today.

jayconrod (commented):
One small comment, looks good though.

Kind of nervous about making a change like this, but I can see how it's necessary, and I'm pretty sure it will be a no-op for `go_binary`, `go_library`, etc.

hugelgupf:
I think I just thought of a way around this, which actually solves another problem as well. I saw that GoSource.deps wants a list of targets, but I supply a list of GoArchives - because of the double provider problem, where if I used the original target, it'd get the original unmodified GoArchive. There's a comment somewhere saying that this happens to work but you want to deprecate GoArchive passing and mandate Targets.

Instead, I need to use a macro to generate a rule for each go_binary that the aspect is supposed to apply to, and then depend on that rule, which produces normal GoSource/GoArchive providers.

hugelgupf:
With a newly generated rule, the name problem would be solved as well (I think), but I *think* this PR might still be useful anyway.

jayconrod:
I wonder if an aspect is strictly necessary? Are you transforming all packages or only main packages? If it's only main packages, you could use a different rule instead of `go_binary` that does the transformation on its `srcs` internally. [`gazelle_binary`](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/gazelle_binary.bzl) is kind of like that.

hugelgupf:
Yeah, I think it is unnecessary when you put it like that. The aspect only applies to the top-level go_binary anyway, and the providers give all necessary information already - the recursion is not really needed (except to collect all transitive deps Targets, but that's already a different aspect).

hugelgupf:
I'm pretty sure the implementation won't be much different from the aspect as it is. Basically just s/ctx.rule/ctx.

jayconrod:
Better to avoid aspects if possible. They add a lot of complication and analysis overhead for large builds.

I think it's still a good idea to fix and submit this PR though. It ought to be possible for a target to create multiple archives.

go/private/actions/archive.bzl
hugelgupf: done.

hugelgupf:
Done, PTAL.

You were right about the aspect; much nicer as a rule. It also made accumulating the Go sources easier -- I was looking at either/both srcs and embed before, and now I can just look at the GoSource provider. In fact, it also fixed cross-compilation. (Though thinking about that now, it probably needed to be go_context(ctx.rule) and not go_context(ctx)? No clue.)

jayconrod (approved):
Looks good, thanks!

Issue #2592, opened 2020-08-07
Build failure in github.com/ulikunitz/xz under coverage


linzhp:
Bazel fails to build `@com_github_ulikunitz_xz//lzma:go_default_test` under `bazel coverage`, although it succeeds in `go test -cover`:

```bash
$ go get github.com/ulikunitz/xz
$ go test -cover github.com/ulikunitz/xz/lzma
ok  	github.com/ulikunitz/xz/lzma	2.455s	coverage: 80.6% of statements
```

### What version of rules_go are you using?

0.23.7

### What version of gazelle are you using?

0.21.1

### What version of Bazel are you using?

3.4.1

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Darwin and Linux amd 64

### What did you do?

Add the following rule to a WORKSPACE file:

```Python
go_repository(
    name = "com_github_ulikunitz_xz",
    importpath = "github.com/ulikunitz/xz",
    sum = "h1:YvTNdFzX6+W5m9msiYg/zpkSURPPtOlzbqYjrFn7Yt4=",
    version = "v0.5.7",
)
```

Run:
```bash
$ bazel coverage @com_github_ulikunitz_xz//lzma:go_default_test
```

### What did you expect to see?

Build and test succeed

### What did you see instead?

```
ERROR: /private/var/tmp/_bazel_zplin/029a80a95b282f3a98937d78e6378829/external/com_github_ulikunitz_xz/lzma/BUILD.bazel:42:8: GoCompilePkg external/com_github_ulikunitz_xz/lzma/go_default_test.internal.a failed (Exit 1) builder failed: error executing command bazel-out/host/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix darwin_amd64 -src external/com_github_ulikunitz_xz/lzma/bintree.go -src ... (remaining 151 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
compilepkg: error running subcommand: exit status 2
/private/var/tmp/_bazel_zplin/029a80a95b282f3a98937d78e6378829/sandbox/darwin-sandbox/5/execroot/__main__/external/com_github_ulikunitz_xz/lzma/bitops.go:7: syntax error: missing import path
[...]

PR #2656, opened 2020-09-19
Redirecting all stdout of subcommands to stderr

linzhp:
**What type of PR is this?**

Bug fix

`go tool compile` writes syntax errors to [stdout](https://github.com/golang/go/blob/a3868028ac8470d1ab7782614707bb90925e7fe3/src/cmd/compile/internal/syntax/parser.go#L361), including the one reported in #2592. Meanwhile, all regular output of subcommands should have been written to files. We should direct all stdout of subcommands to stderr, so we can find all errors in stderr. This mirrors behavior in the [go command](https://go.googlesource.com/go/+/refs/tags/go1.15.2/src/cmd/go/internal/work/exec.go#1958)

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -433,7 +433,8 @@ func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, asmHdrPath
 	args = append(args, "--")
 	args = append(args, srcs...)
 	absArgs(args, []string{"-I", "-o", "-trimpath", "-importcfg"})
-	return goenv.runCommand(args)
+	// "go tool compile" writes syntax errors to stdout. Redirecting them to stderr
+	return goenv.runCommandToFile(os.Stderr, args)
 }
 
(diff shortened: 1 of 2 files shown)


go/tools/builders/compilepkg.go
```diff
@@ -433,7 +433,8 @@ func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, asmHdrPath
 	args = append(args, "--")
 	args = append(args, srcs...)
 	absArgs(args, []string{"-I", "-o", "-trimpath", "-importcfg"})
-	return goenv.runCommand(args)
+	// "go tool compile" writes syntax errors to stdout. Redirecting them to stderr
```
jayconrod, line 436: Instead of changing this file, let's change `env.runCommand`.

Add a comment that this mirrors behavior in the `go` command: https://go.googlesource.com/go/+/refs/tags/go1.15.2/src/cmd/go/internal/work/exec.go#1958

jayconrod (commented):
It's super-weird to me that the compiler writes to stdout, but it looks like `cmd/go` redirects that to stderr, so I guess rules_go should, too.

jayconrod (approved):
Looks good, thanks!

PR #2675, opened 2020-10-12
Exclude external repositories in Nogo

### LINKED ISSUE #2513: nogo: include and exclude targets in configuration
```
 http_archive(
        name = "io_bazel_rules_go",
        sha256 = "87f0fb9747854cb76a0a82430adccb6269f7d394237104a4523b51061c469171",
        urls = [
            "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.23.1/rules_go-v0.23.1.tar.gz",
            "https://github.com/bazelbuild/rules_go/releases/download/v0.23.1/rules_go-v0.23.1.tar.gz",
        ],
    )
```

```
    http_archive(
        name = "bazel_gazelle",
        sha256 = "bfd86b3cbe855d6c16c6fce60d76bd51f5c8dbc9cfcaef7a2bb5c1aafd0710e8",
        urls = [
            "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.21.0/bazel-gazelle-v0.21.0.tar.gz",
            "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.21.0/bazel-gazelle-v0.21.0.tar.gz",
        ],
    )
```
`bazel 3.1.0-homebrew`

Yes

Macos

I am using go modules. 

### What did you do?
`  go_register_toolchains(nogo = "@io_bazel_rules_go//:tools_nogo")`
I want to use nogo. I followed the instructions ( I know it mentions this) and it lines external packages which I have no control over. Is there a way I can limit it my code ? 

Lint only my code 

### What did you see instead?
Lints everything in the external. 
```
external/org_golang_x_tools/go/vcs/vcs.go:532:2: declaration of "metaImport" shadows declaration at line 599
```

tommyknows:
**What type of PR is this?**

Feature

Adds an attribute "importpath" to the nogo rule in order to restrict nogo to only run on a specific Go module. This way, users have the ability to not check external repositories / go libraries that are not in their control.

Fixes #2513 

**Other notes for review**

There's no tests yet. I'm not sure where & how these would be implemented, if at all. Let me know if there's anything i can do for that.

### DIFF
--- a/go/private/rules/nogo.bzl
+++ b/go/private/rules/nogo.bzl
@@ -47,6 +47,8 @@ def _nogo_impl(ctx):
     if ctx.file.config:
         nogo_args.add("-config", ctx.file.config)
         nogo_inputs.append(ctx.file.config)
+    if ctx.attr.importpath:
+        nogo_args.add("-importpath-restriction", ctx.attr.importpath)
     ctx.actions.run(
         inputs = nogo_inputs,
         outputs = [nogo_main],
@@ -91,6 +93,10 @@ nogo = rule(
         ),
         "config": attr.label(
             allow_single_file = True,
+        ),
+        "importpath": attr.string(
+            mandatory = False,
+            doc = "Restrict nogo checks to the given importpath",
         ),
         "_nogo_srcs": attr.label(
             default = "@io_bazel_rules_go//go/tools/builders:nogo_srcs",
--- a/go/tools/builders/generate_nogo_main.go
+++ b/go/tools/builders/generate_nogo_main.go
@@ -50,6 +50,11 @@ var analyzers = []*analysis.Analyzer{
 {{- end}}
 }
 
+// importPathRestriction restricts the nogo checks to a specific import path.
+// With this, nogo can run only on a specific project / import path, instead
+// of all Go code, including external dependencies.
+var importPathRestriction = {{ printf "%q" .ImportPathRestriction}}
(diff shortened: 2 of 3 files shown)


jayconrod:
I don't think this configuration should be done in a new attribute. It should be in the [nogo configuration file](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst#configuring-analyzers) instead.

I think it would be reasonable to add fields like `only_targets` and `exclude_targets`, like the `only_files` and `exclude_files` fields but the keys would be Bazel label patterns (like `@my_repo//...`) instead file regular expressions matching file names.

`//tests/core/nogo/config:config_test` would need an update, as would the documentation.

tommyknows:
I agree, putting this in the config file would probably make more sense.
From what I understand, the config file doesn't support "global" attributes though, is this correct?
It would be nice if it was possible to add `only_targets` and `exclude_targets` globally, instead of needing to configure every analyser separately.

I'm not sure if / when I will have time to look at that though.

jayconrod:
> From what I understand, the config file doesn't support "global" attributes though, is this correct?

That's correct for now, but it could change. I'll post a comment in #2513.

robfig:
Closing, since it sounds like we don't want to do this. Please feel free to reopen if I'm wrong.

PR #2683, opened 2020-10-22
Migrate to the modern linker input API.

benjaminp:
**What type of PR is this?**

Other

Migrates cgo code away from deprecated Bazel cc Starlark apis.

Fixes #

**Other notes for review**

See https://github.com/bazelbuild/bazel/issues/10860.

### DIFF
--- a/go/private/common.bzl
+++ b/go/private/common.bzl
@@ -208,7 +208,7 @@ def has_versioned_shared_lib_extension(path):
 
     return True
 
-MINIMUM_BAZEL_VERSION = "2.2.0"
+MINIMUM_BAZEL_VERSION = "3.0.0"
 
 def as_list(v):
     """Returns a list, tuple, or depset as a list."""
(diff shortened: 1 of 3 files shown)


benjaminp:
I would also need to change the minimum Bazel version to 3.0.0.

go/private/common.bzl
```diff
@@ -208,7 +208,7 @@ def has_versioned_shared_lib_extension(path):
 
     return True
 
-MINIMUM_BAZEL_VERSION = "2.2.0"
+MINIMUM_BAZEL_VERSION = "3.0.0"
```
jayconrod, line 211: Please also update the minimum version in README.rst.

jayconrod (commented):
Change looks good. CI is down right now (bazelbuild/continuous-integration#1055), so I'll check back on this later.

go/private/common.bzl
benjaminp: Okay, I wasn't sure if I should do that, since it specifically references the latest released version of rules_go.

jayconrod: You're right, it's actually better to leave that for the next minor release. Forget my earlier comment.

PR #2696, opened 2020-10-28
go_test: change directory before use init() functions run

dragonsinth:
Fixes #1918 go_test: data not available in init()

**What type of PR is this?**

> Bug fix

**What does this PR do? Why is it needed?**

Forces a new package, `go/tools/test_init` to have its package initializer run before any Go user code can be initialized.  This package sets the correct working directory, PWD, and TMPDIR to match bazel standards.

**Which issues(s) does this PR fix?**

Fixes #1918

PR #2698, opened 2020-10-31
go_test: add tests for #2696

dragonsinth:
**What type of PR is this?**

> Test coverage

Adds test coverage for #2696


dragonsinth:
@jayconrod this should be a simple test-coverage fix

jayconrod (approved):
Thanks for adding this!

PR #2699, opened 2020-11-04
Update macOS -install_name linker flag

keith:

The macOS toolchain is gaining support for params files for link
commands. As part of this we're removing support for arguments with
leading `@`s. This change should be a no-op.

Compatibility with https://github.com/bazelbuild/bazel/pull/12265


jayconrod (approved):
Looks good, thanks!

keith:
Thanks!

PR #2700, opened 2020-11-05
Remove reference to platforms repository

comius:
It is better to use version that comes with bazel, because no version
bumping is needed.

Referring to a fixed version resulted in two broken downstream project:
FlatBuffers and Cloud Robotics core, which depended on rules_go version
0.20.3, which depends on an old version of platforms, whcih is missing
definitions of macos os and arm64 cpus, whcih is needed by the latest
bazel release.

**What type of PR is this?**
Bug fix

It won't cause seemingly unrelated downstream project failures when platforms repository is updated within bazel.

**Other notes for review**
Example failures at: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1746

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -38,21 +38,6 @@ def go_rules_dependencies(is_rules_go = False):
     if getattr(native, "bazel_version", None):
         versions.check(MINIMUM_BAZEL_VERSION, bazel_version = native.bazel_version)
 
-    # Repository of standard constraint settings and values.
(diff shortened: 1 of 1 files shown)


comius:
friendly ping

jayconrod:
Sorry, `@platforms` is needed for compatibility with bazelbuild/bazel#8622.

You don't have to use the `platforms` declaration from rules_go. If you need a higher version, declare it before calling `go_rules_dependencies`.

PR #2706, opened 2020-11-09
README.rst: use `protos` instead of `proto`

Saser:
**What type of PR is this?**
Documentation

The documentation for `go_proto_library` states that the `protos` attribute should be used instead of the old `proto` attribute. Updating the README might help with nudging people to use the new `protos` attribute.

N/A

**Other notes for review**
N/A


jayconrod (approved):

PR #2713, opened 2020-11-12
Support Go 1.15.5, 1.14.12

### LINKED ISSUE #2710: Support Go 1.15.5, 1.14.12

jayconrod:
Fixes #2710

### DIFF
--- a/go/private/sdk_list.bzl
+++ b/go/private/sdk_list.bzl
@@ -12,11 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DEFAULT_VERSION = "1.15.4"
+DEFAULT_VERSION = "1.15.5"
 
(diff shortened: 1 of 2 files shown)


jayconrod:
There's some new cgo flag validation code that's causing problems. Looking into it.

jayconrod:
Specifically, the new cgo validation code forbids flags in the `CGO_LDFLAGS` environment variable that can't also be used in `#cgo CGO_LDFLAGS` directives in source files. Previously, the set of flags allowed in source files were restricted to prevent builds from executing arbitrary code (via compiler plugins and such). But the user has control over the environment variable so anything was allowed there.

Bazel passes lots toolchain-specific flags through `CGO_LDFLAGS` which aren't covered in the security allowlist.

jayconrod:
Upstream issue is golang/go#42565.

PR #2716, opened 2020-11-13
Support go1.15s non Os.Exit TestMain

rabbbit:
Based on https://github.com/golang/go/issues/34129
And in particular https://go-review.googlesource.com/c/go/+/219639/10/src/cmd/go/internal/load/test.go#709

**What type of PR is this?**

 Bug fix

Matches TestMain behaviour of Go.15

According to the docs above, small fixes don't need issues.

**Other notes for review**

I don't know `rules_go` compatibility. This needs Go 1.15 - should I be checking for go version somewhere,
or is following Go head fine?


linzhp:
I was  about to create report and issue and  found this PR. For make it easier to test, here is a small use case:

```
-- WORKSPACE --

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "207fad3e6689135c5d8713e5a17ba9d1290238f47b9ba545b63d9303406209c6",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.24.7/rules_go-v0.24.7.tar.gz",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.24.7/rules_go-v0.24.7.tar.gz",
    ],
)

go_rules_dependencies()

go_register_toolchains()
-- test/BUILD.bazel --
load("@io_bazel_rules_go//go:def.bzl", "go_test")

go_test(
    name = "go_default_test",
    srcs = ["run_test.go"],
)
-- test/run_test.go --
package test

import (
	"fmt"
	"testing"
)

func TestMain(m *testing.M) {
	defer func() {
		fmt.Println("hello")
	}()
	m.Run()
}

func TestHelloWorld(t *testing.T) {
	t.Fatal("fail on purpose")
}
```

tests/core/go_test/README.rst
```diff
@@ -97,3 +98,9 @@ indirect_import_test
 Checks that an external test can import another package that imports the library
 under test. The other package should be compiled against the internal test
 package, not the library under test. Verifies `#1877`_.
+
+testmain_bad_test
+-----------------
+
+Checks that TestMain without calling Os.Exit directly works.
```
jayconrod, line 105: Nit: `os.Exit`

tests/core/go_test/testmain_bad_test.go
```diff
@@ -0,0 +1,13 @@
+// +build !good
```
jayconrod, line 1: Remove this constraint? Not sure why it's here.

Also, please add a copyright notice (copy from another file).

tests/core/go_test/testmain_bad_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+package testmain_bad
+
+import "testing"
+
+func TestMain(m *testing.M) {
```
jayconrod, line 7: It looks like the correct behavior is for this test to always fail. That will break CI, since we essentially run `bazel test //...`.

Please convert this to a `go_bazel_test` that sets up a simple workspace, runs `bazel test` on a target, and verifies that the test fails. See `test_filter_test` for a simple example. Specifically, check that Bazel's exit code is `bazel_testing.TESTS_FAILED`.

jayconrod (commented):
Thanks for working on this. A few comments below.

tests/core/go_test/testmain_bad_test.go
rabbbit: I copied this from `tags_bad_test.go` - I thought that this was how you mark tests as failing (since that test literally calls `t.Fail()`).

I saw your pointer to a proper fix in the comment below, thanks!

tests/core/go_test/testmain_bad_test.go
rabbbit: Yeah, I knew it was supposed to fail, didn't know how to test it (thought the `// +build !good` is the trick, see comment above).

Thanks for the pointer.

rabbbit:
@jayconrod absolutely no rush, but I believe this is ready for another look. I'm not sure what the protocol here is - I guess I'll ping you every week or so ;)

jayconrod (approved):
Looks good, thanks!

Issue #1531, opened 2018-06-07
stdlib not being cached on remote server


kwiesmueller:
We finally got our project to build successfully using bazel, go and docker on cloudbuild but we are still seing very high build times on all cloud builds (up to 6mins) for all kinds of changes (changing just a comment or the readme).

This seems to be caused by the GoStdlib being built every time.
We are using this command to build:
```
bazel run \
	--remote_http_cache=https://storage.googleapis.com/cache-xxx \
	--google_credentials=$(GOOGLE_KEYFILE) \
	--spawn_strategy=remote --genrule_strategy=remote \
	--workspace_status_command=./helpers/status.sh //:push
```
This are the logs:
```
Finished Step #1
Step #1: INFO: Build completed successfully, 118 total actions
Step #1: INFO: Build completed successfully, 118 total actions
Step #1: INFO: 88 processes: 64 remote cache hit, 24 local.
Step #1: INFO: Elapsed time: 224.905s, Critical Path: 77.52s
Step #1: [91 / 101] GoCompile vendor/go.uber.org/zap/internal/color/linux_amd64_stripped/go_default_library~/bitbucket.org/seibert-media/events/vendor/go.uber.org/zap/internal/color.a; 0s remote
Step #1: [83 / 101] GoStdlib external/io_bazel_rules_go/linux_amd64_stripped/stdlib~/pkg; 34s remote
Step #1: [83 / 101] GoStdlib external/io_bazel_rules_go/linux_amd64_stripped/stdlib~/pkg; 10s remote
Step #1: [48 / 50] [Sched] SHA256 cmd/listeners/reseller/image-layer.tar.sha256
Step #1: [18 / 44] [-----] GoCompile external/io_bazel_rules_go/go/tools/builders/linux_amd64_stripped/compile~/go/tools/builders/compile.a [for host]
Step #1: [17 / 44] GoStdlib external/io_bazel_rules_go/linux_amd64_pure_stripped/stdlib~/pkg; 40s remote
Step #1: [17 / 44] GoStdlib external/io_bazel_rules_go/linux_amd64_pure_stripped/stdlib~/pkg; 16s remote
Step #1: [17 / 44] GoStdlib external/io_bazel_rules_go/linux_amd64_pure_stripped/stdlib~/pkg; 8s remote
Step #1: [16 / 44] [-----] GoLink external/io_bazel_rules_go/go/tools/builders/linux_amd64_stripped/stdlib [for host]
[...]

PR #2723, opened 2020-11-18
Use use_default_shell_env whenever possible

mrmeku:
**What type of PR is this?**

>
Bug fix

This PR makes all actions use use_default_shell_env when possible.
This change is needed to increase the likelyhood of cache hits from rule_go actions

Fixes https://github.com/bazelbuild/rules_go/issues/1531


jayconrod:
Sorry, I can't accept this. `use_default_shell_env` makes actions *less* hermetic and *less* likely to be cached. It should only be used when an external tool needs to be invoked with custom flags, or when an internal toolchain can't be configured correctly. #1531 definitely doesn't justify this.

PR #2734, opened 2020-11-19
Remove nacl and 32-bit darwin platforms

### LINKED ISSUE #2727: Confirm toolchains declared for all platforms in Go 1.15

jayconrod:
They are no longer supported by Go.

Fixes #2727


jayconrod:
Abandoning this change for now.

32-bit darwin platforms are still supported by Go 1.14, which will be supported by rules_go for a few more months.

Nacl supposed was removed in Go 1.14, but removing them from the platform list breaks build files generated by Gazelle that mention nacl. Let's wait a while before removing it.

PR #2736, opened 2020-11-20
Relative paths in error messages

### LINKED ISSUE #2061: Feature request: Workspace-relative file paths in compilation errors
I wanted to propose a quality-of-life improvement. Currently, compile errors are passed back to the user verbatim, which on MacOS results in very wide messages due to the sandbox directory.

On the build I just ran, the workspace directory in the sandbox is 104 characters:
```
/private/var/tmp/_bazel_robfig/008081e55e53b945b85110f778538398/sandbox/darwin-sandbox/6/execroot/brain/
```

This adds friction to development:
- Messages often wrap to the next line of my terminal, making them harder to scan / read
- I can not jump directly to the file by copying the path. (Normally I precisely copy the workspace-relative portion of the path). 
- Relatedly, unless you pass --sandbox_debug, those files are already deleted, so you can't even review them (e.g. if they are a third party dependency or generated code).

I propose trimming the paths reported to be workspace relative. That would address the main use case, although figuring out something better for 3rd party deps and generated code would be good too. Maybe a gold plated solution would recognize that case and provide paths in bazel-out (generated) or bazel-WORKSPACE/external (3rd party)

I'm not sure how hard this would be or how to accomplish it. Perhaps the rules could pipe the output through a new Go program that does the filtering?

Any thoughts?

robfig:
**What type of PR is this?**
Feature

Improve the error messages seen by everyday users of rules_go

Fixes #2061

### DIFF
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -133,14 +133,21 @@ func (e *env) runCommand(args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
 	// Redirecting stdout to stderr. This mirrors behavior in the go command:
 	// https://go.googlesource.com/go/+/refs/tags/go1.15.2/src/cmd/go/internal/work/exec.go#1958
-	return runAndLogCommandOutput(cmd, os.Stderr, os.Stderr, e.verbose)
+	buf := &bytes.Buffer{}
+	cmd.Stdout = buf
+	cmd.Stderr = buf
+	err := runAndLogCommand(cmd, e.verbose)
+	fmt.Fprint(os.Stderr, relativizePaths(buf.Bytes()))
+	return err
 }
 
 // runCommandToFile executes a subprocess and writes the output to the given
 // writer.
 func (e *env) runCommandToFile(w io.Writer, args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
-	return runAndLogCommandOutput(cmd, w, os.Stderr, e.verbose)
+	cmd.Stdout = w
+	cmd.Stderr = os.Stderr
+	return runAndLogCommand(cmd, e.verbose)
 }
 
 func absEnv(envNameList []string, argList []string) error {
@@ -152,16 +159,6 @@ func absEnv(envNameList []string, argList []string) error {
 		}
 	}
 	return nil
-}
-
-func runAndLogCommandOutput(cmd *exec.Cmd, stdout, stderr io.Writer, verbose bool) error {
-	var bufout, buferr bytes.Buffer
-	cmd.Stdout = &bufout
-	cmd.Stderr = &buferr
-	err := runAndLogCommand(cmd, verbose)
-	io.Copy(stdout, bytes.NewReader(relativizePaths(bufout.Bytes())))
-	io.Copy(stderr, bytes.NewReader(relativizePaths(buferr.Bytes())))
-	return err
 }
 
 func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
@@ -347,7 +344,7 @@ func relativizePaths(output []byte) []byte {
 	}
 	dirBytes := make([]byte, len(dir), len(dir)+1)
 	copy(dirBytes, dir)
-	if bytes.HasSuffix(dirBytes, []byte{filepath.Separator})  {
+	if bytes.HasSuffix(dirBytes, []byte{filepath.Separator}) {
 		return bytes.ReplaceAll(output, dirBytes, nil)
 	}
 
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -60,6 +60,8 @@ type env struct {
 	workDirPath string
 
 	shouldPreserveWorkDir bool
+
+	execDir string
 }
 
 // envFlags registers flags common to multiple builders and returns an env
@@ -71,6 +73,7 @@ func envFlags(flags *flag.FlagSet) *env {
 	flags.StringVar(&env.installSuffix, "installsuffix", "", "Standard library under GOROOT/pkg")
 	flags.BoolVar(&env.verbose, "v", false, "Whether subprocess command lines should be printed")
 	flags.BoolVar(&env.shouldPreserveWorkDir, "work", false, "if true, the temporary work directory will be preserved")
+	env.execDir, _ = os.Getwd()
 	return env
 }
 
@@ -133,18 +136,20 @@ func (e *env) runCommand(args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
 	// Redirecting stdout to stderr. This mirrors behavior in the go command:
 	// https://go.googlesource.com/go/+/refs/tags/go1.15.2/src/cmd/go/internal/work/exec.go#1958
-	cmd.Stdout = os.Stderr
-	cmd.Stderr = os.Stderr
-	return runAndLogCommand(cmd, e.verbose)
+	var stderr bytes.Buffer
+	cmd.Stdout = &stderr
+	cmd.Stderr = &stderr
+	return runAndLogCommand(cmd, &stderr, e.verbose)
 }
 
 // runCommandToFile executes a subprocess and writes the output to the given
 // writer.
 func (e *env) runCommandToFile(w io.Writer, args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
+	var stderr bytes.Buffer
 	cmd.Stdout = w
-	cmd.Stderr = os.Stderr
-	return runAndLogCommand(cmd, e.verbose)
+	cmd.Stderr = &stderr
+	return runAndLogCommand(cmd, &stderr, e.verbose)
 }
 
 func absEnv(envNameList []string, argList []string) error {
@@ -158,13 +163,15 @@ func absEnv(envNameList []string, argList []string) error {
 	return nil
 }
 
-func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
+func runAndLogCommand(cmd *exec.Cmd, stderr *bytes.Buffer, verbose bool) error {
 	if verbose {
 		fmt.Fprintln(os.Stderr, formatCommand(cmd))
 	}
 	cleanup := passLongArgsInResponseFiles(cmd)
 	defer cleanup()
-	if err := cmd.Run(); err != nil {
+	err := cmd.Run()
+	os.Stderr.Write(relativizePaths(stderr.Bytes()))
+	if err != nil {
 		return fmt.Errorf("error running subcommand %s: %v", cmd.Path, err)
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


go/tools/builders/env.go
```diff
@@ -60,6 +60,8 @@ type env struct {
 	workDirPath string
 
 	shouldPreserveWorkDir bool
+
+	pwd string
```
jayconrod, line 64: Rename to `execDir`.

go/tools/builders/env.go
```diff
@@ -332,6 +336,40 @@ func absArgs(args []string, flags []string) {
 	}
 }
 
+
+// absToRelativePathWriter filters the output stream from a compile command,
+// replacing any absolute paths with relative equivalents.
+//
+// This is done on a best-effort basis, and not much effort is undertaken. If a
+// path is split across multiple Write calls, it will not be relativized.
```
jayconrod, line 344: I don't think this is good behavior. It would be better to buffer all the output, then replace paths all at once.

go/tools/builders/env.go
```diff
@@ -332,6 +336,40 @@ func absArgs(args []string, flags []string) {
...
+type absToRelativePathWriter struct {
+	pwd      string
+	delegate io.Writer
```
jayconrod, line 347: Rename:

`pwd` to `dir`
`delegate` to `w`.

go/tools/builders/env.go
```diff
@@ -332,6 +336,40 @@ func absArgs(args []string, flags []string) {
...
+}
+
+func (wr absToRelativePathWriter) Write(p []byte) (nn int, err error) {
+	if wr.pwd == "" {
+		return wr.delegate.Write(p)
+	}
+	if !strings.HasSuffix(wr.pwd, "/") {
+		wr.pwd += "/"
```
jayconrod, line 355: This won't work on Windows: use `filepath.Separator` instead.

Also, this is going to allocate new memory on every `Write` call since `wr` is passed by value.

go/tools/builders/env.go
```diff
@@ -332,6 +336,40 @@ func absArgs(args []string, flags []string) {
...
+	}
+	return wr.delegate.Write(bytes.ReplaceAll(p, []byte(wr.pwd), nil))
+}
+
+// rel converts absolute paths found in the given output string to relative, if
+// they are within the PWD.
+func rel(output string) string {
```
jayconrod, line 362: Need a more descriptive name, maybe `relativizePaths`. I'd expect `rel` to work like `filepath.Rel` on a single path.

go/tools/builders/env.go
```diff
@@ -332,6 +336,40 @@ func absArgs(args []string, flags []string) {
...
+	dir, err := os.Getwd()
+	if dir == "" || err != nil {
+		return output
+	}
+	if !strings.HasSuffix(dir, "/")  {
+		dir += "/"
```
jayconrod, line 368: This won't work on Windows.

Also, I don't think this will replace the path itself with `.`.

jayconrod (commented):
Thanks for working on this.

Several comments below though.

go/tools/builders/env.go
robfig: Updated to use filepath.PathSeparator.

You're correct - if `Getwd` appeared in an error message with no following separator, it would not be converted to `.`. Do you think that's worth adding a case for? 

It seems unlikely to come up, and I have a personal preference for e.g. `path/to.go` instead of `./path/to.go` which would allow an implementation that "just works" for this case, so I'd be inclined to leave as-is.

go/tools/builders/env.go
robfig: Agreed, updated

go/tools/builders/env.go
robfig: Updated, and avoided extra allocation

go/tools/builders/env.go
robfig: Updated

go/tools/builders/env.go
robfig: That works. In that case, I can consolidate the implementation into just the relativizePaths func.

go/tools/builders/env.go
robfig: Updated

robfig:
Thanks for the review. I made all the updates you requested, with one caveat about the `.` case, and re-tested.

go/tools/builders/env.go
```diff
@@ -133,18 +136,20 @@ func (e *env) runCommand(args []string) error {
...
+	cmd.Stdout = &stderr
+	cmd.Stderr = &stderr
+	return runAndLogCommand(cmd, &stderr, e.verbose)
 }
 
 // runCommandToFile executes a subprocess and writes the output to the given
 // writer.
 func (e *env) runCommandToFile(w io.Writer, args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
+	var stderr bytes.Buffer
```
jayconrod, line 149: Let's make sure this function doesn't change. It needs to produce a file, and we shouldn't alter the content of that file.

Maybe some other function can call `runAndLogCommand` with a `*bytes.Buffer`, then `relativizePaths` on that and copy to `os.Stderr`? Then neither `runAndLogCommand` nor this function need to change.

go/tools/builders/env.go
```diff
@@ -158,13 +163,15 @@ func absEnv(envNameList []string, argList []string) error {
...
-func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
+func runAndLogCommand(cmd *exec.Cmd, stderr *bytes.Buffer, verbose bool) error {
 	if verbose {
 		fmt.Fprintln(os.Stderr, formatCommand(cmd))
 	}
 	cleanup := passLongArgsInResponseFiles(cmd)
 	defer cleanup()
-	if err := cmd.Run(); err != nil {
+	err := cmd.Run()
+	os.Stderr.Write(relativizePaths(stderr.Bytes()))
```
jayconrod, line 173: In general, use `fmt.Fprint` or `io.Copy` when writing to an `*os.File` or any unknown kind of `io.Writer`. The `Write` may write fewer bytes than its given, and it's important to keep calling it until there's nothing left to write.

go/tools/builders/env.go
```diff
@@ -60,6 +60,8 @@ type env struct {
 	workDirPath string
 
 	shouldPreserveWorkDir bool
+
+	execDir string
```
jayconrod, line 64: It doesn't look like this is actually used anymore. Can it be removed?

go/tools/builders/env.go
```diff
@@ -332,6 +339,21 @@ func absArgs(args []string, flags []string) {
 	}
 }
 
+// relativizePaths converts absolute paths found in the given output string to
+// relative, if they are within the PWD.
```
jayconrod, line 343: Replace `PWD` here with `working directory`. `PWD` means "print working directory", and since we're not printing it, I don't think we should use that acronym here.

go/tools/builders/env.go
```diff
@@ -332,6 +339,21 @@ func absArgs(args []string, flags []string) {
...
+	dir, err := os.Getwd()
+	if dir == "" || err != nil {
+		return output
+	}
+	dirBytes := make([]byte, len(dir), len(dir)+1)
+	copy(dirBytes, dir)
+	if !bytes.HasSuffix(dirBytes, []byte{filepath.Separator})  {
+		dirBytes = append(dirBytes, filepath.Separator)
+	}
+	return bytes.ReplaceAll(output, dirBytes, nil)
```
jayconrod, line 354: It looks like this still doesn't convert the working directory to `.`. For completeness, I think it ought to.

(I couldn't respond to the other comment chain; maybe there was a force push?)

go/tools/builders/env.go
robfig: Good point, done.

go/tools/builders/env.go
robfig: This is a lot cleaner, very good idea!

go/tools/builders/env.go
robfig: Good catch, removed

go/tools/builders/env.go
robfig: TIL! I always thought it was "present working directory". Now the distinction between pwd and cwd makes more sense to me. Thanks

go/tools/builders/env.go
robfig: OK, I didn't add that because there didn't seem to be a use case and it complicates the code further, but it was easy enough to add.

Ah, I didn't know that a force push prevents response. In the future I'll avoid that

robfig (commented):
Thanks for the review! Updated.

robfig:
Bump

go/tools/builders/env.go
```diff
@@ -133,18 +133,14 @@ func (e *env) runCommand(args []string) error {
...
 }
 
 // runCommandToFile executes a subprocess and writes the output to the given
 // writer.
 func (e *env) runCommandToFile(w io.Writer, args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
-	cmd.Stdout = w
-	cmd.Stderr = os.Stderr
-	return runAndLogCommand(cmd, e.verbose)
+	return runAndLogCommandOutput(cmd, w, os.Stderr, e.verbose)
```
jayconrod, line 143: Let's be careful that `runCommandToFile` does not relativize the output. It should just call `runAndLogCommand`.

go/tools/builders/env.go
```diff
@@ -133,18 +133,14 @@ func (e *env) runCommand(args []string) error {
 	cmd := exec.Command(args[0], args[1:]...)
 	// Redirecting stdout to stderr. This mirrors behavior in the go command:
 	// https://go.googlesource.com/go/+/refs/tags/go1.15.2/src/cmd/go/internal/work/exec.go#1958
-	cmd.Stdout = os.Stderr
-	cmd.Stderr = os.Stderr
-	return runAndLogCommand(cmd, e.verbose)
+	return runAndLogCommandOutput(cmd, os.Stderr, os.Stderr, e.verbose)
```
jayconrod, line 136: Instead of introducing `runAndLogCommandOutput`, I think it would be simpler to add the relativization here, especially since `runCommandToFile` should not use use it. Something like:

```
buf := &bytes.Buffer{}
cmd := exec.Command(args[0], args[1:]...)
// Redirecting stdout to stderr. This mirrors behavior in the go command:
// https://go.googlesource.com/go/+/refs/tags/go1.15.2/src/cmd/go/internal/work/exec.go#1958			cmd.Stdout = os.Stderr	
cmd.Stdout = buf
cmd.Stderr = buf
err := runAndLogCommand(cmd, e.verbose)
fmt.Fprint(os.Stderr, relativizePaths(buf.Bytes())
return err
```

jayconrod (commented):
Sorry for the delay. I took last week off, and I'm still catching up on everything.

Two more small comments. Looking good though.

go/tools/builders/env.go
robfig: Apologies, missed that requirement. Updated

go/tools/builders/env.go
robfig: LGTM! Updated

robfig:
Updated, thanks for the review!

go/tools/builders/env.go
jayconrod: I was wrong about this: `io.Copy` returns early if a `Write` call doesn't write the entire slice. I was thinking of `Read`, which may return fewer bytes without error. I get them confused, since they're different than the `read` and `write` system calls.

No need to change anything, just wanted to point that out.

jayconrod (approved):
Looks good, thanks for fixing this!

PR #2755, opened 2020-12-04
go_test: combine testinit and wrapper packages into bzltestutil

### LINKED ISSUE #2749: go_test: can't run a test binary wrapped by another test
0.25.0

Commit 1372b0c45c8e83da41c37f28294ec1bc929520ff

3.4.1

Yes

Linux - Fedora 32

### What did you do?
I've got an custom rule that looks like this:

```
go_test(
    name = "test_test",
    srcs = ["integration_test.go"],
    # keep
    tags = ["manual"],
    deps = ["@com_github_lib_pq//:pq"],
)

integration_test(
    name = "integration_test",
    pre_test = [
        "//src/bazel/flyway_migration/test/migration",
    ],
    test = ":test_test",
)
```

It runs the pre_test binaries before executing the test binary in the `test` attribute.
The rule implementation is pretty much an bash script that runs the executables put in the rule attributes.

Everything worked fine on version 0.24.0 of rules_go and but when trying to upgrade to 0.25.0 the rule stops working
and throws the following error when trying to run the go binary put into the `test` attribute.

```
2020/12/03 10:32:15 fork/exec ./src/bazel/flyway_migration/test/test_test_/test_test: no such file or directory
```
A likely candidate is the this commit: https://github.com/bazelbuild/rules_go/commit/9c1568a7f510fe306a59f8d09e91579a71f5e08c#diff-2f4977ebad0d8fd012c0d532023ebe9202e795bf502029f7df454d5fc4f96656

jayconrod:
testinit was a package used to change directories to the package
directory where tests normally run.

The wrapper sources were compiled into the test main package and were
used to re-execute the test binary in order to capture stdout and
stderr and translate verbose test output to XML.

The change combines both into the bzltestutil package. Like testinit,
it's imported by all generated test mains and should be initialized
before any other package outside std.

This fixes an issue where we changed directories before re-executing
the test binary to capture XML output. If os.Args[0] was a relative
path, it would no longer be meaningful.

Fixes #2749


dragonsinth (commented):
👏

PR #2781, opened 2021-01-11
Handle long filenames on Windows in a few more places.

pmuetschard:
C.f. #1382

**What type of PR is this?**
Bug fix

Makes rules_go work on Windows with long filenames.


jayconrod (approved):
Looks good, thanks.

PR #2790, opened 2021-01-19
Adds a --cgo_compiler flag to control the C/C++ compiler used.

pmuetschard:
Adds a config transition that will modify the `--compiler` bazel flag for go rules based upon the `--cgo_compiler` flag. This allows users of rules_go to build non-cgo C/C++ code with a different compiler than cgo code. For example, on Windows users could use MSVC for their non-cgo code and MinGW gcc for the cgo code.

### DIFF
--- a/go/config/BUILD.bazel
+++ b/go/config/BUILD.bazel
@@ -58,6 +58,12 @@ string_list_flag(
     visibility = ["//visibility:public"],
 )
 
+string_flag(
+    name = "cgo_compiler",
+    build_setting_default = "",
+    visibility = ["//visibility:public"],
+)
+
 filegroup(
     name = "all_files",
     testonly = True,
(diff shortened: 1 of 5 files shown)


jayconrod (commented):
Sorry, I'm not comfortable with this setting being part of rules_go. I can't provide support for MSVC toolchains.

There may be another way to solve the problem using platforms and toolchains. If you depend on `cdeps` through a transition on the `--platforms` flag, you could build C/C++ dependencies on a `platform` you define with the `@bazel_tools//tools/cpp:msvc` constraint. That would force a compatible (MSVC) toolchain to be selected.

I'm not sure at all whether the MinGW or MSVC linkers will be able to handle a mixed binary like that though. I'd be surprised if it worked, especially for all link modes.

Issue #2795, opened 2021-01-23
Support Apple Silicon


phst:
### What version of rules_go are you using?

Current head, commit hash da0f36c7a249be313dbb2e000f529f11921f4846

### What version of gazelle are you using?

None

### What version of Bazel are you using?

4.0.0

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

macOS, ARM64

### Any other potentially useful information about your toolchain?

https://github.com/bazelbuild/rules_go/blob/a322b06ae7e1e3d5c2e06c93023b196520b56495/go/private/sdk.bzl#L232 seems to assume that all macOS machines are x86-64, which is no longer the case. This branch likely should check the output of `uname -m` or `arch`, like the Linux branch.

### What did you do?

Download the current master of `rules_go` and run `bazel test //...` on an Apple Silicon machine.

### What did you expect to see?

Tests should succeed.

### What did you see instead?

Toolchain resolution fails:

```
$ bazel test //...
DEBUG: /private/var/tmp/_bazel_p/3d10a82390614c10e7f6b64e3ae2263e/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14: 
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
[...]

PR #2796, opened 2021-01-23
Add minimal support for ARM64 on macOS.

phst:
This still fails several tests that assume x86-64, but it’s at least a starting
point.

**What type of PR is this?**

Feature

This implements basic support for ARM64 macOS systems.

#2795

(This doesn't entirely fix the issue since some of the tests still fail)

**Other notes for review**

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -230,6 +230,15 @@ def _detect_host_platform(ctx):
 
     elif ctx.os.name == "mac os x":
         goos, goarch = "darwin", "amd64"
+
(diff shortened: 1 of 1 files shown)


jayconrod (approved):
Looks good. Thanks for sending this.

I have very limited access to arm64 hardware, and we don't have any systems in CI yet, so I can't verify this. Also, my understanding is that `uname -m` may return `x86_64` when invoked from an `x86_64` process. Not sure whether Bazel runs as x86_64, but something to be aware of.

albertocavalcante:
@phst coming from the future to say thank you for this!

PR #2797, opened 2021-01-23
fix incorrect release date

nicolai86:
**What type of PR is this?**

> Documentation

The release notes incorrectly labeled the latest release as 2020, where it should be 2021.


jayconrod (approved):
Oops, thanks. Still feels like 2020 to me I guess :(

PR #2798, opened 2021-01-25
Fix cc target on macOS with Bazel@HEAD

meteorcloudy:
Context: https://github.com/bazelbuild/bazel/issues/4341#issuecomment-762576370

**What type of PR is this?**

Fix build target with future Bazel version

`supports_dynamic_linker` feature was removed from cc toolchain for macOS
https://github.com/bazelbuild/bazel/issues/4341#issuecomment-758028203


jayconrod (approved):
Looks good, thanks!

PR #2816, opened 2021-02-12
Absolute path for LDFLAGS when building stdlib

### LINKED ISSUE #1357: Cgo and --compiler=gcc
Copied from [bazel-go-discuss](https://groups.google.com/forum/#!topic/bazel-go-discuss/HogSA9AtV-M). cc @nathantchan 

------

One of our verification builds uses gcc, and I currently cannot get past the installation portion of the stdlib rule. It's throwing errors like the following:

```
not-int-const: In function '__cgo_f_1_3':
not-int-const:1:60: error: invalid type argument of unary '*' (have 'int')
not-num-const: In function '__cgo_f_1_4':
not-num-const:1:73: error: expected expression before ';' token
not-str-lit: In function '__cgo_f_1_5':
not-str-lit:1:73: error: expected expression before ';' token
not-type: In function '__cgo_f_2_2':
not-type:2:39: error: '__cgo_undefined__2' undeclared (first use in this function)
not-type:2:39: note: each undeclared identifier is reported only once for each function it appears in
not-type:2:38: error: invalid operands to binary * (have 'int (*)(const struct sockaddr * __restrict__,  socklen_t,  char * __restrict__,  socklen_t,  char * __restrict__,  socklen_t,  int)' and 'const char *')
not-int-const: In function '__cgo_f_2_3':
not-int-const:2:67: error: invalid operands to binary * (have 'int (*)(const struct sockaddr * __restrict__,  socklen_t,  char * __restrict__,  socklen_t,  char * __restrict__,  socklen_t,  int)' and 'int')
not-num-const: In function '__cgo_f_2_4':
not-num-const:2:67: error: incompatible types when initializing type 'double' using type 'int (*)(const struct sockaddr * __restrict__,  socklen_t,  char * __restrict__,  socklen_t,  char * __restrict__,  socklen_t,  int)'
not-str-lit: In function '__cgo_f_2_5':
not-str-lit:2:1: error: invalid initializer
not-int-const: In function '__cgo_f_3_3':
not-int-const:3:65: error: invalid type argument of unary '*' (have 'int')
not-int-const:3:33: error: enumerator value for '__cgo_undefined__3' is not an integer constant
not-num-const: In function '__cgo_f_3_4':
not-num-const:3:78: error: expected expression before ';' token
not-num-const:3:78: error: initializer element is not constant
not-str-lit: In function '__cgo_f_3_5':
not-str-lit:3:78: error: expected expression before ';' token
not-str-lit:3:78: error: invalid initializer
completed: At top level:
completed:1:16: error: '__cgo__2' undeclared here (not in a function)
completed:1:1: error: initializer element is not constant
```

It doesn't seem like setting --compiler=gcc works right now in 0.9.0 of the go rules. Any ideas?

```
$ bazel test --compiler=gcc tests/...
...................
ERROR: No toolchain found for --cpu='k8' --compiler='gcc'. Valid toolchains are: [
  --cpu='armeabi-v7a' --compiler='compiler' --glibc='armeabi-v7a',
  --cpu='ios_x86_64' --compiler='compiler' --glibc='ios',
  --cpu='k8' --compiler='compiler' --glibc='local',
  --cpu='x64_windows' --compiler='cl' --glibc='msvcrt140',
]
```

anrodrig:
When using a custom cpp toolchain that adds libc++.a and libc++abi.a
in link_libs, linking the stdlib fails because it can't find the archives
by their relative paths. Same as issue #1357

**What type of PR is this?**

Bug fix

Takes relative paths in LDFLAGS when building stdlib and makes them absolute. I ran into this because I have relative paths in link_libs in a custom cpp toolchain 

Partially fixes #1357

**Other notes for review**

Sme ldflags are passed like "-flag value". While it's easy to ignore "-flag" we check if "value" is a file that exists when deciding whether or not to make it absolute. This check can be wrong

### DIFF
--- a/go/tools/builders/stdlib.go
+++ b/go/tools/builders/stdlib.go
@@ -90,7 +90,19 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
 
 	// Strip path prefix from source files in debug information.
 	os.Setenv("CGO_CFLAGS", os.Getenv("CGO_CFLAGS")+" "+strings.Join(defaultCFlags(output), " "))
-	os.Setenv("CGO_LDFLAGS", os.Getenv("CGO_LDFLAGS")+" "+strings.Join(defaultLdFlags(), " "))
+	relativeLdflags, _ := splitQuoted(os.Getenv("CGO_LDFLAGS"))
+	var absLdflags strings.Builder
+	for _, f := range relativeLdflags {
+		if strings.HasPrefix(f, "-") {
+			absLdflags.WriteString(f)
+		} else if _, err := os.Stat(f); os.IsNotExist(err) {
+			absLdflags.WriteString(f)
+		} else {
+			absLdflags.WriteString(abs(f))
+		}
+		absLdflags.WriteString(" ")
+	}
+	os.Setenv("CGO_LDFLAGS", absLdflags.String()+" "+strings.Join(defaultLdFlags(), " "))
 
 	// Allow flags in CGO_LDFLAGS that wouldn't pass the security check.
 	// Workaround for golang.org/issue/42565.


achew22:
Seems like a great addition to the codebase. How can we test this to prevent regressions in the future?

anrodrig:
What do you recommend? I was thinking about setting up a integration-like test with a custom cpp toolchain with a dummy link_libs but I'm not sure if thats feasible: I don't know how set crosstool_top within a project (as opposed to using .bazelrc or the command line)

achew22:
Is the toolchain something that has to be in the repo or can we add it as a dev dependency in the `WORKSPACE`?  If the later, we could add an integration test in [`tests/integration`](https://github.com/bazelbuild/rules_go/tree/master/tests/integration), like [this](https://github.com/bazelbuild/rules_go/blob/master/tests/integration/gazelle/gazelle_test.go) where you can invoke a bazel command on a temporary fs.

anrodrig:
I tried two approaches and didn't get anywhere: I started making a real cpp toolchain but realized that making a real cpp toolchain that works on all supported hosts is a lot. Then I considered using BAZEL_LINKLIBS with the autodetected cpp toolchain but if you provide a relative path it's relative to the execroot of the GoStdlib task. This task only depends on the real cpp toolchain, I couldn't figure out a way to smuggle in bogus .a that would could be accessed with a relative path.

I understand not wanting to accept this change without a test but I'm going to throw in the towel: it's easier to change my custom toolchain to use absolute paths itself :) Thanks for your quick help.

achew22:
@anrodrig, thanks so much for taking a stab at it. @jayconrod, I'm comfortable with this if you'd like to merge it but would like to defer to a higher authority.

jayconrod:
Sorry for the latency... I have an embarrassingly long review backlog right now.

I'm a little nervous about absolutizing all paths. There's a good chance it could make the `GoStdLib` action non-reproducible if it ends up embedding a path in the sandbox or on the build machine into an output file.

Could this be narrowed to apply to a specific set of flags that we know are safe?

anrodrig:
For my usecase the parameters I'm passing are relative paths to libc++ and libc++abi in order to statically link against them, so there aren't specific flags that we could allow. The non-reproducability of GoStdLib is a good point; in this case I think it shouldn't be an issue but that does make me more nervous about this patch.

jayconrod:
So could this be changed to only absolutize `-L` flags for example?

anrodrig:
Actually yeah, that works (not sure why we didn't do that before). Something like
`link_libs = ["-Wl,-Bstatic", "-lstdc++", "-Wl,-Bdynamic"],` in the c++ toolchain
does the trick and works with rules_go just fine. I'll close this

PR #2819, opened 2021-02-14
Make go_reset_target forward DefaultInfo.executable and pass go_path into it

### LINKED ISSUE #2689: go_path needs to go through go_reset_transition

```
3f47128dd00917a5a139868766f9ece59f9f4103
```

```
3.7.0
```

Yes

```
macOS / darwin_amd64
```

### What did you do?

The `go_path` binary, when consumed though a platform transition, will match the go platform options (linkmode etc..).

steeve:
**What type of PR is this?**

Bug fix

- Patch `go_reset_target` to properly forward `DefaultInfo.executable` via a symbolic link.
- Patch `go_proto*` rules to now use `ctx.executable`
- Pass `go_path` builder binary through `go_reset_target`

- Fix #2689

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 load(
+    "@bazel_skylib//lib:paths.bzl",
+    "paths",
+)
+load(
     "//go/private:mode.bzl",
     "LINKMODES",
     "LINKMODE_NORMAL",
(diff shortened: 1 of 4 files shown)


jayconrod (approved):
Looks good.

Do you know if `ctx.actions.symlink` works on Windows? In the past, you needed administrative privilege or to have enabled developer mode to create symbolic links. I don't know if that's still true or if `ctx.actions.symlink` creates a different kind of link (junction?).

steeve:
I have no idea, but I saw that: https://www.wintellect.com/non-admin-users-can-now-create-symlinks-windows-10/

steeve:
So from what I can tell, users need to enable Developer mode. I am not sure if it's common practice though. I'd wager it is, given how useful things like symlinks are?

PR #2825, opened 2021-02-20
Include ARM64 in “select” calls.

phst:
These files are architecture-independent, so this works on ARM-64 without
further changes.

**What type of PR is this?**

Bug fix


jayconrod (approved):
Looks good, thanks!

PR #2826, opened 2021-02-20
Fix CPU selector.

phst:
Using the CPU “darwin” doesn’t work correctly on ARM-64.  Use the normal CPU
and operating system constraints from the \@platforms repository instead.

**What type of PR is this?**

Bug fix


jayconrod (approved):
Looks good, thanks!

PR #2827, opened 2021-02-20
Import latest versions of golang.org/x/net.

phst:
This is necessary for ARM-64 support on macOS.

**What type of PR is this?**

Bug fix

**Other notes for review**


jayconrod:
Sorry, PRs that update dependencies are not accepted.

All dependencies are updated regularly before each minor release. I've opened #2829 to track that for the next one, which should be out in a couple weeks.

PR #2828, opened 2021-02-21
Update golang.org/x/sys repository to support ARM-64.

phst:
**What type of PR is this?**

Bug fix

Prepare for ARM64 support on macOS.


jayconrod:
Sorry, PRs that update dependencies are not accepted.

All dependencies are updated regularly before each minor release. I've opened #2829 to track that for the next one, which should be out in a couple weeks.

PR #2830, opened 2021-02-22
Implement “add” and “sub” functions for ARM-64 as well.

phst:
**What type of PR is this?**

Bug fix

A bit of support for ARM-64.


jayconrod (approved):
Looks good, thanks, and sorry for the delay.

PR #2835, opened 2021-03-03
Add prototype gopackagesdriver based on `bazel query`.

### LINKED ISSUE #512: Expose generated Go files to editors and IDEs

ribrdb:
**What type of PR is this?**

Feature

Add a prototype implementation of a gopackagesdriver.

Fixes #512

**Other notes for review**
gopls has recently stopped supporting gopackagesdriver. You need a version earlier than 092357f697aefc45480a378b64b698e063e63da9, or you need to revert that commit and rebuild gopls.

Also, you should make sure your editor excludes the following patterns, or gopls can go crazy scanning them:
```
*.runfiles/**
bazel-{workspace_name}/**
bazel-out/**
```
Don't exclude all of bazel-bin, you do want it to notice when generated files get re-built.


go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+		case patt == ".":
+			d.bazelQueries = append(d.bazelQueries, goFilter(":all"))
+			d.wildcardQuery = true
+		case patt == "./...":
+			d.bazelQueries = append(d.bazelQueries, goFilter("..."))
+			d.wildcardQuery = true
+		case d.sdk.packages[patt]:
+			d.stdlibImports[patt] = true
+		default:
+			// TODO: do we need to check for a package ID instead of import path?
```
jayconrod, line 112: I think we should assume arguments are Bazel labels, not import paths.

We'll need to handle import paths in the standard library, since there's no label for individual packages.

If we must process import paths without wildcards, we should make an attempt to find the directory without scanning the whole repo. `bazel query //...` can be very expensive for large repos.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+		pkgs, roots, err := d.packagesFromQueries(d.bazelQueries)
+		if err != nil {
+			return nil, err
+		}
+		resp.Packages = append(resp.Packages, pkgs...)
+		resp.Roots = append(resp.Roots, roots...)
+	}
+
+	if len(ignoredPkg.IgnoredFiles) > 0 {
+		log.Printf("Ignored %v files", len(ignoredPkg.IgnoredFiles))
```
jayconrod, line 134: Not sure I understand this. Add a comment?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+	if d.cfg.Mode&packages.NeedTypesSizes != 0 {
+		goarch := protocol.GetEnv(&d.cfg, "GOARCH", runtime.GOARCH)
+		// SizesFor always return StdSizesm,
+		resp.Sizes = types.SizesFor("gc", goarch).(*types.StdSizes)
+	}
+
+	return &resp, nil
+}
+
+func (d *bazelDriver) processFileQuery(fp string, ignoredPkg *packages.Package) error {
```
jayconrod, line 157: Add a doc comment? From the signature, it's not clear to me what it means to process a file query or what `ignoredPkg` is.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(fp) == 0 {
+		return fmt.Errorf("\"file=\" prefix given with no query after it")
+	}
+	stdlibRoot := filepath.Join(d.sdk.goroot, "src")
+
+	if filepath.IsAbs(fp) {
+		if strings.HasPrefix(fp, stdlibRoot) {
```
jayconrod, line 164: In general, `strings.HasPrefix` isn't safe to use on paths, since there might be other characters between the prefix and the separator (like `$GOROOT/srcccc`).

Maybe copy in a function like Gazelle's [`pathtools.HasPrefix`](https://github.com/bazelbuild/bazel-gazelle/blob/3fccaeca6a77cc41adcb90c4c8ce0af5c49d2c9d/pathtools/path.go#L32) instead?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+			if rel, err := filepath.Rel(stdlibRoot, fp); err == nil {
+				pkg := filepath.Dir(rel)
+				if d.sdk.packages[pkg] {
+					d.stdlibImports[pkg] = true
+				}
+			}
+			return nil
```
jayconrod, line 171: Should it be an error if the path is inside `$GOROOT` but doesn't match any known package?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+		if rel, err := filepath.Rel(d.workspaceRoot, fp); err == nil {
+			fp = rel
+		}
+	}
+	if prefix := gorootPattern.FindString(fp); prefix != "" {
```
jayconrod, line 177: Not sure I understand this. Please comment?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+	query, err := d.convertFileQuery(fp)
+	if err != nil {
+		return err
+	}
+	d.bazelQueries = append(d.bazelQueries, query)
+	d.fileQueries[fp] = true
+	return nil
+}
+
+func (d *bazelDriver) convertFileQuery(path string) (string, error) {
```
jayconrod, line 192: Please add a doc comment.

jayconrod (commented):
Added a few low level comments. I haven't read through `pkgconv` yet. I hope other maintainers will chime in as well.

My main concern is that the approach based on `bazel query` isn't going to work for more complicated requests, especially those involving cgo, generated code, and custom rules that are compatible with rules_go. This also won't work for a driver that runs inside an action, where we can't invoke Bazel.

The design doc in [Editor and tool integration](https://github.com/bazelbuild/rules_go/wiki/Editor-and-tool-integration) is more the direction I think this should go. The builder can generate package JSON files, either as part of `GoCompilePkg` or another action, then the driver would build the appropriate targets (either using an aspect or an output group), absolutize the paths, then hand those JSON objects back to the caller.

ribrdb:
> Added a few low level comments. I haven't read through `pkgconv` yet. I hope other maintainers will chime in as well.
> 
> My main concern is that the approach based on `bazel query` isn't going to work for more complicated requests, especially those involving cgo, generated code, and custom rules that are compatible with rules_go. This also won't work for a driver that runs inside an action, where we can't invoke Bazel.

Yep, 100% agree. I think bazel aquery might be able to work instead of query, but integrating with the builder seems much better. I'm not sure what you mean about running "inside an action" though.

> The design doc in [Editor and tool integration](https://github.com/bazelbuild/rules_go/wiki/Editor-and-tool-integration) is more the direction I think this should go. The builder can generate package JSON files, either as part of `GoCompilePkg` or another action, then the driver would build the appropriate targets (either using an aspect or an output group), absolutize the paths, then hand those JSON objects back to the caller.

I haven't seen this doc before, I'll try to digest it.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
ribrdb: ignoredPkg can probably be removed. I was trying to get gopls to stop querying about extra files inside the bazel symlinks, but I couldn't figure out anything I could return that would convince gopls not to ask about a file again. Instead I had to configure my editor to exclude things like bazel-bin/**/*.runfiles

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
+// Copyright 2021 The Bazel Go Rules Authors. All rights reserved.
```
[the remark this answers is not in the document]
ribrdb: Why? This is used in other files like go/private/tools/lines_sorted_test.bzl
Wouldn't "The Bazel Authors" refer to https://github.com/bazelbuild/bazel/blob/master/AUTHORS ?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/bazelbuild/bazel-watcher/bazel"
+	"github.com/bazelbuild/bazel-watcher/third_party/bazel/master/src/main/protobuf/blaze_query"
+	"github.com/bazelbuild/rules_go/go/tools/gopackagesdriver/bazelquerydriver/pkgconv"
+	"github.com/bazelbuild/rules_go/go/tools/gopackagesdriver/protocol"
+	"golang.org/x/tools/go/packages"
+)
+
+const fileQueryPrefix = "file="
+
+var gorootPattern = regexp.MustCompile(`^bazel-[^/]+/external/go_sdk\b`)
```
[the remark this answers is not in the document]
ribrdb: Maybe you could configure your editor preferences to set GOROOT for gopls, but that seems difficult.
I couldn't figure out any way to infer this. I only get the rule attribute values from `bazel query`, and I don't see anything useful in @io_bazel_rules_go//:stdlib
aquery also doesn't help, because @io_bazel_rules_go//:stdlib doesn't seem to have any actions

go/private/repositories.bzl
```diff
@@ -120,6 +120,18 @@ def go_rules_dependencies(is_rules_go = False):
         strip_prefix = "rules_cc-b1c40e1de81913a3c40e5948f78719c28152486d",
     )
 
+    # Needed for gopackagesdriver
+    _maybe(
+        http_archive,
+        name = "com_github_bazelbuild_bazel_watcher",
```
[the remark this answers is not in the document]
ribrdb: Would copying the protos and bazel.go here be better? 
If so, should I put them into //third_party or here?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
ribrdb: I'm definitely seeing some queries for import paths from gopls in my logfile.
How would we find the directory without scanning the whole repo?
Maybe we need to maintain a cache on disk of `import path -> package` and `generate filename -> package`?

Note that the first thing gopls does is query for "./..." which does take a long time. But then if you're lucky bazel keeps that loaded so later queries are fast.

go/tools/gopackagesdriver/protocol/protocol.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"golang.org/x/tools/go/packages"
+)
+
+// Request is a JSON object sent by golang.org/x/tools/go/packages
+// on stdin. Keep in sync.
+type Request struct {
+	Command    string            `json:"command"` // FIXME ???
+	Mode       packages.LoadMode `json:"mode"`
+	Env        []string          `json:"env"` // FIXME handle
```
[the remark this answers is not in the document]
ribrdb: Should I start filing bugs for todos in the gopackagesdriver?

go/tools/gopackagesdriver/bazelquerydriver/driver.go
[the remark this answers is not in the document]
jayconrod: It's reported through a provider, but I don't think there's any way to access that from the command line. This is part of why I think build actions (either the normal actions or actions created by an aspect) should generate most of this information: they have much more direct access to it.

It's fine to add a `// TODO(#512)` here for now (and for most other comments here that can't be quickly resolved).

go/private/repositories.bzl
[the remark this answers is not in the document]
jayconrod: Copying the pre-generated .pb.go files is good. They can go anywhere under `go/tools/gopackagesdriver`; nothing else should use them.

go/tools/gopackagesdriver/protocol/protocol.go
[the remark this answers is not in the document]
jayconrod: Sure, for bigger things that might need discussion. Anything small can just use #512 as the blanket issue though.

go/tools/gopackagesdriver/protocol/driver.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+)
+
+// Driver is the gopackagesdriver implementation
+type Driver func(cfg Request, patterns ...string) (*Response, error)
```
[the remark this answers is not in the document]
jayconrod: That's fair. And now that I think about it, we'll probably need different drivers for editors and for tools running within Bazel actions, too.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
jayconrod: Hmm, I imagine we'll need to work with the gopls folks later on to avoid import path queries.

For now, let's add TODOs.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
jayconrod: Sure, that's fine in this case, since we don't need to match the `src` directory itself.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
jayconrod: Makes sense. Wasn't sure.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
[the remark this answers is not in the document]
ribrdb: Can we add Derivita to the bazelbuild/bazel AUTHORS file as well? That would resolve my concerns.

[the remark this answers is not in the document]
jayconrod: Re-reading the internal open source guidelines, my understanding is that "The Bazel Authors" refers to everyone who holds copyright. The AUTHORS file is supposed to be a list of significant contributors and is not supposed to have any implication for copyright ownership.

That is not what rules_go's AUTHORS file says though, and it's certainly incomplete. It seems that the advice from the open source office has changed at some point.

I've opened #2843, and I've added this to my list of things to look at when I have bandwidth for maintenance. It will take time to resolve though.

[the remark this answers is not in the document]
ribrdb: Ok, I've also re-read the CLA. I think that makes sense.

go/tools/gopackagesdriver/bazelquerydriver/driver.go
ribrdb: I added a todo in sdk.go

sluongng:
Hey folks, how is this PR coming along?

@steeve I saw on Twitter you were also working in this area? Perhaps it could be duplicated effort?

steeve:
Ah yes probably! @ribrdb I'm working on a version that's leveraging aspects and output groups. It was a off work weekend hack, but I got it to the point of it working with x/tools/packages: see https://github.com/znly/rules_go/tree/steeve/packagedriver

jayconrod:
I'm going to step back from reviewing this. Unfortunately I have very little time this year available to work on rules_go and Gazelle, so I'm probably blocking progress at this point. I'm happy to answer high level questions or loop in the gopls folks, but the other folks in @bazelbuild/go-maintainers may have more insights, especially @steeve, who's been working on this lately.

One high level comment: I think aspects or output groups are the only way to do this. Each Go package needs a corresponding action that produces a file containing raw package data. That file should contain everything that could go into [`packages.Package`](https://pkg.go.dev/golang.org/x/tools/go/packages#Package), including the AST and type information. The gopackagesdriver should be able to build those, drop data that wasn't requested, absolutize paths, then hand those objects up to the client.

Since those objects will contain information that can only be obtained by reading source files (possibly including generated files), they can't be produced from `bazel query` alone.

It's possible these files could be produced as part of the `GoCompilePkg` action, since that reads those files anyway. If that's done, the files would need to be part of a separate output group so they aren't returned by default when you `bazel build` a library. Alternatively (perhaps more efficiently), they can be produced as part of a separate action, created by an aspect.

steeve:
For the record, I have some prototype of a packages driver that's using aspects and output_groups. It's very done in my spare time, and I'm live tweeting it at https://twitter.com/steeve/status/1373247624929759232?s=20

I'll open a PR, @ribrdb, would love your insight on this, too.

As of last night it's able to generate stdlib info, although I'm not sure why, it's not working in `gopls` yet (even though it reports talking to the driver and no error in the return values).

ribrdb:
@steeve It looks like your aspect is building all the JSON using starlark and then outputting it using `write` actions.
I tried a similar approach last year and it does not scale. Bazel ends up keeping all of the json files in memory, and on a project with lots of dependencies this easily adds up to bazel becoming unusably slow. 
If we use an aspect it needs to `ctx.actions.run` to generate the .json files. For the sdlib itself I expect we could just run `go list`

ribrdb:
Oh, reading through twitter it sounds like you've got a lot more that what's currently on your github. I'd be happy to take a look when it's up.

steeve:
@ribrdb it's up at https://github.com/bazelbuild/rules_go/pull/2858

PR #2854, opened 2021-03-19
Use absolute destination path so that the longpath windows code in st…

### LINKED ISSUE #2853: go_path rule can hit windows longpath issues in certain situations

0.27.0

0.23.0

4.0.0

Yes

Windows 10 enterprise, x86_64

Nothing worth mentioning here

### What did you do?

We have a fairly deeply nested BUILD.bazel file which has a go_path in it, it is as follows in our workspace: `services/hive-controller/backend/BUILD.bazel`. This go_path rule then depends on further fairly deeply nested files, the longest one is: `services/hive-controller/backend/apis/hive/v1/zz_generated.deepcopy.go`

When writing this go_path rule, bazel ends up trying to write to `C:\Users\luna.duclos\_bazel_luna.duclos\ib2x26vf\execroot\embark\bazel-out\x64_windows-fastbuild-ST-5fa8105c1248\bin\services\hive-controller\backend\generate_apis\src\github.com\EmbarkStudios\src\services\hive-controller\backend\apis\hive\v1\zz_generated.deepcopy.go`

This breaks unless the path is passed absolutely to os.Create in go_path.go

A successfully created go_path

### What did you see instead?

```
GoPath: open bazel-out\x64_windows-fastbuild-ST-5fa8105c1248\bin\services\hive-controller\backend\generate_apis\src\github.com\EmbarkStudios\src\services\hive-controller\backend\apis\hive\v1\zz_generated.deepcopy.go: The system cannot find the path specified.
```

luna-duclos:
**What type of PR is this?**

Bug fix

This PR fixes an issue with longpaths on windows when using go_path's create mode.

Fixes #2853 

**Other notes for review**

The standard library's GOPATH handling can only correctly fix absolute paths, as one can see here: https://github.com/golang/go/blob/master/src/os/path_windows.go#L172

Meaning, to take advantage of this, all we need to do is pass in an absolute path.
We do this by calling the already existing abs() helper.

### DIFF
--- a/go/tools/builders/go_path.go
+++ b/go/tools/builders/go_path.go
@@ -154,7 +154,7 @@ func copyPath(out string, manifest []manifestEntry) error {
 		return err
 	}
 	for _, entry := range manifest {
-		dst := filepath.Join(out, filepath.FromSlash(entry.Dst))
+		dst := abs(filepath.Join(out, filepath.FromSlash(entry.Dst)))
 		if err := os.MkdirAll(filepath.Dir(dst), 0777); err != nil {
 			return err
 		}


jayconrod (approved):
Looks good, thanks!

Issue #2443, opened 2020-04-18
Automatic transition from cpu and crosstool_top


steeve:
As of Bazel 3.0, Android and iOS rules do not set constraints (android toolchain can be selected via a constraint, however).

I propose to add an automatic transition step for those, so that the use is seamless. The code is pretty simple and straightforward.

Perhaps could even be added to #2433.

I know ideally those rules shoud set constraints and `rules_go` would follow along, but from what I'd wager, it's really not there yet, so I think it could be a cool stopgap solution.

steeve:
On the other hand, one could get away with `select` + `goos/goarch` now that #2414 is merged. But it wouldn't "just work".

steeve:
Here is a quick draft:
```python
def android_toolchain(crosstool_top, cpu):
    if str(crosstool_top) == "//external:android/crosstool" or crosstool_top.workspace_name == "androidndk":
        _ANDROID_CPUS_TO_TOOLCHAIN.get(cpu)
    return None

def ios_toolchain(crosstool_top, cpu):
    return _IOS_CPUS_TO_TOOLCHAIN.get(cpu)

def detect_mobile_platform(settings, platform, goos, goarch):
    if goos == "auto" and goarch == "auto":
        crosstool_top = settings["//command_line_option:crosstool_top"]
        cpu = settings["//command_line_option:cpu"]
        toolchain = android_toolchain(crosstool_top, cpu) or ios_toolchain(crosstool_top, cpu)
        if toolchain:
            return "@io_bazel_rules_go//go/toolchain:%s_cgo" % toolchain,
    return platform

....
        platform = "@io_bazel_rules_go//go/toolchain:{}_{}{}".format(goos, goarch, "_cgo" if cgo else "")
        platform = detect_mobile_platform(settings, platform)
```

steeve:
When combined with #2445, adding a `go_binary` as a dep to an `android_binary` will correctly produce the multi-arch split:
```
$ bazel build //examples/helloworld:sampleapp
INFO: Analyzed target //examples/helloworld:sampleapp (10 packages loaded, 7133 targets configured).
INFO: Found 1 target...
[257 / 273] 4 actions running
    GoStdlib external/io_bazel_rules_go/android_386_c-shared/stdlib%/pkg; 15s darwin-sandbox
    GoStdlib external/io_bazel_rules_go/android_amd64_c-shared/stdlib%/pkg; 15s darwin-sandbox
[...]

PR #2859, opened 2021-04-03
Add automatic platform detection from inbound crosstool_top and cpu

steeve:
Reopens #2451

When an inbound `crosstool_top` and `cpu` are specified, some platforms
can be automatically inferred/detected.

Some legacy rules still implement configuration splits on
`crosstool_top` and `cpu` instead of platform constaints. This is the
case most notably for rules_apple and rules_android.

Add a generic platform matching system based on `crostool_top`/`cpu`.

Finally, add Apple and Android matchers.

Refs #2443

### DIFF
--- a/go/platform/crosstool.bzl
+++ b/go/platform/crosstool.bzl
@@ -1,0 +1,58 @@
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


gregestren:
I quickly reviewed the old thread. You thought platform mappings were an awkward fit, right? Could a modified version of that fully address this issue? An ideal outcome is to adapt an existing solution to a problem vs. re-invent the wheel multiple times.

jayconrod (approved):
Sorry for the slow review. Looks good.

jayconrod:
You should have merge power btw, so I'll leave it up to you to hit the button!

PR #2863, opened 2021-04-15
nogo: properly check filename for inclusion/exclusion

### LINKED ISSUE #2862: nogo exclude_files doesn't work properly for cgo

v0.27.0

v0.23.0

4.0.0

Yes

darwin/amd64

Nothing special. Using Command Line Tools rather than Xcode.

### What did you do?

Excluding files from nogo analysis with the `exclude_files` config key. The filename in question is something like `external/com_blah/common/abc.go` (a file in a Gazelle `go_repository`), and I added `"external/": ""` to the nogo config file.

Diagnostics from the specified analyzers get excluded when running through this file.

### What did you see instead?

The diagnostics still appear:

```
external/com_blah/common/abc.go:14:23: possible misuse of unsafe.Pointer (unsafeptr)
Target @com_blah//common:common failed to build
```

----

I dug around a bit, and added 

```go
fmt.Fprintf(os.Stderr, "%s(%s)\n", act.a.Name, filename);
```

to right before [line 422 of nogo\_main.go](https://github.com/bazelbuild/rules_go/blob/707c634bf8f7ca5b57c2a923bec16fa62ced1fc7/go/tools/builders/nogo_main.go#L422). Here's what I got in the output logs:

```
compilepkg: unsafeptr(/var/folders/y4/d2xhlc550nv5wxfl3w1mr47xrjmbp2/T/rules_go_work-234795290/abc.cgo1.go)
nogo: errors found by nogo during build-time code analysis:
external/com_blah/common/abc.go:14:23: possible misuse of unsafe.Pointer (unsafeptr)
Target @com_blah//common:common failed to build
```

It appears that this may be a result of using `token.FileSet.File` on [line 397](https://github.com/bazelbuild/rules_go/blob/707c634bf8f7ca5b57c2a923bec16fa62ced1fc7/go/tools/builders/nogo_main.go#L397), which doesn't respect cgo's `//line` adjustment comments, rather than `token.FileSet.Position`.

TimothyGu:
**What type of PR is this?**

Bug fix

Go files could include `//line` comments that indicate the original location of the line. This functionality is used, among other things, cgo's code rewrite step. This commit changes the include/exclude_files logic to use the adjusted/original location, in order to better align with user expectations.

Fixes #2862

**Other notes for review**

### DIFF
--- a/tests/core/nogo/custom/custom_test.go
+++ b/tests/core/nogo/custom/custom_test.go
@@ -79,6 +79,13 @@ go_library(
 )
 
 go_library(
+    name = "has_errors_linedirective",
+    srcs = ["has_errors_linedirective.go"],
+    importpath = "haserrors_linedirective",
+    deps = [":dep"],
+)
+
+go_library(
     name = "no_errors",
     srcs = ["no_errors.go"],
     importpath = "noerrors",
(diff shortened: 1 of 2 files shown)


tests/core/nogo/custom/custom_test.go
```diff
@@ -290,8 +297,25 @@ import (
 )
 
 func Foo() bool { // This should fail foofuncname
-	dep.D()     // This should fail visibility
-	return true // This should fail boolreturn
+	dep.D() // This should fail visibility
+	return true
```
TimothyGu, line 301: We don't have an analyzer named `boolreturn`.

jayconrod (approved):

PR #2865, opened 2021-04-15
Expand Make variables in c/cpp/cxx/clinkopts

### LINKED ISSUE #2864: cppopts doesn't go through Make variable expansion

v0.27.0

v0.23.0

4.0.0

Yes

darwin/amd64

Nothing special, using Command Line Tools rather than Xcode.

### What did you do?

Use `$(GENDIR)` in `cppopts`.

The "Make variable" get expanded to `bazel-out/darwin-fastbuild/bin`.

### What did you see instead?

`$(GENDIR)` is provided verbatim to the Go compiler, without any expansion.

TimothyGu:
**What type of PR is this?**

Bug fix

The documentation for `go_library`, `go_binary`, etc., claims that certain compiler options are subject to "Make variable" substitution. However, this is not the case for the majority of cases: only `gc_goopts` and `gc_linkopts` were handled correctly. This commit fixes that behavior.

Fixes #2864

**Other notes for review**

### DIFF
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -92,10 +92,7 @@ def emit_compilepkg(
     if testfilter:
         args.add("-testfilter", testfilter)
 
-    gc_flags = [
-        go._ctx.expand_make_variables("gc_goopts", f, {})
-        for f in gc_goopts
-    ]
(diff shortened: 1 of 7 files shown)


jayconrod (approved):
Looks good, thanks for fixing this!

PR #2869, opened 2021-04-17
Fix cgo rpath computation for prebuilt dynamic libraries

### LINKED ISSUE #2867: Cgo rpath is incorrectly computed for non-generated cc_imported dynamic libraries

0.27.0

None

4.0.0

Yes

darwin/amd64

Xcode 12.4 on macOS 11.2.1 Big Sur.

### What did you do?

See https://github.com/TimothyGu/rules_go_2867.

Essentially, we have:

* `//import` with a single `cc_import` target with `shared_library = "libhello.dylib"`. libhello.dylib is a precompiled dylib with `LC_ID_DYLIB` set to `@rpath/libhello.dylib`.

* `//src` with a single `go_binary` target that uses a function in libhello.dylib.

`bazel run //src` should run successfully.

### What did you see instead?

`bazel run //src` prints

```
❯ bazel run //src
INFO: Analyzed target //src:src (5 packages loaded, 106 targets configured).
INFO: Found 1 target...
Target //src:src up-to-date:
  bazel-bin/src/src_/src
INFO: Elapsed time: 5.368s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
dyld: Library not loaded: @rpath/libhello.dylib
  Referenced from: /private/var/tmp/_bazel_tigu/.../execroot/__main__/bazel-out/darwin-fastbuild/bin/src/src_/src
  Reason: image not found
```

### Digging deeper

We see that when the `src` binary is getting linked, a rpath is set:

```
❯ bazel aquery 'mnemonic("GoLink", //src)'
...
  Command Line: (exec bazel-out/host/bin/external/go_sdk/builder \
    link \
    [...] \
    -- \
    -extld external/local_config_cc/cc_wrapper.sh \
    '-buildid=redacted' \
    -extldflags '-fobjc-link-runtime -headerpad_max_install_names -no-canonical-prefixes -mmacosx-version-min=11.1 -Wl,-rpath,@loader_path/../../import')
```

(The very last line above includes `-Wl,-rpath,@loader_path/../../import`.)

This rpath is correct if `@loader_path` is taken to be `bazel-bin/src/src_/src.runfiles/__main__/src/src_/src`. Indeed the path

```
bazel-bin/src/src_/src.runfiles/__main__/src/src_/../../import/libhello.dylib
```

reaches libhello.dylib. However, in actuality, the `src` binary in `src.runfiles` is a symbolic link:

```
❯ readlink bazel-bin/src/src_/src.runfiles/__main__/src/src_/src
/private/var/tmp/_bazel_tigu/.../execroot/__main__/bazel-out/darwin-fastbuild/bin/src/src_/src

(i.e., bazel-bin/src/src_/src)
```

and as far as I can tell, dyld sets `@loader_path` to be the "real" path of the executable, obtained after all symlinks are resolved. By this logic we see that 

```
bazel-bin/src/src_/../../import/libhello.dylib
```

is not a valid path to the libhello.dylib file.

### Potential fix

One potential fix is to add another rpath to the linked binary that accommodates for the case where the executable is outside its own .runfiles directory. In this case, if we had

```
-extldflags '-Wl,-rpath,@loader_path/../../import -Wl,-rpath,@loader_path/src.runfiles/__main__/import'
```

then everything would run correctly.

TimothyGu:
**What type of PR is this?**

Bug fix

Previously, we assumed that any dynamically linked library exists in execroot. This is the case if the library is NOT generated (e.g., prebuilt third-party SDKs). On the other hand, the library should always be symlinked in runfiles of the executable, so use that as the basis for rpath computation.

Fixes #2867

**Other notes for review**


jayconrod:
Sorry for the slow review on this. The fix looks good.

The one thing I'm nervous about is checking in binaries as part of the tests. We have no way of reviewing that code, and if it stops working, we have no way of fixing it.

I'm guessing we can't test this fix with binaries that are generated at build time, since they'll be in generated directories?

If that's the case, I'd be fine with marking the test with `tags = ["manual"]` and checking in a bash script that can be run first to generate the binaries, which won't be checked in. I can't really think of a better testing strategy.

cc @steeve also, who knows this code better than I do at this point.

TimothyGu:
@jayconrod Done. Created a `generate_imported_dylib.sh` script for the new tests.

jayconrod (approved):

I think the RBE failure was already corrected on `master`, so I'll go ahead and submit this.

Issue #2131, opened 2019-07-10
go_binary: don't concatenate headers in c-archive, c-shared link modes


jayconrod:
The `go_binary_c_archive_shared` macro declares several additional targets around `go_binary` rules with `linkmode = "c-shared"` or `"c-archive"`. One of these gathers all the .h files in the `cgo_exports` output group and concatenates them into a single header.

This is not what the Go command does. Only the export header for the `main` package should be visible. Incorporating headers from other packages adds confusion.

PR #2874, opened 2021-05-11
Emit correct header in CGo c-archive

fmeum:
**What type of PR is this?**

Bug fix

A go_binary target with linkmode c-archive emits only a single
_cgo_exports.h header instead of all such headers of all transitive
dependencies.

Currently, this header is taken to be the first element in a list
obtained from the depset of these transitive _cgo_exports.h files
via to_list(). While deterministic, this choice is not guaranteed to
give the correct _cgo_exports.h (i.e., the one of the go_binary
itself) as the depset iteration order is not specified.

This commit ensures that the _cgo_exports.h of the target itself is
added as the first element of the `direct` list of a depset with
iteration order specified as `preorder` and will thus always be the
first element when the depset is converted to a list.

The `CcInfo` `hdrs` exposed by a `go_binary` using CGo that itself has dependencies using CGo is essentially arbitrary (but deterministic) and thus may not define the exported symbols or even be empty.


jayconrod:
I'd rather not merge a fix for this without an issue thoroughly explaining the bug being fixed plus tests that verify it works. cgo export headers are not well documented.

The main issue about this is #2131. `go_binary` should only expose one header file, named after the binary. That header should declare exported Go definitions from the `main` package only.

achew22:
+1, I would really only be comfortable merging this if there were a testcase that failed before but passes now

fmeum:
My own example is not public and quite difficult to minimize (cc target depending on go target depending on cc target depending on go target), but #2878 now provides a public example.

I am unfortunately not sure how to write a reliable test for this issue. After all, the problem is that without this fix, `rules_go` depends on the [unspecified default traversal order of a `depset`](https://docs.bazel.build/versions/master/skylark/lib/depset.html#:~:text=Order%20is%20unspecified). This order may not be guaranteed to remain stable across Bazel versions, but any concept for a test I can think of would have to depend on it in order to fail without the current PR. 

@achew22 Do you have an idea how to deal with this situation?

fmeum:
@jayconrod @achew22 Do you want me to add anything before you take another look at this. With https://github.com/bazelbuild/rules_go/issues/2878, there is now an example of a build that fails without this patch.

achew22:
That's great! I think next steps here before merging would be to add a testcase into the repo that demonstrates this behavior so that the purpose of this change is effectively documented. Thanks!

fmeum:
@achew22 Done, I have turned @glukasiknuro's [example](https://github.com/glukasiknuro/rules_go_empty_header) into a testcase that does compile and run with this PR, but fails to compile without it.

Please let me know if you need anything else.

achew22:
Huh, looking at the test output I'm [seeing](https://buildkite.com/bazel/rules-go-golang/builds/3144#d5c5e4e0-87be-4ea5-b5f5-cb1721f5c34c):

```
(20:16:59) ERROR: Analysis of target '//tests/core/c_linkmodes:go_with_cgo_dep_caller' failed; build aborted: Analysis of target '//tests/core/c_linkmodes:go_with_cgo_dep_caller' failed
 ```

Got any theories?

fmeum:
> Huh, looking at the test output I'm [seeing](https://buildkite.com/bazel/rules-go-golang/builds/3144#d5c5e4e0-87be-4ea5-b5f5-cb1721f5c34c):
> 
> ```
> (20:16:59) ERROR: Analysis of target '//tests/core/c_linkmodes:go_with_cgo_dep_caller' failed; build aborted: Analysis of target '//tests/core/c_linkmodes:go_with_cgo_dep_caller' failed
>  ```
> 
> Got any theories?
> ```

Yes, I forgot to load `cc_binary` from `rules_cc`. Should be fixed now. I also added myself to `CONTRIBUTORS`, in case that's required.

fmeum:
@achew22 Sorry, I didn't notice that the CI checks were failing. This PR is ready for review now. The race detector check is failing on Windows though, do you know whether this is related to my changes?

steeve:
> @achew22 Sorry, I didn't notice that the CI checks were failing. This PR is ready for review now. The race detector check is failing on Windows though, do you know whether this is related to my changes?

I don't think it's related to your changes. We have disabled it on `master`, you might want to try to rebase on it.

fmeum:
@steeve That worked, thanks. Should be fully ready now.

fmeum:
@achew22 The tests are all passing now, could you take another look?

fmeum:
I resolved the merge conflict.

borancar:
#fixes 2971

fmeum:
@achew22 Do you think you could take a look? It seems that people keep hitting the cgo bug this fixes.

PR #2877, opened 2021-05-15
Support plugins on Darwin/ARM-64.

phst:
See #2795.

**What type of PR is this?**

Feature

**What does this PR do? Why is it needed?**'

Support plugins on ARM-64/Darwin.

**Other notes for review**


jayconrod (approved):
Looks good, thanks for tracking this down!

PR #2881, opened 2021-05-21
update docs about go_library naming convention

tanyabouman:
**What type of PR is this?**
Documentation

Since `go_default_library` is no longer necessary and Gazelle doesn't generate `go_default_library` targets by default anymore, this updates various pieces of documentation that reference that.  There's also one line of code changed to use the new convention, but this does not affect anything, because the old naming convention is still available.


robfig:
LGTM, thank you! 

@jayconrod want to have a look? Also, I'm not sure why buildkite isn't working...

robfig:
I'm not able to reproduce the buildkite failure:
```
(22:53:27) ERROR: Analysis of target '//tests/legacy/binary_test_outputs:binary_single_output_test' failed; build aborted: Analysis failed
```

`bazel test //tests/legacy/binary_test_outputs:binary_single_output_test` succeeds for me when I patch this PR in.

jayconrod (approved):
Looks good, thanks!

PR #2882, opened 2021-05-25
[bazel_testing] Fix off-by-one error

pcj:
The current implementation always skips the last entry between -begin_files and -end_files


robfig:
Hi @pcj , thanks for the fix!

I'm afraid I'm not really familiar with how this is used. Is there any simple test that could be added or updated to demonstrate the issue & prevent regression? I agree that looking at this one file, it looks like your change is correct, though.

pcj:
I can refactor this part of the code into a utility function and test that.

jayconrod:
The CI failure should be fixed upstream. If you update `master`, it should go green.

PR #2888, opened 2021-06-01
Support multiple keys in x_defs

### LINKED ISSUE #2075: x_defs does't support multiple keys in one value
I have some `x_defs` as described in https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#stamping-with-the-workspace-status-script
```
x_defs = {
  "github.com/example/project.simple": "{BUILD_USER}",
  "github.com/example/project.compound": "{BUILD_USER}@{BUILD_HOST}",
},
```

The `github.com/example/project.simple` variable stamps fine, but the `github.com/example/project.compound` variable does not get stamped. From experimenting it seems like it is not supported to substitute multiple workspace status variables into a single variable, which is unexpected as the documentation mentions you can refer to them when using curly braces.

A reproducible project can be found at https://gist.github.com/robbertvanginkel/97dfe32e795af9c0dee9b5cd6a6d3de3

Versions:
- `rules_go` 0.18.5
- `bazel_gazelle`: N/A
- `bazel`: 0.25.2
- `os`: darwin

kaoet:
**What type of PR is this?**
Feature

Support string interpolation in `x_defs`.

```bazel
go_binary(
  x_defs = {
    "myvar": "{BUILD_USER} built me at {BUILD_TIMESTAMP}",
  }
)
```

Fixes #2075

**Other notes for review**

### DIFF
--- a/go/tools/builders/link.go
+++ b/go/tools/builders/link.go
@@ -25,6 +25,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 )
@@ -36,7 +37,6 @@ func link(args []string) error {
 		return err
 	}
 	builderArgs, toolArgs := splitArgs(args)
-	xstamps := multiFlag{}
 	stamps := multiFlag{}
 	xdefs := multiFlag{}
 	archives := archiveMultiFlag{}
@@ -49,7 +49,6 @@ func link(args []string) error {
 	packageList := flags.String("package_list", "", "The file containing the list of standard library packages")
 	buildmode := flags.String("buildmode", "", "Build mode used.")
 	flags.Var(&xdefs, "X", "A string variable to replace in the linked binary (repeated).")
-	flags.Var(&xstamps, "Xstamp", "Like -X but the values are looked up in the -stamp file.")
 	flags.Var(&stamps, "stamp", "The name of a file with stamping values.")
 	conflictErrMsg := flags.String("conflict_err", "", "Error message about conflicts to report if there's a link error.")
 	if err := flags.Parse(builderArgs); err != nil {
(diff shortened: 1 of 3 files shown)


go/tools/builders/link.go
```diff
@@ -110,33 +109,34 @@ func link(args []string) error {
...
-		}
-	}
 	for _, xdef := range xdefs {
 		pkg, name, value, err := parseXdef(xdef)
 		if err != nil {
 			return err
 		}
-		goargs = append(goargs, "-X", fmt.Sprintf("%s.%s=%s", pkg, name, value))
+		var missingKey bool
+		value = regexp.MustCompile("\\{.*?\\}").ReplaceAllStringFunc(value, func(key string) string {
```
achew22, line 130: Why do you need the `?` following `.*`? `*` is the 0 or more operator so it feels like making that optional is unnecessary. I also think that we would end up in a index panic with this when someone passes in `{}` to be matched.

WDYT about doing something like:

```
		value = regexp.MustCompile(`\{.+\}`).ReplaceAllStringFunc(value, func(key string) string {Write<br class="Apple-interchange-newline">
```

2 changes in there

1. Uses raw string syntax
2. Uses the `+` operator over the `*` operator which would only match if there is one or more entry. That would mean there is always at least a single character inside the `{}` which would prevent the negative index panic I think I see.

kaoet: The question mark here put the `*` in [lazy mode](https://en.wikipedia.org/wiki/Regular_expression#Lazy_matching)

So for a string like `ab{c}d{e}f`, it produces two matches `{c}` and `{e}` instead of one match `{c}d{e}`.

* [x] Raw string
* [x] Use `+` intead of `*`

go/tools/builders/link.go
```diff
@@ -110,33 +109,34 @@ func link(args []string) error {
...
-		}
-	}
 	for _, xdef := range xdefs {
 		pkg, name, value, err := parseXdef(xdef)
 		if err != nil {
 			return err
 		}
-		goargs = append(goargs, "-X", fmt.Sprintf("%s.%s=%s", pkg, name, value))
+		var missingKey bool
+		value = regexp.MustCompile(`\{.+?\}`).ReplaceAllStringFunc(value, func(key string) string {
```
achew22, line 130: I think that `\{.+?\}` would more commonly be written as `\{.*\}`. If you make it that way I'm happy to merge.

achew22 (changes requested):
One small nit but otherwise looks good

go/tools/builders/link.go
kaoet: As I said in the previous comment, `.+?` is not equivalent to `.*` here. The `?` right after `*` doesn't  represent "optional". It actually enable [lazy matching](https://en.wikipedia.org/wiki/Regular_expression#Lazy_matching) here.

achew22: Ah, apologies around that. As is usually the case when I interact with regex I learned something. Thanks for teaching me!

achew22 (approved):

PR #2903, opened 2021-06-30
Upgrade dependencies for v0.28.0

jayconrod:
* Added releaser directives marking automatically upgradeable deps.
* Fixed //tests/core/go_proto_library:proto_package_test: since
  v1.26.0, protoc-gen-go requires that the import path is specified
  via 'option go_package' or an equivalent M flag. The package name is
  no longer derived from the proto package name in either case.
* Fixed //tests/core/nogo/vet:vet_test: the buildtag error message
  text changed in the new version of org_golang_x_tools.


achew22 (approved):
Insofar as anything like this can be reviewed, looks good

jayconrod:
The patches are never going to be readable, but I hope the releaser tool in #2904 will make this easier to verify in the future. I realize that it will error out if it can't upload to GCS, but other than that, it should produce the same diff, modulo test fixes.

PR #2904, opened 2021-06-30
go/tools/releaser: new tool to help with multiple aspects of releases

jayconrod:
For now, this has an "upgrade-dep" subcommand that automates updating
a dependency in WORKSPACE or go/private/repositories.bzl. Upgradeable
dependencies must be marked with a '# releaser:upgrade-dep org repo'
directive that specifies the GitHub organization and repository. You
can run 'bazel run //go/tools/releaser upgrade-dep all' to upgrade all
dependencies.


jayconrod:
For now, this probably won't be useful to anyone other than me, since it requires credentials to upload to mirror.bazel.build. At least it codifies the upgrade process though.

go/tools/releaser/file.go
```diff
@@ -0,0 +1,286 @@
...
+			dir = parent
+		}
+	})
+	return repoRootState.dir, repoRootState.err
+}
+
+// extractArchive extracts a zip or tar.gz archive opened in f, into the
+// directory dir, stripping stripPrefix from each entry before extraction.
+// name is the name of the archive, used for error reporting.
+func extractArchive(f *os.File, name, dir, stripPrefix string) (err error) {
```
achew22, line 77: I would have expected this to return an [fs.FS](https://pkg.go.dev/io/fs#FS)

achew22: Looking at `archive/tar` et al, looks like that might be a tall order. You could use use [`os.DirFS`](https://pkg.go.dev/os#DirFS) to provide this without much extra work.

go/tools/releaser/github.go
```diff
@@ -0,0 +1,89 @@
...
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/google/go-github/v36/github"
+)
+
+func githubListTags(ctx context.Context, gh *github.Client, org, repo string) (_ []*github.RepositoryTag, err error) {
```
achew22, line 27: Have you considered creating a struct that has a `*github.Client` on it and turning this into just `releaserGH.listTags(ctx, org, repo)`?

go/tools/releaser/github.go
```diff
@@ -0,0 +1,89 @@
...
+		if resp.NextPage == 0 {
+			return nil
+		}
+		opt.Page = resp.NextPage
+	}
+}
+
+// githubTokenFlag is used to find a GitHub personal access token on the
+// command line. It accepts a raw token or a path to a file containing a token.
+type githubTokenFlag struct {
```
achew22, line 67: Don't think it matters, but you could also do

`type githubTokenFlag *string`

go/tools/releaser/releaser.go
```diff
@@ -0,0 +1,105 @@
...
+		upgradeDep,
+	}
+}
+
+func run(ctx context.Context, stderr io.Writer, args []string) error {
+	if len(args) == 0 {
+		return errors.New("no command specified. For a list of commands, run:\n\treleaser help")
+	}
+	name, args := args[0], args[1:]
+	if name == "-h" || name == "-help" || name == "--help" {
```
achew22, line 55: I think you should iterate through args and do this check so it doesn't have to be first

```go
for _, arg := range args {
  if arg == "-h" || arg == "-help" || arg == "--help" {
    return helpCmd.run(ctx, stderr, args)
  }
}
```

go/tools/releaser/upgradedep.go
```diff
@@ -0,0 +1,518 @@
...
+Once upgrade-dep has found the URL for the latest version, it will:
+
+* Download the archive.
+* Upload the archive to mirror.bazel.build.
+* Re-generate patches, either by running a command or by re-applying the
+  old patches.
+* Update dependency attributes in WORKSPACE and repositories.bzl, then format
+  and rewrite those files.
+
+Upgradeable dependencies need a comment like '# releaser:upgrade-dep org repo'
```
achew22, line 67: Should this work the other way? Is it possible to default to things being upgradable unless otherwise specified?

achew22, line 67: is it possible to infer org and repo from context?

achew22, line 67: If you do `org` and `repo` in the raw there is no forward compatibility with a different strategy. That may not matter, but my preference would be to do `github.com/org/repo` as a single argument

go/tools/releaser/upgradedep.go
```diff
@@ -0,0 +1,518 @@
...
+	urlsKwarg.Before = []bzl.Comment{{Token: "# " + urlComment}}
+
+	return nil
+}
+
+// parseUpgradeDepDirective parses a '# releaser:upgrade-dep org repo' directive
+// and returns the organization and repository name or an error if the directive
+// was not found or malformed.
+func parseUpgradeDepDirective(call *bzl.CallExpr) (orgName, repoName string, err error) {
+	for _, c := range call.Comment().Before {
```
achew22, line 485: I think right here you could `switch` on known things like `go_repository` or `http_archive` and handle those automatically.

achew22 (approved):

go/tools/releaser/file.go
jayconrod: For this purpose, I really do need to extract the archive though, so that diff, patch, and gazelle can run in the extracted directories.

go/tools/releaser/upgradedep.go
jayconrod: It's probably a good idea eventually, but I'd like to gain more confidence in this approach first. Supporting `go_repository` and `git_repository` first would be good.

This is just meant to support rules_go and Gazelle maintenance so I'm not too worried about changing the directives at this point.

jayconrod: They can probably be inferred at least from `urls`. Still, I'd like to keep it explicit for now until I'm more confident in the approach.

jayconrod: Let's not worry about forward compatibility quite yet. This will likely change in the future, but there aren't many directives to update.

go/tools/releaser/upgradedep.go
jayconrod: Added a TODO.

PR #2905, opened 2021-06-30
Remove ubuntu1604 from presubmit.yml

philwo:
Ubuntu 16.04 is end-of-life, we're going to remove it from Bazel CI.


jayconrod (approved):
Looks good, thanks!

PR #2930, opened 2021-07-28
CGO merged directory should have usable name.

thomas-wk:
cgo2 merges sources and generated code under a common directory, however this directory is anonymous.

This breaks nogo's ability to easily include/exclude files as only the file's basename is retained.

This change embeds cgo/packagePath into the merged directory to make it easy to include/exclude sources being compiled by cgo.

Note: PR #2863 ensured that files //line: directives had their meaningful alt names included in nogo output. This is helpful, however for pure go files that are part of a package that uses cgo, there is nothing that adds //line: directives, and the builder will just copy/link pure files into the into the merged cgo folder.

**What type of PR is this?**

> Bug fix

Makes the merged cgo temp directory include the package path for easy inclusion/exclusion from nogo rules.

Fixes # 2862

**Other notes for review**

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -58,6 +58,15 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 	}
 	defer cleanup()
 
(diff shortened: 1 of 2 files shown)


robfig:
Thank you for the contribution with unit tests! Merging.

PR #2858, opened 2021-04-03
go/tools: add gopackagesdriver

steeve:
This commit introduces an aspect based `GOPACKAGESDRIVER` for `rules_go`.

It is rather clunky, but all VSCode + `gopls` features work with it:
- Completion
- Go to definition
- Go to references
- Quick lookup
- Inline documentation

It does not support static usage at the moment (for instance by running it inside an action for codegen tools), but it could with few additions.

## Usage
The packages driver fundamentally works from the perspective of one or more targets. In order
for it to function properly, the targets must be specified in the packages driver configuration,
with environment variables.

These are:
- `GOPACKAGESDRIVER_BAZEL_TARGETS`: specifies specific targets, `//...` works too (although this is not recommended without tag filters). It is possible to set targets to something other than `go_` rules, as long as the `go_` nodes are reachable via the `deps` attribute in the graph. Such as when `go_binary(linkmode = "c-archive") -> cc_library`.
- `GOPACKAGESDRIVER_BAZEL_QUERY`: runs a `bazel query` to select targets before, common one should be `kind(go_binary, //...)`. This should work for most cases.
- `GOPACKAGESDRIVER_BAZEL_TAG_FILTERS`: makes use of the [`--build_tag_filters`](https://docs.bazel.build/versions/3.2.0/command-line-reference.html#flag--build_tag_filters) option to only build (filter) targets with certain tags. This is useful for when you want each target to define wether or not it should be part of the packages list. For instance, set `GOPACKAGESDRIVER_BAZEL_TARGETS="//..."` and `GOPACKAGESDRIVER_BAZEL_TAG_FILTERS="completion"` to only build targets with the `completion` tag in the whole workspace.

In addition, those environment variables are optional:
- `GOPACKAGESDRIVER_BAZEL`: bazel binary, defaults to `bazel`
- `BUILD_WORKSPACE_DIRECTORY`: directory of the bazel workspace (should be auto detected when using a launcher script because it invokes `bazel run`)

### 1. `gopls`
You'll need `gopls >= v0.6.10` ([released on Apr 13th 2021](https://github.com/golang/tools/releases/tag/gopls%2Fv0.6.10)). If you are using VSCode, it should be automatic.

### 2. Create a launcher script
Create a launcher script, say `tools/gopackagesdriver.sh`:
```bash
#!/usr/bin/env bash
exec bazel run -- @io_bazel_rules_go//go/tools/gopackagesdriver "${@}"
```

### 3. Sample VSCode configuration
[...]

PR #2932, opened 2021-08-03
go/tools/gopackagesdriver: add automatic target detection

steeve:
**What type of PR is this?**

Feature

This PR introduces automatic target detection so that no user input
is required after the `GOPACKAGESDRIVER` setup. This effectively
deprecates the following environment variables:
- `GOPACKAGESDRIVER_BAZEL_TARGETS`
- `GOPACKAGESDRIVER_BAZEL_QUERY`
- `GOPACKAGESDRIVER_BAZEL_TAG_FILTERS`

It works as follows:
- for `importpath` queries, it will `bazel query` a matching `go_library`
    matching it
- for `file=` queries, it will try to find the matching `go_library` in the
    same package
- since it supports multiple queries at the same time, it will `union` those
    queries and query that

Once the `go_library` targets are found, it will try to build them directly,
which should dramatically speed up compilation times, at the loss of transition
support (which wasn't used as much as I thought it would be).

I may reintroduce it in the future via a user-defined flag (to only build the
part of the graph that needs building).

In any case, toolchain or platforms can be switched with the following
environment variables it configuration transitions are needed:
- `GOPACKAGESDRIVER_BAZEL_FLAGS` which will be passed to `bazel` invocations
- `GOPACKAGESDRIVER_BAZEL_QUERY_FLAGS` which will be passed to `bazel query`
    invocations
- `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE` which specifies the scope for `importpath` queries (since `gopls` only issues `file=` queries, so **use if you know what you're doing!**)
- `GOPACKAGESDRIVER_BAZEL_BUILD_FLAGS` which will be passed to `bazel build`
    invocations

Finally, the driver will not fail in case of a build failure, and even so
uses `--keep_going` and will return whatever packages that did build.

It should fix most of the issues from #2858.

**Other notes for review**
I have added a `.vscode` folder for folks who work on `rules_go` with it so that `gopls` now works there too!

### DIFF
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -35,20 +35,25 @@ def _file_path(f):
         prefix = "__BAZEL_OUTPUT_BASE__"
     return paths.join(prefix, f.path)
 
-def _new_pkg_info(archive_data):
+def _go_archive_to_pkg(archive):
     return struct(
-        ID = str(archive_data.label),
-        PkgPath = archive_data.importpath,
-        ExportFile = _file_path(archive_data.export_file),
+        ID = str(archive.data.label),
+        PkgPath = archive.data.importpath,
+        ExportFile = _file_path(archive.data.export_file),
         GoFiles = [
             _file_path(src)
-            for src in archive_data.orig_srcs
+            for src in archive.data.orig_srcs
         ],
         CompiledGoFiles = [
             _file_path(src)
-            for src in archive_data.srcs
+            for src in archive.data.srcs
         ],
     )
+
+def _make_pkg_json(ctx, archive, pkg_info):
+    pkg_json_file = ctx.actions.declare_file(archive.data.name + ".pkg.json")
+    ctx.actions.write(pkg_json_file, content = pkg_info.to_json())
+    return pkg_json_file
 
 def _go_pkg_info_aspect_impl(target, ctx):
     # Fetch the stdlib JSON file from the inner most target
@@ -57,6 +62,7 @@ def _go_pkg_info_aspect_impl(target, ctx):
     deps_transitive_json_file = []
     deps_transitive_export_file = []
     deps_transitive_compiled_go_files = []
+
     for attr in ["deps", "embed"]:
         for dep in getattr(ctx.rule.attr, attr, []):
             if GoPkgInfo in dep:
@@ -76,28 +82,29 @@ def _go_pkg_info_aspect_impl(target, ctx):
                 if not stdlib_json_file:
                     stdlib_json_file = pkg_info.stdlib_json_file
 
-    pkg_json_file = None
+    pkg_json_files = []
     compiled_go_files = []
-    export_file = None
+    export_files = []
+
     if GoArchive in target:
         archive = target[GoArchive]
-        compiled_go_files = archive.data.srcs
-        export_file = archive.data.export_file
-        pkg = struct(
-            ID = str(archive.data.label),
-            PkgPath = archive.data.importpath,
-            ExportFile = _file_path(archive.data.export_file),
-            GoFiles = [
-                _file_path(src)
-                for src in archive.data.orig_srcs
-            ],
-            CompiledGoFiles = [
-                _file_path(src)
-                for src in archive.data.srcs
-            ],
-        )
-        pkg_json_file = ctx.actions.declare_file(archive.data.name + ".pkg.json")
-        ctx.actions.write(pkg_json_file, content = pkg.to_json())
+        compiled_go_files.extend(archive.source.srcs)
+        export_files.append(archive.data.export_file)
+        pkg = _go_archive_to_pkg(archive)
+        pkg_json_files.append(_make_pkg_json(ctx, archive, pkg))
+
+        # if the rule is a test, we need to get the embedded go_library with the current
+        # test's sources. For that, consume the dependency via GoArchive.direct so that
+        # the test source files are there. Then, create the pkg json file directly. Only
+        # do that for embedded targets, and use the importpath go find which.
+        if ctx.rule.kind == "go_test":
+            embedded_targets = [dep[GoArchive].data.importpath for dep in ctx.rule.attr.embed]
+            for archive in target[GoArchive].direct:
+                if archive.data.importpath in embedded_targets:
+                    pkg = _go_archive_to_pkg(archive)
+                    pkg_json_files.append(_make_pkg_json(ctx, archive, pkg))
+                    compiled_go_files.extend(archive.source.srcs)
+                    export_files.append(archive.data.export_file)
 
     # If there was no stdlib json in any dependencies, fetch it from the
     # current go_ node.
@@ -105,10 +112,9 @@ def _go_pkg_info_aspect_impl(target, ctx):
         stdlib_json_file = ctx.attr._go_stdlib[GoStdLib]._list_json
 
     pkg_info = GoPkgInfo(
-        json = pkg_json_file,
         stdlib_json_file = stdlib_json_file,
         transitive_json_file = depset(
-            direct = [pkg_json_file] if pkg_json_file else [],
+            direct = pkg_json_files,
             transitive = deps_transitive_json_file,
         ),
         deps_transitive_json_file = depset(
@@ -122,7 +128,7 @@ def _go_pkg_info_aspect_impl(target, ctx):
             transitive = deps_transitive_compiled_go_files,
         ),
         transitive_export_file = depset(
-            direct = [export_file] if export_file else [],
+            direct = export_files,
             transitive = deps_transitive_export_file,
         ),
         deps_transitive_export_file = depset(
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -18,16 +18,8 @@ load(
     "GoStdLib",
 )
 load(
-    "//go/private:context.bzl",
-    "go_context",
-)
-load(
     "@bazel_skylib//lib:paths.bzl",
     "paths",
-)
-load(
-    "@bazel_skylib//lib:collections.bzl",
-    "collections",
 )
 
 GoPkgInfo = provider()
@@ -40,28 +32,43 @@ def _file_path(f):
         return paths.join("__BAZEL_WORKSPACE__", f.path)
     return paths.join("__BAZEL_EXECROOT__", f.path)
 
+def _new_pkg_info(archive_data):
+    return struct(
+        ID = str(archive_data.label),
+        PkgPath = archive_data.importpath,
+        ExportFile = _file_path(archive_data.export_file),
+        GoFiles = [
+            _file_path(src)
+            for src in archive_data.orig_srcs
+        ],
+        CompiledGoFiles = [
+            _file_path(src)
+            for src in archive_data.srcs
+        ],
+    )
+
 def _go_pkg_info_aspect_impl(target, ctx):
     # Fetch the stdlib JSON file from the inner most target
     stdlib_json_file = None
 
     deps_transitive_json_file = []
     deps_transitive_export_file = []
-    for dep in getattr(ctx.rule.attr, "deps", []):
-        if GoPkgInfo in dep:
-            pkg_info = dep[GoPkgInfo]
-            deps_transitive_json_file.append(pkg_info.transitive_json_file)
-            deps_transitive_export_file.append(pkg_info.transitive_export_file)
-            # Fetch the stdlib json from the first dependency
-            if not stdlib_json_file:
-                stdlib_json_file = pkg_info.stdlib_json_file
+    for attr in ["deps", "embed"]:
+        for dep in getattr(ctx.rule.attr, attr, []):
+            if GoPkgInfo in dep:
+                pkg_info = dep[GoPkgInfo]
+                if attr == "deps":
+                    deps_transitive_json_file.append(pkg_info.transitive_json_file)
+                    deps_transitive_export_file.append(pkg_info.transitive_export_file)
+                elif attr == "embed":
+                    # If deps are embedded, do not gather their json or export_file since they
+                    # are included in the current target, but do gather their deps'.
+                    deps_transitive_json_file.append(pkg_info.deps_transitive_json_file)
+                    deps_transitive_export_file.append(pkg_info.deps_transitive_export_file)
 
-    # If deps are embedded, do not gather their json or export_file since they
-    # are included in the current target, but do gather their deps'.
-    for dep in getattr(ctx.rule.attr, "embed", []):
-        if GoPkgInfo in dep:
-            pkg_info = dep[GoPkgInfo]
-            deps_transitive_json_file.append(pkg_info.deps_transitive_json_file)
-            deps_transitive_export_file.append(pkg_info.deps_transitive_export_file)
+                # Fetch the stdlib json from the first dependency
+                if not stdlib_json_file:
+                    stdlib_json_file = pkg_info.stdlib_json_file
 
     pkg_json_file = None
     export_file = None
@@ -83,6 +90,7 @@ def _go_pkg_info_aspect_impl(target, ctx):
         )
         pkg_json_file = ctx.actions.declare_file(archive.data.name + ".pkg.json")
         ctx.actions.write(pkg_json_file, content = pkg.to_json())
+
         # If there was no stdlib json in any dependencies, fetch it from the
         # current go_ node.
         if not stdlib_json_file:
@@ -113,7 +121,7 @@ def _go_pkg_info_aspect_impl(target, ctx):
         OutputGroupInfo(
             go_pkg_driver_json_file = pkg_info.transitive_json_file,
             go_pkg_driver_export_file = pkg_info.transitive_export_file,
-            go_pkg_driver_stdlib_json_file = depset([pkg_info.stdlib_json_file] if pkg_info.stdlib_json_file else [])
+            go_pkg_driver_stdlib_json_file = depset([pkg_info.stdlib_json_file] if pkg_info.stdlib_json_file else []),
         ),
     ]
 
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -28,9 +28,12 @@ def _is_file_external(f):
     return f.owner.workspace_root != ""
 
 def _file_path(f):
-    if f.is_source and not _is_file_external(f):
-        return paths.join("__BAZEL_WORKSPACE__", f.path)
-    return paths.join("__BAZEL_EXECROOT__", f.path)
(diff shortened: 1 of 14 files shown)


go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,72 +17,104 @@ package main
...
+func (b *BazelJSONBuilder) fileQuery(filename string) string {
+	if filepath.IsAbs(filename) {
+		fp, _ := filepath.Rel(b.bazel.WorkspaceRoot(), filename)
+		filename = fp
+	}
+	return fmt.Sprintf(`some(kind("go_library", same_pkg_direct_rdeps("%s")))`, filename)
+}
+
+func (b *BazelJSONBuilder) packageQuery(importPath string) string {
+	return fmt.Sprintf(`some(kind("go_library", attr(importpath, "%s", //...)))`, importPath)
```
linzhp, line 45: Queries like this would take a very long time to finish in a repository with 100K targets and 2000 third-party libraries, like Uber's Go monorepo

linzhp (commented):
Can we continue to support `GOPACKAGESDRIVER_BAZEL_QUERY` etc? This is important for large repos like Uber's Go monorepo. Although it's huge, most users only work on small subset of targets and only need to index their projects and dependencies. It's quite a waste of time for them to wait for hours for package driver to load `//...`

go/tools/gopackagesdriver/bazel_json_builder.go
steeve: several things we can do:
1. have the scope configurable via environment
2. restrict the scope to the file's package, which is most of the time where the file lives
3. for `importpath` queries, this is trickier as the package could be anywhere in `//...`, however, package queries don't happen when using `gopls`, as it only gives out `file=` ones

steeve: actually `file=` queries use `same_pkg_direct_rdeps` so it shouldn't take long when editing

steeve: ah wait! it's the universe scope of `//...` that loads package anyway. let me try something else.

steeve:
@linzhp I have updated the PR, it shouldn't load the whole `//...` for file queries now

steeve:
@linzhp I have added `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE` which should control the scope when issuing `importpath` queries. That said, you should _not_ need it for `gopls` as it issues `file=` queries, and I have removed the whole universe loading for those.

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,72 +17,102 @@ package main
...
+func (b *BazelJSONBuilder) fileQuery(filename string) string {
+	if filepath.IsAbs(filename) {
+		fp, _ := filepath.Rel(b.bazel.WorkspaceRoot(), filename)
+		filename = fp
+	}
+	return fmt.Sprintf(`some(kind("go_library", same_pkg_direct_rdeps("%s")))`, filename)
+}
+
+func (b *BazelJSONBuilder) packageQuery(importPath string) string {
+	return fmt.Sprintf(`some(kind("go_library", attr(importpath, "%s", %s)))`, importPath, bazelQueryScope)
```
linzhp, line 45: Should be:
```suggestion
	return fmt.Sprintf(`some(kind("go_library", attr(importpath, "%s", deps(%s))))`, importPath, bazelQueryScope)
```
so we can get package information from the dependencies

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,72 +17,102 @@ package main
...
+}
+
+func (b *BazelJSONBuilder) queryFromRequests(requests ...string) string {
+	ret := make([]string, 0, len(requests))
+	for _, request := range requests {
+		if request == "." || request == "./..." {
+			return wildcardQuery
```
linzhp, line 52: gopls sends a `./...` request to package driver when VS Code opens. This wildcardQuery will also load the universe. Maybe we should just load only `bazelQueryScope`?

```suggestion
			return fmt.Sprintf("kind(go_library, %s)", bazelQueryScope)
```

steeve: that's a very good find.
although i'm thinking it may be best to disable `./...` altogether ?

steeve: my reasoning being, is there a valid usage to querying `./...` with `gopls`? 
I think your change may be the most sensible one, perhaps also allowing folks to specify an empty scope to disable it altogether

linzhp: I am fine with either approach. I am not familiar with gopls enough to say whether `./...` is necessary. Disabling `./...` may make gopls faster when VS code first opens

steeve: I think I found a good workaround: by default `./...` will return only stdlib, unless a scope is specified, in which case it will use that. Same for `importpath` queries.

linzhp: That's actually the same as ignoring `./...` because gopls sends `./... builtin` request when VS Code starts.

steeve:
@linzhp I have fixed lots of bugs, can you try this version?

steeve:
Here's a summary of the changes:
- when called without `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE`, return stdlib packages by default (so that vscode doesn't complain)
- add a compiled sources output group to ensure all files are generated and present in the sandbox
- enable querying by `importpath` prefix (`github.com/bazelbuild/rules_go/...`)
- fix path for external generated files
- fix path for external source files

go/tools/gopackagesdriver/bazel.go
```diff
@@ -88,7 +103,13 @@ func (b *Bazel) Build(ctx context.Context, args ...string) ([]string, error) {
 		"--build_event_json_file_path_conversion=no",
 	}, args...)
 	if _, err := b.run(ctx, "build", args...); err != nil {
-		return nil, fmt.Errorf("bazel build failed: %w", err)
+		// Ignore a regular build failure to get partial data.
+		// See https://docs.bazel.build/versions/main/guide.html#what-exit-code-will-i-get on
+		// exit codes.
+		var exerr *exec.ExitError
+		if !errors.As(err, &exerr) || (errors.As(err, &exerr) && exerr.ExitCode() != 1) {
```
linzhp, line 110: equivalent to
```suggestion
		if !errors.As(err, &exerr) || exerr.ExitCode() != 1 {
```

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,80 +17,115 @@ package main
...
+	if filepath.IsAbs(filename) {
+		fp, _ := filepath.Rel(b.bazel.WorkspaceRoot(), filename)
+		filename = fp
+	}
+	return fmt.Sprintf(`some(kind("go_library", same_pkg_direct_rdeps("%s")))`, filename)
+}
+
+func (b *BazelJSONBuilder) packageQuery(importPath string) string {
+	if strings.HasSuffix(importPath, "/...") {
+		importPath = strings.TrimSuffix(importPath, "/...") + "/.*"
```
linzhp, line 43: Should we transform `example.com/pkg/...` to `^example.com/pkg(/.+)?$`, so it includes `example.com/pkg` and `example.com/pkg/foo`, but not `myexample.com/pkg/foo` or `example.com/pkg2`?

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,80 +17,115 @@ package main
...
+				ret = append(ret, fmt.Sprintf(`kind("go_library", %s)`, bazelQueryScope))
+			} else {
+				ret = append(ret, fmt.Sprintf(RulesGoStdlibLabel))
+			}
+		} else if request == "builtin" || request == "std" {
+			ret = append(ret, fmt.Sprintf(RulesGoStdlibLabel))
+		} else if strings.HasPrefix(request, "file=") {
+			f := strings.TrimPrefix(request, "file=")
+			ret = append(ret, b.fileQuery(f))
+		} else if bazelQueryScope != "" {
```
linzhp, line 62: is `bazelQueryScope` mandatory to accept package queries?

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,80 +17,115 @@ package main
...
+		filename = fp
+	}
+	return fmt.Sprintf(`some(kind("go_library", same_pkg_direct_rdeps("%s")))`, filename)
+}
+
+func (b *BazelJSONBuilder) packageQuery(importPath string) string {
+	if strings.HasSuffix(importPath, "/...") {
+		importPath = strings.TrimSuffix(importPath, "/...") + "/.*"
+	}
+	return fmt.Sprintf(`some(kind("go_library", attr(importpath, "%s", deps(%s))))`, importPath, bazelQueryScope)
```
linzhp, line 45:
```suggestion
	return fmt.Sprintf(`kind("go_library", attr(importpath, "%s", deps(%s)))`, importPath, bazelQueryScope)
```
Removing the `some` function would allow us to handle unusual cases nicely:
* When there are more than one `go_library` matching the query, we can return the first one
* When there is no such `go_library`, the `some` function would lead to errors like `ERROR: Evaluation of query failed: argument set is empty`, with exit code 7. It is unclear whether the query has a bug or there is no matching package. If we don't have the `some` function, Bazel query will exit normally with empty result. We can report a better error to gopls for this case.

Same applies to the file query

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,80 +17,115 @@ package main
...
-			"--noimplicit_deps",
-			"--notool_deps",
-			b.query,
-		)
-		if err != nil {
-			return nil, fmt.Errorf("unable to query %v: %w", b.query, err)
-		}
-		buildsArgs = append(buildsArgs, queryTargets...)
+func (b *BazelJSONBuilder) Build(ctx context.Context, mode LoadMode) ([]string, error) {
+	labels, err := b.query(ctx, b.queryFromRequests(b.requests...))
```
linzhp, line 106: Need to check whether `len(labels) == 0`

go/tools/gopackagesdriver/main.go
```diff
@@ -45,77 +49,64 @@ type driverResponse struct {
...
 func main() {
-	if err := run(); err != nil {
+	response, err := run()
+	if err := json.NewEncoder(os.Stdout).Encode(response); err != nil {
+		fmt.Fprintf(os.Stderr, "unable to encode response: %v", err)
+	}
+	if err != nil {
 		fmt.Fprintf(os.Stderr, "error: %v", err)
-		os.Exit(1)
+		os.Exit(0)
```
linzhp, line 110: Why exiting with 0?

linzhp (commented):
It works now. Commented on other minor issues.

go/tools/gopackagesdriver/bazel_json_builder.go
steeve: good call

go/tools/gopackagesdriver/bazel_json_builder.go
steeve: i'd like it to, because else it's very easy to shoot yourself in the foot

go/tools/gopackagesdriver/main.go
steeve: because gopls will check the packages driver exit code, and if there is an error, and fall back to `go list`: https://github.com/golang/tools/blob/8c34cc9cafff8f13c3f3ed95a11de309752c1c60/go/packages/external.go#L88-L90

go/tools/gopackagesdriver/bazel_json_builder.go
steeve: good call

steeve:
@linzhp can you confirm the CLA please?

thinkiny:
can't wait this merge to master

linzhp (commented):
Some tests are failing.

It's strange that googlebot thought I didn't have CLA, given I already had several accepted pull requests to this repo.

@googlebot I consent.

go/tools/gopackagesdriver/bazel_json_builder.go
steeve: good call, i did not yet

steeve:
The test failure is unrelated (race)

steeve:
@linzhp I have added https://github.com/bazelbuild/rules_go/pull/2932/commits/3512e5976cf38889c4a94096fe95d4ac3a038cd7 which makes the gopackagesdriver honor the build tags when listing files. Can you try it ?

steeve:
> can't wait this merge to master

very soon !

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,80 +17,119 @@ package main
...
+	RulesGoStdlibLabel = "@io_bazel_rules_go//:stdlib"
 )
 
-func NewBazelJSONBuilder(bazel *Bazel, query, tagFilters string, targets []string) (*BazelJSONBuilder, error) {
+func (b *BazelJSONBuilder) fileQuery(filename string) string {
+	if filepath.IsAbs(filename) {
+		fp, _ := filepath.Rel(b.bazel.WorkspaceRoot(), filename)
+		filename = fp
+	}
+	return fmt.Sprintf(`kind("go_library", same_pkg_direct_rdeps("%s"))`, filename)
```
linzhp, line 38: This may return empty result when the `filename` is a test file like `foo_test.go`, because no `go_library` would depend on such file

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -17,80 +17,119 @@ package main
...
+				ret = append(ret, fmt.Sprintf(RulesGoStdlibLabel))
+			}
+		} else if request == "builtin" || request == "std" {
+			ret = append(ret, fmt.Sprintf(RulesGoStdlibLabel))
+		} else if strings.HasPrefix(request, "file=") {
+			f := strings.TrimPrefix(request, "file=")
+			ret = append(ret, b.fileQuery(f))
+		} else if bazelQueryScope != "" {
+			ret = append(ret, b.packageQuery(request))
+		}
```
linzhp, line 64:
```suggestion
		result := ""
		if request == "." || request == "./..." {
			if bazelQueryScope != "" {
				result = fmt.Sprintf(`kind("go_library", %s)`, bazelQueryScope)
			} else {
				result = fmt.Sprintf(RulesGoStdlibLabel)
			}
		} else if request == "builtin" || request == "std" {
			result = fmt.Sprintf(RulesGoStdlibLabel)
		} else if strings.HasPrefix(request, "file=") {
			f := strings.TrimPrefix(request, "file=")
			result = b.fileQuery(f)
		} else if bazelQueryScope != "" {
			result = b.packageQuery(request)
		}
		if result != "" {
			ret = append(ret, result)
		}
```

Need to check whether the query result is empty before appending, otherwise later `bazel build` may complain
```
ERROR: Skipping '': the empty string is not a valid target
WARNING: Target pattern parsing failed.
```

linzhp:
> I have added 3512e59 which makes the gopackagesdriver honor the build tags when listing files. Can you try it ?
How should I try it? It seems to be a separate feature. Can you put it in a separate pull request?

linzhp:
> The test failure is unrelated (race)

It failed consistently in recent commits, but not in earlier commits or master. So it may be broken by some recent commit

steeve:
> How should I try it? It seems to be a separate feature. Can you put it in a separate pull request?

Just verify that it correctly work as it did before. It's done behind the scenes.

go/tools/gopackagesdriver/bazel_json_builder.go
steeve: yeah, test files are not done atm, if you have an idea, i'm all ears

steeve:
@linzhp careful:
> leaving a comment that **contains only** `@googlebot I consent`. in this pull request.

steeve:
Jesus everytime I delete the googlebot mesasge (so that only the last remains), it adds another :(

steeve:
I'm _this_ close to making tests work too. The only issue is I can't get the `_test.go` files from within `GoArchive` provider as they are embedded in the `go_test` target, which then compiles them with the `go_default_library`.

linzhp:
@googlebot I consent.

linzhp:
package driver is now broken:

```
Error loading workspace: /Users/zplin/gocode/src/github.com/bazelbuild/rules_go/tools/gopackagesdriver.sh: exit status 127: /Users/zplin/gocode/src/github.com/bazelbuild/rules_go/tools/gopackagesdriver.sh: line 3: exec: bazel: not found : packages.Load error
```

steeve:
I think this is your machine @linzhp `line 3: exec: bazel: not found`, it can't find `bazel`

steeve:
So I got tests working in this latest version.

go/tools/gopackagesdriver/build_context.go
```diff
@@ -0,0 +1,38 @@
...
+	"go/build"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var buildContext = makeBuildContext()
+
+func makeBuildContext() *build.Context {
+	bctx := &build.Context{
```
linzhp, line 13: I suspect these lines made package driver not able to find Bazel in PATH.

steeve:
@linzhp standalone tests are fixed

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -17,90 +17,123 @@ load(
...
+        pkg = _go_archive_to_pkg(archive)
+        pkg_json_files.append(_make_pkg_json(ctx, archive, pkg))
+
+        # if the rule is a test, we need to get the embedded go_library with the current
+        # test's sources. For that, consume the dependency via GoArchive.direct so that
+        # the test source files are there. Then, create the pkg json file directly. Only
+        # do that for direct dependencies that are not defined as deps, and use the
+        # importpath to find which.
+        if ctx.rule.kind == "go_test":
+            deps_targets = [
```
linzhp, line 102: Maybe it's better to find the `go_library` target in the same directory as the `go_test` with regular (not generated) src files. When Gazelle generate `go_test` without `embed`, it puts the `:go_default_library` in `deps`. So we need that from `deps`.

steeve: I don't think it is, as I try to let each target generate their own files, so that the analysis graph remains stable and the resource usage is kept "low".
If the `go_library` is in `deps`, then things will work as expected without all this code. The hard part is handling `embed`. Consuming the "augmented" version requires consuming `GoArchiveData.direct`, so we can't use the `Targets`. Hence this piece.

thinkiny:
awesome

PR #2936, opened 2021-08-10
Make go_test() with embedsrcs work

EdSchouten:
Even though there already is testing coverage for go_test() with
embedsrcs using generated files and such, I can't seem to get one of the
most trivial cases to work. The file to be embedded doesn't get placed
in the sandbox during GoCompilePkg.

This change addresses this by explicitly copying the embedsrcs from the
internal source to the external, which seems to make things work
properly.

**What type of PR is this?**

Bug fix

It fixes go_test() with embedsrcs. It's needed, because I want to use that feature.

I couldn't find any existing issues.

**Other notes for review**

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -76,6 +76,7 @@ def _go_test_impl(ctx):
     )
     external_source = go.library_to_source(go, struct(
         srcs = [struct(files = go_srcs)],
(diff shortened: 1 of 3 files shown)


robfig:
It's surprising that this wasn't caught by the existing embedsrcs_test, which also embeds files outside the directory, but I've confirmed that fact locally. Thanks for the contribution, merging.

PR #2934, opened 2021-08-09
Disable //tests/core/race:race_test on Windows

philwo:
**What type of PR is this?**

Other: Bazel CI configuration change.

**What does this PR do? Why is it needed?**

Disables a test that fails on Windows due to an unfixed upstream bug in Go.

**Which issues(s) does this PR fix?**

Fixes #2911.

**Other notes for review**

PR #2941, opened 2021-08-19
Disable //tests/core/cgo:race_test on Windows

steeve:
Same as in #2934.

Signed-off-by: Steeve Morin <steeve@zen.ly>


linzhp (changes requested):
Already done by #2940

Issue #2949, opened 2021-09-06
Remove dependencie with bazelbuild/rules_cc


limdor:
### What version of rules_go are you using?

master (v0.28.0)

### What version of gazelle are you using?

Not relevant, gazelle does not depend on rules_cc

### What version of Bazel are you using?

Not relevant, it would affect al Bazel versions, but in my case Bazel 4 LTS

### What did you do?
Inspect rules_go dependencies.

Long time ago Bazel was saying that cc_binary, cc_library, cc_test and the other cc_* rules should be imported from rules_cc.
However rules_cc was always saying that there is no need to use them yet.
After several discussions it was clarified that migration to bazelbuild/rules_cc was put on hold and there is not need to the users that they start using it.
One of the reasons why a person would not want to use rules_cc right now is because there is no release and there is no notion of what is a good commit:
https://github.com/bazelbuild/rules_cc/issues/91
https://github.com/bazelbuild/rules_cc/issues/68

The fact that rules_go depends on rules_cc forces any rules_go user to use rules_cc. Considering that rules_docker depends on rules_go, any rules_docker user is also forced to depend on rules_cc.

More information about the discussions:
https://github.com/bazelbuild/rules_cc/issues/86
https://github.com/bazelbuild/rules_cc/issues/92
https://github.com/bazelbuild/buildtools/issues/923
https://github.com/bazelbuild/buildtools/pull/952

### What did you expect to see?

I would request rules_go to stop depending on rules_cc and go back to use the native Bazel cc_* rules until rules_cc creates a release. "ith this way the users of rules_go or rules_docker are also not force to use rules_cc.
If the team of rules_go agrees I can offer myself on preparing a PR for it.

achew22:
This seems like a really good first contribution to the project. I'm really excited that you're looking to get involved. How can I help empower you to make this change? I'm very happy to review any prs you might have on this front

limdor:
For the moment knowing that you are opened to this change is enough. I'll
prepare a PR and if I have issues I'll let you know.

Andrew Z Allen ***@***.***> schrieb am Do., 9. Sep. 2021,
17:17:

limdor:
@achew22 here the PR https://github.com/bazelbuild/rules_go/pull/2950
I also wanted to ask, when is it planned the next release of rules_go? Just asking because to really get rid of the dependency once it is merged I will have to wait for the release of rules_go and then bump the rules_go version of the different rules that depend on rules_go

PR #2950, opened 2021-09-09
Remove dependencie with bazelbuild/rules_cc

limdor:
Long time ago Bazel was saying that cc_binary, cc_library, cc_test and the other cc_* rules should be imported from rules_cc.
However rules_cc was always saying that there is no need to use them yet.
After several discussions it was clarified that migration to bazelbuild/rules_cc was put on hold and there is not need to the users that they start using it.
One of the reasons why a person would not want to use rules_cc right now is because there is no release and there is no notion of what is a good commit:
bazelbuild/rules_cc#91
bazelbuild/rules_cc#68

The fact that rules_go depends on rules_cc forces any rules_go user to use rules_cc. Considering that rules_docker depends on rules_go, any rules_docker user is also forced to depend on rules_cc.

More information about the discussions:
bazelbuild/rules_cc#86
bazelbuild/rules_cc#92
bazelbuild/buildtools#923
bazelbuild/buildtools#952

Fixes https://github.com/bazelbuild/rules_go/issues/2949

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -10,12 +10,10 @@ tasks:
   ubuntu2004:
     # enable some unflipped incompatible flags on this platform to ensure we don't regress.
     build_flags:
-    - "--incompatible_load_cc_rules_from_bzl"
     - "--incompatible_load_proto_rules_from_bzl"
     - "--incompatible_restrict_string_escapes"
     - "--incompatible_enable_cc_toolchain_resolution"
     test_flags:
-    - "--incompatible_load_cc_rules_from_bzl"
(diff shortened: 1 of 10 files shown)


achew22:
Based on @oquenchil 's comments in https://github.com/bazelbuild/rules_cc/issues/91#issuecomment-756673529:

> [...] we deem rules_cc a mistake and we shouldn't have created it before finishing starlarkifying the rules.

it seems that `rules_cc` is not worth the hassle and isn't going to be the thing going forward. I'm going to merge this, removing the dependency on `rules_cc` and when/if that becomes the correct tool to use we can either revert this, or change the repo to adapt.

PR #2952, opened 2021-09-15
Suppress goroot path in tool builds

### LINKED ISSUE #2951: Tool binaries (like builder) are not reproducible and break the cache
`v0.28.0`

`v0.22.3`

`4.0.0`

Yes

Linux, x86

N/A

### What did you do?
Between two workspaces, the builder tool is not reproducible as it includes the full goroot in the binary. This is evident when comparing the execution logs (https://docs.bazel.build/versions/main/remote-execution-caching-debug.html). Because the builder binary can be different in different workspaces, we may lose the ability to make use of remote caching.  This case of non-reproducibility was already fixed in go_binary rules.

Reproduce very quickly like so:

```
git clone https://github.com/bazelbuild/rules_go.git rules_go1
cd rules_go1
bazel build @go_sdk//:builder
sha256sum bazel-bin/external/go_sdk/builder
# <first hash>  bazel-bin/external/go_sdk/builder

git clone https://github.com/bazelbuild/rules_go.git rules_go2
cd rules_go2
bazel build @go_sdk//:builder
sha256sum bazel-bin/external/go_sdk/builder
# <second hash>  bazel-bin/external/go_sdk/builder
```

These binaries should have the same hash and should be identical 

### What did you see instead?
They are not

adobke:
The goroot is encoded in the output binary, but the goroot is different
in different workspaces. As a result tool binaries (such as builder) are
not reproducible. Bazel will have cache misses for build involving
builder even if the resulting output from the build is reproducible.

Similar to go_binary builds, suppress the goroot path in go_tool_binary
output so that the result in reproducible.

**What type of PR is this?**
> Bug fix

Like go_binary output, ensure go_tool_binary output is reproducible

Fixes #2951

### DIFF
--- a/tests/integration/reproducibility/reproducibility_test.go
+++ b/tests/integration/reproducibility/reproducibility_test.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -214,6 +215,23 @@ func Test(t *testing.T) {
 	// Compare dir0 and dir2. They should be different.
 	if err := compareHashes(dirHashes[0], dirHashes[2]); err == nil {
 		t.Fatalf("dir0 and dir2 are the same)", len(dirHashes[0]))
+	}
+
+	// Check that the go_sdk path doesn't appear in the builder binary. This path is different
+	// nominally different per workspace (but in these tests, the go_sdk paths are all set to the same
+	// path in WORKSPACE) -- so if this path is in the builder binary, then builds between workspaces
+	// would be partially non cacheable.
+	builder_file, err := os.Open(filepath.Join(dirs[0], "bazel-out", "host", "bin", "external", "go_sdk", "builder"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer builder_file.Close()
+	builder_data, err := ioutil.ReadAll(builder_file)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Index(builder_data, []byte("go_sdk")) != -1 {
+		t.Fatalf("Found go_sdk path in builder binary, builder tool won't be reproducible")
 	}
 }
 
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -235,6 +235,7 @@ def _go_tool_binary_impl(ctx):
         arguments = [largs],
         inputs = sdk.libs + sdk.headers + sdk.tools + [cout],
         outputs = [out],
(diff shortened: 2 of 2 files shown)


sluongng:
Simple fix, simple test. This change LGTM.

Very interesting find! I am very curious on how you came about and got down to troubleshooting this issue.
I tried to reproduce the issue and diffing the hex dump of the 2 builder binaries but couldnt make much sense of it.

There is a missing link somewhere between the `go_sdk` being encoded inside the binary and the `GOROOT_FINAL` being set.

I know the `GOROOT` string value was picked to be consistent with the go_binary rule, but could you confirm that the value does not matter here as long as it's a fixed non-empty string?

Thanks!

tests/integration/reproducibility/reproducibility_test.go
```diff
@@ -215,6 +216,23 @@ func Test(t *testing.T) {
...
 		t.Fatalf("dir0 and dir2 are the same)", len(dirHashes[0]))
 	}
+
+	// Check that the go_sdk path doesn't appear in the builder binary. This path is different
+	// nominally different per workspace (but in these tests, the go_sdk paths are all set to the same
+	// path in WORKSPACE) -- so if this path is in the builder binary, then builds between workspaces
+	// would be partially non cacheable.
+	builder_file, err := os.Open(filepath.Join(dirs[0], "bazel-out", "host", "bin", "external", "go_sdk", "builder"))
+	if err != nil {
+	  t.Fatal(err);
```
steeve, line 226: gofmt

steeve (changes requested):
Thank you! Minor nitpick regarding gofmt

tests/integration/reproducibility/reproducibility_test.go
adobke: Ah yes of course!

adobke:
> Simple fix, simple test. This change LGTM.
> 
> Very interesting find! I am very curious on how you came about and got down to troubleshooting this issue.
> I tried to reproduce the issue and diffing the hex dump of the 2 builder binaries but couldnt make much sense of it.

Indeed one minor change in the binary seems to end up making the hex dumps totally diverge. The tool that was useful here was to compare the output of `strings` on the each binary which revealed the embedded and different paths. 

> There is a missing link somewhere between the `go_sdk` being encoded inside the binary and the `GOROOT_FINAL` being set.
> 
> I know the `GOROOT` string value was picked to be consistent with the go_binary rule, but could you confirm that the value does not matter here as long as it's a fixed non-empty string?

Confirmed, changing to a random string seemed to make not difference (e.g. all tests pass)

steeve (approved):

steeve:
Thank you! Merging

PR #2956, opened 2021-09-20
tests/core/cgo: do not assume test is being run from main repo

### LINKED ISSUE #2955: @io_bazel_rules_go//tests/core/cgo:opts does not build from another repo

0.28.0

NA

4.2.1

Yes

macOS 11.6
Intel

Default Apple toolchain

### What did you do?

I maintain a C++ toolchain configuration for bazel -- https://github.com/grailbio/bazel-toolchain
Within the tests for this toolchain, I run the cgo tests for rules_go to ensure that my toolchain works with rules_go.

For the build to pass.

### What did you see instead?

In #2865, `//tests/core/cgo:opts` started using manually created include paths for its header files, but assumed that the header files are coming from the main repo. When I build this target as part of my test suite, the build fails because the header files like generated_cppopts.h, etc. are not found.

siddharthab:
Currently, the generated headers import paths assume that `opts_test` is
being run from the main repo, but that may not be the case. When
specifying include paths for header files in bazel, we generally have to
specify both prefixes.

**What type of PR is this?**

Bug fix

The PR is a minor enhancement to the tests such that they can be used by other projects.

Fixes #2955

**Other notes for review**


tests/core/cgo/BUILD.bazel
```diff
@@ -52,16 +52,19 @@ go_library(
     copts = [
         "-DRULES_GO_C",
         "-I$(GENDIR)/tests/core/cgo/generated_copts",
+        "-I$(GENDIR)/external/io_bazel_rules_go/tests/core/cgo/generated_copts",
         "-DDOLLAR_SIGN_C=$$",  # the dollar sign should be escaped
     ],
     cppopts = [
         "-DRULES_GO_CPP",
         "-I$(GENDIR)/tests/core/cgo/generated_cppopts",
+        "-I$(GENDIR)/external/io_bazel_rules_go/tests/core/cgo/generated_cppopts",
```
steeve, line 61: can you use https://docs.bazel.build/versions/main/skylark/lib/native.html#repository_name and https://docs.bazel.build/versions/main/skylark/lib/native.html#package_name perhaps ?

steeve (changes requested):

tests/core/cgo/BUILD.bazel
siddharthab: Thanks. I did not know about these. PTAL.

robfig:
This LGTM. Thank you!

PR #2959, opened 2021-09-23
Update go_googleapis to latest

chaodaiG:
latest cloudbuild golang client isn't compatible with proto from 2021-06-30 any more, causing compilation errors. Updating here so that we can use latest cloudbuild golang client.

Related: https://github.com/kubernetes/test-infra/pull/23693

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -258,13 +258,13 @@ def go_rules_dependencies(is_rules_go = False):
     _maybe(
         http_archive,
         name = "go_googleapis",
-        # master, as of 2021-06-30
+        # master, as of 2021-09-23
         urls = [
-            "https://mirror.bazel.build/github.com/googleapis/googleapis/archive/1c5c56c18ab4e2353f87a9c2e14864f218c5502c.zip",
-            "https://github.com/googleapis/googleapis/archive/1c5c56c18ab4e2353f87a9c2e14864f218c5502c.zip",
+            "https://mirror.bazel.build/github.com/googleapis/googleapis/archive/5eeefdc0e9e6d4789ee99c662c6aa1aa91dfd012.zip",
+            "https://github.com/googleapis/googleapis/archive/5eeefdc0e9e6d4789ee99c662c6aa1aa91dfd012.zip",
         ],
-        sha256 = "fa52a03d9c49c28e475a33f9adfb0551344ca0dbd15168d6862ac32cc8354c1c",
-        strip_prefix = "googleapis-1c5c56c18ab4e2353f87a9c2e14864f218c5502c",
+        sha256 = "762678c68c0dd7cf7ea35120f25be5ad164206cd5d7a6e6f7400bf6fcd4fe8c5",
+        strip_prefix = "googleapis-5eeefdc0e9e6d4789ee99c662c6aa1aa91dfd012",
         patches = [
             # releaser:patch-cmd find . -name BUILD.bazel -delete
             "@io_bazel_rules_go//third_party:go_googleapis-deletebuild.patch",


chaodaiG:
```
Download from https://mirror.bazel.build/github.com/googleapis/googleapis/archive/5eeefdc0e9e6d4789ee99c662c6aa1aa91dfd012.zip failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
```

robfig:
I think that 404 is just a warning, not the error. The issue is:
```
Error in fail: Error applying patch @io_bazel_rules_go//third_party:go_googleapis-deletebuild.patch
```

I believe you have to look to see what that patch file is accomplishing and update it to work with the updated version of the repo.

chaodaiG:
Thank you! Updated following https://github.com/bazelbuild/rules_go/wiki/Updating-dependencies

chaodaiG:
uh. there are more about go_googleapis: https://github.com/bazelbuild/rules_go/wiki/Updating-dependencies#updating-go_googleapis

chaodaiG:
Awesome! test all passed

chaodaiG:
Thank you @robfig

steeve (approved):
thank you for the contribution!

robfig (commented):
Thanks for this update! Just one last request please:

- I see that the timestamps in the patch files were updated from `2000-01-01 00:00:00.000000000 -0000` to `2021-09-23 15:04:46.000000000 -0700`. Do you mind doing a find/replace to set it back? I'll add that step to the wiki.

@jayconrod, please upload https://mirror.bazel.build/github.com/googleapis/googleapis/archive/aa73c560f096416fb96c9a3931ebc3fa2cab72ac.zip .

Thanks!

robfig:
It felt like there should be a better way, but this is the expression I came up with to set the timestamp:
```
diff -urN a b | \
  sed -E 's/^(\+\+\+ .*)20..-..-.. ..:..:............ -....$/\12000-01-01 00:00:00.000000000 -0000/g' \
>rules_go/third_party/com_github_golang_protobuf-gazelle.patch
```

Open to other ideas!

chaodaiG:
> * I see that the timestamps in the patch files were updated from `2000-01-01 00:00:00.000000000 -0000` to `2021-09-23 15:04:46.000000000 -0700`. Do you mind doing a find/replace to set it back? I'll add that step to the wiki.

done

chaodaiG:
> It felt like there should be a better way, but this is the expression I came up with to set the timestamp:
> 
> ```
> diff -urN a b | \
>   sed -E 's/^(\+\+\+ .*)20..-..-.. ..:..:............ -....$/\12000-01-01 00:00:00.000000000 -0000/g' \
> >rules_go/third_party/com_github_golang_protobuf-gazelle.patch
> ```
> 
> Open to other ideas!

not quite sure whether `\d[8]` would work better instead of `........` here. Other than that this regex LGTM

robfig:
Thanks - looks like you can revert your change to third_party/go_googleapis-directives.patch entirely

chaodaiG:
good catch, done

jayconrod:
> @jayconrod, please upload https://mirror.bazel.build/github.com/googleapis/googleapis/archive/aa73c560f096416fb96c9a3931ebc3fa2cab72ac.zip .

Done.

FYI, `bazel run //go/tools/releaser -- upgrade-dep go_googleapis` should be able to generate changes like this. I'll update the wiki next week. Sorry I haven't done that already.

robfig:
Amazing! `releaser` will make things easier. Thanks

Merging. Thanks for the contribution.

Issue #2284, opened 2019-11-18
Building with go-fuzz-build / build for libfuzz


prestonvanloon:
### What version of rules_go are you using?

```python
http_archive(
    name = "io_bazel_rules_go",
    sha256 = "886db2f8d620fcb5791c8e2a402a575bc70728e17ec116841d78f3837a09f69e",
    strip_prefix = "rules_go-9bb1562710f7077cd109b66cd4b45900e6d7ae73",
    urls = ["https://github.com/bazelbuild/rules_go/archive/9bb1562710f7077cd109b66cd4b45900e6d7ae73.tar.gz"],
)
```

### What version of gazelle are you using?

```python
http_archive(
    name = "bazel_gazelle",
    sha256 = "7fc87f4170011201b1690326e8c16c5d802836e3a0d617d8f75c3af2b23180c4",
    urls = [
        "https://storage.googleapis.com/bazel-mirror/github.com/bazelbuild/bazel-gazelle/releases/download/0.18.2/bazel-gazelle-0.18.2.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/0.18.2/bazel-gazelle-0.18.2.tar.gz",
    ],
)
```

### What version of Bazel are you using?

1.0.0

### Does this issue reproduce with the latest releases of all the above?

N/A

### What operating system and processor architecture are you using?

Linux AMD64

### What did you do?

I'm trying to find the ideal way to integrate [go-fuzz](https://github.com/dvyukov/go-fuzz) with rules_go since that is the primary build tooling for our project. 

[...]

PR #2970, opened 2021-09-29
go_test: add compatibility with fuzzing in Go 1.18+

jayconrod:
The GoTestGenTest action now produces testmain.go files compatible
with Go 1.18 and lower versions. In 1.18, the function
testing.MainStart accepts a new parameter, a list of fuzz targets. We
now find those targets and pass them in for Go 1.18 and higher.

Go 1.18 is still months away, and it's possible the API may change
before then. It's likely that FuzzTarget will be renamed. That should be a
minor change though. MainStart is carved out of the Go 1 compatibility
promise and is only meant to be called by code generated by 'go test',
which we're emulating here.

This change does not add full support for fuzzing. It is possible to
start the fuzzing engine by passing --test_arg=-test.fuzz=regexp
--test_arg=-test.fuzzcachedir=dir. However, if a crash is found,
there's nowhere for the fuzzing engine to write it.

For #2284

### DIFF
--- a/tests/core/go_test/fuzz_test.go
+++ b/tests/core/go_test/fuzz_test.go
@@ -1,0 +1,29 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build go1.18
+// +build go1.18
+
+package fuzz_test
+
+import "testing"
+
+func Fuzz(f *testing.F) {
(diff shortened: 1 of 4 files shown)


tests/core/go_test/fuzz_test.go
```diff
@@ -0,0 +1,29 @@
...
+// +build go1.18
+
+package fuzz_test
+
+import "testing"
+
+func Fuzz(f *testing.F) {
+	f.Add("seed")
+	f.Fuzz(func(t *testing.T, s string) {
+		if s != "seed" {
```
achew22, line 25: Is this seed pulled from testdata? Should we add a test for that?

achew22 (approved):

tests/core/go_test/fuzz_test.go
jayconrod: That part won't work yet. Seed inputs may be added using `F.Add` or stored in `testdata/fuzz/$targetname`. The testdata path is relative to the test process's current working directory, but that directory might not actually be present. The `data` attribute must cover those files at least, but on Windows, we'll just get a manifest of runfiles, not an actual directory.

One option would be for the generated testmain to wrap each fuzz target with a function that loads inputs from `testdata` with appropriate runfile dereferencing then calls `F.Add`.

Still need to figure where the test can write failing inputs. That's the main thing. Bazel tests generally shouldn't create new files in the source directory.

PR #2976, opened 2021-10-05
Upgrade Bazel to 4.2.0

robfig:
Upgrade the minimum version of Bazel to 4.2.0, which is required by the removal of `go_tool_library` in #2922.


achew22:
Were we failing before this change?

robfig:
No, but perhaps that points to a missing test of some sort? I'm just basing this on the bug fix that was supposedly needed.

achew22:
To be clear, I'm not blocking at all here. I think the CL is good to go. LGTM + Approval, just wanted to ask if this was coming from an issue filed by the bazel-ci team

robfig:
Nope! The origin is that I was doing this as part of a larger change updating dependencies in advance of the release, but test failures made me want to reduce the size of that commit to ensure I wasn't getting new behavior with a new bazel version. Thanks for the LGTM.

PR #2984, opened 2021-10-13
cover: use CamelCase for coverage variable

### LINKED ISSUE #2982: Staticcheck failed during coverage run

v0.29.0

v0.23.0

v4.2

Yes

Linux AMD64

N/A

### What did you do?

1. Setting up a repo with https://github.com/sluongng/staticcheck-codegen to enable `nogo` to use `staticcheck` analyzers

2. Run `bazel test //...`

3. Run `bazel coverage //...`

Both test and coverage to consistently pass all `nogo` analyzers

### What did you see instead?

While test succeeded, coverage failed with

```
: should not use underscores in Go names; var Cover_my_package_libs_discreterand_0_alias should be CoverMyPackageLibsDiscreterand0Alias (ST1003)
```

This is because the variable rules_go is injecting into the test binary during coverage mode is not in compliance with standard Go practice. https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/bazelbuild/rules_go%24+%28lang:Go+or+lang:Starlark%29+Cover_&patternType=regexp&case=yes

We should be able to fix this easily by changing 'Snake_case' to standard 'CamelCase' for these variables.

sluongng:
During the invocation of 'bazel coverage', rules_go will rewrite the
source files using 'go tool cover -var <cover var name> ...' so that the
variable can be extracted for coverage statistic later. For more
information on this, review the relevant Official Go Team's blog post (1).

rules_go currently injecting a variable using `Cover_snake_case` which
is natural to starlark and python, however this will strip up some
golang static analyzer for not following the `CamelCase` variable naming
convention. In particular, if the generated source for coverage is run
against `staticcheck`'s ST1003 installed as part of `nogo`, it will
fail.

Apply snake_case to CamelCase conversion on all coverage variables.

(1): https://go.dev/blog/cover

**What type of PR is this?**

>
Bug fix
> Feature
> Documentation
> Other

Fixes #2982

**Other notes for review**

I think the code in `go/private/actions/cover.bzl` code path is almost(?) irrelevant right now as updating that starlark files does not reflect much on how coverage is being collected. But I still want to provide a fix there just for completeness sake, just in case somebody else started to rely on the starlark code path one day.

### DIFF
--- a/go/private/actions/cover.bzl
+++ b/go/private/actions/cover.bzl
@@ -26,6 +26,15 @@ def _sanitize(s):
     """Replaces /, -, and . with _."""
     return s.replace("/", "_").replace("-", "_").replace(".", "_")
 
+# convert string from snake_case to CamelCase
+def _snake_to_camel(s):
+    words = s.split("_")
+    out = ""
+    for w in words:
+        out += w[:1].upper()
+        out += w[1:].lower()
+    return out
+
 def emit_cover(go, source):
     """See go/toolchains.rst#cover for full documentation."""
 
@@ -44,7 +53,8 @@ def emit_cover(go, source):
         _, pkgpath = effective_importpath_pkgpath(source.library)
         srcname = pkgpath + "/" + orig.basename if pkgpath else orig.path
 
-        cover_var = "Cover_%s_%s" % (_sanitize(pkgpath), _sanitize(src.basename[:-3]))
+        snake_cover_var = "Cover_%s_%s" % (_sanitize(pkgpath), _sanitize(src.basename[:-3]))
+        cover_var = _snake_to_camel(snake_cover_var)
         out = go.declare_file(go, path = "Cover_%s" % _sanitize(src.basename[:-3]), ext = ".cover.go")
         covered_src_map.pop(src, None)
         covered_src_map[out] = orig
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -250,7 +250,8 @@ func compileArchive(
 			if ext := filepath.Ext(stem); ext != "" {
 				stem = stem[:len(stem)-len(ext)]
 			}
-			coverVar := fmt.Sprintf("Cover_%s_%d_%s", sanitizePathForIdentifier(importPath), i, sanitizePathForIdentifier(stem))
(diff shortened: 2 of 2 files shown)


sluongng:
@robfig 

> it feels unnecessarily complicated to generate a snake case identifier and then run that through a case conversion function

I share this sentiment and was looking for a cheap solution, which is why in #2982 I used a patch that just do a `s/_/Z/` over the variable.

Another alternative might be to avoid having to run this entirely
```
"Cover_%s_%s" % (_sanitize(pkgpath), _sanitize(src.basename[:-3]))
...
fmt.Sprintf("Cover_%s_%d_%s", sanitizePathForIdentifier(importPath), i, sanitizePathForIdentifier(stem))
```

and instead replace `_sanitize` with some SnakeCase conversion function.
However that involve more code changes so I was a bit uncertain to take that route (just in case `pkgpath` or `importPath` or `stem` might contains some natural snake case.

Let me know which direction is preferable. I am more than happy to do a fixup and get this merged 😄

robfig:
Ah, I didn't realize that the `sanitize` functions also introduced underscores. 

I think that there's not much risk in this code churn, so I think it would be fine to fix this up rather than do the minimal patch. It looks like the sanitize functions could be pretty easily modified to use whichever separator character you choose. I think I've seen "Z" used for mangling in the past, and that seems unlikely to violate any checks, so that seems like the best bet?

robfig:
Do you see any reason why the package name/path are passed separately through the sanitize function, instead of the whole result of fmt.Sprintf being passed through it?

robfig:
Even as a "short term" fix, let's replace the _ with Z instead of adding extra complication with case conversions when it's not necessary. Thanks

sluongng:
@robfig I spent a day looking into making nogo ignoring generated source and I don't think it's too hard to achieve. I see that we need to ignore 2 cases of generated source by rules_go:

1. Coverage variable
2. `_empty.go` when no source is available.

Changes can be done in `go/tools/builders/compilepkg.go` and `go/tools/builders/compile.go` to make this happen. 

But I have yet to have time to test my changes locally and these areas don't seem to be well-tested nor are they well documented, so instead I have created https://github.com/bazelbuild/rules_go/pull/2995 as a relatively cheaper fix right now. A future PR that would help nogo ignore rules_go generated code should have an easy time reverting it.

Issue #2437, opened 2020-04-17
Investigate migrating documentation to Stardoc


jayconrod:
[Stardoc](https://github.com/bazelbuild/stardoc) is a documentation generator for Bazel rules, providers, and other definitions.

Currently, rules_go documentation is maintained separately from the rule implementation themselves, so it's easy for it to drift.

We should investigate whether Stardoc is a viable replacement for the current documentation. If it is, we should migrate the current documentation to generated Stardoc.

Analogous Gazelle issue bazelbuild/bazel-gazelle#763

alexeagle:
The main question here is whether to switch to Markdown for this documentation, or write and maintain a custom velocity template so that stardoc can produce the reStructuredText matching the present docs.

alexeagle:
I'm pretty sure we can show all the same content (markdown doesn't have rowspan in tables so the structure will be a bit different, but I think other rulesets have readable docs using markdown). If it were greenfield it would be obvious to use what Stardoc already knows how to write.

However if we change markup languages, we'll have a much harder time reviewing the giant delta to be sure all the docs are the same. With restructuredtext velocity templates we could more-or-less match the current content and just make it the generated output. The tricky bit remaining there is to make the columns align on the generated tables, which in rst actually do have to align as ascii art. I couldn't find an rst formatter that could automatically do that.

alexeagle:
I don't think it should be closed yet, as there are still other .rst files containing manually-curated API docs

PR #2988, opened 2021-10-19
Fix broken bzl_library build graph

cvarier:
**What type of PR is this?**
- Bug fix

Currently, `//go/private:context` is missing a dep on `//go/private/rules:transition`, which is necessary in order to load `@io_bazel_stardoc//stardoc:stardoc.bzl`. This will be used in a subsequent PR for adding targets to generate Stardoc output.

### DIFF
--- a/go/private/BUILD.bazel
+++ b/go/private/BUILD.bazel
@@ -56,6 +56,7 @@ bzl_library(
         ":mode",
         ":providers",
         "//go/platform:apple",
+        "//go/private/rules:transition",
         "@bazel_skylib//lib:paths",
         "@bazel_skylib//rules:common_settings",
         "@bazel_tools//tools/build_defs/cc:action_names.bzl",


Cvarier:
@googlebot I signed it!

robfig:
Thanks for the fix! Can you also add a command that would have identified this problem to our CI script, to prevent regression?

alexeagle:
@robfig something would have to evaluate the starlark load statements to detect a missing edge in the graph, and @Cvarier and I are working on https://github.com/bazelbuild/rules_go/issues/2437 which would add such a regression test.

The only alternative I can think of is to introduce a usage of gazelle in this repo to generate the bzl_library targets, and then on CI assert that gazelle finds no changes to make. But that seems hard per https://github.com/bazelbuild/rules_go/blob/master/BUILD.bazel#L40-L42

robfig:
OK, thanks!

PR #2989, opened 2021-10-23
Skipping linker flag for versioned dylib on Darwin

### LINKED ISSUE #2944: Versioned Darwin dylib not working after 0.26

It was working in v0.25.x, but stopped working after 0.26

d7082fb305422c27d9a33530e6e0f6be463ba178

3.7.0

Yes

Darwin amd64

None

### What did you do?

1. apply this patch to rules_go master:
```diff
diff --git a/tests/core/cgo/BUILD.bazel b/tests/core/cgo/BUILD.bazel
index d205044c..c1e557a5 100644
--- a/tests/core/cgo/BUILD.bazel
+++ b/tests/core/cgo/BUILD.bazel
@@ -162,7 +162,7 @@ go_library(
         # This test exists just for versioned `.so`s on Linux,
         # but we can reuse the above test's dylib so it passes on darwin,
         # where filename suffixes are not used for library version.
-        "@io_bazel_rules_go//go/platform:darwin": [":darwin_imported_dylib"],
+        "@io_bazel_rules_go//go/platform:darwin": [":darwin_imported_versioned_dylib"],
         "//conditions:default": [":linux_imported_versioned_dylib"],
         # TODO(jayconrod): Support windows, skip others.
     }),
@@ -177,6 +177,12 @@ cc_import(
     tags = ["manual"],
 )
 
+cc_import(
+    name = "darwin_imported_versioned_dylib",
+    shared_library = "libimported.dylib.2",
+    tags = ["manual"],
+)
+
 go_test(
     name = "generated_versioned_dylib_test",
     srcs = ["dylib_test.go"],
diff --git a/tests/core/cgo/generate_imported_dylib.sh b/tests/core/cgo/generate_imported_dylib.sh
index 5c272ef4..82f43d9e 100755
--- a/tests/core/cgo/generate_imported_dylib.sh
+++ b/tests/core/cgo/generate_imported_dylib.sh
@@ -11,6 +11,7 @@ case "$(uname -s)" in
     ;;
   Darwin*)
     cc -shared -Wl,-install_name,@rpath/libimported.dylib -o libimported.dylib imported.c
+    cc -shared -Wl,-install_name,@rpath/libimported.dylib.2 -o libimported.dylib.2 imported.c
     ;;
   *)
     echo "Unsupported OS: $(uname -s)" >&2
```
2. Run `tests/core/cgo/generate_imported_dylib.sh`
3. Run `bazel test --test_output=errors tests/core/cgo:versioned_dylib_test`

Test passes

### What did you see instead?
```
dyld: Library not loaded: @rpath/libimported.dylib.2
  Referenced from: /private/var/tmp/_bazel_zplin/b32df389e141e2de05812b7963fff9af/sandbox/darwin-sandbox/24/execroot/io_bazel_rules_go/bazel-out/darwin-fastbuild/bin/tests/core/cgo/versioned_dylib_test_/versioned_dylib_test.runfiles/io_bazel_rules_go/tests/core/cgo/versioned_dylib_test_/versioned_dylib_test
  Reason: image not found
```

linzhp:
**What type of PR is this?**
Bug fix

The linker cannot find the versioned dylib on macOS with `-l :libclntsh.dylib.12.1` like we did for Linux. According to _Mac OS X For Unix Geeks_, 4th Edition, [Chapter 11](https://learning.oreilly.com/library/view/mac-os-x/9780596520625/ch11.html#library_versions), a versioned libMagick library should be named in the form of `libMagick.10.0.7.dylib`, and linked with `-lMagick.10.0.7`. The logic for passing such linker flag is the same as an unversioned shared library.

However, when distributing [Instant Client for macOS](https://www.oracle.com/database/technologies/instant-client/macos-intel-x86-downloads.html), Oracle decided to put the version number at the end of the file name using the same convention as shared libraries on Linux. Fortunately, recent versions of Instant Client also come with an unversioned symbolic link pointing to the versioned dylib. We can use the symlink as an unversioned shared library.

In both cases, we don't need any specific logic to pass linker flags for versioned dylib for Darwin.

Fixes #2944

**Other notes for review**
Alternatively, we can try to figure out how to pass linker options so it can find the versioned dylib files on macOS without the symlink


go/private/rules/cgo.bzl
```diff
@@ -119,25 +119,30 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
...
+                    extension = get_versioned_shared_lib_extension(lib.basename)
+                    if extension.startswith("so"):
+                        # With a versioned .so file, we must use the full filename,
+                        # otherwise the library will not be found by the linker.
+                        libname = ":%s" % lib.basename
+                        clinkopts.extend(["-L", lib.dirname, "-l", libname])
+                        inputs_direct.append(lib)
+                        continue
+                    elif extension.startswith("dylib"):
+                        # With a versioned .dylib file, we can only symlink it as a simple .dylib
```
kmicklas, line 142: Just to be clear, the user has to set up this symlink, right? The comment makes it sound like rules_go will do it.

kmicklas (commented):
Can you test if it works when the file name is in the standard darwin versioned dylib format (`libname.$VERSION.dylib`)? (Maybe there should be a test for this too?)

Also, I think the PR description should be more clear that this is to support the non-standard naming scheme which we found in the Oracle client.

go/private/common.bzl
```diff
@@ -183,21 +183,21 @@ def has_simple_shared_lib_extension(path):
         return True
```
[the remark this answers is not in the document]
linzhp: Why is this useless?

go/private/rules/cgo.bzl
linzhp: Updated the comment

go/private/common.bzl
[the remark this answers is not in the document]
kmicklas: It's redundant.

```
if condition:
    return True
return False
```

go/private/common.bzl
```diff
@@ -183,21 +183,22 @@ def has_simple_shared_lib_extension(path):
...
+    # only iterating to 1 because parts[0] has to be the lib name
     for i in range(len(parts) - 1, 0, -1):
         if not parts[i].isdigit():
             if parts[i] == "dylib" or parts[i] == "so":
-                return True
+                return ".".join(parts[i:])
 
-            # somehting like foo.bar.1.2
-            return False
+            # somehting like foo.bar.1.2 or dylib.1.2
```
kmicklas, line 197: preexisting typo here (somehting)

kmicklas (approved):

PR #2993, opened 2021-10-30
builders/cover: register coverage without changing line numbers

### LINKED ISSUE #2990: Coverage instrumentation breaks line debug info

https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.28.0/rules_go-v0.28.0.zip

https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.23.0/bazel-gazelle-v0.23.0.tar.gz

Build label: 4.1.0

I don't have an easy way to check.

Linux/amd64

-

### What did you do?

Build gVisor with --collect_code_coverage:
https://github.com/google/syzkaller/blob/282f03fbbd76ae15c1ed5e934873fbbc47735176/pkg/build/gvisor.go

Panics and data reports still contain correct line numbers.

### What did you see instead?

Panics and data reports contain contain bogus line numbers.

Here are examples of the same crash on gVisor revision 976ac9710fad without and with code coverage:
https://syzkaller.appspot.com/text?tag=CrashReport&x=1583d069300000
```
	pkg/sentry/socket/unix/transport/queue.go:160
	pkg/sentry/socket/unix/transport/unix.go:690
	pkg/sentry/socket/unix/transport/unix.go:855
	pkg/sentry/socket/unix/transport/connectioned.go:474
```
https://syzkaller.appspot.com/text?tag=CrashReport&x=16d2244d300000
```
	pkg/sentry/socket/unix/transport/queue.go:185
	pkg/sentry/socket/unix/transport/unix.go:762
	pkg/sentry/socket/unix/transport/unix.go:959
	pkg/sentry/socket/unix/transport/connectioned.go:564
```

I've checked on the latest Go toolchain head 087fb86be047e6b520a76f135c607fefaf73c51c by injecting a panic into fmt and running `go test  fmt` and `go test -cover fmt`. Line numbers are correct in both cases. So there is something wrong with how bazel uses Go coverage.
087fb86be047e6b520a76f135c607fefaf73c51c

@avagin @ayushr2 @linuxerwang @jayconrod

robfig:
This copies in Go's cmd/internal/edit package for bytebuffer-based
editing.

Fixes #2990


go/tools/builders/cover.go
```diff
@@ -70,39 +70,57 @@ func instrumentForCoverage(goenv *env, srcPath, srcName, coverVar, mode, outPath
 	if err := goenv.runCommand(goargs); err != nil {
 		return err
 	}
+	coverSrc, err := os.ReadFile(outPath)
```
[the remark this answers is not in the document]
robfig: I had done it this way because it makes the test routine much easier, but I guess that the test is less important. Updated.

go/tools/builders/cover_test.go
```diff
@@ -0,0 +1,130 @@
...
+		varName.Count[:],
+		varName.Pos[:],
+		varName.NumStmt[:])
+}
+`,
+	},
+}
+
+func TestRegisterCoverage(t *testing.T) {
+	const filename = "test_input.go"
```
linzhp, line 109:
```suggestion
	var filename = filepath.Join(t.TempDir(), "test_input.go")
```
The current directory of the test may not be writable, or may happen to already have a file called `test_input.go`. It's better to write to a temp dir. By using `t.TempDir()`, you don't need to do `defer os.Remove(filename)` either.

robfig: Ah, I forgot about `testing.(*T).TempDir`, thanks. Updated

PR #2995, opened 2021-10-31
cover: use Z instead of underscore for variable name separator

### LINKED ISSUE #2982: Staticcheck failed during coverage run

sluongng:
During the invocation of `bazel coverage`, rules_go will rewrite the
source fils using `go tool cover -var ...` so that the variable can be
extracted for coverage statistic after test execution. For more
information on this, review the relevant Official Go Team's blog post (1)

The variable used by rules_go is following the `Cover_snake_case` naming
convention, which is natural to starlark and python, but not Golang.
For this reason, some static analysis could get triggered by the usage
of snake case naming variable while running over coverage source code.

The correct solution for this problem should be to make rules_go's
static analysis framework `nogo` to NOT run on the generated source code
and instead only run on the original source code.  However, replacing
underscore separators with character `Z` is a relatively cheap fix that
would let us unblock static analysis run during `bazel coverage` for
now.  See discussion in (2) for more details.

(1): https://go.dev/blog/cover

(2): https://github.com/bazelbuild/rules_go/pull/2984

What type of PR is this?

Uncomment one line below and remove others.

Bug fix

Feature
Documentation
Other

Fixes #2982

Replaces https://github.com/bazelbuild/rules_go/pull/2984

### DIFF
--- a/go/private/actions/cover.bzl
+++ b/go/private/actions/cover.bzl
@@ -44,7 +44,8 @@ def emit_cover(go, source):
         _, pkgpath = effective_importpath_pkgpath(source.library)
         srcname = pkgpath + "/" + orig.basename if pkgpath else orig.path
 
-        cover_var = "Cover_%s_%s" % (_sanitize(pkgpath), _sanitize(src.basename[:-3]))
+        _cover_var = "Cover_%s_%s" % (_sanitize(pkgpath), _sanitize(src.basename[:-3]))
+        cover_var = _cover_var.replace("_", "Z")
         out = go.declare_file(go, path = "Cover_%s" % _sanitize(src.basename[:-3]), ext = ".cover.go")
         covered_src_map.pop(src, None)
         covered_src_map[out] = orig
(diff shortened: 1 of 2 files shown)


sluongng:
created https://github.com/bazelbuild/rules_go/pull/2996 to propose a longer term fix.

I would still suggest to merge this first as 2996 can take some iterations to get right

robfig:
Thanks

PR #2997, opened 2021-11-01
Stardoc generates modes.md

cvarier:
**What type of PR is this?**

> Documentation

This is a mirror of the content in modes.rst, allowing us to manually inspect the delta between them:
LHS: https://github.com/bazelbuild/rules_go/blob/master/go/modes.rst
RHS: https://github.com/bazelbuild/rules_go/compare/master...aspect-dev:stardoc-gen-modes?expand=1#diff-79f08018fdcff6a7dc885b03481c6074542a4edc5f45d9af25c81594131ae1c2

Technique: 
1. Converted RST to Github-flavored MD using pandoc (https://pandoc.org/try)
2. Filled in gaps by copying doc strings out of existing RST file into the starlark files (converting to MD syntax).

First step in addressing https://github.com/bazelbuild/rules_go/issues/2437
Remaining steps will be to rinse-and-repeat for other .rst files containing rule APIs

**Other notes for review**
Created separate module `mode_docs.bzl` to contain entire docstring - the flags documented in `modes.rst` are located in `go/config/BUILD.bazel`, however stardoc doesn't support documentation generation for named targets in BUILD files. The documentation for these flags is now presented in separate paragraphs for ease of linking via anchors.


go/mode_docs.bzl
```diff
@@ -0,0 +1,200 @@
...
+
+``` bzl
+go_test(
+    name = "go_default_test",
+    srcs = ["lib_test.go"],
+    embed = [":go_default_library"],
+    race = "on",
+)
+```
+"""
```
achew22, line 200: Can you help me understand the value of this `.bzl` file? It doesn't seem that Stardoc is actually adding value here. This file is just directly being projected into the `.md` file without any generated components. Is there a way to get any of these fields out of the `.bzl` files instead of manually documenting them?

achew22 (commented):
FYI, your PR is going to close the associated bug but I think there is more work that will need to be done after this. Can we update the commits to not tag these as fixed?

go/mode_docs.bzl
cvarier: Closing for now, since we don't have a way to support generating documentation within BUILD files. We would have to add support for this within stardoc.

cvarier:
Closing for now, since we don't have a way to support generating documentation within BUILD files. We would have to add support for this within stardoc.

PR #2992, opened 2021-10-28
Stardoc generates extras.md

Cvarier:
**What type of PR is this?**

> Documentation

**What does this PR do? Why is it needed?**
This is a mirror of the content in extras.rst, allowing us to manually inspect the delta between them:
LHS: https://github.com/bazelbuild/rules_go/blob/master/go/extras.rst
RHS: https://github.com/bazelbuild/rules_go/compare/master...aspect-dev:stardoc-chai#diff-facb0b3aafba052f3e5fe16fbe02d2c2f6b2c2c30c443fd08f4617d4c5d38532

As soon as we are satisfied that the generated content is correct, we'll delete extras.rst and update any links to it.
(This could happen as part of this PR, or subsequent)

Technique: 
1. Converted RST to Github-flavored MD using pandoc (https://pandoc.org/try)
2. Filled in gaps by copying doc strings out of existing RST file into the starlark files (converting to MD syntax).

First step in addressing https://github.com/bazelbuild/rules_go/issues/2437
Remaining steps will be to rinse-and-repeat for other .rst files containing rule APIs

**Which issues(s) does this PR fix?**

Fixes https://github.com/bazelbuild/rules_go/issues/2437

PR #3000, opened 2021-11-09
Stardoc generates API docs for core rules

alexeagle:
Replaces hand-maintained core.rst file.

- some prose documentation sections from the core.rst are moved into their own markdown documents
- docs/ folder generated by running stardoc on APIs like `go_binary`
- docstrings for attributes ported back into the starlark code
- verification method is the same side-by-side diff as for #2992


steeve:
This is amazing, thank you!

cvarier:
@googlebot I fixed it.

alexeagle:
@googlebot I fixed it.

cvarier:
@googlebot I consent.

alexeagle:
@steeve this is going stale and will be hard for me to keep it going for long. Can I help get it reviewed?

achew22:
If you sync this up I'm happy to merge without @steeve giving it another look

alexeagle:
https://github.com/bazelbuild/rules_go/pull/3004 landed while this was open, I moved the newly introduced `env` docstring into starlark.

docs/doc_helpers.bzl
```diff
@@ -0,0 +1,84 @@
+load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+
+def stardoc_with_diff_test(
```
alexeagle, line 5: incidentally we have this function in aspect-build/bazel-lib now, and that repo is Apache 2, so we could possibly introduce a dependency on it from rules_go if you want to avoid maintaining this

achew22: I'd like to keep deps in `rules_go` to a bare minimum for end users so I think keeping it in repo is not a terrible situation. Have you considered getting it into the `stardoc` repo? Then using it would be straight forward and wouldn't add any more deps.

alexeagle: I haven't tried, based on Googlers silent on my https://github.com/bazelbuild/bazel-skylib/pulls/alexeagle and especially https://github.com/bazelbuild/bazel-skylib/pull/321 I don't want to spend the time if it's never going to merge.

Thanks for getting this in!

PR #3004, opened 2021-11-12
go_test: Add env attribute

### LINKED ISSUE #2921: no such attribute 'env' in 'go_transition_test' rule
### rules_go version

`v0.27.0`

`v0.23.0`

`4.1.0`

Reproduces with `v0.28.0`

macOS, amd64.

Fairly standard toolchain, using a few custom protobuf compilers for proto+gRPC. This is a set of e2e (external) tests that uses the _actual_ binary as a run-time dependency.

### What did you do?
```
go_test(
    name = "go_default_test",
    srcs = ["svm_test.go"],
    args = ["$(location :svm)"], # actual binary under test
    data = [
        ":svm",
        "//thirdparty/patchcli",
    ],
    env = {
        "SNIPPED_TOOL_PATH": "$(location //thirdparty/snipped_tool)",
    },
    race = "off",
    rundir = ".",
    tags = [
        "functional",
    ],
    deps = ["<snip>"],
)
```
```
load("@bazel_skylib//rules:native_binary.bzl", "native_binary")

native_binary(
    name = "snipped_tool",
    src = select({
        "@platforms//os:windows": ":win64",
        "@platforms//os:macos": ":macos",
        "@platforms//os:linux": ":linux",
    }),
    out = "snipped_tool",
)
```

`$ bazel test //services/svm:go_default_test`

Per [the Bazel common attributes](https://docs.bazel.build/versions/main/be/common-definitions.html#common-attributes-tests), I would expect this to work and this environment variable be available in the test runtime.

### What did you see instead?

```
$ bazel test //services/svm:go_default_test
Another command (pid=42199) is running. Waiting for it to complete on the server (server_pid=39418)...
INFO: Invocation ID: de25d398-863c-4243-9b98-697aea4a6cc5
ERROR: /Users/jhoward/git/patcher/services/svm/BUILD.bazel:61:8: //services/svm:go_default_test: no such attribute 'env' in 'go_transition_test' rule
ERROR: error loading package 'services/svm': Package 'services/svm' contains errors
INFO: Elapsed time: 31.567s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded)
FAILED: Build did NOT complete successfully (1 packages loaded)
```

illicitonion:
**What type of PR is this?**

Feature

This allows setting environment variables before static initialisers
run, and which reference expanded locations.

Partially fixes #2921

### DIFF
--- a/go/tools/bazel/runfiles.go
+++ b/go/tools/bazel/runfiles.go
@@ -51,9 +51,20 @@ func Runfile(path string) (string, error) {
 	}
 
 	// Search manifest if we have one.
-	if entry, ok := runfiles.index[path]; ok {
+	if entry, ok := runfiles.index.GetIgnoringWorkspace(path); ok {
 		return entry.Path, nil
 	}
+
+	if strings.HasPrefix(path, "../") || strings.HasPrefix(path, "external/") {
+		pathParts := strings.Split(path, "/")
+		if len(pathParts) >= 3 {
+			workspace := pathParts[1]
+			pathInsideWorkspace := strings.Join(pathParts[2:], "/")
+			if path := runfiles.index.Get(workspace, pathInsideWorkspace); path != "" {
+				return path, nil
+			}
+		}
+    }
 
 	// Search the main workspace.
 	if runfiles.workspace != "" {
@@ -279,7 +290,7 @@ var runfiles = struct {
 	list []RunfileEntry
 
 	// index maps runfile short paths to absolute paths.
-	index map[string]RunfileEntry
+	index index
 
 	// dir is a path to the runfile directory. Typically this is a directory
 	// named <target>.runfiles, with a subdirectory for each workspace.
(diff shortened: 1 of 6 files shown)


illicitonion:
I think this is ready for review :)

/cc @linzhp @robfig

blico:
Thanks for working on this @illicitonion. 

My understanding is that we could add support for this attribute using the `TestEnvironment` provider avoiding the need for a wrapper executable around `go_test`. https://docs.bazel.build/versions/main/skylark/lib/testing.html#TestEnvironment.

With this approach, we should just need to run whatever pre-processing is needed on the `env` attr (e.g. `ctx.expand_location`) and then construct a `TestEnvironment` provider and return it through `go_test`. 

It would also be good to add some documentation to the `go_test` rule about the new attr, and some test cases verifying `env` is available during `TestMain` and the contents of `env` are expanded.

go/tools/bazel/runfiles.go
```diff
@@ -51,10 +51,21 @@ func Runfile(path string) (string, error) {
 	}
 
 	// Search manifest if we have one.
-	if entry, ok := runfiles.index[path]; ok {
```
blico, line 54: Are you sure all of the changes to this file are needed? I patched the change locally and the tests still passed without the changes to this file.

illicitonion: Yeah, it's necessary on Windows - the way this works on non-Windows is that first we'll stat the file, see the symlink and be happy.

But the way it works on Windows is we parse a metadata file with paths to files written down, and we do a match based on exactly the file name inside whatever workspace the file happens to be in. These changes are necessary for that runfiles lookup to work properly.

I could, if you'd prefer, make the test use a file in the main workspace (rather than from the `go_sdk` workspace), and send a separate PR fixing up the separate-workspace runfiles lookup issue, but figured fixing it all in one was reasonable...

blico: That makes sense, thanks for the clarification. Yeah, I think including the change as part of this PR is reasonable.

Issue #2999, opened 2021-11-09
`go_proto_compiler` causes `protoc` rebuilds by enabling `bootstrap_nogo`


fmeum:
### What version of rules_go are you using?

v0.29.0

### What version of gazelle are you using?

v0.24.0

### What version of Bazel are you using?

Bazel 5.0.0rc1 (also affects 4.2.1)

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Linux

### What did you do?

I am using libraries that themselves use `go_proto_compiler` from `rules_go` and observer that it causes `protoc` to be built in multiple configurations that only differ in the value of `@io_bazel_rules_go//go/private:bootstrap_nogo`. The top-level exec config has this Starlark flag at its default value (`false`), but the `go_reset_target` emitted in the `go_proto_compiler` macro sets it to `true` through `go_reset_transition` (see https://github.com/bazelbuild/rules_go/blob/bd7fbccc635af297db7b36f6c81d0e7db7921cca/go/private/rules/transition.bzl#L239).

### What did you expect to see?

`protoc` is only built once, in the top-level `exec` config.

### What did you see instead?

`protoc` (wastefully) built multiple times.

fmeum:
Setting `bootstrap_nogo` to `False` fails with a cyclic dependency, so fixing this might require duplicating the rule for use in `go_proto_compiler`. I am not sure what the best fix is though as I do not fully understand the transition structure around nogo.

Once this issue has been resolved, I could submit a PR that fixes a similar issue with `go_proto_library`, which depends on `protoc` via its `proto_library` deps and does not use the `go_reset_transition`.

robfig:
`bootstrap_nogo = True` causes any configured nogo to be ignored when building. This could be helpful to avoid having generated code get flagged for any static analysis violations, and it would be unnecessary processing. I'm not exactly sure why the cyclic dependency exists in this case, but the general idea is that the build tools need to be built without nogo so that they can be used to build nogo itself. 

Perhaps the error is that the protoc plugin does need `bootstrap_nogo`, but protoc itself does not and so it's an unnecessary part of the cache key. I'm not sure how to resolve it, but the first step I would take is to make a reproducer that allows easy experimentation with different options. I'm afraid that I and other rules_go maintainers are stretched far too thin at the moment, so I can't promise when I'll look into it. Any contributions would be most welcome.

fmeum:
[...]

PR #3005, opened 2021-11-14
Reset Go settings for protoc dependencies

fmeum:
**What type of PR is this?**

Bug fix

Reduces the number of essentially distinct configurations in which `protoc` as a dependency of `go_proto_library` is analyzed. 

Note that this may not (yet) result in a significant reduction of *rebuilds* since Bazel is not particular friendly to reset transitions at the moment: It encodes some information about the path of transitions taken into the build configuration, which e.g. prevents transitions from ever returning to the top-level target configuration. But this will be solved by https://github.com/bazelbuild/bazel/issues/14023, which is very actively being worked on. In the meantime, this change should at least not regress the current behavior further.

Fixes a part of #2999.

**Other notes for review**

@robfig If you think that this PR is useful, I would also look into a follow-up that resets the non-Go attrs, e.g. `data` and `embedsrcs`, of rules such as `go_library` to their values before `go_transition`. Ideally, `rules_go` settings set by transitions would only ever propagate to rules that provide a Go provider. Since this requires memoizing the state of settings before a transition, this will be a bit more complicated, which is why I would like to get your feedback on this simpler case first.

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -221,7 +221,7 @@ go_transition = transition(
     ]],
 )
 
-_reset_transition_dict = {
+_common_reset_transition_dict = {
     "@io_bazel_rules_go//go/config:static": False,
     "@io_bazel_rules_go//go/config:msan": False,
     "@io_bazel_rules_go//go/config:race": False,
@@ -230,17 +230,22 @@ _reset_transition_dict = {
     "@io_bazel_rules_go//go/config:debug": False,
     "@io_bazel_rules_go//go/config:linkmode": LINKMODE_NORMAL,
     "@io_bazel_rules_go//go/config:tags": [],
-    "@io_bazel_rules_go//go/private:bootstrap_nogo": True,
 }
+
+_reset_transition_dict = dict(_common_reset_transition_dict, **{
+    "@io_bazel_rules_go//go/private:bootstrap_nogo": True,
+})
 
 _reset_transition_keys = sorted([filter_transition_label(label) for label in _reset_transition_dict.keys()])
 
-def _go_reset_transition_impl(settings, attr):
-    """Sets Go settings to default values so tools can be built safely.
+def _go_exec_reset_transition_impl(settings, attr):
+    """Sets most Go settings to default values (use for external Go tools).
 
-    go_reset_transition sets all of the //go/config settings to their default
-    values. This is used for tool binaries like nogo. Tool binaries shouldn't
-    depend on the link mode or tags of the target configuration. This transition
+    go_exec_reset_transition sets all of the //go/config settings to their
+    default values and disables nogo. This is used for Go tool binaries like
+    nogo itself. Tool binaries shouldn't depend on the link mode or tags of the
+    target configuration and neither the tools nor the code they potentially
+    generate should be subject to nogo's static analysis. This transition
     doesn't change the platform (goos, goarch), but tool binaries should also
     have `cfg = "exec"` so tool binaries should be built for the execution
     platform.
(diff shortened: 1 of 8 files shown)


robfig:
Hi @fmeum , thanks for the contribution. We would definitely like to avoid cache thrashing. I don't personally have any knowledge about transitions, but @blico volunteered to study up and review. Apologies for the delay, but be assured we haven't forgotten and would like to get this merged.

fmeum:
No worries, let me know if you need any additional explanations.

fmeum:
I rebased the PR onto master after my other changes were merged.

fmeum:
@blico Friendly ping as the required commit has landed in Bazel: https://github.com/bazelbuild/bazel/commit/52d1d4ae9bfee7379093d46258fa3a56e4d69e61

fmeum:
@robfig I just submitted https://github.com/bazelbuild/rules_go/pull/3108, which is based on this PR and finished the leftover work. Along the way, it fixes a rules_go correctness bug reported in https://github.com/bazelbuild/bazel/issues/14626. Do you think one of the maintainers will have time to review either of those PRs in the near future?

achew22 (approved):

go/private/rules/transition.bzl
```diff
@@ -250,15 +255,44 @@ def _go_reset_transition_impl(settings, attr):
...
     inputs = _reset_transition_keys,
     outputs = _reset_transition_keys,
 )
 
-def _go_reset_target_impl(ctx):
+_non_go_exec_reset_transition_dict = dict(_common_reset_transition_dict, **{
+    "@io_bazel_rules_go//go/private:bootstrap_nogo": False,
+})
+
+_non_go_exec_reset_transition_keys = sorted([filter_transition_label(label) for label in _non_go_exec_reset_transition_dict.keys()])
```
linzhp, line 268: `_non_go_exec_reset_transition_keys` can be constructed from `_reset_transition_dict` too, because it as the same set of keys as `_non_go_exec_reset_transition_dict`. If we do that, then we can don't need `_non_go_exec_reset_transition_dict`, and instead setting `@io_bazel_rules_go//go/private:bootstrap_nogo` in   `_go_non_go_exec_reset_tool_transition_impl`

proto/BUILD.bazel
```diff
@@ -112,6 +116,12 @@ go_proto_compiler(
     ] + WELL_KNOWN_TYPE_RULES.values(),
 )
 
+go_non_go_exec_reset_target(
+    name = "protoc_reset_target_",
```
linzhp, line 120:
```suggestion
    name = "protoc",
```
maybe just name it "protoc"?

go/private/rules/transition.bzl
```diff
@@ -290,25 +324,49 @@ def _go_reset_target_impl(ctx):
...
+        ),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
     },
-    doc = """Forwards providers from a target and applies go_reset_transition.
+    doc = """Forwards providers from a target and applies go_exec_reset_transition.
 
-go_reset_target depends on a single target, built using go_reset_transition.
+go_exec_reset_target depends on a single target, built using go_exec_reset_transition.
```
linzhp, line 364: This doc seems wrong

go/private/rules/transition.bzl
```diff
@@ -290,25 +324,49 @@ def _go_reset_target_impl(ctx):
     )
     return providers
 
-go_reset_target = rule(
-    implementation = _go_reset_target_impl,
+go_exec_reset_target = rule(
```
linzhp, line 327: Maybe keep this `go_reset_target` and name another one `non_go_reset_target`?

linzhp (changes requested):
Passed the tests in Uber's Go monorepo. Left a few stylish comments. Looks good overall. Sorry for taking so long to review.

go/private/rules/transition.bzl
fmeum: I took the chance to improve the docs, but am not sure I understood this particular point.

go/private/rules/transition.bzl
fmeum: I ended up removing `_non_go_exec_reset_transition_keys` as well in favor of `_reset_transition_keys`.

PR #3021, opened 2021-12-03
Add output of bindata rule to runfiles

gregmagolan:
**What type of PR is this?**

Bug fix

The DefaultInfo provided by the `bindata` rule doesn't put its output in runfiles.

A binary rule such as an `sh_binary` that adds on the output of `bindata` to its data would expect the output to be found at `$(rootpath :bindata_target)`.

For example,

```
sh_binary(
    name = "bin_target",
    srcs = ["script.sh"],
    data = [":bindata_target"],
    args = ["$(rootpath :bindata_target)"],
)
```

None filed

**Other notes for review**

### DIFF
--- a/extras/bindata.bzl
+++ b/extras/bindata.bzl
@@ -55,6 +55,7 @@ def _bindata_impl(ctx):
     return [
         DefaultInfo(
             files = depset([out]),
+            runfiles = ctx.runfiles([out]),
         ),
     ]
 


robfig:
Thanks for the fix! Is it feasible to augment the bindata test in tests/legacy/examples/bindata/ so that it would detect this problem, to avoid regressions?

fmeum:
Just as a side remark: This behavior of the native rules, which is inconsistent with the recommended behavior for Starlark rules, will be fixed with https://github.com/bazelbuild/bazel/pull/15052. But this PR is still a good workaround in the meantime.

fmeum:
`bindata` is deprecated and marked for removal in 0.39.0. Going forward, `embedsrcs` is the recommended way to embed files backed by `//go:embed`.

The underlying issue in Bazel is fixed in 6.0.0.

PR #3029, opened 2021-12-14
Always include debug symbols with -c dbg

fmeum:
**What type of PR is this?**

Bug fix

Previously, debugging a `go_test` target was only possible with an explicit `--@io_bazel_rules_go//go/config:debug`.

Fixes https://github.com/bazelbuild/intellij/issues/2313, which I think is more of a `rules_go` issue.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -107,7 +107,11 @@ cgo_context_data_proxy(
 # to depend on all build settings directly.
 go_config(
     name = "go_config",
-    debug = "//go/config:debug",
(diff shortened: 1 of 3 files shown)


robfig:
I'd normally be hesitant to add one-off rules for toggling existing flags or merge a change lacking tests, but this seems reasonable and low-risk and bazel-discuss seemed favorable towards it (https://groups.google.com/g/bazel-discuss/c/1_PC-Pj-nRg). Thanks

PR #3032, opened 2021-12-17
link.bzl: ignore duplicate dep on coverdata

### LINKED ISSUE #3017: go/tools/coverdata: multiple copies of package passed to linker

v0.29.0

Not using Gazelle

4.2.1

Yes

macOS arm64 and Linux amd64

### What did you do?

1. Create a workspace like this:

```
-- WORKSPACE --

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "2b1641428dff9018f9e85c0384f03ec6c10660d935b750e3fa1492a281a53b0f",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.29.0/rules_go-v0.29.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.29.0/rules_go-v0.29.0.zip",
    ],
)

go_rules_dependencies()

go_register_toolchains(version = "1.17.1")
-- fx.go --
package fx

import (
	_ "uber.com/internal"
)
-- fx_test.go --
package fx
-- internal/BUILD.bazel --

go_library(
    name = "go_default_library",
    srcs = ["lib.go"],
    importpath = "uber.com/internal",
    deps = ["@io_bazel_rules_go//go/tools/coverdata"],
)

go_test(
    name = "go_default_test",
    srcs = ["lib_test.go"],
)
-- internal/lib.go --
package internal

import _ "github.com/bazelbuild/rules_go/go/tools/coverdata"
-- internal/lib_test.go --
package internal
-- BUILD.bazel --

go_library(
    name = "go_default_library",
    srcs = ["fx.go"],
    importpath = "code.uber.internal/devexp/code-coverage/cmd/fx",
    deps = ["//internal:go_default_library"],
)

go_test(
    name = "go_default_test",
    srcs = ["fx_test.go"],
)
```
2. Run `bazel build --collect_code_coverage //:go_default_test` or `bazel coverage //:go_default_test`

Build succeeds

### What did you see instead?

```
link: package conflict error: github.com/bazelbuild/rules_go/go/tools/coverdata: multiple copies of package passed to linker:
	@io_bazel_rules_go//go/tools/coverdata:coverdata
	@io_bazel_rules_go//go/tools/coverdata:coverdata
Set "importmap" to different paths or use 'bazel cquery' to ensure only one
package with this path is linked.
Target //:go_default_test failed to build
```

Interestingly, `bazel coverage //internal:go_default_test` succeeded, even it's the package that imports "github.com/bazelbuild/rules_go/go/tools/coverdata"

robfig:
Importing the coverdata package from user code caused the link to fail:

  link: package conflict error: github.com/bazelbuild/rules_go/go/tools/coverdata: multiple copies of package passed to linker:
          @io_bazel_rules_go//go/tools/coverdata:coverdata
          @io_bazel_rules_go//go/tools/coverdata:coverdata

Two different archives were indeed being provided as the object files for that package:

  -arc '@io_bazel_rules_go//go/tools/coverdata:coverdata=github.com/bazelbuild/rules_go/go/tools/coverdata=bazel-out/darwin_arm64-fastbuild/bin/external/io_bazel_rules_go/go/tools/coverdata/coverdata.a'
  -arc '@io_bazel_rules_go//go/tools/coverdata:coverdata=github.com/bazelbuild/rules_go/go/tools/coverdata=bazel-out/darwin_arm64-fastbuild-ST-8f6eb6207b9d/bin/external/io_bazel_rules_go/go/tools/coverdata/coverdata.a'

This is due to a combination of two things:

* coverdata is defined as a `go_tool_library` because it is dependended upon by the toolchain.
  Declaring it as a `go_library` would cause a circular dependency.

* nogo support is implemented with a transition.
  This avoids a circular dependency that requires nogo to build nogo.

Either due to being a `go_tool_library` or due to being a dependent of
the toolchain, coverdata is included on both sides of the transition
when also imported by user code. Since the built package in each
configuration corresponds to different object files, this leads to
the duplication.

In this case, it does not seem important to select a specific
coverdata package (instrumented or not), and it does not appear to be
straightforward to determine which is which. Take the easy way out and
have the policy be: first one in the depset wins.

Since it's not clear if this occurs with other packages (e.g. nogo analyses),
log a warning when a duplicate archive is being ignored.

Fixes #3017


sluongng:
May be it's just me but this is somewhat similar to my fixes for nogo issues in https://github.com/bazelbuild/rules_go/pull/2996/files. Also created a map to de-duplicate things (with a secondary effect which is mapping the system modified source code back to original user's source code).

Not entirely sure whether `go/private/actions/link.bzl` is a more appropriate place to do these or `go/tools/builders/compilepkg.go`.

robfig:
I'm not sure.. I don't think it's possible to solve this in compilepkg.go? The current code already replaces the source, rather than adding a dupe:

```
	if coverMode != "" {
		// ....
		for i, origSrc := range combined {
			// ....
			coverSrc := filepath.Join(workDir, fmt.Sprintf("cover_%d.go", i))
			if err := instrumentForCoverage(goenv, origSrc, srcName, coverVar, coverMode, coverSrc); err != nil {
				return err
			}

			if i < len(goSrcs) {
				goSrcs[i] = coverSrc  // <--- replaces
				nogoSrcsTranslate[coverSrc] = origSrc
			} else {
				cgoSrcs[i-len(goSrcs)] = coverSrc
			}
```

and that confirms what I saw when diagnosing -- each individual compile correctly had a single copy of coverdata, but the eventual link stage had multiple because not all those copies were built with the same configuration.

linzhp:
Tested in Uber's go repo. It works. Thanks for the fix.

fmeum:
@robfig This seems very useful to have. If you are busy, I can apply the fixup suggested by @linzhp and resolve the conflicts.

fmeum:
@linzhp I applied your suggestion and fixed the conflict. Do you think we should merge this or instead discourage use of `coverdata` from outside rules_go?

linzhp (approved):
LGTM, please fix the buildifer issue

PR #3035, opened 2021-12-19
Replace go_name_hack with a usage of Label

fmeum:
**What type of PR is this?**

Other

The Label constructor can be used to transform a label string such as @io_bazel_rules_go//foo:bar into the canonical label referencing this target from the repository in which it is called.

For example, `str(Label("@io_bazel_rules_go//foo:bar"))` returns `"@io_bazel_rules_go//foo:bar"` if rules_go is used as an external repository from another main workspace, but returns `"//foo:bar"` when rules_go is the main repository (e.g., when running the tests in the CI).

It doesn't resolve a bug, but simplifies the rules_go repository setup. This simplification will be especially useful when converting `rules_go` to a Bazel module in the future.

**Other notes for review**

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -33,10 +33,6 @@ load(
     "GoSource",
 )
 load(
-    "@io_bazel_rules_go_name_hack//:def.bzl",
-    "IS_RULES_GO",
-)
-load(
     "@io_bazel_rules_go//go/platform:crosstool.bzl",
     "platform_from_crosstool",
 )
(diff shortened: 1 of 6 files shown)


fmeum:
@robfig Would you be interested in getting rules_go into the [Bazel Central Registry](https://github.com/bazelbuild/bazel-central-registry) as a Bazel module? I could work on it using only patches at first, but of course would prefer to upstream (backwards compatible) changes at some point. This is one of them which I found to be generally useful.

go/private/rules/transition.bzl
```diff
@@ -44,21 +40,16 @@ load(
...
-    like @io_bazel_rules_go//go/config:goos.
+    This works around bazelbuild/bazel#10499 by automatically using the correct
+    way to refer to this repository (@io_bazel_rules_go from another workspace,
+    but only repo-relative labels if this repository is the main workspace).
     """
-    if IS_RULES_GO and label.startswith("@io_bazel_rules_go"):
-        return label[len("@io_bazel_rules_go"):]
-    else:
+    if label.startswith("//command_line_option:"):
+        # This is a special label used to refer to Bazel command-line options in
```
achew22, line 48: Can you add a note about why are these special and when it could be removed? My experience is that  hacks like this get put in place and I forget to write down the condition where it can be removed safely so I carry forward scar tissue.

achew22 (changes requested):
@fmeum thanks so much for the contribution! I'm really excited to have this in place. I've been bitten a number of times by this hack (skylib was made difficult by it) and would love to get rid of it. Just a small comment

go/private/rules/transition.bzl
fmeum: Good point, I expanded the comment. Could you give it another read?

achew22:
Thanks @fmeum!

PR #3043, opened 2022-01-13
compilepkg: fix paths in compile errors

RaduBerinde:
A previous change (18a0557abfe2e1029fa12e35a46e846eb2564292) resulted
in full paths being shown when there are compile errors. The paths
are in a temporary directory and they can be problematic with tools
that interpret the build output (e.g. vim's `:make`).

This change fixes this by switching the current working directory
before calling out to `compile`.

For example - before:
```
/tmp/rules_go_work-3549239678/github.com/cockroachdb/cockroach/pkg/sql/scan.go:91:20: undefined: descpbx
```
After:
```
github.com/cockroachdb/cockroach/pkg/sql/scan.go:91:20: undefined: descpbx
```

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -204,14 +204,19 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 
 	// Copy regular Go source files into the work directory so that we can
 	// use -trimpath=workDir.
-	goBases, err := gatherSrcs(workDir, goSrcs)
+	subDir := filepath.Join(workDir, packagePath)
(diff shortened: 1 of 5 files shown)


RaduBerinde:
Oops, wrong repo, sorry.

achew22:
@RaduBerinde out of curiosity, what was the correct repo? Is this an error in rules_go that you've been running into?

RaduBerinde:
@achew22 this was in CockroachDB's fork (https://github.com/cockroachdb/rules_go/). This is solving a problem that we have only in our fork, caused by an earlier [change](https://github.com/cockroachdb/rules_go/commit/18a0557abfe2e1029fa12e35a46e846eb2564292). That change was a workaround for https://github.com/cockroachdb/cockroach/issues/64379 - let us know if you have any better ideas of how to solve that issue.

PR #3047, opened 2022-01-17
Basic bzlmod setup

fmeum:
**What type of PR is this?**

Feature

This PR adds a MODULE.bazel that makes rules_go usable as a bzlmod module for basic use cases.

The new module definition is verified by a test module that will be used in the presubmit tests of the Bazel Central Registry (BCR).

The following features require more thought and/or work and are not yet supported:
* SDK rules other than go_host_sdk and go_download_sdk.
* non-no-op nogo
* go_proto_library

Work towards #3020.

**Other notes for review**

Based on #3046, so please review that one first.


fmeum:
This is essentially ready for review. The CI fails because the test machines have no Go installed and I chose to use the host SDK as the default toolchain for the new module. Of course we can change that to a hermetic toolchain and/or install Go in the CI.

@achew22 @robfig What is your take on this? Bazel modules are generally expected to be "batteries included" in the sense that they provide a reasonable default toolchain that can then be overriden by the user. In the case of Go, using the host SDK sounds reasonable and would also be in line with how C++ and Java behave with Bazel. However, that would require making a Go SDK available in the CI.

fmeum:
Sorry, I misunderstood the CI failure. Bazel of course has to load the repositories of each toolchain, even if it doesn't end up being selected, so the failure would occur for every user without a local Go SDK. I will look into making this a soft fail, but would already appreciate your feedback on the rest of the PR.

fmeum:
I have put some more thought into how toolchain selection could be improved with bzlmod. I will try to get some feedback on these in the bzlmod community before I continue here.

ar3s3ru:
Hey there @fmeum! Any updates on this PR?
Would be awesome to see `rules_go` support Bzlmod

fmeum:
I haven't continued working on it since `rules_go` without `gazelle` wouldn't be that useful and turning `gazelle` into a bzlmod module would most likely require more progress on https://github.com/bazelbuild/bazel/issues/14590 and https://github.com/bazelbuild/bazel/issues/14580. I will revisit this PR when things have improved.

ar3s3ru:
@fmeum is it really necessary to turn `gazelle` into a bzlmod module for this PR to be merged?
AFAIK the dependency is `gazelle` -> `rules_go`, not the other way around.

What I can see from the failing CI is that the code is expecting a toolchain to be installed in the host machine by default, but we could package a default toolchain instead and make it hermetic, no?

Can't `gazelle` be used still within `WORKSPACE` while these blocking issues are resolved?

achew22:
Personally I agree with @fmeum. `gazelle` provides `go_repository` which is how third-party go code would be loaded into the build graph. I try VERY hard to not add any third-party dependencies on a project, but I'm not sure I've ever made it to production without one. `rules_go` is, sadly, not very useful without `gazelle`.

achew22:
I'm happy to merge this, but what is the story for 3p dependencies declared by rules_go consumers?

fmeum:
@achew22 My current plan is as follows:

1. Get rules_go into the BCR based on this PR. The BCR is still in beta, so we can make changes at any time - there is no need to commit to an API yet.
2. Work on a module extension provided by Gazelle that allows modules to declare their go.mod-style (no http/git/forks/... for starters) dependencies in a way that works well with transitive dependencies. I have a rough prototype of this.
3. Use this module extension in rules_go (cyclic dependencies between modules are permitted) to get the Go modules required for protos to work. That will be the first real test case for these two modules.

After these basics have been found to work, we should start proper discussions on the APIs and features we want to support. I just think it's too early for extended design discussions until there is at least one ruleset handling transitive dependencies in the BCR - currently there are none.

achew22:
Are we dependent on bzlmod being experimental to note that we are going to make backward incompatible changes? Is there any way to put up some blinking hazard lights to notify anyone who may stumble into it that it's definitely not ready for prime time yet?

fmeum:
> Are we dependent on bzlmod being experimental to note that we are going to make backward incompatible changes? Is there any way to put up some blinking hazard lights to notify anyone who may stumble into it that it's definitely not ready for prime time yet?

That's a good point. I added https://github.com/bazelbuild/rules_go/pull/3047/files#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR10. Is is scary enough?

bcr_tests/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
+module(
+    name = "rules_go_bcr_tests",
+    # Test that the default SDK is registered by not registering one from the test module.
+)
+
+bazel_dep(name = "io_bazel_rules_go", version = "", repo_name = "my_rules_go")
+local_path_override(
+    module_name = "io_bazel_rules_go",
```
achew22, line 8: I thought it was going to show up in bzlmod as just `rules_go`?

MODULE.bazel
```diff
@@ -0,0 +1,23 @@
+module(
+    name = "io_bazel_rules_go",
+    version = "0.33.0",
+    compatibility_level = 1,
```
achew22, line 4: Is it okay to put the compatibility level to be the same as the major version, `0`?

bcr_tests/.bazelrc
```diff
@@ -0,0 +1 @@
+build --experimental_enable_bzlmod
```
achew22, line 1: Is there a reason for `bcr_tests` be a separate top level directory in the root rather than live under the general `tests` directory?

go/private/sdk.bzl
```diff
@@ -233,7 +237,7 @@ def _sdk_build_file(ctx, platform):
             "{goos}": goos,
             "{goarch}": goarch,
             "{exe}": ".exe" if goos == "windows" else "",
-            "{rules_go_repo_name}": Label("//go/private:BUILD.sdk.bazel").workspace_name,
```
achew22, line 236: I think the logic behind this was that it would resolve to `io_bazel_rules_go` if you imported it as such, but if people started to import as `rules_go` they would get that. Does this force us to use `io_bazel_rules_go` going forward?

bcr_tests/.bazelrc
fmeum: Moved it, no particular reason - it just needs to be a separate directory.

go/private/sdk.bzl
fmeum: This is a pattern that currently breaks with bzlmod: With repo mappings, which it uses, the value of `workspace_name` will be the *mapped* repo name, which can't be used to form a valid label. This is where https://github.com/bazelbuild/bazel/issues/15593 would help.

As a result, this line either needs to be replaced on master or via a patch in the BCR. I opted for the former as there are other blockers to a non-bzlmod repo name agnostic rules_go (see my other comment).

MODULE.bazel
fmeum: It does work.

MODULE.bazel
```diff
@@ -0,0 +1,23 @@
+module(
+    name = "io_bazel_rules_go",
```
achew22, line 2: Can we use this opportunity to go to just `rules_go`? I think that's allowed for rules that live in the `bazelbuild` org.

achew22 (approved):
LGTM % 1

MODULE.bazel
fmeum: That is something else we can only do via a patch for now: Since there is no `repo_name` attribute for the main module, this has to remain identical to the `name` attribute in our `workspace` declaration. In the BCR, we will patch it to read `rules_go`.

achew22: Is that a limitation of BCR and therefore something we should mostly ignore for now, or is this something where I should do some early legwork to get things sorted out?

fmeum: It's not a limitation of the BCR, rather it was a deliberate choice made by Xudong and Yun: Every Bazel module *except* the root module can specify a repo name that's different from the module name. If we were to change the name to `rules_go` here, the build would fail as the remaining references to `@io_bazel_rules_go` could not be resolved. In the BCR, we patch these to read `@rules_go` and thus can also change the module name to `rules_go`. I don't think we can do anything about this until the migration to `rules_go` has been completed for both `WORKSPACE` and `MODULE.bazel` consumers.

Issue #3036, opened 2021-12-20
Upstream `github.com/phst/runfiles` as `@io_bazel_rules_go//go/runfiles`


fmeum:
rules_go currently doesn't offer a way to discover runfiles following the [standard procedures for runfiles discovery](https://docs.google.com/document/d/e/2PACX-1vSDIrFnFvEYhKsCMdGdD40wZRBX3m3aZ5HhVj4CtHPmiXKDCxioTUbYsDydjKtFDAzER5eg7OjJWs3V/pub). There is https://pkg.go.dev/github.com/bazelbuild/rules_go/go/tools/bazel, but due to it predating the coordinated efforts to provide consistent runfiles libraries for all languages, it uses a non-standard procedure and doesn't interoperate well with the other libraries. It also invokes a bit too much magic that makes it prone to unexpected issues (e.g. it searches the working directory first before looking up the runfiles variables).

A proper replacement has been available as https://pkg.go.dev/github.com/phst/runfiles for quite some time now, so I would propose to upstream it (and would offer to do the work). According to the ["Deploying rules"](https://docs.bazel.build/versions/main/skylark/deploying.html#runfiles-library) guide, this runfiles library should be available under `@io_bazel_rules_go//go/runfiles` (Gazelle could probably be taught to pick it up automatically).

achew22:
Reading through the linked [Bazel Runfiles Libraries](https://docs.google.com/document/d/e/2PACX-1vSDIrFnFvEYhKsCMdGdD40wZRBX3m3aZ5HhVj4CtHPmiXKDCxioTUbYsDydjKtFDAzER5eg7OjJWs3V/pub) doc I think I would like a little bit more authoritative of a citation around the expected structure of the runfiles library. ["Deploying Rules"](https://docs.bazel.build/versions/main/skylark/deploying.html#runfiles-library) is an authoritative source for me and has me convinced that `@io_bazel_rules_go//go/runfiles` is in fact the correct place for the code to live, but if there is a mandated interface, I'd like that interface to be written down and in the mainline bazel docs so we don't end up having to support 3 versions (the original version, the version you wrote, and the version that replaced Laszlo's 2018 design doc).

As a proposed fix for this, could we send a PR to update `https://github.com/bazelbuild/bazel/blob/master/site/docs/skylark/deploying.md` filling in the `Runfiles library` section with the expected API? That way we, the bazel community, will have a canonical resource for discussing this in the future.

What do you think?

[...]

PR #3031, opened 2021-12-16
Update x/tools repository to v0.1.8 #3030

chressie:
Ran the following commands:

```
bazel build go/tools/releaser
sed -i /org_golang_x_tools-public-visibility.patch/d go/private/repositories.bzl
rm third_party/org_golang_x_tools-public-visibility.patch
bazel-bin/go/tools/releaser/releaser_/releaser upgrade-dep org_golang_x_tools
```

**What type of PR is this?**

Bug fix

**What does this PR do? Why is it needed?**

Updates x/tools repository dependency to 0.1.8, which is needed to support new export format version emitted by Go 1.18.

**Which issues(s) does this PR fix?**

Fixes #3030

**Other notes for review**

n/a

PR #3019, opened 2021-11-28
builders/gentestmain: register correct cover mode in the coverage report file when enabling race mode

lbcjbb:
**What type of PR is this?**

Bug fix

**What does this PR do? Why is it needed?**

Fix the cover mode to "atomic" in the coverage data file when the race detector is used.

**Which issues(s) does this PR fix?**

Fixes #3018

PR #3011, opened 2021-11-18
Trim absolute paths in files generated by cgo

fmeum:
**What type of PR is this?**

Bug fix

**What does this PR do? Why is it needed?**

The cgo command generates .c files that contain //line comments
referencing source files by absolute paths into the Bazel cache. As the
contents of these comments end up in the symbol table of the resulting
c-archive, this makes Bazel-built CGo binaries non-reproducible.

This is fixed by passing the -trimpath argument to cgo to trim the
Bazel execroot prefix from these paths.

**Which issues(s) does this PR fix?**

Fixes #3010

PR #3049, opened 2022-01-18
Release rules_go v0.30.0

JamyDev:
Release prep. ran the following tools:
- `bazel sync`
- `bazel run //go/tools/releaser -- upgrade-dep all`
- `bazel sync`
- `bazel build //...`
- `bazel test //...`

# Release Notes

# Changes

- Add first class support for GOOS=ios (#3013)
- Always include debug symbols with -c dbg (#3029)

  Previously, debugging a go_test target was only possible with an explicit --@io_bazel_rules_go//go/config:debug.

- go_test: Add env attribute (#3004)

  This allows setting environment variables before static initialisers run, and which reference expanded locations.

- tests/core/cgo: do not assume test is being run from main repo (#2956)
- reproducibility_test: explicitly build the go_sdk builder (#3026)

# Bug Fixes

- Emit correct header in CGo c-archive (#2874)
- Append path into trimpath if option already exists (#2994)
- Fixed issue with `go_bazel_test` rule on Windows (#3036)
- Skipping linker flag for versioned dylib on Darwin (#2989)
- Trim absolute paths in files generated by cgo (#3011)
- builders/cover: register coverage without changing line numbers (#2993)
- cover: use Z instead of underscore for variable name separator (#2995)
- .bazelci/presubmit.yml: remove old flag (#3023)
- builders/gentestmain: register correct cover mode in the coverage report file when enabling race mode (#3019)
- -fsanitize-ignorelist should have an absolute path (#3001)
- internal changes: #2942, #2987, #2988, #2992, #3000, #3035, #3038, #3046

Thanks to all of the contributors

# Compatibility

- Bazel 4.2.1 or higher is now required. The previous minimum version
  was 4.2.0.

# Updated dependencies

- x/tools: upgraded to v0.1.8 (#3031)

# `WORKSPACE` code

// TBD after release


go/tools/bzltestutil/BUILD.bazel
```diff
@@ -15,14 +15,11 @@ go_tool_library(
 go_test(
     name = "bzltestutil_test",
     srcs = [
-        "init.go",
```
linzhp, line 18: Where are these changes from?

JamyDev: Came in after I ran the depency updater. Guess the gazelle run after it cleared those out.

linzhp:
Repositories like [go/sys](https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl#L99) should be updated to the tip of their master. Maybe there is a bug in the releaser tool?

JamyDev:
Yeah I was surprised it didn't change anything else. Can you try and run it, maybe I don't have the permissions to run it properly.

go/tools/bzltestutil/BUILD.bazel
JamyDev: After my last PR this is definitely from someone not running gazelle after making changes, best to add these in as well since this would always keep showing up.

linzhp: I think rules for this package was written by hand. It was called `go/tools/testinit` initially, and [excluded from Gazelle](https://github.com/bazelbuild/rules_go/blob/master/BUILD.bazel#L36), possibly because of the use of `go_tool_library`. However, Jay forgot to update the `gazelle:exclude` in #2755. Now it seems Gazelle work with this package, so I don't mind having Gazelle to do that rule generation. However, it's better to put it in a different PR together with the `gazelle:exclude` change.

linzhp (approved):

go/tools/bzltestutil/BUILD.bazel
JamyDev: Undone these changes. Making a PR for this later.

PR #3051, opened 2022-01-19
tools/upgradedep: add support for Label based repositories.bzl

### LINKED ISSUE #3050: Dependency upgrader fails to run due to change to Label() dependencies

N/A trying to release rules_go.

v0.22.3

4.2.2

Yes

Debian 10

Just pulled master of rules_go

### What did you do?

Running `bazel run //go/tools/releaser -- upgrade-dep all` results in a `upgrading com_github_golang_protobuf: not all patches are string literals` error, due to the Label change in repositories.bzl.

Working on a fix now so I can release v0.30.0 (#3049). Just wanted to make the issue to track.

JamyDev:
**What type of PR is this?**

Bug fix

upgradedep.go was not working due to the change to Label() based patches in #3035. Modified the parser to support both string and Label() patches.

Added tests to ensure all cases are covered for the additional code.

Fixes #3050

**Other notes for review**


go/tools/releaser/upgradedep_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"io/ioutil"
+	"testing"
+
+	bzl "github.com/bazelbuild/buildtools/build"
+	"github.com/bazelbuild/rules_go/go/tools/bazel"
+)
+
+const (
+	reposFile = "go/tools/releaser/testdata/repositories.bzl"
+	funcName  = "go_rules_dependencies"
```
linzhp, line 14:
```suggestion
	_reposFile = "go/tools/releaser/testdata/repositories.bzl"
	_funcName  = "go_rules_dependencies"
```

[Prefix Unexported Globals with _](https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_)

go/tools/releaser/upgradedep_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func TestPatchItemParser(t *testing.T) {
```
linzhp, line 17: This test adds a lot of complexity just to cover whether whether a patch expression is parsed correctly. Can we avoid all the file loading and rule parsing in the test by doing something like:

```go
patchExpr, err := bzl.Parse("repos.bzl", 
[]byte(`# releaser:patch-cmd gazelle -repo_root . -go_prefix golang.org/x/tools -go_naming_convention import_alias
 Label("//third_party:org_golang_x_tools-gazelle.patch")`))
if err != nil {
   t.Fatalf(...)
}
patchLabelStr, comment, err := parsePatchesItem(patchExpr)
if patchLabelStr != "//third_party:org_golang_x_tools-gazelle.patch" {
  t.Fatalf(...)
}
```

linzhp:
The other way to test this PR is to extract the logic that reads `repositories.bzl` to a separate function, and read the real `repositories.bzl` to test that function. That way, future incompatible changes to `repositories.bzl` will fail that test.

go/tools/releaser/upgradedep_test.go
JamyDev: Oh that makes sense. Didn't immediately think that we don't really need all the wrapping around the test. I'll put something together in the AM.

JamyDev:
I think reading the entire repositories file and doing a dry run of the upgrader would be a good integration test case, but that would bloat this PR, so that's a separate piece of work.

go/tools/releaser/upgradedep.go
```diff
@@ -488,6 +489,45 @@ func upgradeDepDecl(ctx context.Context, gh *githubClient, workDir, name string,
...
+	patchLabelCall, ok := patchLabelExpr.(*bzl.CallExpr)
+
+	if ok {
+		// Verify the identifier, should be Label
+		ident, ok := patchLabelCall.X.(*bzl.Ident)
+		if !ok {
+			return "", nil, fmt.Errorf("invalid identifier while parsing patch label")
+		}
+		if ident.Name != "Label" {
+			return "", nil, fmt.Errorf("invalid patch function: %s", ident.Name)
```
blico, line 503: NIT: `%q` is generally preferred over `%s` for error strings

go/tools/releaser/upgradedep.go
```diff
@@ -433,19 +433,20 @@ func upgradeDepDecl(ctx context.Context, gh *githubClient, workDir, name string,
...
-				return fmt.Errorf("not all patches are string literals")
+			patchLabelValue, comments, err := parsePatchesItem(patchLabelExpr)
+			if err != nil {
+				return err
 			}
-			if !strings.HasPrefix(patchLabel.Value, "@io_bazel_rules_go//third_party:") {
-				return fmt.Errorf("patch does not start with '@io_bazel_rules_go//third_party:': %s", patchLabel)
+
+			if !strings.HasPrefix(patchLabelValue, "//third_party:") {
+				return fmt.Errorf("patch does not start with '//third_party:': %s", patchLabelValue)
```
blico, line 442: NIT: `%q` instead of `%s`

blico:
Looks good besides some mostly stylistic comments.

go/tools/releaser/upgradedep_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("%v", tt.expression), func(t *testing.T) {
+			patchExpr, err := bzl.Parse("repos.bzl", tt.expression)
+			if err != nil {
+				t.Fatalf(err.Error())
+			}
+
+			patchLabelStr, _, err := parsePatchesItem(patchExpr.Stmt[0])
+			if err != nil && err.Error() != tt.error {
```
linzhp, line 72: I prefer to break this into two test tables: `TestPatchItemParser_Success` and `TestPatchItemParser_Error`. That way we can keep the assertions exactly the same for each test table and avoid some branching logic in the test body. In general, we should avoid putting logic in the tests [1][2], because we have tests to cover production logic, but we don't have tests to cover test logic.

[1] https://learning.oreilly.com/library/view/software-engineering-at/9781492082781/ch12.html#donapostrophet_put_logic_in_tests
[2] https://learning.oreilly.com/library/view/software-engineering-at/9781492082781/ch12.html#tests_and_code_sharing_dampcomma_not_dr

JamyDev: Totally fair, was mapping my original case (repos.bzl test file) into a single test, but obviously the landscape changed a little.

PR #3060, opened 2022-01-24
BUILD: cleaned up some old gazelle ignores

JamyDev:
**What type of PR is this?**

Other

Clean up of some old references left behind from #2755.

Fixes up some Gazelle references. Makes it so it doesn't edit this file every time we run gazelle.


JamyDev:
Looks like the build failed on some network issue. Can someone retry it? I don't have the right permissions for it.

PR #3066, opened 2022-02-04
Minor feature: allowing patching the toolchain

amscanne:
**What type of PR is this?**

Feature

This PR adds a trivial feature to go_download_sdk that allows for applying a single patch via the repository_ctx patch method. This is only exposed via the go_download_sdk rule, and not any wrapper rules.

As a trivial feature, there is no issue.


go/private/sdk.bzl
```diff
@@ -121,6 +123,8 @@ _go_download_sdk = repository_rule(
         "sdks": attr.string_list_dict(),
         "urls": attr.string_list(default = ["https://dl.google.com/go/{}"]),
         "version": attr.string(),
+        "patch": attr.label(default = None),
+        "patch_strip": attr.int(default = 0),
```
linzhp, line 127: Let's follow the convention of `go_repository` rule:

* Having `patches` to be a label_list, so it can take multiple patches
* Having `patch_args` to be string_list, and defaults to `["-p0"]`

PR #3068, opened 2022-02-16
Use merge_all() in emit_archive.

moisesvega:
**What type of PR is this?**
Other

This PR updates emit_archive to use the new api `merge_all()` added in [bazel 5.0](https://blog.bazel.build/2022/01/19/bazel-5.0.html#starlark-build-language).

Because of the changes in the Starlak language, it is possible to hit `depth_limit` while using `merge()` inside of a loop, the fix is using `merge_all()` instead of `merge()`. 

This PR includes: 

- Use `merge_all()` in `emit_archive`

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -39,6 +39,8 @@ load(
     "emit_compilepkg",
 )
 
+load("//go/private/skylib/lib:versions.bzl", "versions")
+
 def emit_archive(go, source = None, _recompile_suffix = ""):
     """See go/toolchains.rst#archive for full documentation."""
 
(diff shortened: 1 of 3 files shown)


linzhp:
It appears that we can get Bazel version using `native.bazel_version` like [this](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl#L20). Can you use so we don't need to bump the minimal bazel version now?

go/private/actions/archive.bzl
```diff
@@ -65,10 +67,26 @@ def emit_archive(go, source = None, _recompile_suffix = ""):
...
-        runfiles = runfiles.merge(a.runfiles)
-        if a.source.mode != go.mode:
-            fail("Archive mode does not match {} is {} expected {}".format(a.data.label, mode_string(a.source.mode), mode_string(go.mode)))
+
+    # After bazel version 5.0.0, skylark introduced a new API called merge_all().
+    # There is a recommendation in the release notes to use merge_all() if
+    # we hit the depth limit because of using merge() in a loop, we saw
+    # some repos hitting this issue after upgrading to 5.0.0
+    # To keep it backwards compatible, we will fallback to the old logic if
+    # the version is not at least 5.0.0
```
linzhp, line 76: nit: you can also add a TODO saying that we can remove this check after the minimal version of Bazel becomes 5.0.0, which we will have after integrating with bzlmod

linzhp (approved):

moisesvega:
> It appears that we can get Bazel version using `native.bazel_version` like [this](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl#L20). Can you use so we don't need to bump the minimal bazel version now?

At the moment `native.bazel_version` is not available for rules. [Here](https://github.com/bazelbuild/bazel/issues/8305) is the conversation about allowing native.bazel_version outside of `WORKSPACE`.

For that reason, using "hasattr(runfiles, "merge_all")" is enough to distinguish between skylark API's.

PR #3076, opened 2022-02-25
Prevent downgrading platforms (#3071)

### LINKED ISSUE #3071: Not bundling @platforms, or at least not downgrading it?
Hi Bazel Go folks :)

I ran into a hitch that I thought I should raise for discussion. To make a long story short: My team had stumbled around for a bit before someone helped us realize that, by default, rules_go was downgrading bazelbuild/platforms dependency below the version bundled with Bazel. 

[For more details: _maybe doesn't seem to properly capture the existence of the built in platforms. An older version of rules_go was getting pulled in by the latest gRPC, and that version of rules_go was overriding the latest platforms with a much older, problematic version of platforms. While this can be worked around by manually importing platforms, I suspect I'm not the last person to get tripped up by this. That is, with the current behavior, today's rules_go is likely to be tomorrow's out-of-date version, causing confusion when it breaks unrelated parts of someone else's build.]

I'm  writing to ask if rules_go would consider not downgrading the built in platforms.
Some options that occur:
- If it's okay to have a min bazel version of at least 0.28 for rules_go (readme says "a recent version of Bazel," which that is not), then maybe a good fix would be to just not bundle platforms in https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl, since you can count on having a built in version.
- Otherwise, we could use https://github.com/bazelbuild/rules_go/blob/master/go/private/skylib/lib/versions.bzl to only pull in platforms if it's not bundled--working around the somewhat surprising behavior of `_maybe` not detecting builtins?
- [Or somehow detect the version of the build in platforms and only upgrade?]

Thanks so much for your consideration!
Chris
(ex-Googler)

cpsauer:
**What type of PR is this?**

Bug fix

It proposes one good way of preventing silent, problematic platform downgrade. Please see #3071 for context and other options. (But since it's had been a week without hearing back and others were having the issue, I figured I'd propose the fix that seemed the simplest and best.)

Note: Platforms already long guaranteed to be bundled with Bazel given the minimum Bazel version check of 4.2.1, and declaring an explicit version causes problems.

Fixes #3071

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -38,22 +38,6 @@ def go_rules_dependencies():
     if getattr(native, "bazel_version", None):
         versions.check(MINIMUM_BAZEL_VERSION, bazel_version = native.bazel_version)
 
-    # Repository of standard constraint settings and values.
-    # Bazel declares this automatically after 0.28.0, but it's better to
-    # define an explicit version.
-    # releaser:upgrade-dep bazelbuild platforms
-    _maybe(
-        http_archive,
-        name = "platforms",
-        # 0.0.4, latest as of 2022-01-24
-        urls = [
-            "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.4/platforms-0.0.4.tar.gz",
-            "https://github.com/bazelbuild/platforms/releases/download/0.0.4/platforms-0.0.4.tar.gz",
-        ],
-        sha256 = "079945598e4b6cc075846f7fd6a9d0857c33a7afc0de868c2ccb96405225135d",
-        strip_prefix = "",
-    )
-
     # Needed by rules_go implementation and tests.
     # We can't call bazel_skylib_workspace from here. At the moment, it's only
     # used to register unittest toolchains, which rules_go does not need.


sluongng:
I don't think deleting this would be the right fix.

Instead, the dependencies is loaded with `_maybe()` which will not load if there is an http_archive with the same named defined prior to the loading of `go_rules_dependencies`.

Have you tried defining a custom `http_archive` before `go_rules_dependencies` declaration?

cpsauer:
Hi, @sluongng! Thanks for taking a look.

Well aware of being able to manually override the problematic _maybe with a custom http_archive :) 

Manually importing is explicitly contemplated in the background issue (#3071), along with (in the days since) a stream of other users stumbling over it. The issue is instead that we almost certainly don't want to rely on users having to manually fight a problematic default that silently causes hard to debug problems. Lots more good discussion of that and other options in #3071. 

Are there things we actually need the _maybe(platforms) for that outweigh its problems--or the complexity of a different fix?

[No substitute for reading #3071, but recall that `_maybe()` fails to detect the `@platforms` already bundled with Bazel, so it most often results in a downgrade for up-to-date users...and likely a fairly severe downgrade if ancient rules_go is being pulled in as a transitive dependency from, e.g., gRPC, as it often is. This is really quite hard to debug, because there's no reason users would expect an external repository to subtly break unrelated behavior. Others don't.]

All that said, definitely happy if you guys want to fix #3071 another way! As above, I just tossed this up as the simplest option that fixes things and because the user is guaranteed to have a pretty recent version of platforms anyway. (Not doing this because I need it for myself, but rather because I'd like to help others who will just keep running into this, leaving things better than I found them.)

Thanks so much for your consideration!
Chris
(ex-Googler)

seh:
I had tried pulling in the "platforms" repository myself (per https://github.com/bazelbuild/rules_go/issues/3071#issuecomment-1058194234) and it didn't solve this problem, but perhaps I pulled it in too late, after _rules_go_ had set up its own version. At minimum that shows that it's possible to try and still fail. I agree with @cpsauer that users of _rules_go_ shouldn't have to bear this burden. I expect _rules_go_ should work fine on its own without me needing to patch it in the right way.

cpsauer:
Friendly bump :)

Would love to get another reply--or approval--from someone with merge powers, looking at this with the additional context!

Not wedded to this fix, certainly, just trying to solve this before too many other people get tripped up by it. That said, this does seem like the most direct, sufficient fix.

Tagging @achew22 and @robfig, since it looks like they've most recently merged in non-release changes here. Hope that's okay!

cpsauer:
Another friendly bump :)

linzhp (approved):
Since Bazel already has it, it's one less thing for rules_go to worry about. The only issues is when a new Bazel version is bundled with a `@platforms` not compatible with rules_go, we will have to update rules_go, but I guess that's inevitable anyways.

cpsauer:
Sweet! Thanks so much @linzhp!

PR #3079, opened 2022-03-02
Consider volatile and stable status files separately

### LINKED ISSUE #3078: Consider volatile and stable status files separately

0.30.0

0.22.3

5.0.0

Yes.

Ubuntu 20.04

NA

### What did you do?

Stamped my target with keys only from the volatile status file, but my workspace status script generates both volatile and stable status key values.

No relink triggered if I am not using a stable key from the workspace status files, even if the stable key values change.

### What did you see instead?

A relink is triggered when a stable status key value changes, even when I did not stamp with any stable status key.

siddharthab:
**What type of PR is this?**

Feature

The PR improves local cache hit rate for workspaces that generate stable status variables but may stamp some targets with only volatile status variables.

Fixes #3078

**Other notes for review**

I follow this strategy also in [rules_r](https://github.com/grailbio/rules_r/blob/9df1e588b20004ba4001563ce0490b87edd51624/R/internal/build.bzl#L296).

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -15,6 +15,7 @@
 load(
     "//go/private:common.bzl",
     "as_set",
+    "count_group_matches",
     "has_shared_lib_extension",
 )
 load(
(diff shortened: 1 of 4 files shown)


go/private/actions/link.bzl
```diff
@@ -136,16 +137,27 @@ def emit_link(
...
         builder_args.add("-X", "%s=%s" % (k, v))
+        if go.stamp:
+            stable_vars_count = (count_group_matches(v, "{STABLE_", "}") +
+                                 v.count("{BUILD_EMBED_LABEL}") +
+                                 v.count("{BUILD_USER}") +
+                                 v.count("{BUILD_HOST}"))
+            if stable_vars_count > 0:
+                stamp_x_defs_stable = True
+            if count_group_matches(v, "{", "}") != stable_vars_count:
+                stamp_x_defs_volatile = True
```
sluongng, line 152: I was a bit confused why do we need `count_group_matches` at all and this seems to be overly complicated. But then I think I got the reason:

`x_defs` stamping comes as a key-value pair, something like

```
go_binary(
    name = "cmd",
    srcs = ["main.go"],
    deps = ["//version"],
    x_defs = {"example.com/repo/version.Version": "{STABLE_GIT_COMMIT}"},
)
```

and although most documentation have put value as a variable surrounded with braces `{}`, in truth we do allow multiple variables in 1 value, such as 

```
    x_defs = {"example.com/repo/version.Version": "{STABLE_GIT_COMMIT}_{MY_VOLATILE_VAL}"},
```
is still valid 🤔

This is ultimately following the behavior of the custom go linker wrapper https://sourcegraph.com/github.com/bazelbuild/rules_go@e9a7054ff11a520e3b8aceb76a3ba44bb8da4c94/-/blob/go/tools/builders/link.go#L124-140 which uses the regex `{.+?}`

sluongng (approved):
> The PR improves local cache hit rate for workspaces that generate stable status variables but may stamp some targets with only volatile status variables.

To explain this a bit further, given a workspace with these scenarios:

- Stable status does not change between builds
- Volatile status always change between builds
- `build --stamp` is set for Bazel by default
- Target `//:a` is only stamped with stable variable
- Target `//:b` is only stamped with volatile variable

Previous behavior: both `:a` and `:b` will be invalidate cache between builds as both are getting stamped with a combination of volatile and stable files.

New behavior after this PR: `:a` will NOT be invalidated between builds as it will only get stamped with Stable status and not volatile status.

This LGTM

linzhp (approved):
Tested in Uber's Go monorepo. It works. Thanks for the improvement.

PR #3082, opened 2022-03-08
Resolves #3081: add `analyzer_flags` to nogo config

### LINKED ISSUE #3081: Allow nogo to accept analyzer-specific flags
# Problem

It seems like nogo has no direct means of passing flags to 3rd party analyzers that consume flags. It seems reasonable to be able to pass this information through nogo for each analyzer.

# Solutions

### Config Map

```
"exhaustive": {
    "analyzer_flags": {
        "ignore-enum-members": "_UNSPECIFIED$"
    },
}
```

### Config Array

```
"exhaustive": {
    "analyzer_args": ["-ignore-enum-members": "_UNSPECIFIED$"]
}
```

### nogo Rule Mapping

As suggested in #1867, except since `go_tool_library` is deprecated the arguments will have to be passed through a map or something to the nogo rule itself.

# Notes

Similar ask to #1867 but more straightforward in that no config file needs to be passed in. Though this solution doesn't seem particularly compatible with the use case mentioned in #1867 since the nogo config won't be able to reference Bazel sandboxed data deps, I think the config file use case is more exotic than typical usage patterns of Go analyzers in the wild, many of which consume flags. It would be good to support those common use cases first.

NavneethJayendran:
**What type of PR is this?**

> Feature

Allow nogo config to specify flags to be passed to each analyzer via `analyzer_flags`. The ability to provide flags to each
analyzer is something not currently supported by the driver, which makes it more troublesome to use configurable 
analyzers in a Bazel Go project.

#3081 

**Other notes for review**

Being able to pass in a list of CLI arguments might add greater flexibility, but given the restrictive nature of Go's `FlagSet`
API, it is unlikely that any analyzers in the wild will support positional arguments and the like.

### DIFF
--- a/go/nogo.rst
+++ b/go/nogo.rst
@@ -197,7 +197,7 @@ contain the following key-value pairs:
 +----------------------------+---------------------------------------------------------------------+
 | ``"analyzer_flags"``       | :type:`dictionary, string to string`                                |
 +----------------------------+---------------------------------------------------------------------+
-| Passes on a set of flags as defined by the Go ``flags`` package to the analyzer via the          |
+| Passes on a set of flags as defined by the Go ``flag`` package to the analyzer via the           |
 | ``analysis.Analyzer.Flags`` field. Its keys are the flag names *without* a ``-`` prefix, and its |
 | values are the flag values. nogo will exit with an error upon receiving flags not recognized by  |
 | the analyzer or upon receiving ill-formatted flag values as defined by the corresponding         |
--- a/go/nogo.rst
+++ b/go/nogo.rst
@@ -195,6 +195,14 @@ contain the following key-value pairs:
 | in both ``only_files`` and ``exclude_files``, the analyzer will not emit diagnostics for that    |
 | file.                                                                                            |
 +----------------------------+---------------------------------------------------------------------+
+| ``"analyzer_flags"``       | :type:`dictionary, string to string`                                |
++----------------------------+---------------------------------------------------------------------+
+| Passes on a set of flags as defined by the Go ``flags`` package to the analyzer via the          |
+| ``analysis.Analyzer.Flags`` field. Its keys are the flag names *without* a ``-`` prefix, and its |
+| values are the flag values. nogo will exit with an error upon receiving flags not recognized by  |
+| the analyzer or upon receiving ill-formatted flag values as defined by the corresponding         |
+| ``flag.Value`` specified by the analyzer.                                                        |
++----------------------------+---------------------------------------------------------------------+
 
 Example
 ^^^^^^^
@@ -202,6 +210,8 @@ Example
 The following configuration file configures the analyzers named ``importunsafe``
 and ``unsafedom``. Since the ``loopclosure`` analyzer is not explicitly
 configured, it will emit diagnostics for all Go files built by Bazel.
+``unsafedom`` will receive a flag equivalent to ``-block-unescaped-html=false``
+on a command line driver.
 
 .. code:: json
 
@@ -218,7 +228,10 @@ configured, it will emit diagnostics for all Go files built by Bazel.
         },
         "exclude_files": {
           "src/(third_party|vendor)/.*": "enforce DOM safety requirements only on first-party code"
-        }
+        },
+        "analyzer_flags": {
+            "block-unescaped-html": "false",
+        },
       }
     }
 
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -174,7 +174,17 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
 	for _, a := range analyzers {
 		if cfg, ok := configs[a.Name]; ok {
 			for flagKey, flagVal := range cfg.analyzerFlags {
-				a.Flags.Set(flagKey, flagVal)
+				if err := a.Flags.Set(flagKey, flagVal); err != nil {
+					if strings.HasPrefix(flagKey, "-") {
+						return "", nil, fmt.Errorf(
+							"%s: flag should not begin with '-': %s", a.Name, flagKey)
+					}
+					if flag := a.Flags.Lookup(flagKey); flag == nil {
+						return "", nil, fmt.Errorf("%s: unrecognized flag: %s", a.Name, flagKey)
+					}
+					return "", nil, fmt.Errorf(
+						"%s: invalid value for flag: %s=%s", a.Name, flagKey, flagVal)
+				}
 			}
 		}
 		roots = append(roots, visit(a))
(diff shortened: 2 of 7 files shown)


go/tools/builders/nogo_main.go
```diff
@@ -457,6 +462,9 @@ type config struct {
 	// excludeFiles is a list of regular expressions that match files that an
 	// analyzer will not emit diagnostics for.
 	excludeFiles []*regexp.Regexp
+
+	// analyzerFlags is a map of
+	analyzerFlags map[string]string
```
sluongng, line 467: Nit: missing doc

tests/core/nogo/custom/custom_test.go
```diff
@@ -280,6 +291,48 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+                return nil, nil
+        }
+        if stringSwitch != "yes" {
+                return nil, nil
+        }
+        if intSwitch != 1 {
+                return nil, nil
+        }
+        return nil, errors.New("all switches were set -> fail")
+}
```
sluongng, line 334: Nit: Since we are allowing users to provide the FlagSet as a `map[string]string`, it might worth providing a test to document how nogo would behave when user were to pass some data that is not the correct data type.

For example: Let's have a test where user provided `int-switch: one` to an IntVar flag and expect some error?

tests/core/nogo/custom/custom_test.go
```diff
@@ -427,6 +510,12 @@ func Test(t *testing.T) {
 				// note the cross platform regex :)
 				`.*[\\/]cgo[\\/]examplepkg[\\/]pure_src_with_err_calling_native.go:.*function must not be named Foo \(foofuncname\)`,
 			},
+		}, {
+			desc:        "config_flags_triggering_error",
+			target:      "//:has_errors",
+			wantSuccess: false,
+			config:      "flagger_config.json",
+			includes:    []string{"all switches were set -> fail"},
```
sluongng, line 518: There is some bad copy paste here. Could you please double check?

sluongng (changes requested):
Overall the direction looks good to me but there are some nits and copy pasting errors here and there that need to be fixed.

You would also want to add some documentation in places that mention nogo config json schema such as https://github.com/bazelbuild/rules_go/blob/master/go%2Fnogo.rst

tests/core/nogo/custom/custom_test.go
NavneethJayendran: Ah, I just realized `FlagSet.Set` actually returns an error on failure to parse 🤦 . Let me see what I can do here.

tests/core/nogo/custom/custom_test.go
NavneethJayendran: Sorry what's the bad copy paste here? Is this what's causing the tests to fail on Buildkite?

sluongng: I think your target name is wrong, also wantSuccess value should be changed at least?

tests/core/nogo/custom/custom_test.go
NavneethJayendran: Okay, partly surfacing errors from `FlagSet.Set`. I'm not including the full error text since it's usually not very helpful, but I'm at least surfacing whether the flag is not defined, invalid, or erroneously provided with a leading hyphen. Added a `analyzer_flags` section in the .rst for the nogo config file as well.

tests/core/nogo/custom/custom_test.go
NavneethJayendran: I think I intended on using these values. At any rate, I've separated out these tests into a `flags_test` package to avoid polluting the more intricate tests here.

tests/core/nogo/custom/custom_test.go
FastNav: @sluongng Looks like @NavneethJayendran (AKA me) addressed your review. Mind taking another look?

sluongng (commented):
I have a few nits, but this overall looking really good.

Very nice that CI is also passing 🎉 

Pinging @linzhp @robfig for further reviews

linzhp:
I can take a look later this week or weekend.

go/tools/builders/nogo_main.go
```diff
@@ -172,6 +172,21 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
...
+		if cfg, ok := configs[a.Name]; ok {
+			for flagKey, flagVal := range cfg.analyzerFlags {
+				if err := a.Flags.Set(flagKey, flagVal); err != nil {
+					if strings.HasPrefix(flagKey, "-") {
+						return "", nil, fmt.Errorf(
+							"%s: flag should not begin with '-': %s", a.Name, flagKey)
+					}
+					if flag := a.Flags.Lookup(flagKey); flag == nil {
+						return "", nil, fmt.Errorf("%s: unrecognized flag: %s", a.Name, flagKey)
+					}
```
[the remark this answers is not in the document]
FastNav: I'm more a "better to beg for forgiveness than ask for permission" kind of person, but I'm fine with either here.

go/tools/builders/nogo_main.go
```diff
@@ -172,6 +172,21 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
...
+					return "", nil, fmt.Errorf(
+						"%s: invalid value for flag: %s=%s", a.Name, flagKey, flagVal)
+				}
```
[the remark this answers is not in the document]
FastNav: The message tends to be pretty generic and unhelpful. Maybe it would make sense to include the error if it had something other than that generic message.

go/nogo.rst
```diff
@@ -195,13 +195,23 @@ contain the following key-value pairs:
 | in both ``only_files`` and ``exclude_files``, the analyzer will not emit diagnostics for that    |
 | file.                                                                                            |
 +----------------------------+---------------------------------------------------------------------+
+| ``"analyzer_flags"``       | :type:`dictionary, string to string`                                |
++----------------------------+---------------------------------------------------------------------+
+| Passes on a set of flags as defined by the Go ``flag`` package to the analyzer via the           |
+| ``analysis.Analyzer.Flags`` field. Its keys are the flag names *without* a ``-`` prefix, and its |
+| values are the flag values. nogo will exit with an error upon receiving flags not recognized by  |
```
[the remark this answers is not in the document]
FastNav: I believe there is a requirement [here](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/flag/flag.go;l=862) at flag definition itself. I don't think there is any specific error surfaced by the library when you actually _set_ a flag with a leading `-`. But that's exactly why I'm adding this error message it clarifies that the user should probably try removing the leading `-`.

linzhp:
Thanks for working on this. It's a pretty solid implementation. Thanks for reviewing too @sluongng

go/tools/builders/nogo_main.go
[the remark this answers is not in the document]
FastNav: On second thought, the message helps distinguish between range errors and other errors at least, so it might be worth keeping.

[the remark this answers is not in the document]
linzhp: I don't mind taking another pull request if you want to log the error.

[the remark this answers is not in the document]
FastNav: Ah sorry, I already added it. I just forgot to declare my change of heart here. I'll mark this thread as resolved.

PR #3083, opened 2022-03-08
Fix gopackagesdriver for Go 1.18 by replicating stdlib

### LINKED ISSUE #3080: Go package driver doesn't work with Go 1.18
Current package driver calls `builder stdliblist` to get information for std libraries. `stdliblist` calls `go list`:

https://github.com/bazelbuild/rules_go/blob/e9a7054ff11a520e3b8aceb76a3ba44bb8da4c94/go/tools/builders/stdliblist.go#L148-L149

This caused two problems:

1. essentially, std libraries are compiled twice, one by rules_go and one by `go list`
2. when `stdliblist` calls `go list`, it is in Bazel's sandbox, where all Go files of std libraries are symlinks. One of the Go files in 1.18 embeds a [binary file](https://github.com/golang/go/blob/bf97c99b62fe7d6652cc8c807dbc91998d488a01/src/crypto/elliptic/p256_asm.go#L24), which would break when the binary file is a symlink, leading to errors like:

```
ERROR: /private/var/tmp/_bazel_zplin/db67431f618bf23018777556a2edc2eb/external/io_bazel_rules_go/BUILD.bazel:44:7: GoStdlibList external/io_bazel_rules_go/stdlib_/stdlib.pkg.json failed: (Exit 1): builder failed: error executing command 
  (cd /private/var/tmp/_bazel_zplin/db67431f618bf23018777556a2edc2eb/sandbox/darwin-sandbox/7496/execroot/__main__ && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=12.1 \
    CGO_ENABLED=1 \
    GOARCH=arm64 \
    GOOS=darwin \
    GOPATH='' \
    GOROOT=external/go_sdk \
    GOROOT_FINAL=GOROOT \
    PATH=external/local_config_cc:/bin:/usr/bin \
    XCODE_VERSION_OVERRIDE=13.2.1.13C100 \
    ZERO_AR_DATE=1 \
  bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/go_sdk/builder stdliblist -sdk external/go_sdk -installsuffix darwin_arm64 -out bazel-out/darwin_arm64-fastbuild/bin/external/io_bazel_rules_go/stdlib_/stdlib.pkg.json)
# Configuration: 3082a7bf331eaf230f27e04497eaecbe1891da6c723270a0d9c69754da9505e4
# Execution platform: @local_config_platform//:host

external/go_sdk/src/crypto/elliptic/p256_asm.go:24:12: pattern p256_asm_table.bin: cannot embed irregular file p256_asm_table.bin
```

0.30

5.0.0

### What version of Go are you using?
Go 1.18rc1

Yes

macOS arm64

abhinav:
**What type of PR is this?**

Bug fix

The gopackagesdriver relies on the "stdliblist" action provided by rules_go.
The stdliblist action fails on Go 1.18 with the following error:

```
external/go_sdk/src/crypto/elliptic/p256_asm.go:24:12: pattern p256_asm_table.bin: cannot embed irregular file p256_asm_table.bin
```

We see this failure because Bazel builds a symlink tree of the GOROOT to run
`go list` with. However, since [CL 380475][1], the Go standard library uses
`go:embed` to embed a file in `crypto/elliptic`, but `go:embed` does not
support symlinks.

[1]: https://go.dev/cl/380475

Fix this by having stdliblist copy the relevant portions of the GOROOT to
run `go list` with. This matches [what the stdlib action does][2].

[2]: https://github.com/bazelbuild/rules_go/blob/e9a7054ff11a520e3b8aceb76a3ba44bb8da4c94/go/tools/builders/stdlib.go#L54-L57

Fixes #3080

**Other notes for review**

I've updated the existing stdlib test to also depend on the output of
stdliblist, and verified that that fails with Go 1.18rc1 without this change.

### DIFF
--- a/go/tools/builders/stdliblist.go
+++ b/go/tools/builders/stdliblist.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"flag"
+	"fmt"
 	"go/build"
 	"os"
 	"path/filepath"
@@ -145,6 +146,42 @@ func flatPackageForStd(execRoot string, pkg *goListPackage) *flatPackage {
 	return newPkg
 }
 
+// In Go 1.18, the standard library started using go:embed directives.
+// When Bazel runs this action, it does so inside a sandbox where GOROOT points
+// to an external/go_sdk directory that contains a symlink farm of all files in
+// the Go SDK.
+// If we run "go list" with that GOROOT, this action will fail because those
+// go:embed directives will refuse to include the symlinks in the sandbox.
+//
+// To work around this, cloneRoot creates a copy of external/go_sdk into a new
+// directory "root" while retaining its path relative to the root directory.
+// So "$OUTPUT_BASE/external/go_sdk" becomes
+// "$OUTPUT_BASE/root/external/go_sdk".
+// This ensures that file paths in the generated JSON are still valid.
+//
+// cloneRoot returns the new root directory and the new GOROOT we should run
+// under.
+func cloneRoot(execRoot, goroot string) (newRoot string, newGoroot string, err error) {
+	relativeGoroot, err := filepath.Rel(abs(execRoot), abs(goroot))
+	if err != nil {
+		// GOROOT has to be a subdirectory of execRoot.
+		// Otherwise we're breaking the sandbox.
+		return "", "", fmt.Errorf("GOROOT %q is not relative to execution root %q: %v", goroot, execRoot)
+	}
+
+	newRoot = filepath.Join(execRoot, "root")
+	newGoroot = filepath.Join(newRoot, relativeGoroot)
+	if err := os.MkdirAll(newGoroot, 01755); err != nil {
+		return "", "", err
+	}
+
+	if err := replicate(goroot, newGoroot, replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
+		return "", "", err
+	}
+
+	return newRoot, newGoroot, nil
+}
+
 // stdliblist runs `go list -json` on the standard library and saves it to a file.
 func stdliblist(args []string) error {
 	// process the args
@@ -158,8 +195,8 @@ func stdliblist(args []string) error {
 		return err
 	}
 
-	execRoot := abs(".")
-	if err := replicate(os.Getenv("GOROOT"), execRoot, replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
+	execRoot, goroot, err := cloneRoot(".", os.Getenv("GOROOT"))
+	if err != nil {
 		return err
 	}
 
@@ -169,7 +206,7 @@ func stdliblist(args []string) error {
 		absPaths = append(absPaths, abs(path))
 	}
 	os.Setenv("PATH", strings.Join(absPaths, string(os.PathListSeparator)))
-	os.Setenv("GOROOT", execRoot)
+	os.Setenv("GOROOT", goroot)
 	// Make sure we have an absolute path to the C compiler.
 	// TODO(#1357): also take absolute paths of includes and other paths in flags.
 	os.Setenv("CC", abs(os.Getenv("CC")))
--- a/go/tools/builders/stdliblist.go
+++ b/go/tools/builders/stdliblist.go
@@ -158,18 +158,21 @@ func stdliblist(args []string) error {
 		return err
 	}
 
+	execRoot := abs(".")
+	if err := replicate(os.Getenv("GOROOT"), execRoot, replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
+		return err
+	}
+
 	// Ensure paths are absolute.
 	absPaths := []string{}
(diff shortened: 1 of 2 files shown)


abhinav:
This isn't quite right, I think. Investigating and switching to WIP.

abhinav:
Fixed. The solution is a bit janky: we have to create a copy of external/go_sdk under a `$newdir/external/go_sdk` and flatten the paths relative to `$newdir` to get the same output. This isn't ideal, but I'm not sure how else to get Bazel to create copies of the file instead of symlinks.

go/tools/builders/stdliblist.go
```diff
@@ -145,6 +146,42 @@ func flatPackageForStd(execRoot string, pkg *goListPackage) *flatPackage {
...
+// To work around this, cloneRoot creates a copy of external/go_sdk into a new
+// directory "root" while retaining its path relative to the root directory.
+// So "$OUTPUT_BASE/external/go_sdk" becomes
+// "$OUTPUT_BASE/root/external/go_sdk".
+// This ensures that file paths in the generated JSON are still valid.
+//
+// cloneRoot returns the new root directory and the new GOROOT we should run
+// under.
+func cloneRoot(execRoot, goroot string) (newRoot string, newGoroot string, err error) {
+	relativeGoroot, err := filepath.Rel(abs(execRoot), abs(goroot))
```
linzhp, line 165: If you add `args.add("-sdk", go.sdk.root_file.dirname)` to here: https://github.com/bazelbuild/rules_go/blob/d606ba2c123826f153bbd037cf46e80b5eb8dc3e/go/private/actions/stdlib.bzl#L56-L57

Then you can get a relative path of goroot from execRoot with `goenv.sdk`.

go/tools/builders/stdliblist.go
```diff
@@ -158,19 +195,22 @@ func stdliblist(args []string) error {
 		return err
 	}
 
+	execRoot, goroot, err := cloneRoot(".", os.Getenv("GOROOT"))
```
linzhp, line 198: "execRoot" is a special name in Bazel. It means the "The working directory for all actions" (https://bazel.build/docs/output_directories). Because you copied the symlink tree to to a different place, it's no longer the actual execRoot. Please rename it to something like "newBase" or "cloneBase".  There are several variables/arguments in this file called "execRoot". After this pull request, they are no longer the actual execroot. Please also renaming all of them.

go/tools/builders/stdliblist.go
```diff
@@ -145,6 +146,42 @@ func flatPackageForStd(execRoot string, pkg *goListPackage) *flatPackage {
...
+		return "", "", fmt.Errorf("GOROOT %q is not relative to execution root %q: %v", goroot, execRoot)
+	}
+
+	newRoot = filepath.Join(execRoot, "root")
+	newGoroot = filepath.Join(newRoot, relativeGoroot)
+	if err := os.MkdirAll(newGoroot, 01755); err != nil {
+		return "", "", err
+	}
+
+	if err := replicate(goroot, newGoroot, replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
```
linzhp, line 178: Do you need to replicate `pkg/tool`?

go/tools/builders/stdliblist.go
```diff
@@ -145,6 +146,42 @@ func flatPackageForStd(execRoot string, pkg *goListPackage) *flatPackage {
...
+// under.
+func cloneRoot(execRoot, goroot string) (newRoot string, newGoroot string, err error) {
+	relativeGoroot, err := filepath.Rel(abs(execRoot), abs(goroot))
+	if err != nil {
+		// GOROOT has to be a subdirectory of execRoot.
+		// Otherwise we're breaking the sandbox.
+		return "", "", fmt.Errorf("GOROOT %q is not relative to execution root %q: %v", goroot, execRoot)
+	}
+
+	newRoot = filepath.Join(execRoot, "root")
```
linzhp, line 172: I am a bit nervous about writing things to execRoot. It's supposed to contain a symlink tree intended to be read-only. This may have unexpected result if there happen to be a `root` directory under `execRoot`. Can you copy it to `ioutil.TempDir`?

linzhp (changes requested):

linzhp:
I am also thinking about how to write tests to cover `stdliblist`. The `go_test` will need a `data = "@go_sdk//:files"` and access it using: `filepath.Join(os.Getenv("TEST_SRCDIR"), "go_sdk")`. So there is no longer an "external/go_sdk" section in the abs path. In order for `stdliblist` to work in such test, you will also need to replace [this function](https://github.com/bazelbuild/rules_go/blob/master/go/tools/builders/stdliblist.go#L112-L115) with something like `filepath.Join("__BAZEL_OUTPUT_BASE__/external", p[:strings.Index("/go_sdk/")])`

linzhp:
Superseded by #3157

PR #3096, opened 2022-03-28
Support generated embedsrcs

hugelgupf:
Just like for srcs, rules_go should support generated embedsrcs.

I'm currently running into a situation in which I've generated embed files, but I cannot pass them into `go.new_library` (because those get ignored) or into the attrs for `library_to_source` -- because `library_to_source` attr.embedsrcs is expected to be of type list of Target, and when you generate them, you generate a list of File. There isn't really a way to generate the targets, which is why `srcs` does the same thing of allowing `go.new_library` to specify a list of File and library_to_source attrs to specify targets and combines the two. This extends that to `embedsrcs`.


linzhp:
Can you add tests to `tests/core/go_library` to cover this feature?

PR #3098, opened 2022-03-30
Use output dir for empty packages to be hermetic

joeljeske:
**What type of PR is this?**

> Bug fix

Some packages are not deterministic, rather they have different content hashes when rebuilt. This occurs when rules_go creates an `_empty.go` file to satisfy the compiler. In those cases, the empty file is created in the OS temp dir, which changes across compiler invocations. The path of this file changes the content hash of the resulting package, thus breaking caches and downstream builds. 

This PR changes the location of this tmp file to the output path of the package, such that it is the same output path across compiler invocations. 

Reproduction steps: 

```sh
bazel build //tests/core/go_library:empty 
md5 bazel-bin/tests/core/go_library/empty.a

rm -f  bazel-bin/tests/core/go_library/empty.a

bazel build //tests/core/go_library:empty 
md5 bazel-bin/tests/core/go_library/empty.a
```

**Other notes for review**

I noted a broad use of tmpdir or workdir within rules_go, and this issue may be prevalent in other conditions. If determinism is a priority, perhaps an audit of tmp dir uses may be warranted.

~~It would be useful to have a test that validations reproducibility of this kind, but I'm not aware of a good way to structure that type of test.~~ Reproducibility test has been added

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -178,7 +178,12 @@ func compileArchive(
 	defer cleanup()
 
 	if len(srcs.goSrcs) == 0 {
-		emptyPath := filepath.Join(workDir, "_empty.go")
+		// We need to run the compiler to create a valid archive, even if there's
+		// nothing in it. GoPack will complain if we try to add assembly or cgo
+		// objects.
+		// _empty.go needs to be in a deterministic location (not tmpdir) in order
+		// to ensure deterministic output
+		emptyPath := filepath.Join(filepath.Dir(outPath), "_empty.go")
 		if err := ioutil.WriteFile(emptyPath, []byte("package empty\n"), 0666); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


go/tools/builders/compilepkg.go
```diff
@@ -178,7 +178,12 @@ func compileArchive(
 	defer cleanup()
 
 	if len(srcs.goSrcs) == 0 {
-		emptyPath := filepath.Join(workDir, "_empty.go")
+		// We need to run the compiler to create a valid archive, even if there's
+		// nothing in it. GoPack will complain if we try to add assembly or cgo
+		// objects.
+		// _empty.go needs to be in a deterministic location (not tmpdir) in order
+		// to ensure deterministic output
+		emptyPath := filepath.Join(filepath.Dir(outPath), "_empty.go")
```
joeljeske, line 186: This strategy is already being used here https://github.com/bazelbuild/rules_go/blob/master/go/tools/builders/compile.go#L98

joeljeske:
Fantastic, @achew22, thanks for the tip. I had not seen those tests. I added a simple empty lib test that fails at its commit.

PR #3099, opened 2022-03-31
Remove legacy proto code for Bazel < 0.21

cpsauer:
Hey wonderful Bazel Go folks,

I was browsing the codebase and noticed both that the minimum Bazel version for rules_go is 4.2.1 (in the release notes) and also saw a section of code saying it should be deleted when Bazel minimum version >= 0.21.0. So I figured I should toss up a PR doing just that!

Looks like @jayconrod wrote the code originally. Tagging him in case he wants to weigh in with context. Maybe there's more dead code that should be deleted. But it looks like he might be busy and hasn't committed in a while, so I'd propose that we probably shouldn't wait synchronously on his reply.

Thanks for reading!
Chris
(ex-Googler)

### DIFF
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -143,21 +143,6 @@ def proto_path(src, proto):
     Returns:
         An import path string.
     """
-    if not hasattr(proto, "proto_source_root"):
-        # Legacy path. Remove when Bazel minimum version >= 0.21.0.
-        path = src.path
-        root = src.root.path
-        ws = src.owner.workspace_root
-        if path.startswith(root):
-            path = path[len(root):]
-        if path.startswith("/"):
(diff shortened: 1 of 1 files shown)


achew22:
Seems like a good fix to me. I will merge this on April 1 if no one else objects. If I don't actually do it on that Friday, feel free to ping the PR on Monday and I'll do it. Thanks!

cpsauer:
Sweet! Thanks for a fast reply. Glad it looks to be helpful :)

cpsauer:
Pinging because I happened to think of this and it's past Friday :)

achew22:
Thanks for the reminder and the PR!

cpsauer:
Thank you for being so fast and so great!

PR #3102, opened 2022-04-02
Adding gomock rule

### LINKED ISSUE #3101: Hosting bazel_gomock in rules_go
@jmhodges created [bazel_gomock](https://github.com/jmhodges/bazel_gomock), which is Bazel rules to generate mocks using [golang/mock](https://github.com/golang/mock). However, the project is inactive, with all pull requests from Uber stuck for years. We also have more internal patches for it. These pull requests and patches significantly improve the performance. We would like to contribute these improvement to the open source community.

Can rules_go host the Bazel rules for gomock?

linzhp:
**What type of PR is this?**
Feature

This copies https://github.com/jmhodges/bazel_gomock/blob/4f2ee840432b1a08ccc46ee4f2c1f5a2bad8fade/gomock.bzl with some formatting and the original license as comments. DO NOT USE this version, as there will be some significant changes in the later pull requests.

Fixes #3101

**Other notes for review**

### DIFF
--- a/extras/gomock.bzl
+++ b/extras/gomock.bzl
@@ -1,0 +1,353 @@
+# The MIT License (MIT)
+# Copyright © 2018 Jeff Hodges <jeff@somethingsimilar.com>
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the “Software”), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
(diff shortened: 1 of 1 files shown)


achew22 (changes requested):
If you want to move this into the main repo, you should commit as Jeff so that authorship is preserved and the CLA robot can make sure authorship is correct.

AKA, please don't submit this as is. It will cause a lot of problems and I want to make 1000% sure that Jeff is happy with it. He's the original author so gets to call the shots.

If Jeff doesn't reply or is otherwise unavailable, I think a rewrite of this rule is possible and we could consider merging it in. It might also be worth considering a fork off your own account/corporate account?

linzhp:
I don't think @jmhodges is available, otherwise I would have merged my 4 pull requests in the original repo years ago.

I can explore the option of open source Uber's internal fork, which is a 90% rewrite of the original rules.

linzhp:
@achew22 I changed the author of the commit to be @jmhodges. Can you take another look?

achew22 (approved):

achew22:
Thanks @linzhp for doing that, and thank you @jmhodges for your contribution!

PR #3108, opened 2022-04-08
Trim transitioned Go settings on non-Go dependencies

### LINKED ISSUE #2999: `go_proto_compiler` causes `protoc` rebuilds by enabling `bootstrap_nogo`

### LINKED ISSUE #3120: Genrules in Bazel 6 occasionally try to run shared libraries built by go_binary

0.31.0

None

6.0.0-pre.20220407.1

No, this is a behavior change in Bazel 6, which doesn't occur in Bazel 5, see https://github.com/bazelbuild/bazel/issues/14626.

macOS 12.3.1, ARM

No

### What did you do?

A not very minimal recipe is described in https://github.com/bazelbuild/bazel/issues/14626#issue-1111494401.

Build successful

### What did you see instead?

Build failed as described in https://github.com/bazelbuild/bazel/issues/14626.

https://github.com/bazelbuild/bazel/issues/14626 has lots of information including a root-cause analysis and an open PR.

fmeum:
**What type of PR is this?**

Bug fix

By resetting Go-specific settings changed by go_transition to their previous values when crossing a non-deps dependency edge, e.g. srcs or data, dependencies through those edges are not affected by the change in Go settings. This has two advantages:

* Correctness: Previously, if a go_binary with `linkmode = "c-archive"`  used another go_binary to generate srcs, that go_binary would also be built as a c-archive and thus fail to run during the build. This was observed in https://github.com/bazelbuild/bazel/issues/14626.
* Performance: With the new Bazel flag `--experimental_output_directory_naming_scheme=diff_against_baseline`, transitions can return to the top-level configuration since the Starlark settings that were affected by a transition at some point during the build are no longer tracked explicitly in a setting, but computed as a configuration diff. Resetting the configuration for non-deps dependencies thus prevents redundant rebuilds of non-Go deps caused by changes in Go settings, achieving a version of "configuration trimming" for Go.

Fixes #2999
Fixes #3120
Fixes https://github.com/bazelbuild/bazel/issues/14626

**Other notes for review**

This is based on https://github.com/bazelbuild/rules_go/pull/3005 and completes the work started there. #3005 can be reviewed and merged first if preferred.

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -420,12 +420,22 @@ def _non_go_transition_impl(settings, attr):
     for key, original_key in _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.items():
         original_value = settings[original_key]
         if original_value:
-            # Reset to the original value and clear it.
+            # Reset to the original value of the setting before go_transition.
             new_settings[key] = json.decode(original_value)
-            new_settings[original_key] = ""
         else:
             new_settings[key] = settings[key]
-            new_settings[original_key] = settings[original_key]
+
+        # Reset the value of the helper setting to its default for two reasons:
+        # 1. Performance: This ensures that the Go settings of non-Go
+        #    dependencies have the same values as before the go_transition,
+        #    which can prevent unnecessary rebuilds caused by configuration
+        #    changes.
+        # 2. Correctness in edge cases: If there is a path in the build graph
+        #    from a go_binary's non-Go dependency to a go_library that does not
+        #    pass through another go_binary (e.g., through a custom rule
+        #    replacement for go_binary), this transition could be applied again
+        #    and cause incorrect Go setting values.
+        new_settings[original_key] = ""
 
     return new_settings
 
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -37,18 +37,6 @@ load(
     "platform_from_crosstool",
 )
 
-# A list of rules_go settings that are possibly set by go_transition.
-# Keep their package name in sync with the implementation of
-# _original_setting_key.
-TRANSITIONED_GO_SETTING_KEYS = [
-    "@io_bazel_rules_go//go/config:static",
-    "@io_bazel_rules_go//go/config:msan",
-    "@io_bazel_rules_go//go/config:race",
-    "@io_bazel_rules_go//go/config:pure",
-    "@io_bazel_rules_go//go/config:linkmode",
-    "@io_bazel_rules_go//go/config:tags",
-]
-
 def filter_transition_label(label):
     """Transforms transition labels for the current workspace.
 
@@ -66,13 +54,31 @@ def filter_transition_label(label):
     else:
         return str(Label(label))
 
+# A list of rules_go settings that are possibly set by go_transition.
+# Keep their package name in sync with the implementation of
+# _original_setting_key.
+TRANSITIONED_GO_SETTING_KEYS = [
+    filter_transition_label(key)
+    for key in [
+        "//go/config:static",
+        "//go/config:msan",
+        "//go/config:race",
+        "//go/config:pure",
+        "//go/config:linkmode",
+        "//go/config:tags",
+    ]
+]
+
 def _original_setting_key(key):
     if not "//go/config:" in key:
-        return None
+        fail("_original_setting_key currently assumes that all Go settings live under //go/config, got: " + key)
     name = key.split(":")[1]
-    return filter_transition_label("@io_bazel_rules_go//go/private/rules:original_" + name)
+    return filter_transition_label("//go/private/rules:original_" + name)
 
-_ORIGINAL_SETTING_KEYS = [_original_setting_key(setting) for setting in TRANSITIONED_GO_SETTING_KEYS]
+_SETTING_KEY_TO_ORIGINAL_SETTING_KEY = {
+    setting: _original_setting_key(setting)
+    for setting in TRANSITIONED_GO_SETTING_KEYS
+}
 
 def go_transition_wrapper(kind, transition_kind, name, **kwargs):
     """Wrapper for rules that may use transitions.
@@ -197,23 +203,22 @@ def _go_transition_impl(settings, attr):
         linkmode_label = filter_transition_label("@io_bazel_rules_go//go/config:linkmode")
         settings[linkmode_label] = linkmode
 
-    for key, value in settings.items():
-        original_key = _original_setting_key(key)
-        if not original_key:
-            continue
+    for key, original_key in _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.items():
+        old_value = original_settings[key]
+        value = settings[key]
 
         # If the outgoing configuration would differ from the incoming one in a
         # value, record the old value in the special original_* key so that the
         # real setting can be reset to this value before the new configuration
         # would cross a non-deps dependency edge.
-        if value != original_settings[key]:
+        if value != old_value:
             # Encoding as JSON makes it possible to embed settings of arbitrary
             # types (currently bool, string and string_list) into a single type
             # of setting (string) with the information preserved whether the
             # original setting wasn't set explicitly (empty string) or was set
             # explicitly to its default  (always a non-empty string with JSON
             # encoding, e.g. "\"\"" or "[]").
-            settings[original_key] = json.encode(original_settings[key])
+            settings[original_key] = json.encode(old_value)
         else:
             settings[original_key] = ""
 
@@ -250,7 +255,7 @@ go_transition = transition(
     ] + TRANSITIONED_GO_SETTING_KEYS],
     outputs = [filter_transition_label(label) for label in [
         "//command_line_option:platforms",
-    ] + TRANSITIONED_GO_SETTING_KEYS + _ORIGINAL_SETTING_KEYS],
+    ] + TRANSITIONED_GO_SETTING_KEYS + _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.values()],
 )
 
 _common_reset_transition_dict = dict({
@@ -262,7 +267,7 @@ _common_reset_transition_dict = dict({
     "@io_bazel_rules_go//go/config:debug": False,
     "@io_bazel_rules_go//go/config:linkmode": LINKMODE_NORMAL,
     "@io_bazel_rules_go//go/config:tags": [],
-}, **{setting: "" for setting in _ORIGINAL_SETTING_KEYS})
+}, **{setting: "" for setting in _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.values()})
 
 _reset_transition_dict = dict(_common_reset_transition_dict, **{
     "@io_bazel_rules_go//go/private:bootstrap_nogo": True,
@@ -412,8 +417,7 @@ def _non_go_transition_impl(settings, attr):
     settings such as e.g. race instrumentation.
     """
     new_settings = {}
-    for key in TRANSITIONED_GO_SETTING_KEYS:
-        original_key = _original_setting_key(key)
+    for key, original_key in _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.items():
         original_value = settings[original_key]
         if original_value:
             # Reset to the original value and clear it.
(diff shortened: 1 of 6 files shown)


achew22:
@linzhp can you try this out in your big repo and see if it works as well?

linzhp:
It passed in Uber, but let's merge #3005 first

go/private/rules/transition.bzl
```diff
@@ -173,6 +195,23 @@ def _go_transition_impl(settings, attr):
...
+
+        # If the outgoing configuration would differ from the incoming one in a
+        # value, record the old value in the special original_* setting so that
+        # the real setting can be reset to this value before the new
+        # configuration would cross a non-deps dependency edge.
+        if value != original_settings[setting]:
+            # Encoding as JSON makes it possible to embed settings of arbitrary
+            # types (currently bool, string and string_list) into a string
+            # setting.
+            settings[original_value_setting] = json.encode(original_settings[setting])
```
linzhp, line 211: Why can we not using the original type like:
```suggestion
            settings[original_value_setting] = original_settings[setting]
```

go/private/rules/transition.bzl
```diff
@@ -54,6 +64,14 @@ def filter_transition_label(label):
     else:
         return str(Label(label))
 
+def _original_value_setting(setting):
```
linzhp, line 67:
```suggestion
def _original_setting_key(setting):
```
These are keys to store original settings, right?

go/private/rules/transition.bzl
```diff
@@ -173,6 +195,23 @@ def _go_transition_impl(settings, attr):
         linkmode_label = filter_transition_label("@io_bazel_rules_go//go/config:linkmode")
         settings[linkmode_label] = linkmode
 
+    for setting, value in settings.items():
+        original_value_setting = _original_value_setting(setting)
+        if not original_value_setting:
+            continue
```
linzhp, line 201:
```suggestion
        if not original_value_setting or value == original_settings[setting]:
            continue
```

This would reduce the nesting for the logic below.

go/private/rules/transition.bzl
```diff
@@ -173,6 +195,23 @@ def _go_transition_impl(settings, attr):
         linkmode_label = filter_transition_label("@io_bazel_rules_go//go/config:linkmode")
         settings[linkmode_label] = linkmode
 
+    for setting, value in settings.items():
```
linzhp, line 198:
```suggestion
    for key, value in settings.items():
```

go/private/rules/transition.bzl
```diff
@@ -173,6 +195,23 @@ def _go_transition_impl(settings, attr):
...
+        original_value_setting = _original_value_setting(setting)
```
linzhp, line 199:
```suggestion
        original_key = _original_value_setting(setting)
```

go/private/rules/transition.bzl
```diff
@@ -173,6 +195,23 @@ def _go_transition_impl(settings, attr):
...
+        if not original_value_setting:
+            continue
+
+        # If the outgoing configuration would differ from the incoming one in a
+        # value, record the old value in the special original_* setting so that
```
linzhp, line 204:
```suggestion
        # value, record the old value in the special original_* key so that
```

go/private/rules/transition.bzl
```diff
@@ -54,6 +64,14 @@ def filter_transition_label(label):
     else:
         return str(Label(label))
 
+def _original_value_setting(setting):
+    if not "//go/config:" in setting:
+        return None
+    name = setting.split(":")[1]
+    return filter_transition_label("@io_bazel_rules_go//go/private/rules:original_" + name)
+
+_ORIGINAL_VALUE_SETTINGS = [_original_value_setting(setting) for setting in POTENTIALLY_TRANSITIONED_SETTINGS]
```
linzhp, line 73:
```suggestion
_ORIGINAL_SETTING_KEYS = [_original_value_setting(setting) for setting in POTENTIALLY_TRANSITIONED_SETTINGS]
```

Also, because `_original_value_setting` may return `None`. You need to filter them out

go/private/rules/transition.bzl
```diff
@@ -366,6 +405,39 @@ go_transition.
...
+
+    Examples: This transition is applied to the 'data' attribute of go_binary so
+    that other Go binaries used at runtime aren't affected by a non-standard
+    link mode set on the go_binary target, but still use the same top-level
+    settings such as e.g. race instrumentation.
+    """
+    new_settings = {}
+    for setting in POTENTIALLY_TRANSITIONED_SETTINGS:
+        original_setting = _original_value_setting(setting)
+        original_value = settings[original_setting]
```
linzhp, line 424: `original_setting` can be `None` here

linzhp (changes requested):

go/private/rules/transition.bzl
```diff
@@ -54,6 +64,14 @@ def filter_transition_label(label):
     else:
         return str(Label(label))
 
+def _original_value_setting(setting):
+    if not "//go/config:" in setting:
+        return None
+    name = setting.split(":")[1]
+    return filter_transition_label("@io_bazel_rules_go//go/private/rules:original_" + name)
```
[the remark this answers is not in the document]
fmeum: It converts `@io_bazel_rules_go//go/config:race` to `//go/config:race` when `io_bazel_rules_go` is the main repo and returns it unchanged otherwise. It may not be necessary here since the "original_" settings are never reference on the command-line, but using it for some settings and not others seemed even more confusing to me - it would also affect the way transition inputs/outputs are declared below.

The bug is fixed, but the full fix is only available in Bazel 5+. Once the minimum version has been updated to 5.0.0, `filter_transition_label` can be dropped.

go/private/rules/transition.bzl
```diff
@@ -366,6 +398,39 @@ go_transition.
...
+
+    Examples: This transition is applied to the 'data' attribute of go_binary so
+    that other Go binaries used at runtime aren't affected by a non-standard
+    link mode set on the go_binary target, but still use the same top-level
+    settings such as e.g. race instrumentation.
+    """
+    new_settings = {}
+    for key in TRANSITIONED_GO_SETTING_KEYS:
+        original_key = _original_setting_key(key)
+        original_value = settings[original_key]
```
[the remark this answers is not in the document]
fmeum: I could implement that change, but could you explain why you think it's needed? If `_original_setting_key(key)` is `None` for `key` in `TRANSITIONED_GO_SETTING_KEYS`, then that's a bug in this file. I could possibly see adding a check with a `fail`, but a silent `continue` would - imo - just make these bugs harder to spot.

go/private/rules/transition.bzl
```diff
@@ -173,6 +197,26 @@ def _go_transition_impl(settings, attr):
...
+        if value != original_settings[key]:
+            # Encoding as JSON makes it possible to embed settings of arbitrary
+            # types (currently bool, string and string_list) into a single type
+            # of setting (string) with the information preserved whether the
+            # original setting wasn't set explicitly (empty string) or was set
+            # explicitly to its default  (always a non-empty string with JSON
+            # encoding, e.g. "\"\"" or "[]").
+            settings[original_key] = json.encode(original_settings[key])
+        else:
+            settings[original_key] = ""
```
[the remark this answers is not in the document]
linzhp: Have you tried `None`? According to the [official doc](https://bazel.build/rules/config#defining):
> settings is a dictionary `{String:Object}`

go/private/rules/transition.bzl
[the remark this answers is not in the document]
linzhp: yeah, adding a `fail` with more debuggable message is better. Because `_original_setting_key(key)` can return `None`, it's better to check for that when `None` is not acceptable.

go/private/rules/transition.bzl
```diff
@@ -366,6 +398,39 @@ go_transition.
...
+        else:
+            new_settings[key] = settings[key]
+            new_settings[original_key] = settings[original_key]
+
+    return new_settings
+
+non_go_transition = transition(
+    implementation = _non_go_transition_impl,
+    inputs = TRANSITIONED_GO_SETTING_KEYS + _ORIGINAL_SETTING_KEYS,
+    outputs = TRANSITIONED_GO_SETTING_KEYS + _ORIGINAL_SETTING_KEYS,
```
linzhp, line 431: Do `outputs` need to include `_ORIGINAL_SETTING_KEYS`? After resetting to the original values, do we still need the values for `_ORIGINAL_SETTING_KEYS`?

go/private/rules/transition.bzl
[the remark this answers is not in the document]
fmeum: I added a commit that pushed the conversion into the construction of a dict that contains the mapping and added a fail in the conversion function. What do you think?

go/private/rules/transition.bzl
```diff
@@ -366,6 +400,38 @@ go_transition.
...
+        else:
+            new_settings[key] = settings[key]
+            new_settings[original_key] = settings[original_key]
+
+    return new_settings
+
+non_go_transition = transition(
+    implementation = _non_go_transition_impl,
+    inputs = TRANSITIONED_GO_SETTING_KEYS + _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.values(),
+    outputs = TRANSITIONED_GO_SETTING_KEYS + _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.values(),
```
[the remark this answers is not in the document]
fmeum: By not resetting all settings to their command-line defaults, non-Go dependencies of a transitioned `go_binary` would still differ from top-level targets in the configuration: They would have the top-level config values in their `original_*` settings, whereas top-level (untransitioned) targets do not. This would still solve the correctness issue linked in the PR description, but would not address the performance problem.

go/private/rules/transition.bzl
```diff
@@ -366,6 +403,38 @@ go_transition.
...
+    new_settings = {}
+    for key, original_key in _SETTING_KEY_TO_ORIGINAL_SETTING_KEY.items():
+        original_value = settings[original_key]
+        if original_value:
+            # Reset to the original value and clear it.
+            new_settings[key] = json.decode(original_value)
+            new_settings[original_key] = ""
+        else:
+            new_settings[key] = settings[key]
+            new_settings[original_key] = settings[original_key]
```
linzhp, line 428: How about:
```suggestion
        if original_value:
            new_settings[key] = json.decode(original_value)            
        else:
            new_settings[key] = settings[key]

        new_settings[original_key] = ""
```
because we don't need the backed up original value later

tests/core/transition/hermeticity_test.go
```diff
@@ -123,12 +210,15 @@ func assertDependsCleanlyOnWithFlags(t *testing.T, targetA, targetB string, flag
 	cqueryOut := string(bytes.TrimSpace(out))
 	configHashes := extractConfigHashes(t, cqueryOut)
 	if len(configHashes) != 1 {
-		t.Fatalf(
-			"%s depends on %s in multiple configs with these differences in rules_go options: %s",
-			targetA,
-			targetB,
-			strings.Join(getGoOptions(t, configHashes...), "\n"),
-		)
+		differingGoOptions := getGoOptions(t, configHashes...)
```
linzhp, line 213: Also, in `getGoOptions`, can you run `bazel config --output=json` and parse the json instead?

go/private/rules/transition.bzl
fmeum: Good point, I also added an extensive comment since I remembered another reason for resetting the settings.

tests/core/transition/hermeticity_test.go
```diff
@@ -123,12 +210,15 @@ func assertDependsCleanlyOnWithFlags(t *testing.T, targetA, targetB string, flag
 	cqueryOut := string(bytes.TrimSpace(out))
 	configHashes := extractConfigHashes(t, cqueryOut)
```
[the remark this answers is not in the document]
fmeum: Good point, done. PTAL

linzhp (approved):
Tested again on Uber's Go monorepo: passed

fmeum:
@ripatel-jump Are you sure that this is a regression in 0.38.0? I would think that this has been the behaviour for a long time. Do you happen to have a reproducer I can test this with?

PR #3112, opened 2022-04-11
Use edge transitions not self transitions

### LINKED ISSUE #3103: Can't configure static differently for different platforms

0.31.0

5.1.0

Yes

macOS x86-64

### What did you do?

```starlark
go_binary(
    ...
    static = select({
    }),
)
```

### What did you see instead?

illicitonion:
**What type of PR is this?**

Bug fix

Instead, insert a dummy rule which exists only to transition its
dependencies using an edge transition, which means the appropriate
attributes are resolved before the transition is called.

To achieve this, we symlink the actual built binaries into our
transitioning wrapper rules, and forward any providers needed.

Fixes #3103

**Other notes for review**

The first commit just extracts variables for attr dicts without modifying them, the second contains the meat of the change.


go/private/rules/transition.bzl
```diff
@@ -278,13 +319,13 @@ def _go_reset_target_impl(ctx):
...
         )
-        default_runfiles = default_runfiles.merge(ctx.runfiles([new_executable]))
+        runfiles_wrapper = _add_to_runfiles(ctx, runfiles_wrapper, new_executable)
 
     providers.append(
         DefaultInfo(
             files = default_info.files,
-            data_runfiles = default_info.data_runfiles,
-            default_runfiles = default_runfiles,
+            data_runfiles = runfiles_wrapper.data_runfiles,
```
illicitonion, line 327: Note that previously we didn't add the new executable to `data_runfiles` and now we do - this felt like a positive change, assuming not doing so before was simply oversight, but if there was an important reason for not doing this before, I can revert and add a comment.

fmeum:
@illicitonion I may be missing something, but I think that this can be simplified: Instead of using a wrapper to transition the entire `go_binary`, the `go_transition` could be applied only to the `deps` and `_go_context_data` attribute. This would get rid of the macro and transition wrapper as well as fix some forms of https://github.com/bazelbuild/bazel/issues/14626 (see https://github.com/bazelbuild/rules_go/pull/3108/commits/089f13ea8c2db0065d708586fd58775a334360fc for the very much related complete fix for this issue).

illicitonion:
> Instead of using a wrapper to transition the entire `go_binary`, the `go_transition` could be applied only to the `deps` and `_go_context_data` attribute.

As far as I can tell, that can't change how the binary is actually linked?

My use-case is using a musl toolchain to cross-compile Linux binaries to put in docker images from a macOS host. We want to statically link, and so want to set `static = "on"`, and want our custom C++ toolchain to be used for the linking, so set `pure = "off"`, but we do so via a `select`; effectively:
```
static = select({
}),
pure = select({
  "@io_bazel_rules_go//go/platform:linux": "off",
}),
```

I tried patching in your linked change instead of mine, and didn't end up with the static binary I'd expect. I think this makes sense, because something needs to be able to apply a transition to the `go_binary` itself, which necessitates something depending on it (the transition wrapper), or the current approach (a self transition), except the self-transition approach can't see `select`-ed values.

fmeum:
To clarify what I think may be a misunderstanding: The commit https://github.com/bazelbuild/rules_go/commit/089f13ea8c2db0065d708586fd58775a334360fc, which is part of https://github.com/bazelbuild/rules_go/pull/3108, is not meant to fix the issue you reported as #3103. Rather, it is meant to fix https://github.com/bazelbuild/bazel/issues/14626, which at its root is caused by rules_go applying transitions on Go settings to all its attributes, even those that do not reference Go targets (e.g. `srcs` or `cdeps`).

What I suggested - if it turns out to work - is a way to simultaneously 1) simplify the current PR and 2) reduce, but not completely remove the need for transitions resetting Go settings on non-Go dependency edges such as `srcs` and `cdeps`. 

> I think this makes sense, because something needs to be able to apply a transition to the go_binary itself...

I think that this part is not necessarily true: If I understand the code correctly, the only way in which rules_go accesses the values of the settings set in the transition is through `_go_context_data`. Thus, it should be sufficient if `go_binary` and all its transitive `go_library` dependencies depend on `_go_context_data` in the correct configuration. I think that this is guaranteed as long as both `deps` and `_go_context_data` itself carry `cfg = go_transition` - there should be no need for the `go_binary` itself to be built in this configuration. The only thing that would change if the transitions are moved to the relevant attributes would be the output path of the `go_binary`, but since all Go settings values are hidden away in the opaque `ST` hash suffix anyway, that shouldn't matter.

Sorry for the confusion I caused by bringing up the other PR. I hope that the above managed to clarify the situation and my proposal. Please let me know if I should elaborate on something.

fmeum:
Glad to hear it worked! Since my PR becomes simpler due to yours and yours should be easier to review, I would suggest that we should aim for getting yours merged first. Since I already have to PRs based on each other (https://github.com/bazelbuild/rules_go/pull/3005 and https://github.com/bazelbuild/rules_go/pull/3108) and blocked on a third one (https://github.com/bazelbuild/rules_go/pull/3113), I would just wait until it got merged and rebase.

illicitonion:
Closing in favour of the simpler https://github.com/bazelbuild/rules_go/pull/3116 - thanks @fmeum for the suggestion!

PR #3116, opened 2022-04-12
Transition on edges not self

### LINKED ISSUE #3015: Editor support broken for `gazelle` generated `go_test` for packages containing both internal and external tests
v0.29.0

v0.24.0

4.2.1

I am using the latest versions already.

Intel Mac

I have created the initial setup here: https://gist.github.com/rohitagarwal003/d9087ca9f87271b1b4f1d392a9dea593.

If you download all those files in a directory named `mypkg` and run `bazel run //:gazelle` from there, the updated `BUILD.bazel` will not work correctly with VSCode.

#### `BUILD.bazel` file generated by gazelle that doesn't work with VSCode
```
load("@bazel_gazelle//:def.bzl", "gazelle")

# gazelle:prefix example.com/mypkg
gazelle(name = "gazelle")

go_library(
    name = "mypkg",
    srcs = ["mypkg.go"],
    importpath = "example.com/mypkg",
)

go_test(
    name = "mypkg_test",
    srcs = [
        "mypkg_external_test.go",
        "mypkg_internal_test.go",
    ],
    embed = [":mypkg"],
)
```

error-with-gazelle-generated-rules

Error: **`package mypkg_test; expected mypkg compiler(MismatchedPkgName)`**

#### `BUILD.bazel` file that does work
```

go_library(
    name = "mypkg",
    srcs = ["mypkg.go"],
    importpath = "example.com/mypkg",
)

go_test(
    name = "mypkg_external_test",
    srcs = ["mypkg_external_test.go"],
    deps = [":mypkg"],
)

go_test(
    name = "mypkg_internal_test",
    srcs = ["mypkg_internal_test.go"],
    embed = [":mypkg"],
)
```

no-error-with-manual-rules

---

cc @steeve @linzhp

### LINKED ISSUE #3103: Can't configure static differently for different platforms

illicitonion:
**What type of PR is this?**

Bug fix

We only actually consult configuration data in `_go_context_data`, so we
only actually need to transition on the edges which (transitively) reach
a `_go_context_data`, which is `_go_context_data` itself and `deps`.

Fixes #3103
Fixes #3015

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -392,7 +392,7 @@ def go_context(ctx, attr = None):
     coverdata = None
     nogo = None
     if hasattr(attr, "_go_context_data"):
-        go_context_data = _flatten(attr._go_context_data)
+        go_context_data = _flatten_possibly_transitioned_attr(attr._go_context_data)
         if CgoContextInfo in go_context_data:
             cgo_context_info = go_context_data[CgoContextInfo]
         go_config_info = go_context_data[GoConfigInfo]
@@ -406,7 +406,7 @@ def go_context(ctx, attr = None):
     if hasattr(attr, "_go_config"):
         go_config_info = attr._go_config[GoConfigInfo]
     if hasattr(attr, "_stdlib"):
-        stdlib = _flatten(attr._stdlib)[GoStdLib]
+        stdlib = _flatten_possibly_transitioned_attr(attr._stdlib)[GoStdLib]
 
     mode = get_mode(ctx, toolchain, cgo_context_info, go_config_info)
     tags = mode.tags
@@ -866,12 +866,14 @@ go_config = rule(
 def _expand_opts(go, attribute_name, opts):
     return [go._ctx.expand_make_variables(attribute_name, opt, {}) for opt in opts]
 
+_LIST_TYPE = type([])
+
 # Used to get attribute values which may have been transitioned.
 # Transitioned attributes end up as lists.
 # We never use split-transitions, so we always expect exactly one element in those lists.
 # But if the attribute wasn't transitioned, it won't be a list.
-def _flatten(maybe_list):
-    if type(maybe_list) == "list":
+def _flatten_possibly_transitioned_attr(maybe_list):
+    if type(maybe_list) == _LIST_TYPE:
         if len(maybe_list) == 1:
             return maybe_list[0]
         else:
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -392,12 +392,13 @@ def go_context(ctx, attr = None):
     coverdata = None
     nogo = None
     if hasattr(attr, "_go_context_data"):
-        if CgoContextInfo in attr._go_context_data:
-            cgo_context_info = attr._go_context_data[CgoContextInfo]
-        go_config_info = attr._go_context_data[GoConfigInfo]
-        stdlib = attr._go_context_data[GoStdLib]
-        coverdata = attr._go_context_data[GoContextInfo].coverdata
-        nogo = attr._go_context_data[GoContextInfo].nogo
+        go_context_data = _flatten(attr._go_context_data)
+        if CgoContextInfo in go_context_data:
+            cgo_context_info = go_context_data[CgoContextInfo]
+        go_config_info = go_context_data[GoConfigInfo]
+        stdlib = go_context_data[GoStdLib]
+        coverdata = go_context_data[GoContextInfo].coverdata
+        nogo = go_context_data[GoContextInfo].nogo
     if getattr(attr, "_cgo_context_data", None) and CgoContextInfo in attr._cgo_context_data:
         cgo_context_info = attr._cgo_context_data[CgoContextInfo]
     if getattr(attr, "cgo_context_data", None) and CgoContextInfo in attr.cgo_context_data:
@@ -405,7 +406,7 @@ def go_context(ctx, attr = None):
     if hasattr(attr, "_go_config"):
         go_config_info = attr._go_config[GoConfigInfo]
     if hasattr(attr, "_stdlib"):
-        stdlib = attr._stdlib[GoStdLib]
+        stdlib = _flatten(attr._stdlib)[GoStdLib]
 
     mode = get_mode(ctx, toolchain, cgo_context_info, go_config_info)
     tags = mode.tags
@@ -864,3 +865,15 @@ go_config = rule(
 
 def _expand_opts(go, attribute_name, opts):
     return [go._ctx.expand_make_variables(attribute_name, opt, {}) for opt in opts]
+
+# Used to get attribute values which may have been transitioned.
+# Transitioned attributes end up as lists.
+# We never use split-transitions, so we always expect exactly one element in those lists.
+# But if the attribute wasn't transitioned, it won't be a list.
+def _flatten(maybe_list):
+    if type(maybe_list) == "list":
+        if len(maybe_list) == 1:
+            return maybe_list[0]
+        else:
+            fail("Expected exactly one element in list but got {}".format(maybe_list))
+    return maybe_list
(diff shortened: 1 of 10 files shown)


fmeum:
@illicitonion Looks great and removes a great deal of complexity. The maintainers would likely appreciate a small test (or just a regular target under `tests/`) that proves that `select` can be used in transitioned attributes.

uhthomas:
Should the transition be added to all attributes? This would be a potentially undesirable behaviour change for any attributes which don't include the transition.

It's possible to transition values for almost all attribute types.

uhthomas:
Sorry, I realise I should have added some more context to my previous comment.

It's possible for the transition rule to change the target platform. As such, attributes which don't use this transition will then potentially select on the wrong platform.

Does this make sense?

fmeum:
True, I forgot about the possible platform change when I proposed this in https://github.com/bazelbuild/rules_go/issues/3103#issuecomment-1091408914. Adding the transition to every `label` or `label_list` attribute should preserve the old behavior.

uhthomas:
Don't forget attr.string! They are also selectable. I don't know if other attributes are too.

fmeum:
> Don't forget attr.string! They are also selectable. I don't know if other attributes are too.

You are right, messed up again. It's possible that someone selects essentially any attribute based on the current platform. That's bad. I'll have to think about an alternative, as fixing #3103 also seems desirable. But maybe the wrapper has to be brought back. Sorry @illicitonion for leading you in the wrong direction.

illicitonion:
Makes sense - I've re-opened https://github.com/bazelbuild/rules_go/pull/3130 which adds a wrapper target to do the transitioning, PTAL instead!

fmeum:
@illicitonion Sorry again for the long wait and back and forth on this. If you feel like you don't have the cycles to make the change I mentioned in https://github.com/bazelbuild/rules_go/pull/3130#issuecomment-1144545601, please let me know and I can take over that part.

go/private/context.bzl
```diff
@@ -837,3 +838,15 @@ go_config = rule(
 
 def _expand_opts(go, attribute_name, opts):
     return [go._ctx.expand_make_variables(attribute_name, opt, {}) for opt in opts]
+
+# Used to get attribute values which may have been transitioned.
+# Transitioned attributes end up as lists.
+# We never use split-transitions, so we always expect exactly one element in those lists.
+# But if the attribute wasn't transitioned, it won't be a list.
+def _flatten(maybe_list):
+    if type(maybe_list) == "list":
```
fmeum, line 847: Same as below, could we check against a constant set to `type([])` or use the skylib function?

go/private/context.bzl
```diff
@@ -837,3 +838,15 @@ go_config = rule(
 
 def _expand_opts(go, attribute_name, opts):
     return [go._ctx.expand_make_variables(attribute_name, opt, {}) for opt in opts]
+
+# Used to get attribute values which may have been transitioned.
+# Transitioned attributes end up as lists.
+# We never use split-transitions, so we always expect exactly one element in those lists.
+# But if the attribute wasn't transitioned, it won't be a list.
+def _flatten(maybe_list):
```
fmeum, line 846: I would prefer a less generic name, but have a hard time finding a good one. What do you think of `_flatten_possibly_transitioned_attr`?

go/private/rules/binary.bzl
```diff
@@ -213,6 +213,7 @@ _go_binary_kwargs = {
             doc = """List of Go libraries this package imports directly.
             These may be `go_library` rules or compatible rules with the [GoLibrary] provider.
             """,
+            cfg = go_transition,
```
fmeum, line 216: You should be able to remove the `non_go_transition` configurations: Since the binary rule itself isn't transitioned anymore, there is no need to reset Go settings.

Same for `go_test` below.

go/private/mode.bzl
```diff
@@ -80,8 +80,8 @@ def get_mode(ctx, go_toolchain, cgo_context_info, go_config_info):
     debug = go_config_info.debug if go_config_info else False
     linkmode = go_config_info.linkmode if go_config_info else LINKMODE_NORMAL
     cover_format = go_config_info and go_config_info.cover_format
-    goos = go_toolchain.default_goos
-    goarch = go_toolchain.default_goarch
+    goos = go_toolchain.default_goos if getattr(ctx.attr, "goos", "auto") == "auto" else ctx.attr.goos
```
fmeum, line 83: I'm slightly surprised that this line had to be modified. Could you briefly explain why?

fmeum:
> (Apologies for the delay, I've been out on a lovely relaxing holiday!)

Glad to hear that, relaxed contributors are our favorite contributors :-)

This looks really good overall, only left a few comments. Glad to see the deletion of quite a bit of complex code.

go/private/mode.bzl
illicitonion: Yes! It's this stack of calls, if `goos` is explicitly set on a `go_binary`:
`_go_binary_impl` -> `emit_binary` -> `emit_archive`

In `emit_archive` for the binary, we have this check: `if a.source.mode != go.mode: fail`.

If `goos` is set, `go.mode` is set to that `goos`, but currently the `a.source` uses the overriden `goos`, but `go.mode` uses the platform default (which makes sense - the toolchain is still an exec-native toolchain, we're just telling it to build targeting a different `goos`).

This causes an error, e.g. when building `//tests/core/cross:linux_cross` on a mac, because `go.mode` is set to `darwin_amd64_pure`, causing the check in `emit_archive` to fail.

go/private/rules/library.bzl
```diff
@@ -31,10 +31,6 @@ load(
     "GoLibrary",
     "INFERRED_PATH",
 )
-load(
-    "//go/private/rules:transition.bzl",
-    "non_go_transition",
```
fmeum, line 36: Could you rebase onto master? I fixed a test that should not (rightly) fail without `non_go_transition` on the attrs of go_library (while removing it from go_binary and go_test is fine with edge transitions).

go/private/mode.bzl
fmeum: That makes sense. Can you point to the part of this PR that makes this change necessary? Why do the tests fail only now?

illicitonion: I believe it's the removal of `go_transition_rule` - specifically of setting `cfg = go_transition` in `go_transition_rule`. Before the rule would get transitioned, so would get a new platform:

https://github.com/bazelbuild/rules_go/blob/52a7c74b3975bf9833e4c223370375b8b798a412/go/private/rules/transition.bzl#L177-L187

which would in turn pick a new toolchain, and so have the correct `default_goos`.

Now, instead, we need to decide the goos based on the target, rather than the self-transition, because the self-transition doesn't happen any more.

I _think_ this is the only place that we read these settings assuming they're transitioned, and the tests agree, but it's for sure possible I've missed some others that we will discover :)

go/private/rules/library.bzl
illicitonion: Aha, yep, this test now fails - re-added the `non_go_transition` to `go_library` :)

go/private/rules/transition.bzl
```diff
@@ -224,6 +168,46 @@ def _go_transition_impl(settings, attr):
 
     return settings
 
+def _non_go_transition_impl(settings, attr):
```
fmeum, line 171: Not: Could you clean up the combined diff by moving this definition back to it's original place?

fmeum (approved):

go/private/rules/transition.bzl
illicitonion: Done!

illicitonion:
Ok I think we're good to go on this @achew22 @linzhp!

linzhp (approved):
All tests passed in Uber

PR #3122, opened 2022-04-18
Moving tests for go_embed_data and gomock to tests/extras

linzhp:
**What type of PR is this?**

Refactoring

Both go_embed_data and gomock belong to "extras", so their tests should not be in "tests/core"


.bazelci/presubmit.yml
```diff
@@ -166,17 +166,6 @@ tasks:
     - "-//tests/core/cgo:generated_versioned_dylib_client"
     - "-//tests/core/cgo:generated_versioned_dylib_test"
     - "-//tests/core/cross:proto_test"
-    - "-//tests/core/go_embed_data:go_default_test"
```
achew22, line 169: Is the implication that these were not getting tested at all before?

achew22 (approved):

.bazelci/presubmit.yml
linzhp: Some of the test may be broken but not caught. In fact `//tests/extras/go_embed_data:go_default_test` is still broken, so I am not able to enable it in this PR.

linzhp: Some of the test may be broken but not caught. In fact `//tests/extras/go_embed_data:go_default_test` is still broken, so I am not able to enable it in this PR.

PR #3130, opened 2022-04-25
Use edge transitions not self transitions

### LINKED ISSUE #3103: Can't configure static differently for different platforms

illicitonion:
**What type of PR is this?**

Bug fix

Instead, insert a dummy rule which exists only to transition its
dependencies using an edge transition, which means the appropriate
attributes are resolved before the transition is called.

To achieve this, we symlink the actual built binaries into our
transitioning wrapper rules, and forward any providers needed.

Fixes #3103

**Other notes for review**

The first commit just extracts variables for attr dicts without modifying them, the second contains the meat of the change.

### DIFF
--- a/go/private/rules/BUILD.bazel
+++ b/go/private/rules/BUILD.bazel
@@ -43,6 +43,7 @@ bzl_library(
         "//go/private:providers",
         "//go/private:rpath",
         "//go/private/rules:transition",
+        "@bazel_skylib//lib:dicts",
     ],
 )
 
@@ -132,6 +133,7 @@ bzl_library(
         "//go/private:providers",
         "//go/private/rules:binary",
         "//go/private/rules:transition",
+        "@bazel_skylib//lib:dicts",
         "@bazel_skylib//lib:structs",
     ],
 )
@@ -148,6 +150,7 @@ bzl_library(
         "//go/private:mode",
         "//go/private:platforms",
         "//go/private:providers",
+        "//go/private/skylib/lib:dicts",
     ],
 )
 
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -29,8 +29,11 @@ load(
 )
 load(
     "//go/private/rules:transition.bzl",
-    "go_transition_rule",
+    "ForwardingPastTransitionProvider",
+    "forward_through_transition_impl",
+    "go_transition",
     "non_go_transition",
+    "transition_attrs",
 )
 load(
     "//go/private:mode.bzl",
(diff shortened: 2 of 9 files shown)


achew22:
@fmeum, how much does this overlap with your other transition PRs?

fmeum:
> @fmeum, how much does this overlap with your other transition PRs?

It does overlap in terms of merge conflicts, but the goals are completely separate. When in doubt, merge this PR first as it is certainly more complicated than mine and would require more effort to rebase.

linzhp:
@achew22 let me know when you finish your review and want me to test in Uber.

illicitonion:
@achew22 I've rebased this past https://github.com/bazelbuild/rules_go/pull/3143 - would you have some time to take a look? It's gotten smaller as a result of the rebasing :)

fmeum (commented):
Only while reviewing did I gain some much-needed clarity on the way to configurable attributes. @illicitonion Sorry for all the back and forth on this!

Currently, no attributes are configurable since we use a self transition. With this PR, all attributes would be configurable, with potentially confusing or at least unclear semantics in the case where `goos`/`goarch` is set and there is also a select on a constraint setting.

At the same time, while technically sound, this PR adds quite a bit of complexity: A new wrapper (around a wrapper), including the need to forward providers and symlink executables.

@uhthomas @linzhp Given the fact that the alternate approach of https://github.com/bazelbuild/rules_go/pull/3116 would achieve mostly the same by *deleting complex code*, I would like to hear your thoughts on the following plan:

1. Merge https://github.com/bazelbuild/rules_go/pull/3116, but with a small addition: Keep a macro wrapper around `go_binary` that `fail`s the build if `goos`/`goarch` are set and any other attribute is a `select` expression. Since `select` expressions didn't work before, this does not affect backwards compatibility - it merely restricts a new feature to an (IMO) reasonable subset of its full scope.
2. Think about ways to replace `goos`/`goarch` in a way that doesn't complicate `go_binary`. For example, we could introduce a `go_switch_os_arch` rule that handles the platform transition for a single `go_binary` in its attributes.
3. Deprecate and eventually remove `goos`/`goarch` as Bazel platform support matures further.

fmeum:
@illicitonion I discussed this with @achew22 and @linzhp and we arrived at the conclusion that merging #3116 with the small addition noted in https://github.com/bazelbuild/rules_go/pull/3130#pullrequestreview-986692846 is the preferred solution for both rules_go users and maintainers: It prevents users from introducing confusing interactions with platforms and essentially implements a new feature by deleting complicated code, which is nice.

illicitonion:
Closing in favour of #3116

PR #3133, opened 2022-04-28
Generate docs for gomock

linzhp:
**What type of PR is this?**
Documentation

Generate docs for the newly added `gomock` rule. For this purpose, the signature of `gomock` macro was refactored to spell out all parameters, instead of using a `kargs`.


go/private/BUILD.bazel
```diff
@@ -137,6 +137,7 @@ bzl_library(
     visibility = [
         "//go:__subpackages__",
         "//proto:__pkg__",  # keep
+         "//extras:__pkg__",
```
achew22, line 140: Can you format this?

achew22 (approved):

PR #3138, opened 2022-05-04
Add staticcheck analyzers into rules_go

### LINKED ISSUE #3129: Include staticcheck analyzers inside rules_go
I have created https://github.com/sluongng/staticcheck-codegen and successfully use it in some small code base.

I want to know if there is any interest from the current maintainers of rules_go to have this repo to be part of rules_go instead of living outside of a separate repository.

Nowaday in the Golang ecosystem, `staticcheck` has become the defacto standard linter and including the checks in rules_go should theoratically improve discoverability and maintainability. This should help complement the existing `TOOLS_NOGO` defined at https://github.com/bazelbuild/rules_go/blob/c1426043e79efaef956201a4f80b1652d123c2ce/go/def.bzl#L78-L119

My plan is to have all the analyzers to be under `@rules_go//go/nogo_analyzers/staticcheck/...` and the code generator binary to be a part of `@rules_go//go/tools/staticcheck-codegen`. All definitions will continued be exposed via `//go/def.bzl`.

WDYT?

sluongng:
Extract all analyzers from staticcheck for easier consumption with
rules_go's nogo.

This follows the design suggested in https://github.com/bazelbuild/rules_go/issues/3129 by @uhthomas and export a set of analyzers extracted from staticcheck.

The current code can be built and is minimally tested without much documentation.

Fixes #3129


WORKSPACE
```diff
@@ -122,6 +122,76 @@ go_repository(
...
+)
+
+go_repository(
+    name = "com_github_yuin_goldmark",
+    importpath = "github.com/yuin/goldmark",
+    sum = "h1:/vn0k+RBvwlxEmP5E7SZMqNxPhfMVFEJiykr15/0XKM=",
+    version = "v1.4.1",
+)
+
+# go_repository(
```
uhthomas, line 146: ?

WORKSPACE
```diff
@@ -122,6 +122,76 @@ go_repository(
     version = "v0.0.0-20210628180205-a41e5a781914",
 )
 
+go_repository(
```
uhthomas, line 125: These will need to be in the go_repositories macro.

go/analyzer/staticcheck/BUILD.bazel
```diff
@@ -0,0 +1,185 @@
+# gazelle:ignore
```
uhthomas, line 1: Is this needed? If it is, could we prefer # keep instead of ignoring entire directories.

go/analyzer/staticcheck/util/util.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+var StaticcheckAnalyzers = [][]*lint.Analyzer{
+	quickfix.Analyzers,
+	simple.Analyzers,
+	staticcheck.Analyzers,
+	stylecheck.Analyzers,
+}
+
+var analyzerMap, AnalyzerSlice = func() (map[string]*analysis.Analyzer, []*analysis.Analyzer) {
```
uhthomas, line 22: why is the slice public?

go/analyzer/staticcheck/util/util.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"golang.org/x/tools/go/analysis"
+	"honnef.co/go/tools/analysis/lint"
+	"honnef.co/go/tools/quickfix"
+	"honnef.co/go/tools/simple"
+	"honnef.co/go/tools/staticcheck"
+	"honnef.co/go/tools/stylecheck"
+)
+
+var StaticcheckAnalyzers = [][]*lint.Analyzer{
```
uhthomas, line 15: these don't need to be public, and I think I'm general it's a good idea to keep the package scope slim. Could we move this into the function?

go/analyzer/staticcheck/util/util.go
linzhp, line 22: We may not need `AnalyzerSlice`, as the only place that uses it is the `all_checks_printer`, which can also iterate through the `analyzerMap`.

go/analyzer/staticcheck/template/analyzer.go
```diff
@@ -0,0 +1,11 @@
+package template
+
+import "github.com/bazelbuild/rules_go/go/analyzer/staticcheck/util"
+
+var (
+	// Value to be added during stamping
+	name = "dummy-value-please-replace-using-x_defs"
```
uhthomas, line 7: Does it make sense to leave this blank and then fail at build time if it's blank? There are tests to prevent this, but I really don't like magic strings.

go/analyzer/staticcheck/cmd/all_checks_printer/main.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/bazelbuild/rules_go/go/analyzer/staticcheck/util"
+)
+
+func main() {
+	for _, a := range util.AnalyzerSlice {
```
uhthomas, line 10: Ah this is where it's used, I see. I still don't know why though, if all the analysers are in that map. It feels weird to have both, can we stick to just one?

go/analyzer/staticcheck/BUILD.bazel
```diff
@@ -0,0 +1,185 @@
...
+[
+    go_library(
+        name = analyzer,
+        srcs = ["//go/analyzer/staticcheck/template:analyzer.go"],
+        importpath = "github.com/bazelbuild/rules_go/go/analyzer/staticcheck/" + analyzer,
+        visibility = ["//visibility:public"],
+        x_defs = {"name": analyzer},
+        deps = ["//go/analyzer/staticcheck/util"],
+    )
+    for analyzer in ANALYZERS
```
uhthomas, line 175: We can reduce the build graph quite significantly if we don't even declare analysers we're not using. It may make more sense (and be be a bit cleaner) to use a macro with `maybe`?

uhthomas (changes requested):
I like where this is going, and I think a unit test is a great idea.

Though I think we could do with some documentation.

Would you be able to help understand how you envision these analysers be consumed? I don't see an obvious was of doing so in this PR.

I think we're also in a good position to make the implementation nicer than just loading a list of target labels and shoving them into nogo. It could be possible to add a new attribute (similar to `vet`) which allows staticcheck to be enabled. This would also have to account for enabling just S, ST, QF, etc. or filtering out even further. Needs some thought.

go/analyzer/staticcheck/BUILD.bazel
uhthomas: Though maybe it doesn't matter too much. Either way I'd advice against liar comprehension as it's not very easy to read or maintain. Especially as this comprehension is duplicated already.

go/analyzer/staticcheck/util/util.go
uhthomas: The only problem with the map is that its order is not deterministic. As such, the output (could) be different every time.

WORKSPACE
sluongng: I think a full fledge `deps.bzl` is needed for these `go_repository` targets, and optionally opt-in for those who are interested in using staticcheck? wdyt?

go/analyzer/staticcheck/util/util.go
sluongng: Will remove slice entirely next iteration

go/analyzer/staticcheck/util/util.go
sluongng: I think it's a bit easier to read to have it in a separate package scoped var and encourage others to modify.

But I will just move them into the function as you suggested for now, no strong opinion here.

go/analyzer/staticcheck/cmd/all_checks_printer/main.go
sluongng: agree

WORKSPACE
uhthomas: https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl

go/analyzer/staticcheck/BUILD.bazel
sluongng: So in rules_go, we would want to build all and test all analyzers that we are providing to end users.

But end users will only use a portion of these analyzers and thus, not having to build all of these analyzers in their buildgraph.

I think it will be more clear once we provide a macro that teach others on how to consume the analyzers here 🤔

sluongng:
Changes since last review:
- Rename cmd to `list_analyzers`
- Get rid of separate analyzer slice, `list_analyzers` will sort map's key at run time
- Adjust export scope of variables
- Adjust dummy value to be more readable
- Renamed template package to 'analyzer'

TODO:
- Move go_repository to a separate deps.bzl macro
- Add macro to help consume these analyzers
- Add doc on how to consume this
- Add nogo test which consume this

achew22:
@sluongng this is a great addition to the rules_go ecosystem! The only thing I'm concerned about is bringing in more third party dependencies. For the most part rules_go only depends upon go tools and libraries (with the notable exception of protobuf and related libraries). I would love to see this made available but I'm concerned about expanding the import surface area.

So much of the work you're doing is really valuable I would hate for it to slip away into some personal repo and never see the light of day. What would you think about us creating a repo underneath https://github.com/bazel-contrib that has extras for rules_go, sort of like https://github.com/bazel-contrib/rules_jvm? Presuming that org is okay with it, I would be happy to help you get it set up and it would provide a place for all this value in the ecosystem to congregate and allowing people to a la cart adopt things like static check?

uhthomas:
I feel that it would be valuable to have staticcheck be a part of rules_go and nogo. It's already included with gopls and doesn't require anything more than some go_repository targets. It should also help with the adoption of nogo, especially for users who already depend on golangci-lint.

sluongng:
Recent Change:
- Moved all go_repository dependencies to a new deps.bzl file with a small `maybe` wrapper
- Moved go files in template dir to parent dir in hope that it will fix CI on `@io_bazel_rules_go//go/toolchain:windows_amd64_cgo` platform

sluongng:
> @sluongng this is a great addition to the rules_go ecosystem! The only thing I'm concerned about is bringing in more third party dependencies. For the most part rules_go only depends upon go tools and libraries (with the notable exception of protobuf and related libraries). I would love to see this made available but I'm concerned about expanding the import surface area.
> 
> So much of the work you're doing is really valuable I would hate for it to slip away into some personal repo and never see the light of day. What would you think about us creating a repo underneath https://github.com/bazel-contrib that has extras for rules_go, sort of like https://github.com/bazel-contrib/rules_jvm? Presuming that org is okay with it, I would be happy to help you get it set up and it would provide a place for all this value in the ecosystem to congregate and allowing people to a la cart adopt things like static check?

Hey thanks for taking a look at this.

I am currently already having a similar setup in my separate repo https://github.com/sluongng/staticcheck-codegen. If rejected from rules_go, I will probably run all of these under a repo of my own for faster iteration and then donate it to bazel-contrib once it has reached maturity. 

Although, I would much prefer if we can have this in rules_go. The problems I ran into are:
a. Hard for folks to discover this repo to use it with nogo
b. Hard to write tests in rules_go that would make use of an external dependencies

(b) is particularly annoying to deal with when I was trying to setup tests for https://github.com/bazelbuild/rules_go/pull/2996 and not having the analyzers available at hand.

In the near future, here are some of the changes I envision we doing:
- Add staticcheck analyzers to rules_go
- Fix all analyzer compatibility with rules_go (#2996)
- Add golangci-lint analyzers to rules_go (instead of using a list of names, we can use a list of func in https://github.com/sluongng/nogo-analyzer-golangci-lint/blob/master/main.go#L28-L86) 
- Refactor nogo to accept more config flag with sane default. (i.e. being able to ignore external package for all analyzers could be a sane default config to have)

These changes require throrough testing between nogo source code and the analyzers used. So I would much prefer if we can keep these analyzer in the same source tree as rules_go if possible.

---

As for dependencies importing bloat, my current approach would be for each `//go/analyzer/{staticcheck, golangci-lint, ...}` there would be a separate set of `deps.bzl` and `def.bzl`. Users who wish to adopt any of the analyzer can load the repository macro in `deps.bzl` by themselves. These will be included in rules_go's `WORKSPACE` by default (to help facilitate testing) but not in `@io_bazel_rules_go//go:deps.bzl` where folks load the depenencies by default.

In short, the analyzer will need to be opted-in by users and their dependencies will not be loaded by default.

sluongng:
Recent change:
- Created `deps.bzl` and `def.bzl` for end user to easily consume the rule
- Added files targets needed for the `bazel_testing` integration test (i created these targets manually, is there something like kazel to create them automatically?)
- Setup a basic integration test(copied from `vet` tests and simplified) where staticcheck ST1000 was used to detect package has no documentation

TODO:
- Refactor test so that we can swap out different checks in table testing
- Add some more tests
- Add a README explaining usages and example

tests/core/nogo/staticcheck/staticcheck_test.go
```diff
@@ -0,0 +1,113 @@
...
+-- BUILD.bazel --
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_tool_library", "nogo")
+load("@io_bazel_rules_go//go/analyzer/staticcheck:def.bzl", "staticcheck_analyzer")
+
+nogo(
+    name = "nogo",
+		deps = staticcheck_analyzer(["ST1000"]),
+    visibility = ["//visibility:public"],
+)
+
```
sluongng, line 37: FYI: This block shows the UX I am aiming for:

`staticcheck_analyzer` being a macro that return a list of analyzers that can be used in `nogo` deps attribute.

This being a list also means that folks can compose it freely in the future. Example would be:

```
nogo(
    name = "nogo",
		deps = TOOLS_NOGO + staticcheck_analyzer(STATICCHECK_ALL_ANALYZER) + golangci_lint_analyzer(["gofumpt"]),
)
```

sluongng:
> > when I was trying to setup tests for #2996 and not having the analyzers available at hand.
> 
> Have you tried the analyzers under `@org_golang_x_tools//go/analysis/passes`? On the other hand, it's also a good idea to write a simple analyzer for tests. Reusing existing analyzers adds unnecessary dependencies to the test. For example, the existing analyzer may break for some unrelated issue, making the test fail.

The issues in #2996 are very specific to staticcheck analyzers interact with rules_go generated code: usage of snake case variable and _empty.go file does not have package doc.

I can try copy paste/rewrite those specific analyzers and use them in rules_go, but this current PR is a lot nicer.

sluongng:
Change log:
- Refactor integration test setup to handle multiple analyzers with nogo on/off
- Added a test for another analyzer SA4000
- Renamed testdata files and packages to better reflect test case

TODO:
- Add a README file explaining use case
- Fix window CI (or disable testing for this on window) --> looking for helps/pointers

sluongng:
```
open bazel-out\x64_windows-fastbuild\bin\go\analyzer\staticcheck\_empty.go: The system cannot find the file specified.
```

I suspect that the problem is within this block
https://github.com/bazelbuild/rules_go/blob/a2677c2c0af38a851bf381e013c45a27642892ac/go/tools/builders/compilepkg.go#L187-L204

Because we are declaring multiple `go_library` in the same bazel package, it seems like the value of `emptyPath` is the same between each target, which means that same `_empty.go` file was being used for different `go_library` compilation. At the end of each compilation, the `defer` statement kicked in and deleted the file and caused issue for other `go_library` that are still depending on it.

I have a hard time understanding why this issue happen on Window only.
A hypothesis would be that on Linux and MacOS, bazel was using a different sandbox strategy and thus the created file was not shared between compilation. On Window, the sandbox strategy is none(?) thus this conflict is occuring? 🤔 

Not sure whats the right fix for this 😮‍💨

go/analyzer/staticcheck/cmd/list_analyzers/main.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/bazelbuild/rules_go/go/analyzer/staticcheck/util"
+)
+
+func main() {
+	names := []string{}
```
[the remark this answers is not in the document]
sluongng: Good spot.

Will fix to use the `make` suggestion here.

I think the `maps` package is something quite new after 1.18 so I will not use that for now.

sluongng:
I created a separate (draft) PR to investigate Window CI issue https://github.com/bazelbuild/rules_go/pull/3145

This PR should be ready for review/merge when the maintainers have the time to take a look 🙏

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
+# Staticcheck nogo Analyzers
+
+This package is provided as an opt-in part of rules_go.
+
+Projects who have the need to use staticcheck analyzers as part of rules_go's nogo framework may use this package to enable the checks over existing code base
```
uhthomas, line 5: Missing a period.

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## How to use
+
+1. Please refer to how to use **nogo** in https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst
```
uhthomas, line 9: "Please refer to the nogo documentation"?

"to how to use" reads weird.

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+2. Enable staticcheck dependencies in your WORKSPACE.
```
uhthomas, line 12: Enable -> Load?

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+load("@io_bazel_rules_go//go/analyzer/staticcheck:deps.bzl", "staticcheck_deps")
+
+staticcheck_deps()
+```
+
+Because this make uses of `go_repository` rules of Gazelle, you should only load it after `gazelle_dependencies`.
```
uhthomas, line 20: Is this true? I think I saw some documentation from Gazelle that they would like dependencies to be declared after.

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Note that loading `staticcheck_deps()` is completely optional, advance users may want to manage their own staticcheck dependencies separately in your WORKSPACE file.
```
uhthomas, line 22: advanced*

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+3. Configure `nogo` target in a `BUILD` file
```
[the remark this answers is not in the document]
uhthomas: Or even just keep it as "configure nogo". There is existing documentation on this. I think it makes sense for this document to expand on it, not repeat it.

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+]
+
+nogo(
+    name = "nogo",
+		deps = staticcheck_analyzers([STATICHECK_ANALYZERS]),
+    visibility = ["//visibility:public"],
+)
+```
+
+Note here that `staticcheck_analyzers` return a list which can be combined with other analyzers such as the default TOOLS_NOGO analyzers
```
uhthomas, line 42: Missing a period.

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+nogo(
+    name = "nogo",
+		deps = TOOLS_NOGO + staticcheck_analyzers([STATICHECK_ANALYZERS]),
+    visibility = ["//visibility:public"],
+)
+```
+
+4. Setup nogo config file.
```
uhthomas, line 52: "Configure nogo"?

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+User often do not want to run static analysis over external dependencies / packages by default.
+Therefore, we provide [sample_config_json.txt](./sample_config_json.txt) with all `external` packages excluded from `nogo` checks by default. This file is not meant to be consumed directly, but for convenience copy-pasting.
```
uhthomas, line 55: This isn't nice. Can't we make this a bit better? The JSON file becomes huge and unmaintainable by a human, without the necessary tooling to machines to work with it.

go/analyzer/staticcheck/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Please refer to [nogo's documentation](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst) for more information on how to setup json config file
```
uhthomas, line 57: Missing a period.

go/analyzer/staticcheck/cmd/list_analyzers/main.go
```diff
@@ -0,0 +1,22 @@
...
+	names := make([]string, 0, len(util.Analyzers))
+
+	for a := range util.Analyzers {
+		names = append(names, a)
+	}
+
+	sort.Strings(names)
+
+	for _, name := range names {
+		fmt.Printf("%s\n", name)
```
uhthomas, line 20: Why not just Println?

go/analyzer/staticcheck/def.bzl
```diff
@@ -0,0 +1,193 @@
+# All analyzers in staticcheck.
+#
+# Generated this list by running
```
uhthomas, line 3: Generate *

go/analyzer/staticcheck/def.bzl
```diff
@@ -0,0 +1,193 @@
...
+    "ST1017",
+    "ST1018",
+    "ST1019",
+    "ST1020",
+    "ST1021",
+    "ST1022",
+    "ST1023",
+]
+
+def staticcheck_analyzers(analyzers, prefix_path = "@io_bazel_rules_go//go/analyzer/staticcheck"):
```
uhthomas, line 165: What if we just want all of them?

uhthomas: Could we make the default all? Or maybe even make it opt out rather than opt in? I don't see many use cases where people will want to disable many of these checks. Even if they do, I think it's safer to eagerly enable them rather than make it easy to forget.

go/analyzer/staticcheck/deps.bzl
```diff
@@ -0,0 +1,76 @@
...
+        sum = "h1:kQgndtyPBW/JIYERgdxfwMYh3AVStj88WQTlNDi2a+o=",
+        version = "v0.6.0-dev.0.20220106191415-9b9b3d81d5e3",
+    )
+    go_repository(
+        name = "org_golang_x_net",
+        importpath = "golang.org/x/net",
+        sum = "h1:OfiFi4JbukWwe3lzw+xunroH1mnC1e2Gy5cxNJApiSY=",
+        version = "v0.0.0-20211015210444-4f30a5c0130f",
+    )
+    go_repository(
```
uhthomas, line 47: I think we already have some of these deps declared elsewhere?

go/analyzer/staticcheck/README.md
sluongng: I think it was true for an earlier iteration on my local laptop, but not true for current version in this PR.
In particular, we can switch the test setup around like this:

```
--- a/tests/core/nogo/staticcheck/staticcheck_test.go
+++ b/tests/core/nogo/staticcheck/staticcheck_test.go
@@ -73,13 +73,13 @@ func alwaysTrue(a int) string {
 }
 `,
                WorkspaceSuffix: `
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
-
-gazelle_dependencies()
-
 load("@io_bazel_rules_go//go/analyzer/staticcheck:deps.bzl", "staticcheck_deps")

 staticcheck_deps()
+
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+
+gazelle_dependencies()
```

And things still work for  `bazel test //tests/core/nogo/staticcheck/...`  🤔

I am not quite sure what was the issue I ran into during development but I cannot reproduce it now so I would be more than happy to drop this sentence in the doc.

go/analyzer/staticcheck/README.md
sluongng: I wanted to break it down into 2 separate steps: 1 is the nogo 'build target' and 1 is the json config file for nogo.

Not sure how to best express this, or should I combine them into 1 step?

uhthomas:
@linzhp I think that's a good idea. The Go analysis framework itself expects one analyzer per package and so I don't see why static check would deviate.

go/analyzer/staticcheck/deps.bzl
sluongng: It should not matter much as the `go_repository` here is actually wrapped by `maybe()` so if it's already loaded once somewhere prior, it will not be loaded here again.

sluongng:
> Can you upstream the change to the [original repo](https://github.com/dominikh/go-tools) of staticcheck, so nogo can use it just like analyzers in `@org_golang_x_tools//go/analysis/passes` without any changes in rules_go? I feel it's better for each linter to implement a standard interface, instead of making rules_go a central repo to implement wrappers for all linters.

First, I want to emphasize that this does not change 'much' of the existing rules_go but much more of a new feature added on top of existing functionality. `nogo`, before and after this PR, would still be using `@org_golang_x_tools//go/analysis/passes`. This PR provides some syntatic sugar on top of existing mechanism but not modify/replace it.

Second, from the linter package maintainer perspective, they might also push back on changing their code base to support Bazel's use case. Either that or it would take significant effort for the existing maintainers to learn bazel + rules_go to maintain this wrapper layer. Here are some prime examples:
- https://github.com/dominikh/go-tools/issues/865#issuecomment-725308811
- https://github.com/golangci/golangci-lint/issues/1473#issuecomment-1113964790

Third, the main reason why I prefer having it in rules_go and not in a separate repo (such as https://github.com/sluongng/staticcheck-codegen) is because testing both `nogo` code and the analyzers wrapper code can be done a lot easier when they live in the same repo. For example: in this PR I was able to make use of `go_bazel_test` quite easily by copying from `vet_test`. I also don't have to bother with setting up CI infrastructure in a brand new repo. I expect future tests/code improvement for `nogo` would also benefits from these.

---

With all that said, if there is a strong feeling from the maintainer not wanting to have this as part of rules_go, then I will move this to a separate repo and slow roll it step-by-step 😞 
Let me know what you think.

linzhp:
> First, I want to emphasize that this does not change 'much' of the existing rules_go but much more of a new feature added on top of existing functionality. 

The main concern @achew22 and I have is that this PR expands rules_go's set of features and dependencies: 
* rules_go now has nogo wrappers to staticcheck. People may ask: is it OK to have nogo wrappers of golangci-lint too? How about this cool linter that "everyone" is using? It would be hard for us to say no at that point. We would rather linters to conform to the standard set by `golang.org/x/tools/go/analysis/passes`.
* the dependencies added by this PR are still rules_go's dependencies. If we provide staticcheck wrappers as a feature, we will have to maintain all dependencies of this feature, including upgrading them and resolving conflicts. Additional dependencies always makes releasing rules_go harder.

> Second, from the linter package maintainer perspective, they might also push back on changing their code base to support Bazel's use case. Either that or it would take significant effort for the existing maintainers to learn bazel + rules_go to maintain this wrapper layer.

The requirement to conform to `golang.org/x/tools/go/analysis/passes`'s standard is not specific to Bazel. The linter maintainers don't need to learn Bazel. According to https://github.com/dominikh/go-tools/issues/865#issuecomment-725308811, all you need to do is to split each Analyzer into its own package, which may have been done in https://github.com/dominikh/go-tools/commit/9d57f241ec5d006d31d4e2c5e7459156ba22ca9e, or at least it's the direction author is moving to.

> With all that said, if there is a strong feeling from the maintainer not wanting to have this as part of rules_go

We really need appreciate contributions, but don't want to set bad precedence or make future maintenance harder.

sluongng:
@linzhp Thanks for the feedbacks. I agree with the concerns.

I think I will close this PR and related issue as the result and will attempt to move the PR content into a stand alone for now.

Once thats done, I will comeback to https://github.com/bazelbuild/rules_go/pull/2996 and use the new repo inside a `go_bazel_test`.

go/analyzer/staticcheck/README.md
sluongng: Further testing showed that the order here does matter as loading gazelle deps first would load a diff version of staticcheck

dominikh:
> According to https://github.com/dominikh/go-tools/issues/865#issuecomment-725308811, all you need to do is to split each Analyzer into its own package, which may have been done in https://github.com/dominikh/go-tools/commit/9d57f241ec5d006d31d4e2c5e7459156ba22ca9e, or at least it's the direction author is moving to

Just to clarify on that: we've split fact-producing analyses into separate packages, because there are few of them and because they have exported APIs that benefit from cleaner naming by having dedicated packages. I don't think we'll ever do the same for all analyses, as that would result in 150+ packages and be far too annoying to work with. The structure you support is fine if you have to work on tens of analyses. At hundreds of analyses, it would have the same effect as creating one package per function in any other code base.

uhthomas:
I agree that hundred of packages is not great. Maybe the Go analyser API should change instead? Accepting one package with an `var Analyzers []*analysis.Analyzer` seems like a small and effective change.

dominikh:
go/analysis doesn't have any API concerning this. The "one package per analyzer in a tree called `passes`" layout is just a convention that x/tools / vet use. There is nothing in go/analysis itself that consumes this directory tree. Analysis runners such as `multichecker` just accept a slice of `*analysis.Analyzer`. They have no idea where they come from.

uhthomas:
Makes sense, maybe nogo needs to change.

Thanks for your input @dominikh :)

Issue #3136, opened 2022-05-03
dl.google.com is currently a single point of failure


fishy:
### What version of rules_go are you using?

`v0.31.0`

### What version of gazelle are you using?

`v0.24.0`

### What version of Bazel are you using?

`5.1.1`

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

linux/amd64

### What did you do?

There was (I assume) a bad deploy that broke dl.google.com yesterday afternoon around 3:45pm (pacific time) that was fixed ~15min later. during that ~15min, all builds on our CI/CD system failed with:

```
190 | WARNING: Download from https://dl.google.com/go/go1.18.1.linux-amd64.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 502 Bad Gateway
191 | ERROR: An error occurred during the fetch of repository 'go_sdk':
192 | Traceback (most recent call last):
193 | File "/root/.cache/bazel/_bazel_root/fc07cdbdb3ccc5391e01bb1a31f63d3c/external/io_bazel_rules_go/go/private/sdk.bzl", line 100, column 16, in _go_download_sdk_impl
194 | _remote_sdk(ctx, [url.format(filename) for url in ctx.attr.urls], ctx.attr.strip_prefix, sha256)
195 | File "/root/.cache/bazel/_bazel_root/fc07cdbdb3ccc5391e01bb1a31f63d3c/external/io_bazel_rules_go/go/private/sdk.bzl", line 205, column 21, in _remote_sdk
196 | ctx.download(
197 | Error in download: java.io.IOException: Error downloading [https://dl.google.com/go/go1.18.1.linux-amd64.tar.gz] to /root/.cache/bazel/_bazel_root/fc07cdbdb3ccc5391e01bb1a31f63d3c/external/go_sdk/go_sdk.tar.gz: GET returned 502 Bad Gateway
```

We happened to be trying to do a deploy around the time and have to keep retrying it until dl.google.com was fixed to finally be able to continue the deploy.

For context, we have this part related to `go_sdk` in our `WORKSPACE` file:

```bazel
GO_VERSION = "1.18.1"

...

go_register_toolchains(version = GO_VERSION)
```

Here are some ideas/suggestions/feature requests to make dl.google.com no longer a single point of failure:

1. Make the go_sdk cache-able by remote cache

[...]

PR #3139, opened 2022-05-04
Add an example for go_download_sdk.sdks

fishy:
Fixes https://github.com/bazelbuild/rules_go/issues/3136.

**What type of PR is this?**

Documentation

Fixes #

**Other notes for review**

### DIFF
--- a/go/toolchains.rst
+++ b/go/toolchains.rst
@@ -296,7 +296,20 @@ This downloads a Go SDK for use in toolchains.
         name = "go_sdk",
         goos = "linux",
         goarch = "amd64",
-        version = "1.12.5",
+        version = "1.18.1",
+        sdks = {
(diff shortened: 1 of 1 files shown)


go/toolchains.rst
```diff
@@ -296,7 +296,12 @@ This downloads a Go SDK for use in toolchains.
         name = "go_sdk",
         goos = "linux",
         goarch = "amd64",
-        version = "1.12.5",
+        version = "1.18.1",
+        sdks = {
+            # You can get the expected filenames and checksums from https://go.dev/dl/
```
achew22, line 301: Can you make it a bit more clear that these are not necessary all the time and further elaborate on what the circumstances are when you'd want to override these

fishy: Updated the comment. PTAL.

linzhp (approved):

PR #3141, opened 2022-05-04
Officially deprecate `go_embed_data`

achew22:
rules_go has had a built in way to handle embedded for a while. Let's deprecate the non-standard way.

### DIFF
--- a/extras/embed_data.bzl
+++ b/extras/embed_data.bzl
@@ -34,6 +34,8 @@ go_embed_data_dependencies()
 """
 
 def _go_embed_data_impl(ctx):
+    print("Embedding is now better handled by using rules_go's built in https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/embedding.md functionality. The `bindata` rule is deprecated and will be removed in rules_go version 0.35.")
+
     go = go_context(ctx)
     if ctx.attr.src and ctx.attr.srcs:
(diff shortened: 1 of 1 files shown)


linzhp (approved):

extras/embed_data.bzl
```diff
@@ -34,6 +34,8 @@ go_embed_data_dependencies()
 """
 
 def _go_embed_data_impl(ctx):
+    print("Embedding is now better handled by using rules_go's built in https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/embedding.md functionality. The `bindata` rule is deprecated and will be removed in rules_go version 0.35.")
```
ribrdb, line 37: This link doesn't actually say anything about how to embed data. Is there any documentation for the official way to do this?

chickenandpork: This unanswered comment does highlight one frustration:
 - we cannot use the old way (where there exist example code across the web)
 - there seems to be no example of the new way (or even what we'd google to find it)
 
 It would be really awesome if when one method is deprecated in favour of another, documentation and/or examples show developers how to use the better way.  Additional examples in the `examples` directory -- built during CI to confirm that they're accurate/functional -- would be really awesome but I know it's a stretch to get test coverage and documentation.

PR #3143, opened 2022-05-09
Let go_binary's executable bit depend on `linkmode`

fmeum:
**What type of PR is this?**

Bug fix

If a go_binary's `linkmode` is neither `normal` nor `pie`, the resulting
file can't be executed. In this case, `executable` should be set to
`False` on the go_binary rule so that Bazel doesn't create a runfiles
tree for the targets and does not include e.g. a resulting static
library in the runfiles of dependants.


achew22:
Thanks for your contribution!

PR #3145, opened 2022-05-09
compilepkg: fix race when run without sandbox

### LINKED ISSUE #3144: window: race condition when declare multiple go_library in the same directory

Latest

Irrelevant

Latest

yes

rules_go Window CI

Via our CI setup

### What did you do?

Setup is described in https://github.com/bazelbuild/rules_go/pull/3138#issuecomment-1119433664

where we have something like this in the same BUILD file
```
[
    go_library(
        name = analyzer,
        srcs = ["analyzer.go"],
        importpath = "github.com/bazelbuild/rules_go/go/analyzer/staticcheck/" + analyzer,
        x_defs = {"name": analyzer},
        deps = ["//go/analyzer/staticcheck/util"],
    )
    for analyzer in ANALYZERS
]

[
    go_test(
        name = analyzer + "_test",
        srcs = ["analyzer_test.go"],
        embed = [analyzer],
    )
    for analyzer in ANALYZERS
]
```

Window CI to successfully passed

### What did you see instead?

Window CI reporting this error
```
open bazel-out\x64_windows-fastbuild\bin\go\analyzer\staticcheck\_empty.go: The system cannot find the file specified.
```

Most likely caused by this block https://github.com/bazelbuild/rules_go/blob/a2677c2c0af38a851bf381e013c45a27642892ac/go/tools/builders/compilepkg.go#L187-L204

Here the `outpath` value of each go_library target is the same `bazel-out\x64_windows-fastbuild\bin\go\analyzer\staticcheck` thus a compilation of one go_library was deleting the same `_empty.go` file of another compilation. This is most likely caused by Window lack of sandboxing support for these action to run in parallel https://github.com/bazelbuild/bazel/issues/5136

sluongng:
compilepkg: fix race when run without sandbox

On platforms without sandbox execution support such as Windows, multiple
go_test targets declared in the same package would cause a race condition.

```
[
    go_library(
        name = "same_package_{}".format(i),
        srcs = ["same_package.go"],
        importpath = "github.com/bazelbuild/rules_go/tests/core/go_test/same_package_{}".format(i),
        x_defs = {
            "name": "{}".format(i),
        },
    )
    for i in range(1, 80)
]

[
    go_test(
        name = "same_package_{}_test".format(i),
        srcs = ["same_package_test.go"],
        embed = [":same_package_{}".format(i)],
    )
    for i in range(1, 80)
]
```

For each of go_test targets, there would be internal archive and
external archive that needs to compile from source files.  However, due
to testfilter, external archive would end up without any Go source and
thus compilepkg needs to generate a dummy source file `_empty.go` to
trick 'go tool compile'.

This `_empty.go` file is generated in the output path of the Bazel
package thus multiple go_test targets in the same package would generate
this file using the same path.  On a sandbox supported platform,
this is not a problem as the full path is prefixed with the sandbox root
dir.  However, without sandbox, the full path is the same for each
go_test's external archive compilation, leading to a race condition
where multiple compilations running in parallel would compete to
create/delete the same file.

Fix this by using the unique importPath to create a directory to store
the _empty.go file.  This would effectively give each go_test compilation
it's own _empty.go file in non-sandbox environment.

Worth to note that in current solution, we still have yet to solve
problem when 'importPath' is not provided. For example, in a setup like
this where `go_library` does not exist, `importPath` will be blank and
thus causes similar issues on non-sandbox platforms.

```
[
    go_test(
        name = "same_package_{}_test".format(i),
        srcs = ["same_package_test.go"],
    )
    for i in range(1, 80)
]
```

Although this is a valid use case, we recommend avoid setting up tests
this way. Instead, you could setup a single go_test target with multiple
shards to achieve similar result:

```
go_test(
    name = "same_package_{}_test".format(i),
    srcs = ["same_package_test.go"],
    shard_count = 80,
)
```

For more information, please review Bazel Test Sharding documentation(1)

(1): https://docs.bazel.build/versions/main/test-encyclopedia.html#test-sharding

Fix #3144

### DIFF
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -99,9 +99,10 @@ def emit_compilepkg(
         args.add("-cover_format", go.cover_format)
         args.add_all(cover, before_each = "-cover")
     args.add_all(archives, before_each = "-arc", map_each = _archive)
-    args.add("-label_name", go.label.name)
     if importpath:
         args.add("-importpath", importpath)
+    else:
+        args.add("-importpath", go.label.name)
     if importmap:
         args.add("-p", importmap)
     args.add("-package_list", go.package_list)
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -99,6 +99,7 @@ def emit_compilepkg(
         args.add("-cover_format", go.cover_format)
         args.add_all(cover, before_each = "-cover")
     args.add_all(archives, before_each = "-arc", map_each = _archive)
+    args.add("-label_name", go.label.name)
     if importpath:
         args.add("-importpath", importpath)
     if importmap:
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -194,11 +194,11 @@ func compileArchive(
 		// otherwise platforms without sandbox support may race to create/remove
 		// the file during parallel compilation.
 		emptyDir := filepath.Join(filepath.Dir(outPath), sanitizePathForIdentifier(importPath))
-		err := os.Mkdir(emptyDir, 0700)
-		if err != nil {
-			return err
+		if err := os.Mkdir(emptyDir, 0700); err != nil {
+			return fmt.Errorf("could not create directory for _empty.go: %v", err)
 		}
 		defer os.RemoveAll(emptyDir)
+
 		emptyPath := filepath.Join(emptyDir, "_empty.go")
 		if err := os.WriteFile(emptyPath, []byte("package empty\n"), 0666); err != nil {
 			return err
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -42,7 +42,7 @@ func compilePkg(args []string) error {
 	goenv := envFlags(fs)
 	var unfilteredSrcs, coverSrcs, embedSrcs, embedLookupDirs, embedRoots multiFlag
 	var deps archiveMultiFlag
-	var importPath, packagePath, nogoPath, packageListPath, coverMode string
+	var labelName, importPath, packagePath, nogoPath, packageListPath, coverMode string
 	var outPath, outFactsPath, cgoExportHPath string
 	var testFilter string
 	var gcFlags, asmFlags, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags quoteMultiFlag
@@ -63,6 +63,7 @@ func compilePkg(args []string) error {
 	fs.Var(&objcFlags, "objcflags", "Objective-C compiler flags")
 	fs.Var(&objcxxFlags, "objcxxflags", "Objective-C++ compiler flags")
 	fs.Var(&ldFlags, "ldflags", "C linker flags")
+	fs.StringVar(&labelName, "label_name", "", "The label's name")
 	fs.StringVar(&nogoPath, "nogo", "", "The nogo binary. If unset, nogo will not be run.")
 	fs.StringVar(&packageListPath, "package_list", "", "The file containing the list of standard library packages")
 	fs.StringVar(&coverMode, "cover_mode", "", "The coverage mode to use. Empty if coverage instrumentation should not be added.")
@@ -123,6 +124,7 @@ func compilePkg(args []string) error {
 
 	return compileArchive(
 		goenv,
+		labelName,
 		importPath,
 		packagePath,
 		srcs,
@@ -152,6 +154,7 @@ func compilePkg(args []string) error {
 
 func compileArchive(
 	goenv *env,
+	labelName string,
 	importPath string,
 	packagePath string,
 	srcs archiveSrcs,
@@ -193,7 +196,7 @@ func compileArchive(
 		// to ensure deterministic output. The location also needs to be unique
 		// otherwise platforms without sandbox support may race to create/remove
 		// the file during parallel compilation.
-		emptyDir := filepath.Join(filepath.Dir(outPath), sanitizePathForIdentifier(importPath))
+		emptyDir := filepath.Join(filepath.Dir(outPath), labelName)
 		err := os.Mkdir(emptyDir, 0700)
 		if err != nil {
 			return err
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -42,7 +42,7 @@ func compilePkg(args []string) error {
 	goenv := envFlags(fs)
 	var unfilteredSrcs, coverSrcs, embedSrcs, embedLookupDirs, embedRoots multiFlag
 	var deps archiveMultiFlag
-	var labelName, importPath, packagePath, nogoPath, packageListPath, coverMode string
+	var importPath, packagePath, nogoPath, packageListPath, coverMode string
 	var outPath, outFactsPath, cgoExportHPath string
 	var testFilter string
 	var gcFlags, asmFlags, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags quoteMultiFlag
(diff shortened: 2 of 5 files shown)


sluongng:
This PR is ready for review.

The detail contexts is shared in https://github.com/bazelbuild/rules_go/issues/3144

sluongng:
Merged latest master into this PR to resolve the upstream merge-conflict

sluongng:
@achew22 @linzhp could you guys please take a look at this PR?

The fix is relatively small and there are tests included.

linzhp (commented):
Thanks for the fix. It also make the Windows CI on rules_go more stable. Minor suggestions

go/tools/builders/compilepkg.go
```diff
@@ -188,19 +188,29 @@ func compileArchive(
...
+		err := os.Mkdir(emptyDir, 0700)
+		if err != nil {
+			return err
+		}
+		defer os.RemoveAll(emptyDir)
+		emptyPath := filepath.Join(emptyDir, "_empty.go")
 		if err := os.WriteFile(emptyPath, []byte("package empty\n"), 0666); err != nil {
 			return err
 		}
+		defer os.Remove(emptyPath)
```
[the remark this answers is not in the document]
sluongng: agree, will remove this.

go/tools/builders/compilepkg.go
```diff
@@ -188,19 +188,29 @@ func compileArchive(
...
 		// nothing in it. GoPack will complain if we try to add assembly or cgo
 		// objects.
+		//
 		// _empty.go needs to be in a deterministic location (not tmpdir) in order
-		// to ensure deterministic output
-		emptyPath := filepath.Join(filepath.Dir(outPath), "_empty.go")
+		// to ensure deterministic output. The location also needs to be unique
+		// otherwise platforms without sandbox support may race to create/remove
+		// the file during parallel compilation.
+		emptyDir := filepath.Join(filepath.Dir(outPath), sanitizePathForIdentifier(importPath))
```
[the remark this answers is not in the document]
sluongng: The call chain from go_test rule down to this is very long winded and the 'name' attribute is not being passed down through the call stack:

```
go_test
--> test.bzl -> _go_test_impl() -> go.archive()
--> archive.bzl -> emit_archive()
--> compilepkg.bzl -> emit_compilepkg()
--> compilepkg.go
```

Now the problem is that `go.archive()` is used in many places.
I can try refactoring it to start passing down the stack the value of `ctx.label.name`, but that would require a lot of refactoring 🤔 
And im not yet fully sure on the implications of doing that just yet.

If this is the direction that you are ok with, then I can explore that refactoring in a separate PR to reduce the diff noise? WDYT? 🤔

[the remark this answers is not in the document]
sluongng: Ah nvm, I can use `go.label.name` which is a lot cleaner. Refactoring only need to happen from `compilepkg.bzl` down below which is quite clean.

Let me give your suggestion a try.

sluongng:
Looks like all CI have passed @linzhp PTAL 🙏

go/tools/builders/compilepkg.go
```diff
@@ -416,8 +428,8 @@ func compileArchive(
 	// Compile the .s files.
 	if len(srcs.sSrcs) > 0 {
 		includeSet := map[string]struct{}{
-			filepath.Join(os.Getenv("GOROOT"), "pkg", "include"): struct{}{},
-			workDir: struct{}{},
+			filepath.Join(os.Getenv("GOROOT"), "pkg", "include"): {},
+			workDir: {},
```
sluongng, line 432: This was auto formatted by my editor, let me know if it needs to be excluded from the PR

go/tools/builders/compilepkg.go
```diff
@@ -176,8 +179,8 @@ func compileArchive(
 	outPath string,
 	outXPath string,
 	cgoExportHPath string,
-	coverFormat string) error {
-
+	coverFormat string,
+) error {
```
sluongng, line 183: Same here, auto formatted by editor

go/private/actions/compilepkg.bzl
```diff
@@ -99,6 +99,7 @@ def emit_compilepkg(
         args.add("-cover_format", go.cover_format)
         args.add_all(cover, before_each = "-cover")
     args.add_all(archives, before_each = "-arc", map_each = _archive)
+    args.add("-label_name", go.label.name)
     if importpath:
         args.add("-importpath", importpath)
```
linzhp, line 104:
```suggestion
    if importpath:
        args.add("-importpath", importpath)
    else:
        args.add("-importpath", go.label.name)
```

How about this? So we don't need to add another option to compilepkg.go

fmeum: @sluongng This PR looks almost ready to be merged except for this comment. I'm planning to prepare the 0.33.0 release soon and could include this fix.

sluongng: Tbh, I am not quite sure on this suggestion.
It probably will work for the particular code block that I am interested in, but I am not sure if elsewhere in compilepkg.go do the code expect `importpath` variable to be empty in some rare scenarios.

As i did not have enough time to read through everything, I pushed up a fix according to the suggestion anyway.
I would trust CI result a lot more than myself. 😆

sluongng: ok CI seems to be all green, I think this PR is good to merge @fmeum 🙏

fmeum (approved):
@linzhp I don't have write access yet, could you do the merge if you are satisfied with the PR?

sluongng:
FWIW, I think using target's name in the action input might be an anti pattern.
It will become part of the key calculated in action cache, so similar actions would not re-use cached action if they have different names:

```
[
    go_test(
        name = "test_{}".format(i),
        srcs = ["my_test.go"],
    )
    for i in range(1, 80)
]
```

In such (rare?) case, the 80 test binary compilation/link might run not once, but 80 times.

I am convinced that this is not an issue worth solving right now 🤔 
The use case is very rare that it might not matter at all.

But thinking about the (unrelated) use case of constructing a synthetic tests for RBE, where I want the build graph to be widen to benefit from parallelism, made me think back about this PR and how it might have unintentionally enabled such use case.

fmeum:
Avoiding this would be harder than expected though: The command line is part of the action key computation and necessarily contains the path of the output of the compile/link actions, which is already derived from the test name. If I'm not missing anything, the recent change thus should make a difference here.

sluongng:
I quickly created a test like this:
```
bazel aquery '//tests/core/go_test:same_package_1_test' > /tmp/same_package_1.txt
bazel aquery '//tests/core/go_test:same_package_2_test' > /tmp/same_package_2.txt

diff -u /tmp/same_package_*

# less noisy
diff -u /tmp/same_package_* | grep -vE '(1|2)_test'
```

Then I ran the test before and after reverting this commit.
```
# check out my commit
> git checkout 1e9931b0c2ec4c89f6f651dc9cfccda75d95e1d5

# run test

# revert //go dir
> git checkout 1e9931b0c2ec4c89f6f651dc9cfccda75d95e1d5^ -- go
> git status
HEAD detached at 1e9931b0
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   go/private/actions/compilepkg.bzl
        modified:   go/tools/builders/compilepkg.go
```

And the result seems to be that `ActionKey` values were always updated as target's name has always been part of the Command Line input.

I event tried doing this on top of my patch

```
master ~/work/misc/rules_go> git diff
diff --git a/tests/core/go_test/BUILD.bazel b/tests/core/go_test/BUILD.bazel
index 829ca8d5..3e58db06 100644
--- a/tests/core/go_test/BUILD.bazel
+++ b/tests/core/go_test/BUILD.bazel
@@ -199,6 +199,7 @@ go_library(
     go_test(
         name = "same_package_{}_test".format(i),
         srcs = ["same_package_test.go"],
+        importpath = "blah/blah",
     )
     for i in range(1, 80)
 ]
```

And then ran the same test: same result, ActionKey hash always changed.

---

So it's safe to conclude that this PR did not introduce a regression. 😇

I was being paranoid for no reason. Sorry for the noise 🙇

PR #3148, opened 2022-05-11
Restore original paniconexit0 flag after calling lcovAtExitHook

linzhp:
**What type of PR is this?**
Bug fix

This PR adds a field `OriginalPanicOnExit` to `go/tools/bzltestutil. LcovTestDeps`, so it can restore the original paniconexit bit after calling `lcovAtExitHook()`. This allows testing Exec command using [this pattern](https://jamiethompson.me/posts/Unit-Testing-Exec-Command-In-Golang/).

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -193,7 +193,7 @@ func main() {
 	}
 {{if eq .CoverFormat "lcov"}}
 	panicOnExit0Flag := flag.Lookup("test.paniconexit0").Value
-	testDeps.OriginalPanicOnExit, _ = strconv.ParseBool(panicOnExit0Flag.String())
+	testDeps.OriginalPanicOnExit = panicOnExit0Flag.(flag.Getter).Get().(bool)
 	// Setting this flag provides a way to run hooks right before testing.M.Run() returns.
 	panicOnExit0Flag.Set("true")
 {{end}}
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -192,12 +192,10 @@ func main() {
 		flag.Lookup("test.failfast").Value.Set("true")
 	}
 {{if eq .CoverFormat "lcov"}}
-	panicOnExit0Flag := flag.Lookup("test.paniconexit0")
-	if panicOnExit0Flag != nil && panicOnExit0Flag.Value.String() != "false" {
-		testDeps.OriginalPanicOnExit = true
-	}
+	panicOnExit0Flag := flag.Lookup("test.paniconexit0").Value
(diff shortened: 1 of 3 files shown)


go/tools/builders/generate_test_main.go
```diff
@@ -191,14 +191,15 @@ func main() {
...
-	// Setting this flag serves two purposes:
-	// 1. It attains parity with "go test", which enables this feature by default.
-	//    https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/cmd/go/internal/test/test.go;l=1331-1337
-	// 2. It provides a way to run hooks right before testing.M.Run() returns.
-	flag.Lookup("test.paniconexit0").Value.Set("true")
-
-	{{if ne .CoverMode ""}}
+{{if eq .CoverFormat "lcov"}}
+	panicOnExit0Flag := flag.Lookup("test.paniconexit0")
+	if panicOnExit0Flag != nil && panicOnExit0Flag.Value.String() != "false" {
```
achew22, line 196: Can we check that it explicitly == "true"? Or maybe it would be better to do

```go
if panicOnExit0Flag != nil {
  testDeps.OrigionalPAnicOnExit = panicOnExit0Flag.Value.String()
}
```

go/tools/builders/generate_test_main.go
```diff
@@ -191,14 +191,13 @@ func main() {
...
-	// Setting this flag serves two purposes:
-	// 1. It attains parity with "go test", which enables this feature by default.
-	//    https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/cmd/go/internal/test/test.go;l=1331-1337
-	// 2. It provides a way to run hooks right before testing.M.Run() returns.
-	flag.Lookup("test.paniconexit0").Value.Set("true")
-
-	{{if ne .CoverMode ""}}
+{{if eq .CoverFormat "lcov"}}
+	panicOnExit0Flag := flag.Lookup("test.paniconexit0").Value
+	testDeps.OriginalPanicOnExit, _ = strconv.ParseBool(panicOnExit0Flag.String())
```
abhinav, line 196: FWIW, you don't have to re-parse.
flag.Bool is backed by a [boolValue](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/flag/flag.go;l=126-140;drc=f12461cb0f3878b6d332f4be3a02fa44cc744df6),
which always [implements](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/flag/flag.go;l=140;drc=f12461cb0f3878b6d332f4be3a02fa44cc744df6) the [flag.Getter](https://pkg.go.dev/flag#Getter) interface.

So you can just upcast panicOnExit0Flag to Getter and take the parsed value out.

```suggestion
	testDeps.OriginalPanicOnExit = panicOnExit0Flag.(Getter).Get().(bool)
```

abhinav (approved):

PR #3151, opened 2022-05-12
Don't include non-executable go_binary in dependent's runfiles

fmeum:
**What type of PR is this?**

Bug fix

If a go_binary is built with a non-executable link mode such as
`c-archive`, its dependents currently pick up a runfile dependency on it
since its DefaultInfo specifies the resulting archive as an executable.
This is unnecessary as the dependent should be able to decide whether to
include the file (e.g. dynamic linking) or not (e.g. static linking).

With this commit, the executable field of the DefaultInfo is only
populated if the go_binary is built with an executable link mode.

Follow-up to #3143

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -20,7 +20,6 @@ load(
 )
 load(
     "//go/private:mode.bzl",
-    "LINKMODE_C_SHARED",
     "LINKMODE_NORMAL",
     "LINKMODE_PLUGIN",
     "extld_from_cc_toolchain",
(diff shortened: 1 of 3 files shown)


fmeum:
I had to work around a Bazel bug, but will not make any more changes now.

fmeum:
The CI failure might be due to https://github.com/bazelbuild/bazel/issues/10254. I will wait for feedback on that issue.

fmeum:
@achew22 @linzhp I don't know much about the macOS linker, but https://github.com/bazelbuild/bazel/issues/10254#issuecomment-1125218345 gave me confidence to try deleting the special casing that previously broke my PR. All tests pass with the custom logic removed, but please run this by someone who can confirm it's safe.

kmicklas (approved):

achew22 (approved):
LGTM, but can we test it in the big repo @linzhp

linzhp:
@kmicklas did you already tested it in Uber?

Issue #3119, opened 2022-04-14
depend on rules that don't have mandatory providers: 'GoLibrary'


Kruemelmann:
### What version of rules_go are you using?
v0.31.0

### What version of gazelle are you using?
v0.24.0

### What version of Bazel are you using?
5.1.1

### Does this issue reproduce with the latest releases of all the above?
yes

### What operating system and processor architecture are you using?
on both:
* Linux / x86_64
* Darwin / x86_64

### Any other potentially useful information about your toolchain?
* using rules_nodejs rule to depend on

### What did you do?
I am building a frontend (in my case react using the rules_nodejs) and after this i want to embed its output into a go binary.
So in my understanding i need to depend on the rule that builds my frontend and after this i can use it in embedsrc.

```python
go_library(
    name = "example",
    srcs = ["main.go"],
    embedsrcs = [
        "//ui/example:build_folder",  # keep <- so that gazelle does not kick it out
    ],
    importpath = "github.com/example/example",
    visibility = ["//visibility:public"],
    deps = [
        "//ui/example:build", # keep <- so that gazelle does not kick it out
    ]
)
```
The project where i got this problem is on github but if thats needed i can provide a minimal example project to reproduce.

### What did you expect to see?
I expect the build to first build the gui and after this build use the output and embed it into my go binary.

### What did you see instead?
I got this error:
```
ERROR: /Users/.../example/BUILD.bazel:24:11: in deps attribute of go_library rule //:example: '//ui/example:build' does not have 
mandatory providers: 'GoLibrary'. Since this rule was created by the macro 'go_library_macro', the error might have been caused 
by the macro implementation
```
which in my understanding means that i can only depend on rules related to rules_go.

Kruemelmann:
Maybe i am missunderstanding something. Sorry if thats the case just close this issue

achew22:
Hurm... Is `//ui/example:build` a `go_library`? If not then I don't think this should work that way. `embedsrcs` (just like `data`) is about taking a series of bytes and making them available to your binary. `deps` is how you specify a library that you're going to add an `import` statement in a `.go` file for.

Could you elaborate a bit on what `//ui/example:build` is? Maybe you could point to a repo that's public where this is happening?

Kruemelmann:
hey @achew22 many thanks for your fast reply on my issue (more a question:) )

[...]

PR #3153, opened 2022-05-13
Fix gopackagesdriver for Go 1.18 by replicating stdlib and add test for stdliblist.go

### LINKED ISSUE #3080: Go package driver doesn't work with Go 1.18

xytan0056:
**What type of PR is this?**

This PR addresses comments for #3083

It fixes `go list` invocations from bazel for go1.18, where std library uses `go:embed`
see #3083

Fixes #3080 

**Other notes for review**
updated the patch in our repo with this patch

1. verified that the targets that failed without the patch can build successfully
2. verified a rebuild of all our repo behaved as expected

### DIFF
--- a/go/tools/builders/BUILD.bazel
+++ b/go/tools/builders/BUILD.bazel
@@ -34,7 +34,10 @@ go_test(
         "stdliblist_test.go",
     ],
     data = ["@go_sdk//:files"],
-    rundir = ".",
+    # -sdk in stdliblist needs to be a relative path, otherwise it breaks
+    # assumptions of cloning go_sdk, thus we need to set up the test in a
+    # way that go_sdk is under the directory where test is run.
+    rundir = "..",
     deps = ["//go/tools/bazel"],
 )
 
--- a/go/tools/builders/BUILD.bazel
+++ b/go/tools/builders/BUILD.bazel
@@ -23,6 +23,19 @@ go_test(
     ],
 )
 
+go_test(
+    name = "stdliblist_test",
+    size = "small",
+    data = ["@go_sdk//:files"],
+    srcs = [
+        "stdliblist.go",
+        "stdliblist_test.go",
+        "env.go",
+        "flags.go",
+        "replicate.go",
+    ],
+)
+
 filegroup(
     name = "builder_srcs",
     srcs = [
--- a/go/tools/builders/BUILD.bazel
+++ b/go/tools/builders/BUILD.bazel
@@ -26,14 +26,14 @@ go_test(
 go_test(
     name = "stdliblist_test",
     size = "small",
-    data = ["@go_sdk//:files"],
     srcs = [
-        "stdliblist.go",
-        "stdliblist_test.go",
         "env.go",
         "flags.go",
         "replicate.go",
+        "stdliblist.go",
+        "stdliblist_test.go",
     ],
+    data = ["@go_sdk//:files"],
 )
 
 filegroup(
--- a/go/tools/builders/BUILD.bazel
+++ b/go/tools/builders/BUILD.bazel
@@ -34,6 +34,8 @@ go_test(
         "stdliblist_test.go",
     ],
     data = ["@go_sdk//:files"],
+    rundir = ".",
+    deps = ["//go/tools/bazel"],
 )
 
 filegroup(
--- a/go/tools/builders/stdliblist.go
+++ b/go/tools/builders/stdliblist.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"flag"
+	"fmt"
 	"go/build"
 	"os"
 	"path/filepath"
@@ -165,9 +166,6 @@ func flatPackageForStd(cloneBase string, pkg *goListPackage) *flatPackage {
 // under.
 func cloneGoRoot(execRoot, relativeGoroot, cloneBase string) (newGoRoot string, err error) {
 	goroot := filepath.Join(execRoot, relativeGoroot)
-	if err != nil {
-		return "", err
-	}
 	newGoRoot = filepath.Join(cloneBase, relativeGoroot)
 	if err := os.MkdirAll(newGoRoot, 01755); err != nil {
 		return "", err
@@ -197,13 +195,11 @@ func stdliblist(args []string) error {
 	if err != nil {
 		return err
 	}
-	defer func() {
-		cleanup()
-	}()
+	defer func() { cleanup() }()
 
 	cloneGoRoot, err := cloneGoRoot(goenv.wd, goenv.sdk, cloneBase)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to clone new go root %v", err)
 	}
 
 	// Ensure paths are absolute.
--- a/go/tools/builders/stdliblist.go
+++ b/go/tools/builders/stdliblist.go
@@ -158,12 +158,12 @@ func flatPackageForStd(cloneBase string, pkg *goListPackage) *flatPackage {
 // To work around this, cloneGoRoot creates a copy of external/go_sdk into a new
 // cloneBase directory while retaining its path relative to the root directory.
 // So "$OUTPUT_BASE/external/go_sdk" becomes
-// tmp_directory/external/go_sdk", which will be set at GOROOT later.
+// {cloneBase}/external/go_sdk", which will be set at GOROOT later.
 // This ensures that file paths in the generated JSON are still valid.
 //
 // cloneGoRoot returns the new GOROOT we should run
 // under.
-func cloneGoRoot(execRoot, relativeGoroot string, cloneBase string) (newGoRoot string, err error) {
+func cloneGoRoot(execRoot, relativeGoroot, cloneBase string) (newGoRoot string, err error) {
 	goroot := filepath.Join(execRoot, relativeGoroot)
 	if err != nil {
 		return "", err
--- a/go/tools/builders/stdliblist.go
+++ b/go/tools/builders/stdliblist.go
@@ -18,8 +18,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"flag"
-	"fmt"
 	"go/build"
+	"io/ioutil"
 	"os"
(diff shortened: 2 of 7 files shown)


go/tools/builders/stdliblist.go
```diff
@@ -145,6 +149,35 @@ func flatPackageForStd(execRoot string, pkg *goListPackage) *flatPackage {
...
+// cloneGoRoot returns the new GOROOT we should run
+// under.
+func cloneGoRoot(execRoot, relativeGoroot, cloneBase string) (newGoRoot string, err error) {
+	goroot := filepath.Join(execRoot, relativeGoroot)
+	newGoRoot = filepath.Join(cloneBase, relativeGoroot)
+	if err := os.MkdirAll(newGoRoot, 01755); err != nil {
+		return "", err
+	}
+
+	if err := replicate(goroot, newGoRoot, replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
```
xytan0056, line 174: "pkg/tool" is needed because go list would fail without it
```
go: no such tool "compile"
go: no such tool "compile"
go: no such tool "compile"
go: no such tool "compile"
go: no such tool "compile"
stdliblist: error running subcommand external/go_sdk/bin/go: exit status 2
```
maybe it's used in internal packages [such as cfg](https://github.com/golang/go/blob/go1.18/src/cmd/go/internal/cfg/cfg.go#L397) which is then used in go commands

linzhp: Good finding. Yes, `go list` needs to call `compile`. Can you add comments to explain this finding?

go/tools/builders/stdliblist_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/bazelbuild/rules_go/go/tools/bazel"
+)
+
+func Test_stdliblist(t *testing.T) {
+	testDir := t.TempDir()
+	f, _ := ioutil.TempFile(testDir, "*")
```
linzhp, line 17: You don't need to create a temp file in a temp dir. `outJSON = filepath.Join(testDir, "out.json")` should be enough

go/tools/builders/stdliblist_test.go
```diff
@@ -0,0 +1,58 @@
...
+	// (go.sdk.root_file.dirname), thus setting wd to
+	// {runfile directory} so that go_sdk is discoverable
+	// {runfile directory} is the parent directory of bazel.RunfilesPath()
+	runFilesPath, err := bazel.RunfilesPath()
+	if err != nil {
+		t.Error("failed to find runfiles path")
+	}
+	test_args := []string{
+		fmt.Sprintf("-out=%s", f.Name()),
+		fmt.Sprintf("-sdk=%s", "go_sdk"),
```
linzhp, line 31: Can you pass the absolute path to go_sdk here? It should be available in `$TEST_SRCDIR/external/go_sdk`. With the absolute path, you don't need to pass `-wd`, right?

go/tools/builders/stdliblist_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func Test_stdliblist(t *testing.T) {
+	testDir := t.TempDir()
+	f, _ := ioutil.TempFile(testDir, "*")
+
+	// test files are at run file directory, but this test is run at
+	// {runfile directory}/bazel.TestWorkspace()
+	// since -sdk is assumed to be a relative path to execRoot
+	// (go.sdk.root_file.dirname), thus setting wd to
```
linzhp, line 22: Can make`-sdk` to take either relative path to current directory (execRoot in bazel action) or an absolute path?

go/tools/builders/stdliblist.go
xytan0056: done

go/tools/builders/stdliblist_test.go
xytan0056: done

go/tools/builders/stdliblist_test.go
xytan0056: `-sdk` is passed as relative path in stdlib.bazl, and stdliblist.go has this assumption too.
Also it looks like we can't get absolute path easily #2713

https://github.com/xytan0056/rules_go/blob/8b0dde29fb0dd2a7a7b35d6e06a43093f8716bc2/go/private/actions/stdlib.bzl#L55-L58

image

go/tools/builders/stdliblist_test.go
xytan0056: `-sdk` passed by `go.sdk.root_file.dirname` in stdlib.bzl, which is a relative path. See below comment

go/tools/builders/BUILD.bazel
```diff
@@ -23,6 +23,24 @@ go_test(
...
+        "flags.go",
+        "replicate.go",
+        "stdliblist.go",
+        "stdliblist_test.go",
+    ],
+    data = ["@go_sdk//:files"],
+    # -sdk in stdliblist needs to be a relative path, otherwise it breaks
+    # assumptions of cloning go_sdk, thus we need to set up the test in a
+    # way that go_sdk is under the directory where test is run.
+    rundir = "..",
```
xytan0056, line 40: this is an easy way to avoid having to break assumption of go_sdk to be absolute path.
by setting it this way, the structure will become
```
/a/b/c/runfiles/   <- this will be where the test is run with rundir being ..
    go_sdk         <- this will be where the data files are
    workspace      <- this will be where the test is run with rundir being . 
```
then naturally `go_sdk` is the relative path.

go/tools/builders/stdliblist_test.go
xytan0056: Actually -sdk needs to be a relative path, because during cloning, we need to **retain the same relative path of the original go_sdk**, i.e. `/old/root/`**`rel_go_sdk`**`/src/package` -> `/new/root/`**`rel_go_sdk`**`/src/package`

It cannot be an absolute path, because if so, the only way to get `rel_go_sdk` from goroot is `filepath.Rel(".", env.go_sdk)`, but this doesn't work for tests:
consider the following test file structure
```
/a/b/c/runfiles
/a/b/c/runfiles/go_sdk <- data files
/a/b/c/runfiles/workspace <- where test is run with rundir being "." in go_test rule
```
the relative path of go_sdk is `../go_sdk`;
Let's make `cloneBase := /tmp/123`;
Then in order to retain the same relative path ,`newGoRoot`  becomes `/tmp/123/../go_sdk`, which is `/tmp/go_sdk`;
Then `go list` under `newGoRoot` returns `pkg.Dir` being `/tmp/go_sdk/src/package`, which does not share a prefix with `cloneBase` and thus cannot be replaced with `__BAZEL_OUTBASE_LABEL__`
```
func labelledPath(cloneBase, p string) string {
        // ../go_sdk/src/package 
	dir, _ := filepath.Rel(cloneBase, p) 

        // __BAZEL_OUTPUT_BASE__/../go_sdk/src/package => go_sdk/src/package (wrong)
	return filepath.Join("__BAZEL_OUTPUT_BASE__", dir)  
}
```

The culprit here is `..` in the relative path. Instead of maniplulating the path, a simple way is to set `rundir=".."` in the go_test rule, so that all assumptions naturally fall into places.

linzhp: Why not just cloning from `/old/root/external/go_sdk` to `/new/root/external/go_sdk`?

linzhp: Never mind. https://github.com/xytan0056/rules_go/pull/1

xytan0056:
close in favor of #3119 3157

Issue #3168, opened 2022-06-01
Deprecate the unused compiling API


achew22:
Tracking bug for deprecating the unused compiling API that was removed when Jay switched over to having all the compilation/linking. We are going to add `print` warning statements into the `compile`, `compilePackage` et al to ask people to come check in on this bug.

@fmeum

fmeum:
Submitted #3172 to add deprecation warnings. #3173 is a CI test run to check that removing these actions is actually safe.

PR #3172, opened 2022-06-03
Deprecate legacy action APIs

fmeum:
The asm, compile, cover, and pack actions on go_context are no longer
used by rules_go itself and not covered by any tests. Unless there is
a very strong use case for them in our user base and someone steps up
to contribute test coverage, these are marked for removal in version
0.36.0.

This commit `print`s a warning if any of the deprecated actions is
invoked.

Note: `cover` has already been removed in 0.32.0.

Related to #3168


fmeum:
@achew22

PR #3173, opened 2022-06-03
Delete legacy actions API

### LINKED ISSUE #3168: Deprecate the unused compiling API

fmeum:
Fixes #3168


fmeum:
Merging for 0.37.0, one later than we promised.

PR #3174, opened 2022-06-03
Fix linker error if go_library with cgo references unresolved symbol

fmeum:
**What type of PR is this?**

Bug fix

Fix linker error if go_library with cgo references unresolved symbol
Before this commit, if a go_library referenced an unresolved C symbol,
the build would fail with an error about an undefined symbol since
compilepkg.go tries to link all object files into an executable to be
consumed by cgo's -dynimport argument.

Since this library is never used for any purpose other than having its
symbols parsed by the cgo command, retry with additional linker flags
intended to make the linker ignore unresolved symbols. If they remain
missing in the real link step, they will still be reported as unresolved
there.

Fixes https://github.com/golang/go/issues/25832 for rules_go.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -197,7 +197,30 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 	args = append([]string{cc, "-o", mainBin, mainObj}, cObjs...)
 	args = append(args, combinedLdFlags...)
 	if err := goenv.runCommand(args); err != nil {
-		return "", nil, nil, err
+		// If linking the binary for cgo fails, this is usually because the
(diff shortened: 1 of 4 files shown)


linzhp:
I think this makes the `go_library` target harder to use. Anyone who wants to import `:use_external_symbol` has to provide `external_symbol` somewhere. It sounds like a bad practice that we should discourage

fmeum:
@linzhp I agree the test itself is quite pathological, I just wanted to avoid having to add more source files to it and thus (ab)used CGo directives to get an unresolved symbol. 

There are realistic use cases of this pattern though: For example, at my company, we develop fuzzers based on LLVM's libFuzzer and would like to interact with its API from `go_library` targets via CGo. However, we can only link against the actual libFuzzer library from a `cc_binary` that bundles in the CGo libraries. Without allowing for external symbols to remain undefined (as is quite common for `cc_library`), this isn't possible.

PR #3179, opened 2022-06-05
Update Go to 1.18.3

fmeum:


sluongng:
From the log it seems like we need to upgrade/downgrade `org_golang_x_tools` ?

fmeum:
> From the log it seems like we need to upgrade/downgrade `org_golang_x_tools` ?

Yes, but afaik that requires updating the version of x_mod used by Gazelle first, which in turn requires a Gazelle release.

sluongng:
created https://github.com/bazelbuild/bazel-gazelle/issues/1283 to request for a newer Gazelle release

linzhp:
Can we make rules_go manage its own version of x_mod so this doesn't require Gazelle release?

fmeum:
We probably can by reordering WORKSPACE imports. I can give it a try next week. Worth noting that this painpoint will naturally go away with bzlmod.

sluongng:
I did a quick brain dump over in https://github.com/bazelbuild/rules_go/pull/3220

There seems to be an issue with a test in `x/tools` that im not yet sure how to solve.

```
Executing tests from @org_golang_x_tools//go/types/objectpath:objectpath_test
--- FAIL: TestGenericPaths (0.00s)
    objectpath_go118_test.go:90: Object(type comparable interface{comparable}) error was "predeclared type comparable interface{comparable} has no path", want "predeclared type comparable interface{} has no path"
```

But the rest should work.

sluongng:
> @org_golang_x_tools//go/types/objectpath:objectpath_test

Seems like 1.18.3 was ahead of go/x/tools v0.1.11 as the fix for this test is in master https://cs.opensource.google/go/x/tools/+/845bb90a1df4ce4789c794479a29a4ea07e12414

So it's safe to ignore the test 🤔

PR #3175, opened 2022-06-05
Prepare release v0.33.0

fmeum:
# Release Notes

## Breaking changes

* [cover](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#cover) has been removed in v0.32.0. Please [leave a comment](https://github.com/bazelbuild/rules_go/issues/3168) if you are affected by this.

## Deprecations

* The [asm](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#asm), [compile](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#compile), and [pack](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#pack) action generators provided by [go_context](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#go_context) are deprecated and planned for removal in version v0.36.0. Please leave a comment on the [tracking bug](https://github.com/bazelbuild/rules_go/issues/3168) if [archive](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#archive) and [link](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#link) are not suitable replacements for your use cases.

## Bug Fixes

* @sluongng fixed a race condition that could cause non-sandboxed builds of `go_test` targets to fail (https://github.com/bazelbuild/rules_go/pull/3145) 
* @abhinav made `//go:embed` work with `go_path` (https://github.com/bazelbuild/rules_go/pull/3163)
* @xytan0056 made gopackagesdriver work with Go 1.18 (https://github.com/bazelbuild/rules_go/pull/3157)
* @nickgooding ensured that `gomock` can be used with any Gazelle naming convention (https://github.com/bazelbuild/rules_go/pull/3155)
* `go_library` targets using CGo can now reference unresolved symbols (https://github.com/bazelbuild/rules_go/pull/3174) 

Thanks to all of the contributors!

## Compatibility

The minimum required version of Bazel remains at 4.2.1.

## Updated dependencies

* Updated `org_golang_x_sys`, `org_golang_x_xerrors`, `org_golang_google_genproto`, `go_googleapis` to their most recent commit as of 2022-06-05

As always, you can use higher versions of rules_go's dependencies by declaring them in WORKSPACE before calling go_rules_dependencies. Lower versions may work but are not supported.

## WORKSPACE code

// TBD after release

PR #3180, opened 2022-06-06
releaser: fix scrubbing timestamp from patch files

sluongng:
Make sure that we scrub the time zone correctly for the line started with `---` as well.

This should keep the release PR a lot smaller and easier to review.
See sample of current release PR in https://github.com/bazelbuild/rules_go/pull/3175#issuecomment-1147463053 where this problem occur.


sluongng:
cc: @fmeum @linzhp QOL for maintainers 😄

linzhp (approved):
Thanks!

PR #3186, opened 2022-06-07
Link in native libraries of transitive dependencies in archive mode

fmeum:
**What type of PR is this?**

Bug fix

When linking a go_binary in one of the native archive modes, all
native libraries of transitive dependencies should be included in the
resulting CcInfo, not just the libraries obtained from the cdeps.

This commit adds a `cc_info` field to the `GoSource` provider that
includes the result of merging the `CcInfo`s of all `deps` and `cdeps`
for all Go libraries, not just for those using cgo themselves.


go/private/rules/binary.bzl
```diff
@@ -163,7 +163,7 @@ def _go_binary_impl(ctx):
             cc_import_kwargs["alwayslink"] = True
         ccinfo = new_cc_import(go, **cc_import_kwargs)
         ccinfo = cc_common.merge_cc_infos(
-            cc_infos = [ccinfo] + [d[CcInfo] for d in source.cdeps],
+            cc_infos = [ccinfo, source.cc_info],
```
achew22, line 166: Much cleaner 👍

go/providers.rst
```diff
@@ -204,6 +204,10 @@ method. In general, only rules_go should need to build or handle these.
 +--------------------------------+-----------------------------------------------------------------+
 | The exposed cc headers for these sources.                                                        |
 +--------------------------------+-----------------------------------------------------------------+
+| :param:`cc_info`               | :type:`CcInfo`                                                  |
++--------------------------------+-----------------------------------------------------------------+
+| The result of merging the ``CcInfo``s of all `deps` and `cdeps`                                  |
```
achew22, line 209: Out of curiosity, is this autogenerated? If not is there any way to declare the outputs of our rules that would result in Stardoc being able to pick this up?

achew22 (approved):
LGTM, can we have @linzhp try this in a bigger repo and see what happens before merging?

go/providers.rst
fmeum: I added this manually. I think that only the rules and extra docs are generated automatically, everything else is handwritten. In fact, I think that Stardoc can only produce Markdown, not RST.

linzhp:
Assigned @kmicklas to review and test in Uber.

go/private/context.bzl
```diff
@@ -280,6 +283,19 @@ def _collect_runfiles(go, data, deps):
...
     return runfiles
 
+def _collect_cc_infos(deps, cdeps):
+    cc_infos = []
+    for dep in cdeps:
+        if CcInfo in dep:
+            cc_infos.append(dep[CcInfo])
+    for dep in deps:
+        # dep may be a struct, which doesn't support indexing by providers.
+        if is_struct(dep):
```
kmicklas, line 293: Just so I understand, can you give an example of this?

fmeum: The deps may be providers rather than targets, see the note on `attr` at https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#library_to_source.

kmicklas (approved):

PR #3187, opened 2022-06-07
Do not print to stderr if cgo linking succeeds after retry

fmeum:
**What type of PR is this?**

Bug fix

The rety logic introduced by 6ab73ec5b still emitted the stderr of the
failing compiler execution before the retry. Now, this output is cached
and only emitted if the retry also fails.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -23,6 +23,7 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


go/tools/builders/cgo2.go
```diff
@@ -196,7 +197,8 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 	mainBin := filepath.Join(workDir, "_cgo_.o") // .o is a lie; it's an executable
 	args = append([]string{cc, "-o", mainBin, mainObj}, cObjs...)
 	args = append(args, combinedLdFlags...)
-	if err := goenv.runCommand(args); err != nil {
+	originalErrBuf := &bytes.Buffer{}
```
linzhp, line 200: nit: prefer `var originalErrBuf bytes.Buffer` ([style guide](https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs)) and use `&originalErrBuf` as needed. It also more explicit when you need a pointer instead of a value.

linzhp (approved):

linzhp (approved):

PR #3190, opened 2022-06-08
Use param files with go-protoc

### LINKED ISSUE #3188: protoc argument list too long

v0.32.0

v0.25.0

5.1.0

Yes.

Linux x86

### What did you do?

In a large project with several transitive protobuf dependencies, did a `bazel build ...:xxx_go_proto` (a go_proto_library target).

Since changing the OS limits is not feasible, expect protoc to be invoked with `@<filename>` option so that long argument list length is not a problem

### What did you see instead?

bazel build of a go_proto_library target fails with the below error:

```
2022/06/07 16:53:18 error running protoc: fork/exec bazel-out/k8-opt-exec-2B5CBBC6/bin/rules/proto/protoc: argument list too long
```

The length of the args is about 220kB. It fails on linux, while it works on macOS; at this size.

### Proposed fix

Use `@<filename>` option instead of passing all the args in command line, only if the length is >100k. `@<filename>` should not be used for all cases as file IO would slow down the build. Though 100k seems to work it is, kind of, arbitrary. Any better suggestion is welcome.

fmeum:
**What type of PR is this?**

Bug fix

Makes the `GoProtocGen` action use parameter files if the argument list gets too long.

Fixes #3188

**Other notes for review**

Tested manually by temporarily setting `use_always = True` and running the tests.

### DIFF
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -176,7 +176,9 @@ func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
 // expandParamsFiles looks for arguments in args of the form
 // "-param=filename". When it finds these arguments it reads the file "filename"
 // and replaces the argument with its content.
-func expandParamsFiles(args []string) ([]string, error) {
+// It returns the expanded arguments as well as a bool that is true if any param
+// files have been passed.
+func expandParamsFiles(args []string) ([]string, bool, error) {
 	var paramsIndices []int
 	for i, arg := range args {
 		if strings.HasPrefix(arg, "-param=") {
@@ -184,7 +186,7 @@ func expandParamsFiles(args []string) ([]string, error) {
 		}
 	}
 	if len(paramsIndices) == 0 {
-		return args, nil
+		return args, false, nil
 	}
 	var expandedArgs []string
 	last := 0
@@ -195,12 +197,12 @@ func expandParamsFiles(args []string) ([]string, error) {
 		fileName := args[pi][len("-param="):]
 		fileArgs, err := readParamsFile(fileName)
 		if err != nil {
-			return nil, err
+			return nil, true, err
 		}
 		expandedArgs = append(expandedArgs, fileArgs...)
 	}
 	expandedArgs = append(expandedArgs, args[last:]...)
-	return expandedArgs, nil
+	return expandedArgs, true, nil
 }
 
 // readParamsFiles parses a Bazel params file in "shell" format. The file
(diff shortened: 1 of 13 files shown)


fmeum:
@linzhp Could you try this at Uber?

sateeshpnv:
Always using the param file can cause the build to slow down because of the file IO overhead. May not be visible in small projects. However, with 100's of dependencies it can get amplified.

In any case, I verified with this fix and it still fails with "argument list too long"

fmeum:
@sateeshpnv `use_param_files` without further keyword arguments only tells Bazel how to use a param file, it doesn't force it to use param files in all cases. That is what the (unset by default) parameter `use_always` is for.

Could you share logs of the failed command line, ideally with `-s`?

sateeshpnv:
It fails with

```
  bazel-out/k8-opt-exec-2B5CBBC6/bin/external/io_bazel_rules_go/go/tools/builders/go-protoc/go-protoc-bin '-param=bazel-out/k8-fastbuild/bin/idl/code.uber.internal/rds/bliss-hydrator/user_context/user_contextpb_go_proto_/gogoproto/code.uber.internal/rds/bliss-hydrator/user_context/user_context.pb.go-0.params')

...

2022/06/08 10:32:09 error running protoc: fork/exec bazel-out/k8-opt-exec-2B5CBBC6/bin/rules/proto/protoc: argument list too long
```

Content of the params file:
```
-protoc
bazel-out/k8-opt-exec-2B5CBBC6/bin/rules/proto/protoc
-importpath
...
```

Though param file is used by go-protoc-bin, I guess the problem is - the native protoc binary is still executed with a very long argument list.

Please see the #3191, which is working; not sure if that is the most elegant way, though.

sateeshpnv:
@fmeum ,

It works now. Thank you.

go/tools/builders/env.go
```diff
@@ -176,15 +176,15 @@ func runAndLogCommand(cmd *exec.Cmd, verbose bool) error {
 // expandParamsFiles looks for arguments in args of the form
 // "-param=filename". When it finds these arguments it reads the file "filename"
 // and replaces the argument with its content.
-func expandParamsFiles(args []string) ([]string, error) {
+func expandParamsFiles(args []string) ([]string, bool, error) {
```
linzhp, line 179: Can you also update the doc of this function for the additional `bool` return value?

linzhp (approved):
Because @sateeshpnv is from Uber, I believe he has tested this PR in Go monorepo :)

go/tools/builders/env.go
fmeum: Done.

linzhp (approved):

PR #3198, opened 2022-06-14
runfiles: remove deprecated api

sluongng:
These API have been deprecated since
Jun 3 2019 with 4442d82a001f378d0605cbbca3fb529979a1c3a6.

It has been 3 years since then, let's clean up the deprecation.


fmeum (approved):
Thanks for the cleanup. Thus has no chance to work on Windows, so let's get it out.

PR #3206, opened 2022-06-17
Fix failing open hermeticity test

fmeum:
Due to an incorrect JSON field name used in a test helper, the test
did not fail when a target is built in two configurations that differ
in rules_go settings.


fmeum:
@linzhp Could you take a look? This bug causes the test to pass even when it shouldn't for https://github.com/bazelbuild/rules_go/pull/3116.

PR #3207, opened 2022-06-18
Allow gomock to take Bazel common attributes

linzhp:
**What type of PR is this?**

Bug fix

Add a `kwargs` to gomock so it can take common attributes such as `visibility`


achew22 (approved):
I think you need to regenerate the docs because the signature changed, but otherwise LGTM

PR #1131, opened 2017-12-12
Fix aspect based proto builds

ianthehat:
This changes the structure of our providers.
It is now much more principled about configuration specific and transitive
providers, separating them cleanly.
It also allows the GoLibray provider to supply a function that can be used to
produce the configuration specific version (GoSource). This allows us to fix the
proto library dependancies when built in aspect mode.

PR #3219, opened 2022-06-28
new_library: remove unused resolver

sluongng:
`_testmain_library_to_source` was introduced in 1c4f6fd1 and last usage
of this resolver was in 56e5592c.

  ```
  > git log -S'_testmain_library_to_source' --oneline
  56e5592c Move the test library rule to be go_test internal actions (#1267)
  9031d58e Add GoContext (#1140)
  1c4f6fd1 Fix aspect based proto builds (#1131)
  ```

Let's remove the unused resolver and replace the documentation with the
resolver that is actually being used.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -54,9 +54,6 @@ load(
     "structs",
 )
 
-def _testmain_library_to_source(go, attr, source, merge):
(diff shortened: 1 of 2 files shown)


sluongng:
cc: @ianthehat FYI

linzhp (approved):

PR #3212, opened 2022-06-22
nogo: instantiate type info for generic types when running under Go >=1.18

farhaven:
This mirrors what golang.org/x/tools/go/packages does when loading packages.

Should fix #3211 (and probably #3164 as well).

**What type of PR is this?**

> Bug fix

**What does this PR do? Why is it needed?**

This PR adds an initialization that is required for the package loader in nogo to load instantiation information for generic types. This takes a lot of inspiration from what [the package loader in golang.org/x/tools](https://github.com/golang/tools/blob/master/go/packages/packages.go#L952) does.

**Which issues(s) does this PR fix?**

Fixes #3211 and #3164

**Other notes for review**

My company signed the CLA, but it looks like that fact has to percolate through something at Google.

PR #3231, opened 2022-07-07
asm: Pass package path with -p

### LINKED ISSUE #3199: Unable to build golang.org/x/sys with Go 1.19 beta1

v0.33.0

v0.25.0

```
% bazel version
Build label: 5.1.1
Build time: Fri Apr 8 15:49:48 2022 (1649432988)
Build timestamp: 1649432988
Build timestamp as int: 1649432988
```

Yes.

```
% uname -sm
Linux x86_64
```

### What did you do?

Attempt to build a binary that imports golang.org/x/sys with Go 1.19 beta1.

A successful build.

### What did you see instead?

```
ERROR: /home/user/src/tmp/hello/BUILD.bazel:11:10: GoLink hello/hello_/hello failed: (Exit 1): builder failed: error executing command bazel-out/k8-opt-exec-2B5CBBC6/bin/external/go_sdk/builder link -sdk external/go_sdk -installsuffix linux_amd64 -arc ... (remaining 17 arguments skipped)

2022/06/14 16:03:32 link: unlinkable object (from package golang.org/x/sys/unix) - compiler requires -p flag
link: error running subcommand external/go_sdk/pkg/tool/linux_amd64/link: exit status 1
Target //hello:hello failed to build
```

---

To reproduce, here's a copy of all files and the directory layout:

```
-- BUILD.bazel --
load("@bazel_gazelle//:def.bzl", "gazelle")

# gazelle:prefix example.com/demo
gazelle(name = "gazelle")

-- WORKSPACE --

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "685052b498b6ddfe562ca7a97736741d87916fe536623afb7da2824c0211c369",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.33.0/rules_go-v0.33.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.33.0/rules_go-v0.33.0.zip",
    ],
)

http_archive(
    name = "bazel_gazelle",
    sha256 = "5982e5463f171da99e3bdaeff8c0f48283a7a5f396ec5282910b9e8a49c0dd7e",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.25.0/bazel-gazelle-v0.25.0.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.25.0/bazel-gazelle-v0.25.0.tar.gz",
    ],
)

load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")

go_repository(
    name = "org_golang_x_sys",
    importpath = "golang.org/x/sys",
    sum = "h1:Zu/JngovGLVi6t2J3nmAf3AoTDwuzw85YZ3b9o4yU7s=",
    version = "v0.0.0-20220610221304-9f5ed59c137d",
)

go_rules_dependencies()

go_register_toolchains(version = "1.19beta1")

gazelle_dependencies()

# Workaround for https://github.com/bazelbuild/bazel-gazelle/issues/1217
go_repository(
    name = "org_golang_x_xerrors",
    importpath = "golang.org/x/xerrors",
    sum = "h1:go1bK/D/BFZV2I8cIQd1NKEZ+0owSTG1fDTci4IqFcE=",
    version = "v0.0.0-20200804184101-5ec99f83aff1",
)

-- go.mod --
module example.com/demo

go 1.18

require golang.org/x/sys v0.0.0-20220610221304-9f5ed59c137d

-- go.sum --
golang.org/x/sys v0.0.0-20220610221304-9f5ed59c137d h1:Zu/JngovGLVi6t2J3nmAf3AoTDwuzw85YZ3b9o4yU7s=
golang.org/x/sys v0.0.0-20220610221304-9f5ed59c137d/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=

-- hello/BUILD.bazel --

go_library(
    name = "hello_lib",
    srcs = ["hello.go"],
    importpath = "example.com/demo/hello",
    deps = ["@org_golang_x_sys//unix"],
)

go_binary(
    name = "hello",
    embed = [":hello_lib"],
)

-- hello/hello.go --
package main

import (
	"fmt"

	_ "golang.org/x/sys/unix"
)

func main() {
	fmt.Println("hello")
}
```

I got successful builds for the following cases:

- `:hello_lib`
- replace `/x/sys/unix` with a no-op local dependency `//foo:foo_lib` 
- replace `/x/sys/unix` with a simple external dependency (go.uber.org/atomic in this case)

So this is specifically an issue with how x/sys/unix is linked.

@sywhang thinks this is caused by https://github.com/golang/go/commit/e8534640557939755917edc3e6afc16d93b2e70b.

fmeum:
This is required to produce linkable objects as of Go 1.19.

Fixes #3199

### DIFF
--- a/go/tools/builders/asm.go
+++ b/go/tools/builders/asm.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -60,7 +61,10 @@ func asm(args []string) error {
 	}
 
 	// Build source with the assembler.
-	return asmFile(goenv, source, asmFlags, outPath)
+	// Note: As of Go 1.19, this would require a valid package path to work.
+	// But since this functionality is only used by the deprecated action API,
+	// we won't fix this.
+	return asmFile(goenv, source, "", asmFlags, outPath)
 }
 
 // buildSymabisFile generates a file from assembly files that is consumed
@@ -137,12 +141,34 @@ func buildSymabisFile(goenv *env, sFiles, hFiles []fileInfo, asmhdr string) (str
 	return symabisName, err
 }
 
-func asmFile(goenv *env, srcPath string, asmFlags []string, outPath string) error {
+func asmFile(goenv *env, srcPath, packagePath string, asmFlags []string, outPath string) error {
 	args := goenv.goTool("asm")
 	args = append(args, asmFlags...)
+	// The package path has to be specified as of Go 1.19 or the resulting
+	// object will be unlinkable, but the -p flag is also only available
+	// since Go 1.19.
+	if packagePath != "" && isGo119OrHigher() {
+		args = append(args, "-p", packagePath)
+	}
 	args = append(args, "-trimpath", ".")
 	args = append(args, "-o", outPath)
 	args = append(args, "--", srcPath)
 	absArgs(args, []string{"-I", "-o", "-trimpath"})
 	return goenv.runCommand(args)
+}
+
+var goMinorVersionRegexp = regexp.MustCompile(`^go1\.(\d+)`)
+
+func isGo119OrHigher() bool {
+	match := goMinorVersionRegexp.FindStringSubmatch(runtime.Version())
+	if match == nil {
+		// Developer version or something with an unparseable version string,
+		// assume Go 1.19 or higher.
+		return true
+	}
+	minorVersion, err := strconv.Atoi(match[1])
+	if err != nil {
+		return true
+	}
+	return minorVersion >= 19
 }
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -474,7 +474,7 @@ func compileArchive(
 		}
 		for i, sSrc := range srcs.sSrcs {
 			obj := filepath.Join(workDir, fmt.Sprintf("s%d.o", i))
-			if err := asmFile(goenv, sSrc.filename, asmFlags, obj); err != nil {
+			if err := asmFile(goenv, sSrc.filename, packagePath, asmFlags, obj); err != nil {
 				return err
 			}
 			objFiles = append(objFiles, obj)


go/tools/builders/asm.go
```diff
@@ -137,9 +140,12 @@ func buildSymabisFile(goenv *env, sFiles, hFiles []fileInfo, asmhdr string) (str
 	return symabisName, err
 }
 
-func asmFile(goenv *env, srcPath string, asmFlags []string, outPath string) error {
+func asmFile(goenv *env, srcPath, packagePath string, asmFlags []string, outPath string) error {
 	args := goenv.goTool("asm")
 	args = append(args, asmFlags...)
+	if packagePath != "" {
+		args = append(args, "-p", packagePath)
+	}
```
sluongng, line 148: This would make rules_go incompatible with lower version of go sdk (before 1.19) because `go tool asm` have yet to added this flag.

Do we need to do this in a separate file dedicated to 1.19?

fmeum: Good catch. I pushed a new commit where I only add the flag with Go 1.19+. For some reason, I do get duplicate definition build errors though - I'm probably using `//go:build` wrong. Do you know how I can fix it?

sywhang:
@abhinav is on vacation till a couple weeks later, so I took over the testing here at Uber. Can confirm that the fix works (minus the duplicate build errors part).

go/tools/builders/asm_go1.go
```diff
@@ -0,0 +1,8 @@
+//go:build !go1.19
```
btwiuse, line 1: If we use pre go1.18 to build, will the `//go:build` directive still work?

sluongng:
> For some reason, I do get duplicate definition build errors though - I'm probably using //go:build wrong. Do you know how I can fix it?

I tried to reproduce this with normal `go build` command

```
~/test> cat main.go
package main

import "fmt"

func main() {
    fmt.Println(Version)
}

~/test> cat lib_go1.go
//go:build !go1.19
// +build !go1.19

package main

var Version = "1.18"

~/test> cat lib_go119.go
//go:build go1.19
// +build go1.19

package main

var Version = "1.19"

~/test> cat go.mod
module temp

go 1.18

~/test> go version
go version go1.18.3 darwin/arm64
```

And then run this
```
~/test> GOCACHE=$(mktemp -d) go build -x .

...
/opt/homebrew/Cellar/go/1.18.3/libexec/pkg/tool/darwin_arm64/compile -o $WORK/b001/_pkg_.a -trimpath "$WORK/b001=>" -p main -lang=go1.18 -complete -buildid UPD40Vf15HmsGAFN49LI/UPD40Vf15HmsGAFN49LI -goversion go1.18.3 -shared -c=4 -nolocalimports -importcfg $WORK/b001/importcfg -pack ./lib_go1.go ./main.go
...
```

Note how `lib_go1.go` was included and not `lib_go119.go`.

So it seems like `go tool compile` is depending on `go build` to filter out the source file before hand?
And we might need to replicate that logic with rules_go?

sluongng:
It seems like this will affect https://github.com/bazelbuild/rules_go/pull/3212 as well.

There are a few ways we can go about this:

- Build tooling to support different implementation for different goSdk versions

- Commit to a non-backward compatible upgrade

The second option seems to be a lot easier:

- Make v0.34.0 to support Go 1.18 minimum, include https://github.com/bazelbuild/rules_go/pull/3212 in the release

- Make v0.35.0 to support Go 1.19 minimum, include this PR in the release.

sluongng:
cc: @thempatel who have been doing some work around build constraints over in Gazelle repo.

sluongng:
I think thats the sensible solution, but I don't like it personally. 😆 
It's possible, does not means we should.

Reason is simply "additional complexity" on top of existing platform + cgo on/off matrixes.
And to support this long term, the parser inside gazelle would also need to handle this properly. (yes, we already DO have a parser, see https://github.com/bazelbuild/bazel-gazelle/pull/1243).
Between both rules_go and gazelle, the active maintainers is quite limitted so I would vote to keep it simple so that we can focus on more important things.

Imo we only need to provide a release cadence where 1 version that is backward compatible with the old GoSDK. Meaning that the moment 1.19 hits release, we can safely say that 1.17 is no longer supported and move to 1.18, then shortly move to 1.19 in the version after that, with possible cherry-pick backporting of critical bugs via patch releases.

Folks who wish to adopt latest Go version earlier (i.e. during RC releases) should have the capacity to patch rules_go with the content of this PR to make it work.

fmeum:
Sounds reasonable to me in general.

But don't we have the special problem here that we need this change to support 1.19 but the change fails with versions before 1.19? This would be much simpler if `asm` had `-p` already in 1.18 as then we could just add support for 1.19 at the same time we drop support for 1.17.

sluongng:
See https://github.com/bazelbuild/rules_go/wiki/Deprecation-schedule#supported-go-versions

> The Go Team provides support for the current and previous minor versions of Go. For example, if Go 1.15 is the current version, then 1.15 and 1.14 are supported. Only supported versions receive bug fixes and security updates, so users are strongly encouraged to update to those versions.
> 
> Each minor version of rules_go only supports the minor versions of Go that were supported at the time of release. rules_go may break compatibility with older versions.

So I would suggest these:

1. We remove the multi-version support and leave this PR open for now. 
2. Add a description at top guiding people on how to patch their rules_go if they wish to use 1.19 early.
3. Complete and merge https://github.com/bazelbuild/rules_go/pull/3212 and create a new release for 1.18 compatibility. In which we announce next release to be on the same date with Go 1.19 
4. When Go 1.19 get released, merge this and release a new rules_go version targeting 1.19 onward.


I would love to hear other maintainers opinion on this as well. 🤔

fmeum:
@sywhang @sluongng I pushed a new version with a runtime check that should allow us to merge this right away. PTAL.

go/tools/builders/asm.go
```diff
@@ -137,12 +141,34 @@ func buildSymabisFile(goenv *env, sFiles, hFiles []fileInfo, asmhdr string) (str
...
+		// Developer version or something with an unparseable version string,
+		// assume Go 1.19 or higher.
+		return true
+	}
+	minorVersion, err := strconv.Atoi(match[1])
+	if err != nil {
+		return true
+	}
+	return minorVersion >= 19
+}
```
sluongng, line 174: A quick test on this approach for those who wonder "how would this work with patch versions"

```
~/test> cat main.go
package main

import (
    "fmt"
    "regexp"
    "runtime"
)

func main() {
    fmt.Println("version:", runtime.Version())

    goMinorVersionRegexp := regexp.MustCompile(`^go1\.(\d+)`)
    matches := goMinorVersionRegexp.FindStringSubmatch(runtime.Version())
    for i, m := range matches {
        fmt.Printf("match no %d: %s\n", i, m)
    }
}
~/test> go run main.go
version: go1.18.4
match no 0: go1.18
match no 1: 18
```

So it works :)

sluongng (approved):
Look obviously correct to me. 👍

go/tools/builders/asm.go
fmeum: Yeah, that's why I went with a regexp instead of simple split logic - I was worried of patch and rc/beta versions.

martinbaillie:
Would it be possible to cut an official release that has this in it? 🙏🏻

sywhang:
https://github.com/bazelbuild/rules_go/releases/tag/v0.34.0 has this fix in it.

martinbaillie:
So it does. My apologies!

PR #3232, opened 2022-07-11
bzlmod: Add support for gomock

fmeum:
This introduces a dependency on the new gazelle module, which requires
using Bazel at HEAD for c068b0d2347905c5e9be793b741f693c7edb6e53.


docs/go/extras/extras.md
```diff
@@ -98,7 +98,7 @@ If `source` is given, the mocks are generated in source mode; otherwise in refle
 | <a id="gomock-package"></a>package |  the name of the package the generated mocks should be in. If not specified, uses mockgen's default. See [mockgen's -package](https://github.com/golang/mock#flags) for more information.   |  <code>""</code> |
 | <a id="gomock-self_package"></a>self_package |  the full package import path for the generated code. The purpose of this flag is to prevent import cycles in the generated code by trying to include its own package. See [mockgen's -self_package](https://github.com/golang/mock#flags) for more information.   |  <code>""</code> |
 | <a id="gomock-aux_files"></a>aux_files |  a map from source files to their package path. This only needed when <code>source</code> is provided. See [mockgen's -aux_files](https://github.com/golang/mock#flags) for more information.   |  <code>{}</code> |
-| <a id="gomock-mockgen_tool"></a>mockgen_tool |  the mockgen tool to run.   |  <code>"@com_github_golang_mock//mockgen"</code> |
+| <a id="gomock-mockgen_tool"></a>mockgen_tool |  the mockgen tool to run.   |  <code>Label("//extras/gomock:mockgen")</code> |
```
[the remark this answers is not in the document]
fmeum: I think that they shouldn't have to: This is only the default value of a parameter and ultimately resolves to the same target in a repository provided by rules_go (both in the WORKSPACE and bzlmod setup). I may be missing something here, but the actual value should only matter if someone wanted to dig deeper and understand what exact tool we are using here, in which case they would have to look up the alias target definition within rules_go.

fmeum:
@linzhp Could you check that this doesn't break gomock for WORKSPACE users?

Now that the rules_go/gazelle module interaction seems to work well, I'll try to get go_proto_library working next.

achew22 (approved):

extras/gomock/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+alias(
+    name = "gomock",
+    actual = "@com_github_golang_mock//gomock",
```
linzhp, line 3: What's the purpose of introducing these aliases?

fmeum: With bzlmod, modules can only see repositories they explicitly depend on in their `MODULE.bazel`. Without there's aliases, rules_go users would need to depend on Gazelle themselves and import the `@com_github_golang_gomock` repo via a `use_repo` directive just to use the rules_go provided `gomock` macro.

That seemed very tedious, so I introduced these aliases instead.

linzhp (approved):
Passed tests in Uber

PR #3240, opened 2022-07-15
gopackagesdriver: Descend into go_proto_compiler's deps

### LINKED ISSUE #3239: gopackagesdriver skips protobuf well known types
When using gopackagesdriver, protobufs mostly work. But any field using one of google's well known types shows up as an error in the editor.

```proto
syntax = "proto3";

import "google/protobuf/wrappers.proto";

message Foo {
  string ok = 1;
  google.protobuf.Int32Value not_ok = 2;
}
```

I noticed that the .pkg.json file isn't getting generated for @org_golang_google_protobuf//types/known/wrapperspb
I believe this dependency comes from the deps of the go_proto_compiler used. So I tried adding "compiler" and "compilers" to the  go_pkg_info_aspect attr_aspects, but this didn't make any difference.
If I manually add "@org_golang_google_protobuf//types/known/wrapperspb" as a dep of my go_proto_library it will work, but then gazelle will remove this. Can we either get gazelle to add these implicit deps or get the aspect to understand them?

fmeum:
**What type of PR is this?**

Bug fix

This lets the aspect generate .pkg.json files for well-known type
targets such as `@org_golang_google_protobuf//types/known/wrapperspb`.

Fixes #3239

### DIFF
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -24,6 +24,16 @@ load(
 
 GoPkgInfo = provider()
 
+DEPS_ATTRS = [
+    "deps",
+    "embed",
+]
+
+PROTO_COMPILER_ATTRS = [
+    "compiler",
+    "compilers",
+]
+
 def _is_file_external(f):
     return f.owner.workspace_root != ""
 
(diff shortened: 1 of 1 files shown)


fmeum:
The BCR module got broken by bazel head, will fix it soon.

linzhp (approved):

PR #3242, opened 2022-07-19
stdlib: distinct GoStdlib actions

sluongng:
GoStdlib actions provide a set of compiled archives of Go's standard
library for downstream binaries to link against.  In order to support
multiple different compilation and link modes that rules_go users might
use, GoStdlib needs to be prepared in a combination of different
settings.  These combinations include, but not limited to, the
following settings:

- Different toolchains for different OS / CPU architectures
- With / without CGO enabled with different link modes
- With / without race detection enabled

Each unique combination that is dependent upon will generate a new
GoStdlib action in the build.  These actions are displayed to the user's
Bazel build progress outputs like this:

  ```
  GoStdlib stdlib_/pkg; 3s darwin-sandbox
  GoStdlib stdlib_/pkg; 0s darwin-sandbox
  GoStdlib stdlib_/pkg; 1s darwin-sandbox
  ```

This confuses users who are not familiar with rules_go internal into
thinking that the same action are being wrongly executed multiple times
wastefully.  As the result, troubleshooting the Action Graph became
harder and one must rely on the detail outputs in the Build Events
outputs to tell the differences between each actions.

Let's make the mnemonic of GoStdlib actions a bit more descriptive
regarding what's it's doing by including some of the settings mentioned
above into the mnemonic itself.  As the result, we should be able to
see more distinct actions in the Bazel build outputs depending on which
setting was used.

As a result, the actions mnemonic can be distinctly recognized:

  ```
  # Build and output Build Event JSON file
  rules_go> bazel clean
  rules_go> bazel build --build_event_json_file=bes.json --build_event_publish_all_actions //...
  rules_go> grep 'GoStdlib' bes.json |\
              jq 'select(.id | keys[0] == "actionCompleted") | .action.type + " " + .action.label' |\
              sort |\
              uniq -c |\
              sort -n

  # Before this patch
     1 "GoStdlibList //:stdlib"
    14 "GoStdlib //:stdlib"

  # After this patch
    1 "GoStdlibDarwinAmd64 //:stdlib_pure"
    1 "GoStdlibDarwinArm64CGoLinkCArchive //:stdlib_cgo"
    1 "GoStdlibDarwinArm64CGoLinkCShared //:stdlib_cgo"
    1 "GoStdlibDarwinArm64CGoLinkPie //:stdlib_cgo"
    1 "GoStdlibDarwinArm64CGoLinkPlugin //:stdlib_cgo"
    1 "GoStdlibLinuxAmd64 //:stdlib_pure"
    1 "GoStdlibList //:stdlib_cgo"
    1 "GoStdlibPlan9Amd64 //:stdlib_pure"
    1 "GoStdlibPlan9Arm //:stdlib_pure"
    1 "GoStdlibWindowsAmd64 //:stdlib_pure"
    2 "GoStdlibDarwinArm64RaceCGo //:stdlib_cgo"
    3 "GoStdlibDarwinArm64 //:stdlib_cgo"
  ```

Note how this still not yet fully de-duplicate the mnemonic.  This is
likely due to the fact that different version of `builder` binaries was
used. But it's unclear to me why that is and whether these actions are
truly duplicated?

  ```
  > grep 'GoStdlib' bes.json |\
      jq 'select(.id | keys[0] == "actionCompleted") |\
            select( .action.type == "GoStdlibDarwinArm64") |\
            select(.action.label == "//:stdlib_cgo") |\
            .action.commandLine[0]'
  "bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-36745079df94/bin/external/go_sdk/builder"
  "bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-6ca1a6e450d1/bin/external/go_sdk/builder"
  "bazel-out/darwin_arm64-opt-exec-2B5CBBC6-ST-149cbabf6624/bin/external/go_sdk/builder"
  ```


fmeum:
My personal opinion is that we should diversify `progress_message` rather than `mnemonic` (e.g. "Compiling Go stdlib with race..."), but I'm also open to doing the latter.

We should definitely try to use the various reset transitions, maybe we can get rid of some variations entirely.

sluongng:
> My personal opinion is that we should diversify `progress_message` rather than `mnemonic` (e.g. "Compiling Go stdlib with race..."), but I'm also open to doing the latter.

I think the differences between `progress_message` and `mnemonic` is the ability to programatically filter `mnemonic()` function when using `bazel aquery`, which is quite useful when you want to investigate the Action Graph.

```
> bazel aquery 'mnemonic("^GoStdlib$", deps(//...))' --include_artifacts=false

# VS

> bazel aquery 'mnemonic("^GoStdlibDarwinArm64CGoLinkCShared$", deps(//...))' --include_artifacts=false
```

So I would prefer to use `mnemonic` here unless there is a strong objection.

---

> We should definitely try to use the various reset transitions, maybe we can get rid of some variations entirely.

I am not too familiar with Bazel Transition (only read through the text, have yet to write one).

However I do know `bazel aquery`:

```
> bazel aquery 'mnemonic("^GoStdlibDarwinArm64$", deps(//...) union //...)' --include_artifacts=false | grep Configuration
  Configuration: darwin_arm64-fastbuild-ST-149cbabf6624
# Configuration: 6e115a77edf6ce619ff5d91a261a256a6c21ef061ac79c78941193d9a45f1ce4
  Configuration: darwin_arm64-fastbuild-ST-36745079df94
# Configuration: b1c88c3680c6ca6f4aa7023d7781b6ee0eb46b2b04793edc9efe451e21aa9b5e
  Configuration: darwin_arm64-fastbuild-ST-6ca1a6e450d1
# Configuration: 1e4194f1fc97994902ed50903f376d088556af91ec542c8c1c28ea8d6ce5bda7

> bazel config 6e115a77edf6ce619ff5d91a261a256a6c21ef061ac79c78941193d9a45f1ce4 > cfg1.txt
> bazel config b1c88c3680c6ca6f4aa7023d7781b6ee0eb46b2b04793edc9efe451e21aa9b5e > cfg2.txt
> bazel config 1e4194f1fc97994902ed50903f376d088556af91ec542c8c1c28ea8d6ce5bda7 > cfg3.txt

> diff -u cfg1.txt cfg3.txt
... <omit hash differences>
@@ -21,7 +21,7 @@
 }
 FragmentOptions com.google.devtools.build.lib.analysis.config.CoreOptions {
   action_env: []
-  affected by starlark transition: [//go/config:pure]
+  affected by starlark transition: [//go/config:pure, //go/private/rules:original_pure]
   allow_analysis_failures: false
   analysis_testing_deps_limit: 800
   archived_tree_artifact_mnemonics_filter: -(?:(?>.*))
@@ -387,4 +387,5 @@
 }
 FragmentOptions user-defined {
   //go/config:pure: true
+  //go/private/rules:original_pure: false
 }
 
 > diff -u cfg1.txt cfg2.txt
... <omit hash differences>
@@ -21,7 +21,7 @@
 }
 FragmentOptions com.google.devtools.build.lib.analysis.config.CoreOptions {
   action_env: []
-  affected by starlark transition: [//go/config:pure]
+  affected by starlark transition: [//go/config:pure, //go/private/rules:original_pure, //go/private:request_nogo]
   allow_analysis_failures: false
   analysis_testing_deps_limit: 800
   archived_tree_artifact_mnemonics_filter: -(?:(?>.*))
@@ -387,4 +387,6 @@
 }
 FragmentOptions user-defined {
   //go/config:pure: true
+  //go/private/rules:original_pure: false
+  //go/private:request_nogo: true
 }
```

So I think you are right, transition are causing the duplicated GoStdlib actions. No clue on how to fix it though.

I wonder if I could tap into the list of current active transition(s) any how within the Starlark API to add that info into action's `progress_message`? 🤔 

That would be a solid way to solve this for good.

sluongng:
@achew22 I don't see how this would impact Bazel aspects. AFAIK you don't use aspects with action's mnemonic but with rules' attribute.

I did a quick search for impact via both of the top opensource code search engine and found nothing.
- https://sourcegraph.com/search?q=context:global+GoStdlib&patternType=literal&case=yes
- https://cs.github.com/?q=%22GoStdlib%22

Though due to Hyrum's Law, it's hard to definitely say for sure that non of the downstream consumer will be impacted.
Do you have any suggestion regarding who would be considered/classified as `major` consumer of rules_go?

Perhaps Uber's folks would want to weight into this? cc: @linzhp 

---

With that said, your comment reminded me that this PR is missing a doc change in https://github.com/bazelbuild/rules_go/blob/efc3212592320c1ab7f986c9a7882770ee06ad3b/go/toolchains.rst 

I will be addressing this in the next iteration of this PR.

sluongng:
Scratch that, I don't think we have a doc specifically for `Action's mnemonic` of `GoStdlib`.

We do have docs for `ToolchainInfo` as well as a `Provider` with similar name: `GoStdLib`. Confusing eh? 😂

linzhp:
> Do you have any suggestion regarding who would be considered/classified as major consumer of rules_go?
> Perhaps Uber's folks would want to weight into this? 

Sorry for the delayed reply. Uber doesn't filter GoStdLib with mnemonic. So I am OK with this.

sluongng:
I ran into this class of issue again today when enabling multiple platforms compilation for the same go_binary.

```
[
  go_binary(..., pure="on")
  for os, arch in (...)
]
```

This causes the same target `//lib/a:a_library` to be compiled multiple times under the same mnemonic + target pattern combination.
It would improve user experience quite a bit if we attach platform information to the mnemonic (or have Bazel prints out target platform information somehow).

fmeum:
@sluongng What do you think of starting with a custom progress message including information about the target platform and mode?

sluongng:
Let's close this for now as there has been pretty good effort into deduplicating stdlib.

We could create a different PR if we want the relevant information in the progress message.

PR #3257, opened 2022-08-01
README: Fix sha256 of io_bazel_rules_go

philwo:
While trying out the instructions, I noticed that the sha256 of the v0.32.0.zip as served by mirror.bazel.build and github.com is different from the one shown in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -276,7 +276,7 @@ Add the ``bazel_gazelle`` repository and its dependencies to your
 
     http_archive(
         name = "io_bazel_rules_go",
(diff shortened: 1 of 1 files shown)


philwo:
Oh, also I just realized that the README still refers to 0.32.0 while the latest version is 0.34.0? Should we just update these snippets in general?

fmeum:
Look like the gazelle section didn't get updated to v0.34.0. Could you do that in this PR?

fmeum:
Closing as we are now on 0.36.0

PR #3260, opened 2022-08-02
Add support for --@io_bazel_rules_go//go/toolchain:sdk_version flag.

JamesMBartlett:
**What type of PR is this?**
Feature

**What does this PR do? Why is it needed?**
This PR adds a build flag `--@io_bazel_rules_go//go/toolchain:sdk_version`, that allows changing which go SDK version to use to build targets. Implementation details:
- Each go_sdk now has config_settings that specify different values of `sdk_version`. Eg. a `go1.17.3` SDK would declare a config_setting for `sdk_version="1"`, `sdk_version="1.17"`, `sdk_version="1.17.3"`, and `sdk_version=""`.
- Then each toolchain that is registered for an SDK will have a config_setting_group constraint that combines the above config_settings. In the example given above, the go1.17.3 SDK would have a toolchain that only gets selected if sdk_version is 1, 1.17, 1.17.3 or the empty string (default).
- If sdk_version flag is not specified, then all SDKs can be selected because of the inclusion of the empty string config_setting in the config_setting_group.

**Which issues(s) does this PR fix?**
Partially #3202 
Second diff coming to fully close that issue.

**Other notes for review**
See #3202 for more discussion about this feature.

Working with my employer on signing the CLA at the moment.

Signed-off-by: James Bartlett <jamesbartlett@pixielabs.ai>

PR #3261, opened 2022-08-02
Add `go_cross_binary` rule for cross-compilation.

### LINKED ISSUE #3202: [FR/Proposal] Support use of multiple SDK versions in the same bazel workspace
I didn't see an FR issue template and the bug template didn't seem relevant, so sorry for not using the template.

### Use Case
[Pixie](https://github.com/pixie-io/pixie) would like to be able to build go libraries/binaries with different versions of the go SDK. Pixie has to work with multiple versions of user's go binaries, so we use go binaries built by different versions of go as data inputs to our tests to ensure our code can interact with go binaries regardless of what version built them. 

### Proposal
I've created a patch of rules_go, that allows specifying `gosdk` (as a label to the GoSDK provider) on any `go_binary` or `go_test` target. Then using a bazel transition all of that targets dependencies will be built with the given SDK. By default the first `go_download_sdk/go_host_sdk/go_.*_sdk` rule will become the default SDK to use for all rules that don't explicitly specify `gosdk`.

I've attached a link to the patch below, if you are OK with this sort of change I can add tests, open a PR and continue review there, but I figured I should start a discussion here first since its a non-trivial change. (Note that this patch is based off of v0.32, but I can easily rebase for a PR)

[Patch](https://github.com/pixie-io/pixie/blob/6de646d2a772ae4ef4fac621044750653bed095b/bazel/external/rules_go.patch)

You can also see this patch in action in our build files [here](https://github.com/pixie-io/pixie/blob/6de646d2a772ae4ef4fac621044750653bed095b/src/stirling/obj_tools/testdata/go/BUILD.bazel#L40)

Also, I'm happy to work on a different solution to building with multiple SDK versions, if people are unhappy with this solution.

### Related Issues
This feature request is similar to #2527.

JamesMBartlett:
**What type of PR is this?**
Feature

This PR adds a `go_cross_binary` rule, that allows cross compiling `go_binary` generated targets to different platforms and/or different versions of the go SDK. See #3202 for the desired use case this supports. Implementation details:
- The `go_cross_binary` rule wraps a `go_binary` and transitions on `//command_line_option:platforms` and `@io_bazel_rules_go//go/toolchain:sdk_version` based on the `platform` and `sdk_version` attrs of the `go_cross_binary` rule. 
- Adds testing in `tests/core/cross` that ensures that `go_cross_binary` targets get built for the correct platform and sdk_version.

Fixes #3202

**Other notes for review**
This is the second of two diffs to support the use case in #3202. The first is here #3260. So once the other one lands this will not include the first commit.

Working with my employer at the moment to sign the CLA.

### DIFF
--- a/go/private/rules/cross.bzl
+++ b/go/private/rules/cross.bzl
@@ -1,0 +1,141 @@
+# Copyright 2014 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load(
+    "//go/private/rules:transition.bzl",
+    "go_cross_transition",
+)
+load(
+    "//go/private:providers.bzl",
+    "GoArchive",
+    "GoLibrary",
+    "GoSource",
+)
+
+def _is_windows(ctx):
+    return ctx.configuration.host_path_separator == ";"
+
+WINDOWS_ERR_SCRIPT = """
+@echo off
+>&2 echo {}
+exit /b 1
+"""
+UNIX_ERR_SCRIPT = """
+>&2 echo '{}'
+exit 1
+"""
+
+def _error_script(ctx):
+    errmsg = 'cannot run go_cross target "{}": underlying target "{}" is not executable'.format(
+        ctx.attr.name,
+        ctx.attr.target.label,
+    )
+    if _is_windows(ctx):
+        error_script = ctx.actions.declare_file("error.bat")
+        ctx.actions.write(error_script, WINDOWS_ERR_SCRIPT.format(errmsg), is_executable = True)
+        return error_script
+
+    error_script = ctx.actions.declare_file("error")
+    ctx.actions.write(error_script, UNIX_ERR_SCRIPT.format(errmsg), is_executable = True)
+    return error_script
+
+def _go_cross_impl(ctx):
+    old_default_info = ctx.attr.target[DefaultInfo]
+    old_executable = old_default_info.files_to_run.executable
+
+    new_default_info = None
+    if old_executable:
+        # Bazel requires executable rules to created the executable themselves,
+        # so we create a symlink in this rule so that it appears this rule created its executable.
+        new_executable = ctx.actions.declare_file(ctx.attr.name)
+        ctx.actions.symlink(output = new_executable, target_file = old_executable)
+        new_default_info = DefaultInfo(
+            files = depset([new_executable]),
+            runfiles = old_default_info.default_runfiles,
+            executable = new_executable,
+        )
+    else:
+        # There's no way to determine if the underlying `go_binary` target is executable at loading time
+        # so we must set the `go_cross` rule to be always executable. If the `go_binary` target is not
+        # executable, we set the `go_cross` executable to a simple script that prints an error message
+        # when executed. This way users can still run a `go_cross` target using `bazel run` as long as
+        # the underlying `go_binary` target is executable.
+        error_script = _error_script(ctx)
+
+        # See the implementation of `go_binary` for an explanation of the need for default vs data runfiles here.
+        new_default_info = DefaultInfo(
+            files = depset([error_script] + old_default_info.files.to_list()),
+            default_runfiles = old_default_info.default_runfiles,
+            data_runfiles = old_default_info.data_runfiles.merge(ctx.runfiles([error_script])),
+            executable = error_script,
+        )
+
+    providers = [
+        ctx.attr.target[provider]
+        for provider in [
+            GoLibrary,
+            GoSource,
+            GoArchive,
+            OutputGroupInfo,
+            CcInfo,
+        ]
+        if provider in ctx.attr.target
+    ]
+    return [new_default_info] + providers
+
+_go_cross_kwargs = {
+    "implementation": _go_cross_impl,
+    "attrs": {
+        "target": attr.label(
+            doc = """Go binary target to transition to the given platform and/or sdk_version.
+            """,
+            providers = [GoLibrary, GoSource, GoArchive],
+            mandatory = True,
+        ),
+        "platform": attr.label(
+            doc = """The platform to cross compile the `target` for.
+            If unspecified, the `target` will be compiled with the
+            same platform as it would've with the original `go_binary` rule.
+            """,
+        ),
+        "sdk_version": attr.string(
+            doc = """The golang SDK version to use for compiling the `target`.
+            Supports specifying major, minor, and/or patch versions, eg. `"1"`,
+            `"1.17"`, or `"1.17.1"`. The first Go SDK provider installed in the
+            repo's workspace (via `go_download_sdk`, `go_wrap_sdk`, etc) that
+            matches the specified version will be used for compiling the given
+            `target`. If unspecified, the `target` will be compiled with the same
+            SDK as it would've with the original `go_binary` rule.
+            Transitions `target` by changing the `--@io_bazel_rules_go//go/toolchain:sdk_version`
+            build flag to the value provided for `sdk_version` here.
+            """,
+        ),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+    },
+    "cfg": go_cross_transition,
+    "doc": """This wraps an executable built by `go_binary` to cross compile it
+    for a different platform, and/or compile it using a different version
+    of the golang SDK.<br><br>
+    **Providers:**
+    <ul>
+      <li>[GoLibrary]</li>
+      <li>[GoSource]</li>
+      <li>[GoArchive]</li>
+    </ul>
+    """,
+}
+
+go_cross = rule(executable = True, **_go_cross_kwargs)
--- a/go/private/rules/cross.bzl
+++ b/go/private/rules/cross.bzl
@@ -1,4 +1,4 @@
-# Copyright 2014 The Bazel Authors. All rights reserved.
+# Copyright 2022 The Bazel Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/go/private/rules/cross.bzl
+++ b/go/private/rules/cross.bzl
@@ -138,4 +138,4 @@ _go_cross_kwargs = {
     """,
 }
 
-go_cross = rule(executable = True, **_go_cross_kwargs)
+go_cross_binary = rule(executable = True, **_go_cross_kwargs)
--- a/tests/core/cross/BUILD.bazel
+++ b/tests/core/cross/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_cross", "go_library", "go_test")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_cross_binary", "go_library", "go_test")
 load("@io_bazel_rules_go//go/tools/bazel_testing:def.bzl", "go_bazel_test")
 load(":def.bzl", "no_context_info")
 
@@ -40,19 +40,19 @@ go_binary(
     deps = [":platform_lib"],
 )
 
-go_cross(
+go_cross_binary(
     name = "windows_go_cross",
     platform = "@io_bazel_rules_go//go/toolchain:windows_amd64",
     target = ":native_bin",
 )
 
-go_cross(
+go_cross_binary(
     name = "linux_go_cross",
     platform = "@io_bazel_rules_go//go/toolchain:linux_amd64",
     target = ":native_bin",
 )
 
-go_cross(
+go_cross_binary(
     name = "darwin_go_cross",
     platform = "@io_bazel_rules_go//go/toolchain:darwin_amd64",
     target = ":native_bin",
@@ -90,7 +90,7 @@ go_test(
 )
 
 go_test(
-    name = "go_cross_test",
+    name = "go_cross_binary_test",
     size = "small",
     srcs = ["cross_test.go"],
     args = [
--- a/tests/core/cross/BUILD.bazel
+++ b/tests/core/cross/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_cross", "go_library", "go_test")
 load("@io_bazel_rules_go//go/tools/bazel_testing:def.bzl", "go_bazel_test")
 load(":def.bzl", "no_context_info")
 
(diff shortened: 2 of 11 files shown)


go/private/rules/cross.bzl
```diff
@@ -0,0 +1,74 @@
...
+     "GoLibrary",
+     "GoSource",
+     "GoArchive",
+)
+
+def _go_cross_impl(ctx):
+    old_default_info = ctx.attr.target[DefaultInfo]
+    old_executable = old_default_info.files_to_run.executable
+    if not old_executable:
+        fail('target must have executable')
```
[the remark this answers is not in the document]
JamesMBartlett: Do you have ideas on how switch to between a executable and non executable version of `go_cross`. I see how `go_binary` does it in wrappers.bzl. However, it seems to me like there’s no way for a `go_cross_wrapper` macro to access the attributes of the `go_binary` specified by the `target` attribute. I could use `native.existing_rule` but as far as I can tell that would only work if the rule the `target` Label refers to is in the same package as the `go_cross` rule.

go/private/rules/cross.bzl
```diff
@@ -0,0 +1,74 @@
...
+    new_executable = ctx.actions.declare_file(ctx.attr.name)
+    ctx.actions.symlink(output=new_executable, target_file=old_executable)
```
sluongng, line 28: Im a bit confused by this as it reads as:
- declare a new binary executable file
- the new binary executable file is a symlink to the old executable

If you don't mind, I would love to get 1-2 line of comments on top of this explaining what's happening. 🙏

tests/core/cross/BUILD.bazel
```diff
@@ -33,6 +33,31 @@ go_binary(
...
+    target = ":native_bin",
+    platform = "@io_bazel_rules_go//go/toolchain:linux_amd64", 
+)
+
+go_cross(
+    name = "darwin_go_cross",
+    target = ":native_bin",
+    platform = "@io_bazel_rules_go//go/toolchain:darwin_amd64",
+)
+
```
sluongng, line 60: Nit:

This makes sense from a technical implementation perspective. The ability to compose your requirement platforms and use them in `go_cross` is huge.

But from an end-user perspective, it's confusing to read and to use.

Let's say that I setup my targets like this:

```
go_binary(
    name = "native_bin",
    srcs = ["main.go"],
    pure = "on",
    deps = [":platform_lib"],
)

go_cross(
    name = "darwin_go_cross",
    target = ":native_bin",
    platform = "@io_bazel_rules_go//go/toolchain:darwin_amd64",
)
```

which is very typical for normal development nowaday.

What happen when I call `bazel build :all` on a Mac? Would both binaries be identical SHA256 wise?

Compare this to the current setup where we have

```
go_binary(
    name = "linux_bin",
    srcs = ["main.go"],
    goos = "linux",
    deps = [":platform_lib"],
)

go_binary(
    name = "macos_bin",
    srcs = ["main.go"],
    goos = "darwin",
    deps = [":platform_lib"],
)
```

Then I would argue that the current setup is cleaner to read / consume for end-user.

So my question here is: is this the best we could do in term of UX for go_cross?
Perhaps, there is a macro design that could help abstract away go_cross + go_binary into 1 single declaration in a BUILD file? 🤔

tests/core/cross/BUILD.bazel
```diff
@@ -33,6 +33,31 @@ go_binary(
     deps = [":platform_lib"],
 )
 
+go_binary(
+    name = "native_bin",
+    srcs = ["main.go"],
+    pure = "on",
```
sluongng, line 39: Is `pure` needed here? If yes, I think it should be at least documented.

tests/core/cross/BUILD.bazel
JamesMBartlett: I would argue that in use cases that have multiple platforms/sdk version combinations its much easier/cleaner to use the `go_cross` setup, especially if the binary has a lot of options eg:
```
go_binary(
  name = "linux_bin",
  goos = "linux",
  srcs = ["main.go"],
  deps = [":lib"],
  gc_goopts = [
    "-N",
    "-l",
    "-more-options-here",
    "-lots-of-options",
    ...
  ],
  gc_linkopts = [
    "-more-opts-in-links",
    "-opt",
    ...
  ],
  data = [
    ":data_dependencies",
    ":more_deps",
    ":makes_the_options_longer",
    ...
  ],
)

go_binary(
  name = "macos_bin",
  srcs = ["main.go"],
  goos = "darwin",
  deps = [":lib"],
  gc_goopts = [
    "-N",
    "-l",
    "-more-options-here",
    "-lots-of-options",
    ...
  ],
  gc_linkopts = [
    "-more-opts-in-links",
    "-opt",
    ...
  ],
  data = [
    ":data_dependencies",
    ":more_deps",
    ":makes_the_options_longer",
    ...
  ],
)

go_binary(
  name = "windows_bin",
  srcs = ["main.go"],
  goos = "windows",
  deps = [":lib"],
  gc_goopts = [
    "-N",
    "-l",
    "-more-options-here",
    "-lots-of-options",
    ...
  ],
  gc_linkopts = [
    "-more-opts-in-links",
    "-opt",
    ...
  ],
  data = [
    ":data_dependencies",
    ":more_deps",
    ":makes_the_options_longer",
    ...
  ],
)
```
versus
```
go_binary(
  name = "native_bin",
  srcs = ["main.go"],
  deps = [":lib"],
  gc_goopts = [
    "-N",
    "-l",
    "-more-options-here",
    "-lots-of-options",
    ...
  ],
  gc_linkopts = [
    "-more-opts-in-links",
    "-opt",
    ...
  ],
  data = [
    ":data_dependencies",
    ":more_deps",
    ":makes_the_options_longer",
    ...
  ],
)

go_cross(
  name = "linux_bin",
  target = ":native_bin",
  platform = "//path_to:my_linux_platform",
)
go_cross(
  name = "macos_bin",
  target = ":native_bin",
  platform = "//path_to:my_macos_platform",
)
go_cross(
  name = "windows_bin",
  target = ":native_bin",
  platform = "//path_to:windows_platform",
)
```

But I will defer to @fmeum

tests/core/cross/BUILD.bazel
JamesMBartlett: It is only required in this test so that there's no dependency on darwin/linux/windows cpp toolchains. However, for an arbitrary user of `go_cross` the binary is not required to be pure

go/private/rules/cross.bzl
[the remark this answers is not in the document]
JamesMBartlett: I updated the diff to have a `go_cross` wrapper macro that tries to find the target in the local package, but if it fails it suggests that the user explicitly call `go_cross_executable` or `go_cross_non_executable`. This feels a little gross to me, so let me know if you have a better suggestion.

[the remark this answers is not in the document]
fmeum: Thanks for trying this out. It also feels rather invasive to me, so I would prefer a different solution.

Simple idea first: How common is it for `go_cross` targets to be `bazel run`? With `platform` set, it would be relatively unlikely that the target platform matches host.

Next simplest idea: Unconditionally mark the rule as executable, but write and advertise a bash/bat script or simple Go binary that prints an error message when executed in case the link mode doesn't provide you with an actual `executable`.

tests/core/cross/BUILD.bazel
fmeum: I like the design of `go_cross` as it is, but we could think about introducing a `go_cross_binary` macro wrapper that automatically adds `tags = ["manual"]` to the `go_binary`. @sluongng What do you think? This could be done in a follow-up PR though.

sluongng: yup, my thought exactly.

I think `go_cross` is a very nice to have _knob_ for us to build upon.
But it does not comes with a _sane_ default UX that would be needed to make it easier for end users for happy cases.

Im perfectly fine with a macro being introduced in a separate PR 🙏  Thanks!

go/private/rules/cross.bzl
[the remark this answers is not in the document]
JamesMBartlett: Since one can specify a `go_cross` rule without a platform and with just an `sdk_version`, I decided to go for the second option. This way users can still `bazel run` a `go_cross` rule that just changes the sdk version.

go/private/rules/cross.bzl
```diff
@@ -0,0 +1,141 @@
...
+        )
+    else:
+        # There's no way to determine if the underlying `go_binary` target is executable at loading time
+        # so we must set the `go_cross` rule to be always executable. If the `go_binary` target is not
+        # executable, we set the `go_cross` executable to a simple script that prints an error message
+        # when executed. This way users can still run a `go_cross` target using `bazel run` as long as
+        # the underlying `go_binary` target is executable.
+        error_script = _error_script(ctx)
+
+        # See the implementaion of `go_binary` for an explanation of the need for default vs data runfiles here.
```
fmeum, line 76:
```suggestion
        # See the implementation of `go_binary` for an explanation of the need for default vs data runfiles here.
```

go/private/rules/cross.bzl
```diff
@@ -0,0 +1,141 @@
...
+)
+load(
+    "//go/private:providers.bzl",
+    "GoArchive",
+    "GoLibrary",
+    "GoSource",
+)
+
+def _is_windows(ctx):
+    return ctx.configuration.host_path_separator == ";"
```
fmeum, line 27: This is problematic since it checks for the OS of the host platform, not the target platform. You could instead use `ctx.target_platform_has_constraint` and implicitly depend on `@platforms//os:windows".

Thinking about it some more, I'm not sure whether `.bat` files can simply be executed on Windows without a launcher. Have you been able to test that?

JamesMBartlett: I assumed that `bazel run` would run things on the host platform and not the target platform, is that assumption wrong?

I haven't tested it manually, however, I think the `non_executable_test` shouldn't pass on the windows CI if it doesn't work.

fmeum: `bazel run` only runs on the host, but the `executable` may also be (accidentally) run when used as e.g. a tool during the build or called from a test. If it turns out to be too difficult, this is not absolutely necessary though.

Agreed, if this passed on Windows CI, it's probably fine.

JamesMBartlett: So should I just use `target_platform_has_constraint` and not worry about the case where someone is trying to `bazel run` something for a different target than the host?

fmeum: Good point. Given that this rules is all about cross-compilation, a helpful error with `bazel run` is probably more important - we can keep this as is.

fmeum:
@achew22 This generally looks good to me. Are you also fine with the new API?

achew22 (commented):
One really dumb high level question. Is `go_cross` always going to wrap a `go_binary` or can it wrap a `go_test` (or other)? If it's really only ever intended to be used to wrap a `go_binary` what would we think about calling it `go_cross_binary`?

achew22:
Other than that question, this looks more than good. Thanks so much for putting in so much hard work @JamesMBartlett!

joeljeske:
> One really dumb high level question. Is `go_cross` always going to wrap a `go_binary` or can it wrap a `go_test` (or other)?

Does this PR include a strategy for setting the sdk for a go_tests? I would really love to see that, especially for testing a specific sdk version.

sluongng:
> Does this PR include a strategy for setting the sdk for a go_tests? I would really love to see that, especially for testing a specific sdk version.

@joeljeske  See https://github.com/bazelbuild/rules_go/commit/bc60909ffec7494dfa6eb7304ce84ec738a2b9b9

joeljeske:
Thanks! From what I can tell, running a go_test with a different SDK version requires passing different bazel cli flags, right? 

Given that go_cross enables transitioning a go_binary the SDK using attrs, do you think we should enable a similar behavior for go_test? It seems like it would be important to be able to run tests under the non-default sdk in the same invocation as building binaries. 

It would be important that one could use go_cross with a target of a test, or if one could set sdk version for a go_test directly. 

What do you think?

achew22:
I love the excitement around this feature, but let's leave the feature requests for different conversations. This PR is useful as it stands right now, we're just haggling over naming. Besides, I think the bikeshed should be blue -- it's obviously the best color.

joeljeske:
Very good, thank you!

achew22 (approved):
I ❤️ this PR. Thanks @JamesMBartlett. Leaving it for @fmeum to take one last look since he has a knack for the subtle.

JamesMBartlett:
I changed the name to `go_cross_binary` since separate rules are necessary to support `go_binary` and `go_test` crossing. And I opened another PR for the `go_cross_test` rule #3274

tests/core/cross/README.rst
```diff
@@ -30,6 +31,15 @@ files have a ``goos`` suffix, so they will only be built on the right platform.
 If the wrong source file is used or if all files are filtered out, the
 `go_binary`_ will not build.
 
+go_cross_test
+-------------
+
+Indentical test to ``cross_test`` except tests using a `go_cross`_ rule wrapping a `go_binary`_ instead of the ``goos`` and ``goarch`` attributes on a `go_binary`_.
```
fmeum, line 37: This needs updating now that `go_cross` is `go_cross_binary`.

go/private/rules/cross.bzl
```diff
@@ -0,0 +1,141 @@
+# Copyright 2014 The Bazel Authors. All rights reserved.
```
fmeum, line 1: Nit: Could you update the dates in the new files?

fmeum (approved):

fmeum:
Thanks again for the great contribution, @JamesMBartlett!

jfirebaugh:
Just curious, what's the advantage of `go_cross_binary` versus using the `--platforms` flag? The latter seems like the [officially recommended](https://bazel.build/concepts/platforms#goal) way to do cross compiling with Bazel. (Although `--platforms` doesn't seem to be very well supported by rules_go.)

JamesMBartlett:
`go_cross_binary` does use Bazel's `--platforms` flag. The purpose of `go_cross_binary` is to (in Bazel lingo) transition a `go_binary` rule from the default value of the `--platforms` flag to the specified platform in the rule. More about transitions [here](https://bazel.build/extending/config#user-defined-transitions). This can be useful, when managing many different platforms. For instance, at [Pixie](https://github.com/pixie-io/pixie) our use case was to have multiple go binaries built with different platforms/sdk version all as data dependencies to a test, something like:
```
go_binary(
  name = "binary",
  ...
)
go_cross_binary(
  name = "binary_amd64",
  platform = "//path/to/amd/platform",
)
go_cross_binary(
  name = "binary_arm",
  platform = "//path/to/arm/platform",
)

some_test_rule(
  name = "test using binaries",
  data = [":binary_amd64", ":binary_arm"],
  ...
)
```
Passing the `--platforms` flag on the command line causes all rules (without transitions) to use that platforms flag, which would make this use case difficult.

Also, correct me if I'm wrong @fmeum but I believe `rules_go` supports the `--platforms` flag, with some caveats around setting up crosstool files if you're using cgo. See: https://github.com/bazelbuild/rules_go#how-do-i-cross-compile

jfirebaugh:
Thank you for the explanation, that makes sense.

I encountered two issues with rules_go's support of the `--platforms` flag:

* I was trying to use a self-defined "linux_amd64" platform rather than `--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64` (I was unaware of the latter at the time). This leads to a "No matching toolchains" error. Possibly related to https://github.com/bazelbuild/rules_go/issues/3269.
* https://github.com/bazelbuild/rules_go/issues/3305

JamesMBartlett:
Don't really have context to help with 2, but for your first issue I think you just need to specify a cgo constraint on your custom platform (`"@io_bazel_rules_go//go/toolchain:{cgo_off,cgo_on}"`). I don't know the details of how to do it with cgo on, but with `cgo_off` I think it should Just Work:tm:

PR #3264, opened 2022-08-09
Add dependency to rules_proto in MODULE.bazel.

ntnj:
rules_proto is being used here: https://github.com/bazelbuild/rules_go/blob/master/proto/def.bzl#L43, but it is not declared.

I'm getting an error when trying to use go_proto_library: `Unable to find package for @rules_proto//proto:defs.bzl: The repository '@rules_proto' could not be resolved: Repository '@rules_proto' is not visible from repository '@rules_go.0.33.0'.`

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -11,6 +11,7 @@ print("WARNING: The rules_go Bazel module is still highly experimental and subje
 
 bazel_dep(name = "bazel_skylib", version = "1.2.0")
 bazel_dep(name = "platforms", version = "0.0.4")
+bazel_dep(name = "rules_proto", version = "4.0.0")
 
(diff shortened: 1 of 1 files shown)


fmeum:
@ntnj Did you check that this is all it takes to make `go_proto_library` work with bzlmod? I worked on that briefly and found other things to be missing as well.

fmeum:
Could you open an issue instead? I can keep you updated on my progress to getting `go_proto_library` to work with bzlmod there.

ntnj:
Created https://github.com/bazelbuild/rules_go/issues/3265

PR #3266, opened 2022-08-09
Export support files needed for using Go's WebAssembly (WASM) output.

ralimi:
**What type of PR is this?**

Feature

Export support files needed for using Go's WebAssembly (WASM) output.  In particular, misc/wasm/wasm_exec.js contains glue code for making use of WASM output within Javascript.

No issue filed. Per contribution guidelines, this looked too minor to file an issue. Please let me know if an issue should be filed.

**Other notes for review**

Without this change, my project (https://github.com/google/chrome-ssh-agent) needs to apply a patch as part of importing the rules_go repository.  This seems general-purpose enough that it should be applied in rules_go itself.

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -74,6 +74,9 @@ filegroup(
 )
 
 exports_files(
-    ["lib/time/zoneinfo.zip"],
+    glob([
+        "lib/time/zoneinfo.zip",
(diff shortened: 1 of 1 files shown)


fmeum:
Small world, I still remember contributing to Chrome Secure Shell. :-)

Thanks for the contribution, I agree this is small enough to just merge.

fmeum (approved):

PR #3272, opened 2022-08-18
migrating to rbe_preconfig

aranguyen:
**What type of PR is this?**
 Other

We had a flag flip to enforce the use of constraints values from @platforms. `rules_go` still uses the older `bazel_toolchains` that links to RBE which caused the failure when the flag is flipped.
Please see https://github.com/bazelbuild/continuous-integration/issues/1404

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -54,21 +54,20 @@ llvm_toolchain(
 )
 
 http_archive(
-    name = "bazel_toolchains",
-    sha256 = "179ec02f809e86abf56356d8898c8bd74069f1bd7c56044050c2cd3d79d0e024",
-    strip_prefix = "bazel-toolchains-4.1.0",
-    urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/releases/download/4.1.0/bazel-toolchains-4.1.0.tar.gz",
-        "https://github.com/bazelbuild/bazel-toolchains/releases/download/4.1.0/bazel-toolchains-4.1.0.tar.gz",
-    ],
+    name = "bazelci_rules",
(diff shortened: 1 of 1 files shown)


fmeum (approved):

aranguyen:
@fmeum do you know when is the next release and whether this commit would be included?

fmeum:
@aranguyen It would be included. We don't have a particular date set for the next release yet, but could certainly prepare one if needed. Does the downstream pipeline run against rules_go HEAD or the latest release?

fmeum:
Since this commit only touches `WORKSPACE`, I actually don't think it would even be affecting any release.

aranguyen:
`rules_sass` currently depends on `0.16.5` release. I was trying to upgrade it to this commit but it seems to complain about missing BUILD file. To my surprise the downloaded tar.gz file do have the BUILD file. Please see error in https://github.com/bazelbuild/rules_sass/pull/143

fmeum:
rules_go's `WORKSPACE`, which is the only file this PR touched, is not evaluated when it is loaded by other workspaces. 

Why do you want to update the version of rules_go used by rules_sass? rules_sass currently uses a pretty ancient version of rules_go, so I would expect updating to be somehwat difficult.

aranguyen:
I want to update it because the ancient `rules_go` that `rules_sass` use do not use constraint values from `@platforms`. But you're saying that since my change in `rules_go` do not touch anything else beside `WORKSPACE`, this should not matter? I'm a bit confused on this point.

aranguyen:
ok there are other places in the old `rules_go` that uses `@bazeltools//platforms:xxx` but that is gone for `0.34.0`. After updating to `0.34.0`, there are no other other usage of `@bazeltools//platforms:xxx` in `rules_sass`. So your point about the WORKSPACE file is correct. Thanks for the discussion!

PR #3276, opened 2022-08-25
Remove unused env_execute

kmicklas:
**What type of PR is this?**

> Other

This appears completely unused, and it's in a private package. It confused me a bit when searching around for env var handling.

Fixes #

**Other notes for review**


linzhp:
It seems some checks are failing. Could you take a look?

fmeum:
@kmicklas Try rebasing on master, that should fix the pipeline.

PR #3281, opened 2022-09-06
Fix non_executable_test to work with `@//` labels

Wyverald:
Needed to flip https://github.com/bazelbuild/bazel/issues/16196

### DIFF
--- a/tests/core/cross/non_executable_test.go
+++ b/tests/core/cross/non_executable_test.go
@@ -15,11 +15,13 @@
 package non_executable_test
 
 import (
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Thanks, the change looks good.

I could see this error message benefiting from the same improvement as in https://github.com/bazelbuild/bazel-skylib/pull/394/files#diff-0d521430fd46dbc443c2f620b0ab8503a1226c3ef004f3b058ef1e870eab2e96.

PR #3282, opened 2022-09-08
Getting architecture from JRE

### LINKED ISSUE #3086: no matching toolchains found for types @io_bazel_rules_go//go:toolchain on DARWIN ARM64

v0.30.0

v0.22.3

4.2.2

Yes

darwin arm64

### What did you do?

```
http_archive(
    name = "io_bazel_rules_go",
    sha256 = "d6b2513456fe2229811da7eb67a444be7785f5323c6708b38d851d2b51e54d83",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.30.0/rules_go-v0.30.0.zip",
    ],
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies", "go_download_sdk")

go_download_sdk(
    name = "go_sdk",
    version = "1.17.7",
    goos = "darwin",
    goarch = "arm64",
)

go_rules_dependencies()
go_register_toolchains()
```

Successful Execution: `INFO: Build completed successfully,`

### What did you see instead?

```
ERROR: While resolving toolchains for target //980-commons/src/main/proto/io/harness/security:980_commons_security_go_proto: no matching toolchains found for types @io_bazel_rules_go//go:toolchain
ERROR: Analysis of target '//980-commons/src/main/proto/io/harness/security:980_commons_security_go_proto' failed; build aborted: no matching toolchains found for types @io_bazel_rules_go//go:toolchain
INFO: Elapsed time: 15.284s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (8 packages loaded, 217 targets configured)

```

linzhp:
**What type of PR is this?**
Bug fix

This PR makes rules_go gets the host architecture repository_ctx, which in turn get it from JVM. This is because Bazel determines default host platform from its JVM https://github.com/bazelbuild/bazel/blob/a380a157f658cc392f66f86b6aba340c461dc408/src/main/java/com/google/devtools/build/lib/bazel/repository/LocalConfigPlatformFunction.java#L63 and https://github.com/bazelbuild/bazel/blob/577cfbc8a2b0841cd67a918febe9e3c5fcb3882e/src/main/java/com/google/devtools/build/lib/util/CPU.java#L56

If rules_go determines the host platform by calling `uname`, we may run into an interesting situation when users accidentally runs an ARM64 version of Bazel binary, with comes with ARM64 JVM, in a terminal with Rosetta translation:

* rules_go downloads an AMD64 version of Go SDK, because uname returns "amd64"
* Bazel tries to find a Go toolchain for ARM64, because JVM tells Bazel that it's running in ARM64

Leading to:

```
no matching toolchains found for types @io_bazel_rules_go//go:toolchain
```

Fixes #3086

**Other notes for review**
Linux and Windows shouldn't have this problem, but getting arch from repository_ctx is much simpler than their current implementation. So I change all of them together. I don't have access to all these OS/Arch combination to verify. So I just optimistically assume that they would work. We can always override the arch names as needed in the future.

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -258,53 +258,15 @@ def _sdk_build_file(ctx, platform, version):
     )
 
 def _detect_host_platform(ctx):
-    if ctx.os.name == "linux":
-        goos, goarch = "linux", "amd64"
-        res = ctx.execute(["uname", "-p"])
-        if res.return_code == 0:
-            uname = res.stdout.strip()
-            if uname == "s390x":
-                goarch = "s390x"
-            elif uname == "i686":
-                goarch = "386"
-
-        # uname -p is not working on Aarch64 boards
-        # or for ppc64le on some distros
-        res = ctx.execute(["uname", "-m"])
-        if res.return_code == 0:
-            uname = res.stdout.strip()
-            if uname == "aarch64":
-                goarch = "arm64"
-            elif uname == "armv6l":
-                goarch = "arm"
-            elif uname == "armv7l":
-                goarch = "arm"
-            elif uname == "ppc64le":
-                goarch = "ppc64le"
-
-        # Default to amd64 when uname doesn't return a known value.
-
-    elif ctx.os.name == "mac os x":
-        goos, goarch = "darwin", "amd64"
-
-        res = ctx.execute(["uname", "-m"])
-        if res.return_code == 0:
-            uname = res.stdout.strip()
-            if uname == "arm64":
-                goarch = "arm64"
-
-        # Default to amd64 when uname doesn't return a known value.
-
-    elif ctx.os.name.startswith("windows"):
+    goos = ctx.os.name
+    if goos == "mac os x":
+        goos = "darwin"
+    elif goos.startswith("windows"):
         goos = "windows"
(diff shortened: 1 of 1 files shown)


fmeum (commented):
I'm all for this change, but it would require us to raise the minimum Bazel version requirement (I believe to 5.2.0 or 5.3.0). Since 5.3.0 would buy us many other nice things, that could be worth it, but I can't tell whether it's too early for that change.

linzhp:
5.1 is sufficient for this to work: https://docs.bazel.build/versions/5.1.0/skylark/lib/repository_os.html

I am still debating with myself whether to bump the minimal version (given that 5.1.0 is half year old), or check the existence of repository_os.arch and keep the old logic if the attribute is not available.

fmeum:
Should have known that since I added it myself ;)

I'm personally in favor of bumping the minimum version as we don't know of any other important bugs that would require fixing for people stuck on Bazel 4. But I also never maintained a large monorepo.

linzhp:
Uber's Go monorepo has been using Bazel 5.2 for quite a while. So it won't help discover any recent rules_go bug with Bazel 4 either.

fmeum (approved):

PR #3285, opened 2022-09-09
go_path: support go:embed of generated files

### LINKED ISSUE #3284: go_path: Support embedsrcs of generated files
v0.34.0

v0.24.0

Bazelisk version: development
Build label: 5.1.0
Build time: Thu Mar 24 14:10:21 2022 (1648131021)
Build timestamp: 1648131021
Build timestamp as int: 1648131021

Yes.

Darwin x86_64

### What did you do?
Added a Go library that uses go:embed as a dependency to a go_path rule. The go:embed file was generated with a copy_file rule.

Files embedded in Go library to be copied into the generated GOPATH.

### What did you see instead?
fail("Path '%s' is not beneath '%s'" % (path, start))
Error in fail: Path 'bazel-out/k8-fastbuild/bin/foo/bar' is not beneath 'foo'

Exact filenames edited

S-Chan:
When embedded files are generated, relativize fails because the file path starts with the output directory. Without the change in path.bzl, the test case fails as follows:

Error in fail: Path 'bazel-out/darwin-fastbuild/bin/tests/core/go_path/pkg/lib/renamed_embedded_src.txt' is not beneath 'tests/core/go_path/pkg/lib'

**What type of PR is this?**

Bug fix

Fixes #3284

**Other notes for review**

### DIFF
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -83,7 +83,8 @@ def _go_path_impl(ctx):
         for f in pkg.embedsrcs:
             if src_dir == None:
                 fail("cannot relativize {}: src_dir is unset".format(f.path))
-            dst = pkg.dir + "/" + paths.relativize(f.path, src_dir)
+            embedpath = paths.relativize(f.path, f.root.path)
+            dst = pkg.dir + "/" + paths.relativize(embedpath, src_dir)
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
     if ctx.attr.include_pkg:
         for pkg in pkg_map.values():
--- a/tests/core/go_path/cmd/bin/BUILD.bazel
+++ b/tests/core/go_path/cmd/bin/BUILD.bazel
@@ -9,6 +9,17 @@ go_binary(
 )
 
 go_binary(
+    name = "pie",
+    srcs = ["bin.go"],
+    importpath = "example.com/repo/cmd/bin",
+    visibility = ["//visibility:public"],
+    linkmode="pie",
+    deps = [
+        "//tests/core/go_path/pkg/lib:go_default_library"
+    ]
+)
+
+go_binary(
     name = "cross",
     srcs = ["bin.go"],
     goarch = "arm",
--- a/tests/core/go_path/cmd/bin/BUILD.bazel
+++ b/tests/core/go_path/cmd/bin/BUILD.bazel
@@ -13,7 +13,7 @@ go_binary(
     srcs = ["bin.go"],
     importpath = "example.com/repo/cmd/bin",
     visibility = ["//visibility:public"],
-    linkmode="pie",
+    linkmode = "pie",
     deps = [
         "//tests/core/go_path/pkg/lib:go_default_library"
     ]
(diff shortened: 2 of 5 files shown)


go/private/tools/path.bzl
```diff
@@ -81,9 +81,12 @@ def _go_path_impl(ctx):
             dst = pkg.dir + "/" + f.basename
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
         for f in pkg.embedsrcs:
+            embedpath = f.path
             if src_dir == None:
                 fail("cannot relativize {}: src_dir is unset".format(f.path))
-            dst = pkg.dir + "/" + paths.relativize(f.path, src_dir)
+            if f.dirname.startswith(ctx.bin_dir.path):
```
fmeum, line 87: It's subtle, but `ctx.bin_dir` is the bin dir of the rule, which may not be the same as the bin dir of the files in `embedsrcs` when there are transitions involved. You should be able use the path of [File.root](https://bazel.build/rules/lib/File#root) instead.

tests/core/go_path/pkg/lib/BUILD.bazel
```diff
@@ -10,13 +11,20 @@ go_library(
...
         "embedded_src.txt",
+        "renamed_embedded_src.txt",
         "template/index.html.tmpl",
     ],
     importpath = "example.com/repo/pkg/lib",
     visibility = ["//visibility:public"],
     deps = [":transitive_lib"],
 )
 
+copy_file(
```
fmeum, line 22: A simple test case you could add that would cover the case of a transition would be a `go_binary` built with a non-default `linkmode`.

go/private/tools/path.bzl
S-Chan: Thanks for the tip! Changed to use f.root instead.

tests/core/go_path/pkg/lib/BUILD.bazel
S-Chan: Added a test case that only passed with the f.root.path route you suggested. I'm not that familiar with the testing convention, so any tips to simplify the test setup would be appreciated.

tests/core/go_path/cmd/bin/BUILD.bazel
```diff
@@ -8,6 +8,17 @@ go_binary(
     visibility = ["//visibility:public"],
 )
 
+go_binary(
+    name = "pie",
+    srcs = ["bin.go"],
+    importpath = "example.com/repo/cmd/bin",
+    visibility = ["//visibility:public"],
+    linkmode="pie",
```
fmeum, line 16: Nit:
```suggestion
    linkmode = "pie",
```

tests/core/go_path/pkg/lib/BUILD.bazel
fmeum: Thanks for making the changes, this already looks pretty good. 

Could we add one more test case where we turn this around and add a transitioned Go binary ("pie" or "cross" should work) as an `embedsrc` here?

fmeum: Thinking about it more, I don't think this test case would actually be relevant here.

fmeum (approved):

fmeum (approved):

PR #3288, opened 2022-09-10
Preparing 0.35 release

linzhp:
Update deps and bump the version, preparing for 0.35 release


MODULE.bazel
```diff
@@ -1,7 +1,7 @@
 module(
     name = "rules_go",
     repo_name = "io_bazel_rules_go",
-    version = "0.34.0",
+    version = "0.35.0",
```
fmeum, line 4: Sorry I have gotten to this yet, but how should we deal with the dependencies declared in `MODULE.bazel`? I haven't included updating them in the release flow yet. It's also not as important since MVS ensures that users can easily get newer version themselves if needed, and compatibility is tested with the `WORKSPACE` setup.

fmeum (approved):

MODULE.bazel
linzhp: Good question. We should include that in the releaser tool. Maybe we should use the versions resolved by MVS as the source of truth, instead of unconditionally upgrading everything to the latest.

PR #3290, opened 2022-09-12
bzlmod: Use register_toolchains

fmeum:
`toolchains_to_register` is deprecated.


fmeum:
@achew22 Going to merge this right away as we might otherwise break a downstream pipeline without need.

PR #3300, opened 2022-09-16
also write test stderr to json (xml) output file

### LINKED ISSUE #3298: test stderr is only in test.log, not test.xml
I'm not sure if this is the desired behavior, but it is pretty surprising: only `stdout` output from the test binary is in the `test.xml` produced by `rules_go`. The `stderr` is ONLY in `test.log` (which also contains the `test.xml` `stdout` output). For us at `cockroach` we parse `test.xml` to get test results, and it's often the case that the `test.xml` says there's a failure but contains no relevant information because many tests/frameworks/etc. only write error output to `stderr` rather than `stdout`. This means if you want the full or relevant output you MUST look in `test.log`, which may be confusing.

I have a candidate fix at [https://github.com/rickystewart/rules_go/tree/stderrxml](this branch), and can file a PR if desired.

rickystewart:
Without this, people are forced to look in the test.log file for output to stderr, rather than test.xml.

Allow finding all relevant test output in `test.xml` instead of requiring a look in `test.log`.

Fixes #3298.

### DIFF
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 // TestWrapperAbnormalExit is used by Wrap to indicate the child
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Thanks!

Issue #3302, opened 2022-09-20
Support host SDK in toolchain:sdk_version command-line flag


ash2k:
### What version of rules_go are you using?

v0.35.0

### What version of gazelle are you using?

v0.27.0

### What version of Bazel are you using?

5.3.0

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

macOS M1

### Any other potentially useful information about your toolchain?

I want to use a particular SDK version by default (it is defined first) but also be able to switch the whole build to use host SDK using a flag (or some other way?).

### What did you do?

WORKSPACE:
```
go_download_sdk(
    name = "go_sdk",
    version = "1.18.5",
)

go_host_sdk(
    name = "host_sdk",
)

go_rules_dependencies()

go_register_toolchains()

gazelle_dependencies(
    go_sdk = "go_sdk",
)
```

```
bazel build --@io_bazel_rules_go//go/toolchain:sdk_version=host_sdk //some-target
```

### What did you expect to see?

Target built with the host Go SDK.

### What did you see instead?

```
ERROR: somepath/BUILD.bazel:16:23: While resolving toolchains for target //some-target: no matching toolchains found for types @io_bazel_rules_go//go:toolchain
ERROR: Analysis of target '//some-target' failed; build aborted: 
```

fmeum:
This could be achieved relatively easily by adding a `config_value` for `host_sdk` (or `host`, to avoid duplicating `sdk`) [here](https://github.com/bazelbuild/rules_go/blob/4d4a2454c588dab27853ccffb03fcb9205001192/go/private/BUILD.sdk.bazel#L100) conditionally if the [_sdk_build_file](https://github.com/bazelbuild/rules_go/blob/4d4a2454c588dab27853ccffb03fcb9205001192/go/private/sdk.bzl#L234) function is called from [_go_host_sdk_impl](https://github.com/bazelbuild/rules_go/blob/4d4a2454c588dab27853ccffb03fcb9205001192/go/private/sdk.bzl#L30).

@ash2k or anyone else interested in picking this up, I can help in case you get stuck.

ash2k:
I'll open a PR. Thanks for the instructions!

PR #3303, opened 2022-09-20
Allow to select Go SDK version from declared SDKs

ash2k:
**What type of PR is this?**

Feature

See linked issue.

Fixes https://github.com/bazelbuild/rules_go/issues/3302.

**Other notes for review**

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -31,7 +31,7 @@ def _go_host_sdk_impl(ctx):
     goroot = _detect_host_sdk(ctx)
     platform = _detect_sdk_platform(ctx, goroot)
     version = _detect_sdk_version(ctx, goroot)
-    _sdk_build_file(ctx, platform, version)
+    _sdk_build_file(ctx, platform, version, "host")
     _local_sdk(ctx, goroot)
 
 _go_host_sdk = repository_rule(
@@ -231,7 +231,7 @@ def _local_sdk(ctx, path):
     for entry in ["src", "pkg", "bin", "lib"]:
         ctx.symlink(path + "/" + entry, entry)
 
-def _sdk_build_file(ctx, platform, version):
+def _sdk_build_file(ctx, platform, version, sdk_type = "remote"):
     ctx.file("ROOT")
     goos, _, goarch = platform.partition("_")
 
@@ -254,6 +254,7 @@ def _sdk_build_file(ctx, platform, version):
             "{minor_version}": str(minor),
             "{patch_version}": str(patch),
             "{prerelease_suffix}": prerelease,
+            "{sdk_type}": sdk_type,
         },
     )
 
--- a/go/toolchains.rst
+++ b/go/toolchains.rst
@@ -76,6 +76,8 @@ the flag ``--@io_bazel_rules_go//go/toolchain:sdk_version="version"`` where
 ``"version"`` is the SDK version you would like to build with, eg. ``"1.18.3"``.
 The SDK version can omit the patch, or include a prerelease part, eg. ``"1"``,
 ``"1.18"``, ``"1.18.0"``, and ``"1.19.0beta1"`` are all valid values for ``sdk_version``.
+When ``go_host_sdk`` is used, ``"version"`` can be set to ``host`` to refer to the host Go SDK.
+It can also be set ``remote`` to match any non-host version.
 
 The toolchain
 ~~~~~~~~~~~~~
--- a/tests/core/go_download_sdk/go_download_sdk_test.go
+++ b/tests/core/go_download_sdk/go_download_sdk_test.go
@@ -90,7 +90,7 @@ go_download_sdk(
 		{
 			desc: "multiple_sdks",
 			rule: `
-load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk")
+load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk", "go_host_sdk")
 
 go_download_sdk(
     name = "go_sdk",
@@ -107,7 +107,8 @@ go_download_sdk(
 `,
 			optToWantVersion: map[string]string{
 				"": "go1.16",
-				"--@io_bazel_rules_go//go/toolchain:sdk_version=1":   "go1.16",
+				"--@io_bazel_rules_go//go/toolchain:sdk_version=remote": "go1.16",
+				"--@io_bazel_rules_go//go/toolchain:sdk_version=1":      "go1.16",
 				"--@io_bazel_rules_go//go/toolchain:sdk_version=1.17":   "go1.17",
 				"--@io_bazel_rules_go//go/toolchain:sdk_version=1.17.0": "go1.17",
 				"--@io_bazel_rules_go//go/toolchain:sdk_version=1.17.1": "go1.17.1",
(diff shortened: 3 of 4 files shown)


go/toolchains.rst
```diff
@@ -76,6 +76,7 @@ the flag ``--@io_bazel_rules_go//go/toolchain:sdk_version="version"`` where
 ``"version"`` is the SDK version you would like to build with, eg. ``"1.18.3"``.
 The SDK version can omit the patch, or include a prerelease part, eg. ``"1"``,
 ``"1.18"``, ``"1.18.0"``, and ``"1.19.0beta1"`` are all valid values for ``sdk_version``.
+You can also refer to a Go SDK by name of the target used to declare it.
```
fmeum, line 79: This is a smart trick. :)
There are two problem with it though which makes me think we should solve this differently:
- With Bzlmod, canonical repository names as returned by `ctx.attr.name` will become an implementation detail, they shouldn't (and can't easily) be passed as command-line arguments.
- This essentially shortcuts ordinary toolchain resolution to "just use this exact toolchain", which I think is better solved using `--extra_toolchains`.

How about adding a `config_setting` that matches by SDK type (`host`/`download` or perhaps `local`/`remote` to match the Java toolchains)?

ash2k: I've added `host` to be consistent with `go_register_toolchains(version = "host")`.

go/private/sdk.bzl
```diff
@@ -231,7 +231,7 @@ def _local_sdk(ctx, path):
     for entry in ["src", "pkg", "bin", "lib"]:
         ctx.symlink(path + "/" + entry, entry)
 
-def _sdk_build_file(ctx, platform, version):
+def _sdk_build_file(ctx, platform, version, extra_sdk_version = None):
```
fmeum, line 234: We could set this to `remote` by default, which would make it possible to extract the `config_setting` into `BUILD.sdk.bazel`.

go/toolchains.rst
```diff
@@ -76,6 +76,8 @@ the flag ``--@io_bazel_rules_go//go/toolchain:sdk_version="version"`` where
 ``"version"`` is the SDK version you would like to build with, eg. ``"1.18.3"``.
 The SDK version can omit the patch, or include a prerelease part, eg. ``"1"``,
 ``"1.18"``, ``"1.18.0"``, and ``"1.19.0beta1"`` are all valid values for ``sdk_version``.
+When ``go_host_sdk`` is used, ``"version"`` can be set to ``host`` to refer to the host Go SDK.
+It can also be set ``remote`` to match any non-host version.
```
fmeum, line 80:
```suggestion
It can also be set to ``remote`` to match any non-host version.
```

tests/core/go_download_sdk/go_download_sdk_test.go
```diff
@@ -107,7 +107,7 @@ go_download_sdk(
 `,
 			optToWantVersion: map[string]string{
 				"": "go1.16",
-				"--@io_bazel_rules_go//go/toolchain:sdk_version=1":   "go1.16",
+				"--@io_bazel_rules_go//go/toolchain:sdk_version=1":      "go1.16",
```
fmeum, line 110: Looks like we are missing a test case for the new feature now.

fmeum (commented):
This looks pretty good now, thanks.

tests/core/go_download_sdk/go_download_sdk_test.go
ash2k: I'm not sure how to test that without depending on the host SDK. Any ideas?

fmeum: Good point. I think it's fine to test the `remote` case.

linzhp:
Right now, we [have](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#registering-a-custom-sdk): 
> It must still be named go_sdk, but this is a temporary limitation that will be removed in the future.

Can we instead remove this limitation, so we can name Go SDK with different names and switch with `--@io_bazel_rules_go//go/toolchain:sdk=host_sdk`, `--@io_bazel_rules_go//go/toolchain:sdk=experimental_sdk`, which matches the target name of `go_download_sdk`, `go_wrap_sdk`, `go_host_sdk` and `go_local_sdk`:

```
go_wrap_sdk(
    name = "experimental_sdk",
    root_file = "@other_repo//go:README.md",
)

go_host_sdk(
    name = "host_sdk",
    root_file = "@other_repo//go:README.md",
)
```

fmeum (approved):

fmeum:
@linzhp Wouldn't `--extra_toolchains` cover the case of selecting a particular toolchain "by name"? Matching by name rather than by label will not work well with Bzlmod, so I'm hesitant to introduce this now.

ash2k:
Can we merge this please and consider using SDK names for the flag separately? I just really need this feature ASAP. Thank you!

zakcutner:
> @linzhp Wouldn't `--extra_toolchains` cover the case of selecting a particular toolchain "by name"? Matching by name rather than by label will not work well with Bzlmod, so I'm hesitant to introduce this now.

@fmeum Sorry to ping you on an old PR, I wanted to check if I was understanding your suggestion properly before opening a new issue. I'm facing the same problem where I want to force a different toolchain to be used by passing a flag, and I'm not sure `--extra_toolchains` would cover this use case.

```bazel
# WORKSPACE

...

go_download_sdk(
    name = "go_sdk",
    version = "1.21.1",
)

go_wrap_sdk(
    name = "patched_go_sdk",
    register_toolchains = False,
    root_file = "@patched_go//:README.md",
)

...
```
I think `bazel build --extra_toolchains=@patched_go_sdk_toolchains//:go_darwin_amd64 ...` does not quite work how I would need, since it would enable both `go_sdk` and `patched_go_sdk` (and Bazel would choose somehow between the two) rather than enforcing that `patched_go_sdk` is used over `go_sdk`?

In addition, I think there's no way to have a flag with this method that works in general across platforms that I could add to my project's `.bazelrc` file, i.e. I would need to specify `--extra_toolchains=@patched_go_sdk_toolchains//:go_linux_arm64` on ARM64 Linux?

fmeum:
> I think bazel build --extra_toolchains=@patched_go_sdk_toolchains//:go_darwin_amd64 ... does not quite work how I would need, since it would enable both go_sdk and patched_go_sdk (and Bazel would choose somehow between the two) rather than enforcing that patched_go_sdk is used over go_sdk?

Enabling (i.e., "registering") both `go_sdk` and `patched_go_sdk` is totally fine. Bazel tries all toolchains in a specified order and those listed in `--extra_toolchains` always come first. Assuming both `go_sdk` and `patched_go_sdk` match the required platform constraints, `patched_go_sdk` will be preferred with that command-line flag (and not used without it).

> In addition, I think there's no way to have a flag with this method that works in general across platforms that I could add to my project's .bazelrc file, i.e. I would need to specify --extra_toolchains=@patched_go_sdk_toolchains//:go_linux_arm64 on ARM64 Linux?

You should be able to specify `--extra_toolchains=@patched_go_sdk_toolchains//:all` to register all the different platform variants of the given toolchain.

zakcutner:
@fmeum Thanks so much for the quick response, that works great for my use case!

PR #3311, opened 2022-10-05
synchronize when writing to the JSON converter

rickystewart:
We have sporadically seen issues where tests appear to fail with `No pass/skip/fail event found for test`. This is apparently because the test output from `go test` is interspersed with normal `stderr` output from the test process (notice how the `PASS:` line is split by stderr output):

```
--- PASS: TeI220916 15:19:35.406930 702 3@pebble/event.go:649  [pebble,s?] 2  [JOB 1] MANIFEST created 000001
I220916 15:19:35.407183 702 3@pebble/event.go:677  [pebble,s?] 3  [JOB 1] WAL created 000002
...
I220916 15:19:35.414755 702 sql/colexec/colexectestutils/utils.go:691  [-] 22  batchSize=1/sel=false
stWindowFramer/mode=Groups/start=OffsetFollowing/end=UnboundedFollowing#01/count=17 (0.00s)
```

Here we avoid this by adding extra synchronization to ensure we only write a whole line from one of the two streams at a time.

**What type of PR is this?**

> Bug fix

See https://github.com/bazelbuild/rules_go/issues/3298#issuecomment-1268560258.

### DIFF
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -15,6 +15,7 @@
 package bzltestutil
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Thanks for the contribution!

PR #3328, opened 2022-10-20
feat(pkg-drv): allow setting custom aspect and kinds

JamyDev:
**What type of PR is this?**

Feature

adds two new env variables for the packages driver:
- `GOPACKAGESDRIVER_BAZEL_KINDS`: Tells the packages driver to also query for the specified kinds alongside `go_default_(binary|library|test)`
- `GOPACKAGESDRIVER_BAZEL_ASPECT`: Lets the consumer define a custom aspect, to support querying these other kinds

Combining these two, users can add support for custom rules which consume and emit Go files in a nonstandard way. At Uber, for example, we have a code generator uses a Go file as its source and generates a matching Go file with inlined content. 

Having both of these Go files in the resulting `GoArchive` will not work for the Go compiler (since they define the same variables/methods/structs/..), which in turn means that the GoPackagesDriver doesn't know about this source Go file (only the generated output).

**Other notes for review**

An alternative approach would have been to call this custom aspect as an additional aspect, but I wasn't sure how the packages driver would handle multiple `GoPkgInfo` outputs.

### DIFF
--- a/go/tools/gopackagesdriver/bazel_json_builder.go
+++ b/go/tools/gopackagesdriver/bazel_json_builder.go
@@ -30,16 +30,15 @@ const (
 	RulesGoStdlibLabel = "@io_bazel_rules_go//:stdlib"
 )
 
+var _defaultKinds = []string{"go_library", "go_test", "go_binary"}
+
 func (b *BazelJSONBuilder) fileQuery(filename string) string {
 	if filepath.IsAbs(filename) {
 		fp, _ := filepath.Rel(b.bazel.WorkspaceRoot(), filename)
 		filename = fp
 	}
-	addtlKinds := ""
-	if len(additionalKinds) > 0 {
-		addtlKinds = "|" + strings.Join(additionalKinds, "|")
-	}
-	return fmt.Sprintf(`kind("go_library|go_test|go_binary%s", same_pkg_direct_rdeps("%s"))`, addtlKinds, filename)
+	kinds := append(_defaultKinds, additionalKinds...)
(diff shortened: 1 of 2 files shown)


go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -35,7 +35,11 @@ func (b *BazelJSONBuilder) fileQuery(filename string) string {
 		fp, _ := filepath.Rel(b.bazel.WorkspaceRoot(), filename)
 		filename = fp
 	}
-	return fmt.Sprintf(`kind("go_library|go_test|go_binary", same_pkg_direct_rdeps("%s"))`, filename)
+	addtlKinds := ""
+	if len(additionalKinds) > 0 {
+		addtlKinds = "|" + strings.Join(additionalKinds, "|")
+	}
+	return fmt.Sprintf(`kind("go_library|go_test|go_binary%s", same_pkg_direct_rdeps("%s"))`, addtlKinds, filename)
```
fmeum, line 42: Could you extract the default kinds in a slice and then do a single join to get the `|`-separated list?

fmeum (approved):
CI failure is unrelated and will be fixed by https://github.com/bazelbuild/rules_go/pull/3329

fmeum:
The failure should go away if you rebase on master.

go/tools/gopackagesdriver/bazel_json_builder.go
JamyDev: Sounds good will do shortly.

JamyDev: Updated

PR #3329, opened 2022-10-20
bzlmod: Allow yanked versions in CI

fmeum:
Versions yanked from the BCR would otherwise non-hermetically fail our pipelines.


fmeum:
FYI @achew22 @linzhp

PR #3330, opened 2022-10-21
fix(packages-driver): allow defining additional aspects instead

JamyDev:
**What type of PR is this?**

Other

In #3328 I made it so defining an aspect would overwrite the default one defined by rules_go. This means the custom aspect would need to duplicate the rules_go code and potentially manually merge any changes we do in rules_go every release.

Talking with @linzhp , he indicated that you can pass more than one aspect to `--aspects`, so in this case when additional aspects are defined they will be passed in and can export similar info to the default aspect.

### DIFF
--- a/go/tools/gopackagesdriver/bazel_json_builder.go
+++ b/go/tools/gopackagesdriver/bazel_json_builder.go
@@ -119,11 +119,13 @@ func (b *BazelJSONBuilder) Build(ctx context.Context, mode LoadMode) ([]string, 
 		return nil, fmt.Errorf("found no labels matching the requests")
 	}
 
+	aspects := append(additionalAspects, rulesGoAspect)
+
 	buildArgs := concatStringsArrays([]string{
 		"--experimental_convenience_symlinks=ignore",
 		"--ui_event_filters=-info,-stderr",
 		"--noshow_progress",
-		"--aspects=" + customAspect,
+		"--aspects=" + strings.Join(aspects, ","),
 		"--output_groups=" + b.outputGroupsForMode(mode),
 		"--keep_going", // Build all possible packages
 	}, bazelFlags, bazelBuildFlags, labels)
--- a/go/tools/gopackagesdriver/bazel_json_builder.go
+++ b/go/tools/gopackagesdriver/bazel_json_builder.go
@@ -119,7 +119,7 @@ func (b *BazelJSONBuilder) Build(ctx context.Context, mode LoadMode) ([]string, 
 		return nil, fmt.Errorf("found no labels matching the requests")
 	}
 
-	aspects := append(additionalAspects, rulesGoAspect)
+	aspects := append(additionalAspects, goDefaultAspect)
 
 	buildArgs := concatStringsArrays([]string{
 		"--experimental_convenience_symlinks=ignore",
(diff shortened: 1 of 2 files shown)


go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -119,11 +119,13 @@ func (b *BazelJSONBuilder) Build(ctx context.Context, mode LoadMode) ([]string,
 		return nil, fmt.Errorf("found no labels matching the requests")
 	}
 
+	aspects := append(additionalAspects, rulesGoAspect)
```
fmeum, line 122: AFAIK the order may be quite significant here. 

@linzhp Is this the correct one (and if so, why?)?

JamyDev: This was a syntax simplification assuming the order didn't matter, if it does I'd be happy to change it so rules_go is first.

fmeum: I'm not saying this order is wrong, I simply don't know the aspect well enough to decide this. If I understood the interaction correctly, the latter aspect would see providers added by the former, but maybe that's just not how it works. :-)

linzhp: If the additional aspects need some information from each other, or from the default aspect, they can specify `required_providers` in their aspect definition, then Bazel would run those aspect in the order to satisfy the requirement.

@JamyDev did you test the these aspects don't override each other when that have the same output group and provider?

go/tools/gopackagesdriver/main.go
```diff
@@ -52,13 +52,14 @@ var (
 	// It seems https://github.com/bazelbuild/bazel/issues/3115 isn't fixed when specifying
 	// the aspect from the command line. Use this trick in the mean time.
 	rulesGoRepositoryName = getenvDefault("GOPACKAGESDRIVER_RULES_GO_REPOSITORY_NAME", "@io_bazel_rules_go")
+	rulesGoAspect         = rulesGoRepositoryName + "//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect"
```
linzhp, line 55:
```suggestion
	goDefaultAspect         = rulesGoRepositoryName + "//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect"
```

nit: this aspect is less about rules_go, more about getting default Go pkg info

go/tools/gopackagesdriver/bazel_json_builder.go
JamyDev: There is possibility for overlap if you output to the same `.pkg.json` file as the default aspect. The packagesdriver [only considers the .pkg.json files](https://github.com/bazelbuild/rules_go/blob/master/go/tools/gopackagesdriver/bazel_json_builder.go#L136-L140), so long as there's no overlap in these files it's fine.  In our (Uber's) case that's not a problem, we output to `ruleName.pkg.json` (just like the default aspect does) and only output anything when we encounter our custom rule. 

Now if the order in the argument here is actually followed this means that if I write a bad custom aspect, it'll get overridden by the rulesGo aspect, unless I directly declare it as a `required_provider`.

linzhp (approved):
Do you think there ever is a scenario where the `goDefaultAspect` is not wanted? If so, maybe we can instead have `GOPACKAGESDRIVER_BAZEL_ASPECTS`, which defaults to `//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect`, and ask users to set it to `//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect,//some/other%aspect` if they want aspects in addition to the default aspect, and set it to `//some/other%aspect` if they don't want to default aspect

JamyDev:
> Do you think there ever is a scenario where the `goDefaultAspect` is not wanted?

Maybe but then we'd also have to define each of the go targets individually instead of only additional ones. At that point it might make sense to just fork GoPackagesDriver, no?

linzhp:
yeah, that's right.

PR #3333, opened 2022-10-25
Add support for parsing Go development versions

### LINKED ISSUE #3331: Go version not correctly recongised

997520fc8d75fddc0ee51cec62833e748bb06000

When detecting the version of Go, rules_go expects the format to be of the form `go version go1.18.3 linux/amd64` in [`sdk.bzl`](https://github.com/bazelbuild/rules_go/blob/997520fc8d75fddc0ee51cec62833e748bb06000/go/private/sdk.bzl#L320). It then checks the third word of this output to determine the version.

However, development Go versions will be of the form `go version devel go1.18.3 linux/amd64`, so the Go version is not correctly determined in this case. I would suggest potentially checking the fourth word for the version instead if the third word  is `devel`.

zakcutner:
**What type of PR is this?**

Bug fix

The previous code assumed that the output of `go version` was of the form:
```text
go version go1.18.3 linux/amd64
```

However, it can also potentially be of the form:
```text
go version devel go1.19-fd1b5904ae Tue Mar 22 21:38:10 2022 +0000 linux/amd64
```

Therefore, rather than reading the third argument to find the version (1.18.3 in this case), we instead now read the fourth argument if the third argument is `devel`.

Fixes #3331

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -317,11 +317,21 @@ def _detect_sdk_version(ctx, goroot):
     if result.return_code != 0:
         fail("Could not detect SDK version: '%s version' exited with exit code %d" % (go_binary_path, result.return_code))
 
-    # go version output is of the form "go version go1.18.3 linux/amd64"
+    # go version output is of the form "go version go1.18.3 linux/amd64" or "go
+    # version devel go1.19-fd1b5904ae Tue Mar 22 21:38:10 2022 +0000
+    # linux/amd64". See the following links for how this output is generated:
+    # - https://github.com/golang/go/blob/2bdb5c57f1efcbddab536028d053798e35de6226/src/cmd/go/internal/version/version.go#L75
+    # - https://github.com/golang/go/blob/2bdb5c57f1efcbddab536028d053798e35de6226/src/cmd/dist/build.go#L333
+    #
+    # Read the third word, or the fourth word if the third word is "devel", to
+    # find the version number.
     output_parts = result.stdout.split(" ")
-    if len(output_parts) < 3 or not output_parts[2].startswith("go"):
+    if len(output_parts) > 2 and output_parts[2].startswith("go"):
+        version = output_parts[2][len("go"):]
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Thanks for the fix!

Looks good to me, I will merge after you have signed the CLA.

go/private/sdk.bzl
```diff
@@ -317,11 +317,17 @@ def _detect_sdk_version(ctx, goroot):
     if result.return_code != 0:
         fail("Could not detect SDK version: '%s version' exited with exit code %d" % (go_binary_path, result.return_code))
 
-    # go version output is of the form "go version go1.18.3 linux/amd64"
+    # go version output is of the form "go version go1.18.3 linux/amd64" or "go
+    # version devel go1.19-fd1b5904ae Tue Mar 22 21:38:10 2022 +0000
+    # linux/amd64". Read the third word, or the fourth word if the third word is
+    # "devel" to find the version number.
```
fmeum, line 323: Given that the logic got more complex, can we cite these authorative sources in the comment?
https://github.com/golang/go/blob/2bdb5c57f1efcbddab536028d053798e35de6226/src/cmd/go/internal/version/version.go#L75
https://github.com/golang/go/blob/2bdb5c57f1efcbddab536028d053798e35de6226/src/cmd/dist/build.go#L333

zakcutner: No problem, I've updated the PR—let me know if that looks good for you!

fmeum (approved):

go/private/sdk.bzl
fmeum: Looks good! Only the CLA remains, let me know if you run into issues with it.

zakcutner: Thanks! Sorry, this is my first time contributing so I need to get added to the right group within my company. I'll let you know when that's done

zakcutner:
@fmeum I've got the CLA sorted so I think we're ready to merge 😄

PR #3336, opened 2022-10-31
Match GOARCH ppc64 to @platforms//cpu:ppc as well.

### LINKED ISSUE #3335: Fails to match host Go toolchain on linux/ppc64

Git master

None

5.3.2

Yes?

linux/ppc64

Go 1.19.2 toolchain for linux/ppc64 built from source

### What did you do?

Basic usage with `go_register_toolchains(version = "host")` and `go_binary` targets.
The same WORKSPACE and BUILD files have been tested on other platforms such as darwin/arm64 and windows/amd64.
See: https://github.com/mahiuchun/adventofcode-2018

It works like other platforms.

### What did you see instead?

It errors with something like `While resolving toolchains for target //day01:part2: no matching toolchains found for types @io_bazel_rules_go//go:toolchain`

mahiuchun:
**What type of PR is this?**

Bug fix

It matches GOARCH ppc64 to something so that rules_go would work on linux/ppc64.

Fixes #3335

**Other notes for review**

ppc64 is the big endian version of 64-bit PowerPC.
While it is not very common these days, I think it makes sense to support ppc64 as long as Go itself supports ppc64.

### DIFF
--- a/go/private/platforms.bzl
+++ b/go/private/platforms.bzl
@@ -30,6 +30,7 @@ BAZEL_GOARCH_CONSTRAINTS = {
     "amd64": "@platforms//cpu:x86_64",
     "arm": "@platforms//cpu:arm",
     "arm64": "@platforms//cpu:aarch64",
+    "ppc64": "@platforms//cpu:ppc",
     "ppc64le": "@platforms//cpu:ppc",
     "s390x": "@platforms//cpu:s390x",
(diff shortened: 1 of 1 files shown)


fmeum:
I obviously can't test this, bit it looks reasonable.

fmeum (approved):

PR #3338, opened 2022-10-31
gopackagesdriver: default to NotHandled:true for package queries

### LINKED ISSUE #3334: Gopackagesdriver: tell language server to fall back when bazel query fails

0.35.0

N/A

```
Bazelisk version: development
Build label: 5.3.2
Build time: Wed Oct 19 18:29:41 2022 (1666204181)
Build timestamp: 1666204181
Build timestamp as int: 1666204181
```

Yes

macOS 12.6, x86_64

Using GOPACKAGESDRIVER with 
```
Version: 1.72.2
Commit: d045a5eda657f4d7b676dedbfa7aab8207f8a075
Date: 2022-10-12T22:16:30.254Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Darwin x64 21.6.0
Sandboxed: No
```

### What did you do?

Set up a simple `rules_go` repo and add a file not yet tracked by Bazel (`untracked-file` branch)
https://github.com/ian-h-chamberlain/gopackagesdriver-repro/tree/untracked-file

I don't necessarily expect completions, etc. to work on the new file, but I also don't expect completion to block for a very long time. In this particular case, `xyz.go` is not part of any Bazel target, but the default `gopls` package `List` implementation can handle it and provide completions where `gopackagesdriver` cannot.

### What did you see instead?

Asking for completions takes quite a long time and never actually works correctly in `xyz.go`.

I believe this happens because `gopackagesdriver` responds to the language server that it _can_ handle queries about the file, even though `bazel query` has errors:
```json
{"NotHandled":false,"Sizes":{"WordSize":8,"MaxAlign":8},"Packages":[]}
```

I believe that simply by responding with `"NotHandled":true` when `bazel query` fails, the language server will fallback to its default implementation and completion, etc. will work mostly correctly.
The `main` branch includes a proof-of-concept `gopackagesdriver.sh` that responds this way for any file with `xyz` in the name:
https://github.com/ian-h-chamberlain/gopackagesdriver-repro/tree/main

Hopefully this is an easy fix that will make things work nicer when a file isn't yet included in Bazel rules so that `gopls` can fall back to its default behavior!

ian-h-chamberlain:
**What type of PR is this?**
Bug fix / minor enhancement (?)

Default the driver response with `NotHandled` set to `true` so that if any errors are encountered during `bazel query` or otherwise, the language server will fallback to its default implementation.

Fixes #3334 by telling gopls to fallback if the gopackagesdriver can't `bazel query` about the requested file.

**Other notes for review**
No other change is needed because any driver response with actual contents will be populated with `false` here: https://github.com/bazelbuild/rules_go/blob/master/go/tools/gopackagesdriver/json_packages_driver.go#L54

### DIFF
--- a/go/tools/gopackagesdriver/main.go
+++ b/go/tools/gopackagesdriver/main.go
@@ -62,7 +62,7 @@ var (
 	additionalAspects     = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_ADDTL_ASPECTS"))
 	additionalKinds       = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_KINDS"))
 	emptyResponse         = &driverResponse{
-		NotHandled: false,
+		NotHandled: true,
 		Sizes:      types.SizesFor("gc", "amd64").(*types.StdSizes),
 		Roots:      []string{},
 		Packages:   []*FlatPackage{},


fmeum (approved):
@linzhp Looks like a strict improvement to me, but please take a look as well.

JamyDev (approved):

PR #3340, opened 2022-11-04
Add 'gc_goopts' flag to command line to allow gc_goopts to propagate

### LINKED ISSUE #2402: Set gc_goopts for all targets in a build

0.22.1

0.20.0

2.1.1

Yes

Linux amd64 

Trying to build a go_binary as a c-archive with all dependencies build with gc_goopts=-d=libfuzzer

### What did you do?

So I tried setting `gc_goopts="-d=libfuzzer"` on my go_binary target, but this flag is only applied at compile time to that specific target. So I need to apply this flag to all libraries / dependencies up to this target. 

I have tried a simple bazel build setting flag like `--define=gc_goopts=-d=libfuzzer`, but this is not supported in rules_go. I've considered a transition rule, but I'm not sure it could be applied here.

What do you recommend as the best approach to support this functionality?

This is part of https://github.com/bazelbuild/rules_go/issues/2284, which I managed to get working, but missing some coverage symbols since `-d=libfuzzer` is not applied to all targets.

LWarrens:
**What type of PR is this?**
Feature

Appends '-gcflags' options to all executions of compilepkg triggered by a call containing the 'gc_goopts' flag

Fixes #2402


linzhp:
`gcc` has a flag `-D` to define macros, which is why there is a `define` attribute in cc rules. Go compiler (gc) has no such flag. Instead, you are trying to pass `gc_goopts` to the compiler, why don't you call it `gc_goopts`?

LWarrens:
changed the name to gc_goopts_global

gc_goopts exists in bazel as various rule attributes that get passed around the context but aren't persistent.
gc_goopts_global only exists as a flag reference and value accessed through BuildSetting provider; the single labeled build setting serves as a single source of truth as well - so the same name with the "_global" suffix seems fitting.

go/private/actions/compilepkg.bzl
```diff
@@ -117,8 +122,9 @@ def emit_compilepkg(
         outputs.append(out_cgo_export_h)
     if testfilter:
         args.add("-testfilter", testfilter)
-
```
linzhp, line 120: Why deleting this empty line?

linzhp:
Also, please make sure CI passes.

go/config/BUILD.bazel
```diff
@@ -74,3 +74,9 @@ filegroup(
     srcs = glob(["**"]),
     visibility = ["//visibility:public"],
 )
+
+string_flag(
```
linzhp, line 78: Have you tested the case when you pass multiple flags? `gc_goopts` and `gc_linkopts` are similar enough that I don't want to make one of them `string_list_flag`, the other `string_flag`. People reading the code in the future will be confused.

linzhp: Because of the call to `_quote_opts`, I suspect you won't be able to pass multiple gc flags if you make `gc_goopts` as `string_flag`. Can you verify?

https://github.com/bazelbuild/rules_go/blob/3d19fb53089ddbaf3732c4083cd1a9e2066f08b1/go/private/actions/compilepkg.bzl#L132

LWarrens: I did initially - not since appending to mode though, I thought _quote_opts would mess with things too but gc_flags is a list that includes race and msan conditions
iirc it shows up as -gcflags '--first=1 --second=2'
going to try to see output now

LWarrens: I can switch it to string_list flag it does contain escaped quotes in this case
`bazel-out/k8-opt-exec-2B5CBBC6/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix linux_amd64 -src tests/core/go_binary/hello.go -embedroot '' -embedroot bazel-out/k8-fastbuild/bin -embedlookupdir tests/core/go_binary -importpath hello_nopie_bin -p main -package_list bazel-out/k8-opt-exec-2B5CBBC6/bin/external/go_sdk/packages.txt -o bazel-out/k8-fastbuild/bin/tests/core/go_binary/hello_nopie_bin.a -x bazel-out/k8-fastbuild/bin/tests/core/go_binary/hello_nopie_bin.x -gcflags ''\''-m=2 -m=3'\''' -asmflags '' -cppflags '-iquote .' ...`

linzhp: Did you try passing multiple real gc flags and see it gc is OK with it? The command you shared only shows the args Bazel calls `builder compilepkg`, but it doesn't show how `builder` actually calls to gc. Like `compile --first=1 --second=2` is different than `compile '--first=1 --second=2'`.

LWarrens: It was wrong initially, after changing to string flag I input "-m=2 -c=1"

go/private/context.bzl
```diff
@@ -391,6 +391,7 @@ def go_context(ctx, attr = None):
     stdlib = None
     coverdata = None
     nogo = None
+    gc_goopts = None
```
linzhp, line 394: Now that gc_goopts is read from `ctx.attr.gc_goopts[BuildSettingInfo].value`, You don't need this anymore, right?

linzhp (commented):
Sorry, one last thing that I missed earlier, then we are good to go

linzhp:
Also, please update the PR title and descriptions to match the final approach.

JamyDev (approved):

go/private/context.bzl
LWarrens: That's right, done!

PR #3341, opened 2022-11-06
Move new runfiles library to `//go/runfiles`

fmeum:
Follows https://bazel.build/rules/deploying#runfiles_library.


fmeum:
FYI @sluongng

linzhp:
Since go/tools/bazel/runfiles is newly created and it's not released, why not moving the library to `go/runfiles`?

achew22:
+1 let's move it instead of aliasing. The imports will be easier to work with as well

fmeum:
Sounds good, will do. Should I keep the tests in the same Bazel package or move them to `tests`?

achew22 (approved):

achew22:
Wrt tests that's dealers choice. I have a slight bias to move but don't mind

fmeum:
@achew22 I moved the tests under `tests/`.

Issue #3315, opened 2022-10-07
linker warnings on macOS XCode 14: `ld: warning: -undefined dynamic_lookup may not work with chained fixups`


garymm:
Note I reported a similar issue for `cc_library` [here](https://github.com/bazelbuild/bazel/issues/16413).

If we apply the same fix, then it would just be to remove this line:
https://github.com/bazelbuild/rules_go/blob/d517cd9bec43f0541b421315fe633cdd0f204270/go/tools/builders/cgo2.go#L215

### What version of rules_go are you using?

v0.34.0

### What version of gazelle are you using?

v0.26.0

### What version of Bazel are you using?

5.3.1

### Does this issue reproduce with the latest releases of all the above?

yes

### What operating system and processor architecture are you using?

macOS 12.6, arm64

### Any other potentially useful information about your toolchain?

XCode Command Line tools v14

### What did you do?

Build anything that depends on CGo

### What did you expect to see?

No warnings

### What did you see instead?

```
NFO: From GoLink ...:
ld: warning: -undefined dynamic_lookup may not work with chained fixups
```

garymm:
OK I looked at this more closely and it looks like this code path is only ever done as a fallback, so it's not actually resulting in warnings in the normal case. The warnings I was seeing my build are due to the cc_library stuff that's built in to bazel and is fixed by https://github.com/bazelbuild/bazel/pull/16414.

Either the fallback should be in place for MacOS and Linux, or for neither. I'm going to close this.

PR #3345, opened 2022-11-10
don't set -undefined dynamic_lookup on darwin, ios

garymm:
**What type of PR is this?**

Bug fix

Don't set -undefined dynamic_lookup when building cgo dependencies on darwin, ios.
Needed because that option is deprecated in XCode 14+.
See discussion [here](https://github.com/bazelbuild/bazel/issues/16413).

Fixes https://github.com/bazelbuild/rules_go/issues/3315

**Other notes for review**

Not sure about testing.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -64,7 +64,7 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 	// that a fragment of the path is stable and human friendly enough to be
 	// referenced in nogo configuration.
 	workDir = filepath.Join(workDir, "cgo", packagePath)
-	if err := os.MkdirAll(workDir, 0700); err != nil {
+	if err := os.MkdirAll(workDir, 0o700); err != nil {
 		return "", nil, nil, err
 	}
 
@@ -112,7 +112,7 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 		}
 	} else {
 		srcDir = filepath.Join(workDir, "cgosrcs")
-		if err := os.Mkdir(srcDir, 0777); err != nil {
+		if err := os.Mkdir(srcDir, 0o777); err != nil {
 			return "", nil, nil, err
 		}
 		copiedSrcs, err := gatherSrcs(srcDir, cgoSrcs)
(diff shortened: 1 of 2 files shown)


fmeum (approved):

fmeum:
Thanks. If the existing tests pass, that should be good enough.

garymm:
@fmeum I basically had to disable the relevant test on MacOS for this to pass.

From reading [this](https://openradar.appspot.com/radar?id=5536824084660224), it seems like we have to choose between supporting this feature (allowing undefined symbols at link time) and having a build with no warnings.

For my uses, I want a build with no warnings, so I'd be happy to have this merged, but might be prudent to wait to see if Apple replies to that.
LMKWYT.

fmeum:
Reading through the linked issue, we could alternatively explicitly enumerate all undefined symbols. But given that that results in the same linker input, is highly backwards incompatible and not even known to work (has Apple clarified the "may not work" anywhere?), I think it's too early to make this change.

fmeum:
@keith Do you happen to have more insight into what exactly "may not work" with undefined symbols? I saw you recently removed this flag from Bazel's CC toolchain?

keith:
I haven't seen any concrete issues. But I also think in most cases it isn't actually needed anyways, so unless there is some very specific use case it likely isn't the right default.

fmeum:
@linzhp What do you think, should we remove this flag and document in the release notes that users have to add it manually if needed?

linzhp:
Dumb question: if "-undefined dynamic_lookup may not work with chained fixups", why don't we disable chained fixups? It seems "-undefined dynamic_lookup" is needed for dynamic libraries.

keith:
yea disabling chained fixups is definitely an option for now (although I imagine in the future it will be a dyld requirement). I imagine apple's ideal here is that `-undefined dynamic_lookup` should not be required for dynamic libraries and that instead you should setup your dependency tree such that each library can at least see that other libraries with those symbols will exist

linzhp:
I am concerned that we have to disable a cgo test in macOS for this change. Removing a warning at the cost of not supporting a cgo scenarios doesn't seem worthwhile.

keith:
we can add this flag directly to the places that need it. but that would likely still be preferred to having it set globally

garymm:
Some context from [here](https://github.com/python/cpython/issues/97524#issuecomment-1442449271):
It seems on the newest XCode (14.3), setting `-undefined dynamic_lookup` will *not* result in a warning but *will* result in disabling a code size optimization (chained fixups). I guess my vote would be to merge this PR and let people add this flag back where it's needed, but I guess either way the warning will go away with XCode 14.3.

linzhp:
OK. Can you fix the test for macOS?

garymm:
@linzhp is there something other than what runs in the GitHub checks? If so, please tell me how to run it.
Because all GitHub checks pass.

garymm:
Actually let me see if I can re-enable the test I disabled but set the linker flags on it.

garymm:
OK I looked at this more closely and it looks like this code path is only ever done as a fallback, so it's not actually resulting in warnings in the normal case. The warnings I was seeing my build are due to the cc_library stuff that's built in to bazel and is fixed by https://github.com/bazelbuild/bazel/pull/16414.

So I actually don't think this PR makes sense now. Either the fallback should be in place for MacOS and Linux, or for neither. I'm going to close this and the attached issue.

PR #3348, opened 2022-11-14
Declare toolchains in a separate repository

### LINKED ISSUE #3196: Repositories declared with go_wrap_sdk, etc. are downloaded, even when not used

Latest master (8d6b21cbd13131e05be1626376c9af160517c264)

Latest master (32dab97bac88b1a35653d33a034b38e55d757983)

```
> bazelisk version
Bazelisk version: v1.11.0
WARNING: Ignoring JAVA_HOME, because it must point to a JDK, not a JRE.
Build label: 5.2.0
Build target: bazel-out/darwin_arm64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Jun 7 16:11:09 2022 (1654618269)
Build timestamp: 1654618269
Build timestamp as int: 1654618269
```

Reproduces with bazel at commit `6aa210d4e2a1537eb504a5ca8fc044b79a0b7ddc` (`USE_BAZEL_VERSION=last_green bazelisk build ...`)

macOS 12.4 (Monterey), arm64

We have multiple `go_*sdk` repository declarations with different names.  This is discussed further below.

### What did you do?

We currently have two different golang SDKs available to download for our developers:

- A custom-built Linux one, downloaded by an internal rule and wrapped with `go_wrap_sdk`
- An unmodified macOS one, downloaded using `go_download_sdk`.

Due to the fact that these are downloaded in different ways, we cannot unify behind a single `go_sdk` repository.  I understand that there is a documented [limitation](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#registering-a-custom-sdk) that all go SDK repositories must be named `go_sdk`, but in practice, I have found this to not be the case.

One characteristic of our custom downloader (for the Linux instance) refuses to download binaries on macOS systems due to platform incompatibility.  This is simulated via the code in https://gist.github.com/nacl/34d33864fc4a31a933eafc8a18e7954e.  Note that `i386_linux_go_sdk` is intentionally given a URL that should not resolve.

It may need to be modified to support the platform you are currently on.

The issue can be reproduced by running:

```
bazel build @bazel_gazelle//repo
```

Bazel would not attempt to fetch the `i386_linux_go_sdk` repo, and only downloads the `macos_go_sdk`.

### What did you see instead?

Bazel attempts to fetch `i386_linux_go_sdk`, and the build fails.

My prior investigation into this problem found that this is due to the fact that `go_download_sdk`, `go_wrap_sdk` and friends always declare and register all toolchains in their macro implementations.  Unfortunately, Bazel seems to expect to have all repositories referenced in `register_toolchains` calls available prior to toolchain resolution, so this combination leads to all of these go repositories being downloaded regardless of whether the toolchain is actually needed.

As an internal workaround, we have patched rules_go to make toolchain registration optional, and then manually declare and register the relevant toolchains in our main repository.  There may be a more elegant solution to this problem.

jfirebaugh:
**What type of PR is this?**

Other (performance improvement)

Declares toolchains in a separate repository. For instance, if the SDK repository is `@go_sdk`, the toolchain rules will be defined in `@go_sdk_toolchains`. This avoids eagerly fetching `@go_sdk`, fixing #3196.

One caveat: `:sdk_version_setting` needs to be defined in the same repo as the toolchain rule, e.g. in `@go_sdk_toolchains`. I tried `target_settings = "@go_sdk//:sdk_version_setting"`, but that causes `@go_sdk` to be eagerly fetched again.

This means that `@go_sdk_toolchains` needs to know the SDK version. There are two possibilities:
* The user specifies the version up front, for instance via `go_register_toolchains(version = ...)`. In that case, `@go_sdk_toolchains` uses this version.
* The user does not specify the version up front. `go_wrap_sdk`, `go_download_sdk`, and `go_local_sdk` do not require a `version` attribute. In this case, we have no choice but to inspect the SDK itself to obtain the version, forcing an eager load. However, we can at least add an optional `version` attribute to these rules, so that users can avoid the eager fetch if they wish.

In all cases, when a `version` attribute is provided by the user, the implementation (lazily) validates that it matches the actual SDK version.

Fixes #3196.

**Other notes for review**

As discussed on Slack: https://bazelbuild.slack.com/archives/CDBP88Z0D/p1668022802395059

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -22,7 +22,7 @@ use_repo(
 go_sdk = use_extension("//go:extensions.bzl", "go_sdk")
 go_sdk.download(name = "go_default_sdk", version = "1.18.3")
 use_repo(go_sdk, "go_default_sdk")
-register_toolchains("@go_default_sdk//:all")
+register_toolchains("@go_default_sdk_toolchains//:all")
 
 bazel_dep(name = "gazelle", version = "0.27.0")
 go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -21,7 +21,7 @@ use_repo(
 
 go_sdk = use_extension("//go:extensions.bzl", "go_sdk")
 go_sdk.download(name = "go_default_sdk", version = "1.18.3")
-use_repo(go_sdk, "go_default_sdk", "go_default_sdk_toolchains")
+use_repo(go_sdk, "go_default_sdk_toolchains")
 register_toolchains("@go_default_sdk_toolchains//:all")
 
 bazel_dep(name = "gazelle", version = "0.27.0")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -21,7 +21,7 @@ use_repo(
 
 go_sdk = use_extension("//go:extensions.bzl", "go_sdk")
 go_sdk.download(name = "go_default_sdk", version = "1.18.3")
-use_repo(go_sdk, "go_default_sdk")
+use_repo(go_sdk, "go_default_sdk", "go_default_sdk_toolchains")
 register_toolchains("@go_default_sdk_toolchains//:all")
 
 bazel_dep(name = "gazelle", version = "0.27.0")
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -127,9 +127,6 @@ TOOLS_NOGO = [
 # check this to determine compatibility.
 RULES_GO_VERSION = "0.35.0"
 
-declare_go_toolchains = _declare_go_toolchains
-declare_bazel_toolchains = _declare_bazel_toolchains
-
 def declare_toolchains(host, sdk, builder, sdk_version_setting):
     host_goos, _, host_goarch = host.partition("_")
     _declare_go_toolchains(
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -41,6 +41,7 @@ load(
 )
 load(
     "//go/private:go_toolchain.bzl",
+    _declare_go_toolchains = "declare_go_toolchains",
     _declare_toolchains = "declare_toolchains",
     _go_toolchain = "go_toolchain",
 )
@@ -126,6 +127,7 @@ TOOLS_NOGO = [
 # check this to determine compatibility.
 RULES_GO_VERSION = "0.35.0"
 
+declare_go_toolchains = _declare_go_toolchains
 declare_toolchains = _declare_toolchains
 go_context = _go_context
 go_embed_data = _go_embed_data
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -41,8 +41,8 @@ load(
 )
 load(
     "//go/private:go_toolchain.bzl",
+    _declare_bazel_toolchains = "declare_bazel_toolchains",
     _declare_go_toolchains = "declare_go_toolchains",
-    _declare_toolchains = "declare_toolchains",
     _go_toolchain = "go_toolchain",
 )
 load(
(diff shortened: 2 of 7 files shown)


go/def.bzl
```diff
@@ -126,6 +127,7 @@ TOOLS_NOGO = [
 # check this to determine compatibility.
 RULES_GO_VERSION = "0.35.0"
 
+declare_go_toolchains = _declare_go_toolchains
 declare_toolchains = _declare_toolchains
```
fmeum, line 131: `declare_toolchains` isn't documented anywhere, but it is in a "public" file. Could you make it a macro that just calls to two macros under `go/private`, one that defines `toolchain`s and one that defines `go_toolchains`?

go/private/sdk.bzl
```diff
@@ -129,8 +140,82 @@ _go_download_sdk = repository_rule(
     },
 )
 
+def _version_constants(version):
```
fmeum, line 143: I would prefer a more descriptive name (e.g. `define_version_constants`)

fmeum:
In order to fix the failing test, update the `register_toolchains` line in `MODULE.bazel` to `@go_default_sdk_toolchains//:all`.

go/def.bzl
fmeum: Mostly yes, but could you also move `declare_X_toolchains` out of `go/defs.bzl`? I would rather choose not to export them publicly, the are only intended to be used by our own generated SDK build files for now. We can always make them public API later if users need them.

jfirebaugh:
Any hints about how to fix the remaining CI failure?

fmeum:
Adding `"go_default_sdk_toolchains"` as a positional argument to https://github.com/bazelbuild/rules_go/blob/3d19fb53089ddbaf3732c4083cd1a9e2066f08b1/MODULE.bazel#L24 should fix it.

MODULE.bazel
```diff
@@ -21,7 +21,7 @@ use_repo(
 
 go_sdk = use_extension("//go:extensions.bzl", "go_sdk")
 go_sdk.download(name = "go_default_sdk", version = "1.18.3")
-use_repo(go_sdk, "go_default_sdk")
+use_repo(go_sdk, "go_default_sdk", "go_default_sdk_toolchains")
```
fmeum, line 24: I actually think we should be able to get rid of the `"go_default_sdk"` here. Could you give that a try?

fmeum (approved):

go/toolchains.rst
```diff
@@ -225,6 +235,10 @@ SDK version to use (for example, :value:`"1.15.5"`).
 | Normally this is set to a Go version like :value:`"1.15.5"`. It may also be                      |
 | set to :value:`"host"`, which will cause rules_go to use the Go toolchain                        |
 | installed on the host system (found using ``GOROOT`` or ``PATH``).                               |
+|                                                                                                  |
+| If ``version`` is specified and is not set to :value:`"host"`, the SDK will be lazily fetched    |
```
fmeum, line 239: Sorry for not seeing this earlier, but I think we can clarify this a bit more. How about:

"If version ..., the SDK will only be fetched when the build uses a Go toolchain and [toolchain resolution](https://bazel.build/extending/toolchains#toolchain-resolution) results in this SDK being chosen. Otherwise, it will be fetched unconditionally.

I could see "lazily" and "eagerly" being slightly confusing here as "lazily" can very well mean "never" and "eagerly" can mean "always". I would also like to avoid referring to the "analysis phase" if we don't have to.

fmeum:
@linzhp Could you test this at Uber?

fmeum (approved):

linzhp:
Will test this next week

linzhp:
Tests passed

PR #3359, opened 2022-11-23
bzlmod: Add missing `strip_prefix` field to `source.template.json`

fmeum:
This field is required even if its value is empty.


fmeum:
FYI @achew22

PR #3360, opened 2022-11-24
Update docs regarding vendored proto files

garymm:
**What type of PR is this?**

Documentation

https://github.com/bazelbuild/bazel/issues/3867 is fixed, so
instructions for vendored proto files need to be updated.

### DIFF
--- a/proto/core.rst
+++ b/proto/core.rst
@@ -13,7 +13,6 @@ Go Protocol buffers
 .. _Bourne shell tokenization: https://docs.bazel.build/versions/master/be/common-definitions.html#sh-tokenization
 .. _gogoprotobuf: https://github.com/gogo/protobuf
 .. _compiler.bzl: compiler.bzl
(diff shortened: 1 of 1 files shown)


fmeum:
> proto_library is now provided by rules_proto, not built-in.

Which Bazel version contains this change?

garymm:
@fmeum not sure. The docs [here](https://bazel.build/reference/be/protocol-buffer#proto_library) say:

> Deprecated. Please use https://github.com/bazelbuild/rules_proto instead.

fmeum:
I think that `proto_library` is still part of Bazel. This deprecation was added since the long-term plan is to load `proto_library` from `rules_proto`. But that would require having everyone load it from `rules_proto` explicitly.

Could we split this PR in two parts? The other half seems completely non-controversial.

garymm:
What is the objection to directing people to use the non-deprecated proto_library?

fmeum (approved):
Thanks!

PR #3370, opened 2022-11-30
go link: use external linker when in race mode

motiejus:
As of clang 15.0.3 (via zig v0.10), when building with `race = "on"` on x86_64 Linux, we observe the following:

    runtime/cgo(.text): relocation target memset not defined

Using an external linker when in race mode fixes that. Previously `linkmode=external` was set for windows-only race builds, now lets just apply it indiscriminately.

From my past experience, reporting a Go linker error when the external linker works has a high chance to get the ticket closed as unactionable; so it makes sense to just use an external linker, when it works.

> Bug fix

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -89,12 +89,14 @@ def emit_link(
     tool_args = go.tool_args(go)
 
     # Add in any mode specific behaviours
-    tool_args.add_all(extld_from_cc_toolchain(go))
+    extld = extld_from_cc_toolchain(go)
+    tool_args.add_all(extld)
     if go.mode.race:
         tool_args.add("-race")
     if go.mode.msan:
         tool_args.add("-msan")
     if ((go.mode.static and not go.mode.pure) or
+        (go.mode.race and extld) or
         go.mode.link != LINKMODE_NORMAL or
         go.mode.goos == "windows" and (go.mode.race or go.mode.msan)):
         # Force external linking for the following conditions:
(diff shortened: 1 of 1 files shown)


go/private/actions/link.bzl
```diff
@@ -96,16 +96,18 @@ def emit_link(
         tool_args.add("-msan")
     if ((go.mode.static and not go.mode.pure) or
         go.mode.link != LINKMODE_NORMAL or
-        go.mode.goos == "windows" and (go.mode.race or go.mode.msan)):
+        go.mode.race or
+        go.mode.goos == "windows" and go.mode.msan):
```
fmeum, line 100:
```suggestion
        (go.mode.goos == "windows" and go.mode.msan)):
```
Makes precedence more obvious

go/private/actions/link.bzl
```diff
@@ -96,16 +96,18 @@ def emit_link(
...
         # * Non-normal build mode: may not be strictly necessary, especially
         #   for modes like "pie".
-        # * Race or msan build for Windows: Go linker has pairwise
-        #   incompatibilities with mingw, and we get link errors in race mode.
-        #   Using the C linker avoids that. Race and msan always require a
-        #   a C toolchain. See #2614.
+        # * Msan build for Windows: Go linker has pairwise incompatibilities
+        #   with mingw. Using the C linker avoids that. Race and msan always
+        #   require a C toolchain. See #2614.
+        # * Race builds (previously Windows, later +linux): we get linker
```
fmeum, line 109: Let's document the current state only, not the history of this part of the code. We can always rely on git history for that. 

Please also add an example of the error you get with the internal linker.

motiejus: Done, both suggestions.

fmeum (approved):

fmeum:
@linzhp Looks good to me, I will leave merging to you

motiejus:
Addressed @linzhp 's concerns raised offline.

fmeum (approved):

PR #3371, opened 2022-12-02
fix(packagesdriver): bazelFlags should prefix the command

JamyDev:
**What type of PR is this?**

Bug fix

General bazel flags (eg `--output_base`) are required to be set before the action/command you'd like to run.

This PR adds the bazelFlags that are defined by the `GOPACKAGESDRVER_BAZEL_FLAGS` env variable into the Bazel struct, so we can prefix them ahead of the command to be ran. `GOPACKAGESDRVER_BAZEL_BUILD_FLAGS` and `GOPACKAGESDRVER_BAZEL_QUERY_FLAGS` are still appended at the end.

```
▶ bazel info --output_base=/some/path 
ERROR: --output_base=/some/path :: Unrecognized option: --output_base=/home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f_pkgdrv

▶ bazel --output_base=/some/path info 
bazel-bin: /some/path/execroot/__main__/bazel-out/k8-fastbuild/bin
bazel-genfiles: /some/path/execroot/__main__/bazel-out/k8-fastbuild/bin
bazel-testlogs: /some/path/execroot/__main__/bazel-out/k8-fastbuild/testlogs
...
output_base: /some/path
output_path: /some/path/execroot/__main__/bazel-out
...
```

N/A

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/bazel.go
+++ b/go/tools/gopackagesdriver/bazel.go
@@ -34,10 +34,10 @@ const (
 )
 
 type Bazel struct {
-	bazelBin      string
-	workspaceRoot string
-	bazelFlags    []string
-	info          map[string]string
+	bazelBin          string
+	workspaceRoot     string
+	bazelStartupFlags []string
+	info              map[string]string
 }
 
 // Minimal BEP structs to access the build outputs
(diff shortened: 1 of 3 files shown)


achew22 (approved):

go/tools/gopackagesdriver/bazel.go
```diff
@@ -36,6 +36,7 @@ const (
 type Bazel struct {
 	bazelBin      string
 	workspaceRoot string
+	bazelFlags    []string
```
linzhp, line 39: Nit: following the convention of [bazelrc](https://bazel.build/run/bazelrc), it's more clear to call them `startupFlags` or `startupOptions`.

PR #3372, opened 2022-12-02
fix(gpd): Write large target patterns to file

JamyDev:
**What type of PR is this?**

Bug fix

In rare cases (querying an entire large repo for it's packages) our query will return too many targets to fit inside the arguments length for your system. 

Bazel added `--target_pattern_file` (https://github.com/bazelbuild/bazel/pull/10856) for this exact scenario. This implements that by creating a temp file with all the target patterns.

Marking draft as I want to test this out a little more.

N/A

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/bazel_json_builder.go
+++ b/go/tools/gopackagesdriver/bazel_json_builder.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -166,7 +167,24 @@ func (b *BazelJSONBuilder) Build(ctx context.Context, mode LoadMode) ([]string,
...
+
+	if len(labels) < 100 {
+		buildArgs = append(buildArgs, labels...)
+	} else {
+		// To avoid hitting MAX_ARGS length, write labels to a file and use `--target_pattern_file`
+		targetsFile, err := ioutil.TempFile("", "gopackagesdriver_targets_")
+		if err != nil {
+			return nil, fmt.Errorf("unable to create target pattern file: %w", err)
+		}
+		targetsFile.WriteString(strings.Join(labels, "\n"))
```
linzhp, line 180: Can you write to the file line by line instead? It would have much smaller memory footprint

JamyDev: Done

go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -177,7 +178,15 @@ func (b *BazelJSONBuilder) Build(ctx context.Context, mode LoadMode) ([]string,
...
 			return nil, fmt.Errorf("unable to create target pattern file: %w", err)
 		}
-		targetsFile.WriteString(strings.Join(labels, "\n"))
+		writer := bufio.NewWriter(targetsFile)
+		defer writer.Flush()
+		for _, l := range labels {
+			writer.WriteString(l+"\n")
+		}
+		err = writer.Flush()
+		if err != nil {
```
linzhp, line 187:
```suggestion
		if err := writer.Flush(); err != nil {
```
nit

linzhp (approved):

PR #3379, opened 2022-12-05
add cpu type loongarch64

yangjiwei-source:
add cpu type loongarch64


fmeum:
@yangjiwei-source I don't see this arch in https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD. You may have to add it there first and wait for a release.

PR #3385, opened 2022-12-06
always produce .a files at the beginning of a build

matloob:
This change updates rules_go in anticipation for Go 1.20, which will ship without .a files for the standard library. Now, instead of conditionally building the standard library if the configuration differs from the default, the standard library will always be built. In addition, the builder will be built using 'go build' instead of individual invocations of go tool compile/link to avoid needing to bootstrap the standard library before producing the builder.

**What type of PR is this?**

>
> Other

It updates rules_go to run a go install to produce stdlib .a files (which won't be shipped with 1.20). It's necessary to get rules_go working with 1.20.

Fixes #

**Other notes for review**


achew22:
In no way objecting to this change since it or something like it is necessary, but out of curiosity, what is the performance implication (or benefit) or switching from `"{go} tool compile ..."` to `{go} build ..."`?

go/private/rules/binary.bzl
```diff
@@ -415,60 +416,32 @@ def _go_tool_binary_impl(ctx):
...
-    )
-
     return [DefaultInfo(
         files = depset([out]),
         executable = out,
     )]
 
 go_tool_binary = rule(
     implementation = _go_tool_binary_impl,
+    cfg = go_tool_transition,  # For fmeum: Is this correct? do we need go_reset_target?
```
fmeum, line 444: `non_go_tool_transition` may perform better here (the bootstrap rules don't use nogo, so the setting shouldn't matter), but better verify that with `cquery`.

matloob:
> In no way objecting to this change since it or something like it is necessary, but out of curiosity, what is the performance implication (or benefit) or switching from `"{go} tool compile ..."` to `{go} build ..."`?

On Go 1.20+ the `go build` will build the necessary parts of the standard library needed to build the builder program which the `go tool compile` commands didn't need to do because the stdlib .a files existed in the distribution. On Go 1.19 and earlier there shouldn't be much of a performance difference over the go command analysis overhead, which is minimal, because the standard library will be found in the pkg/ directories.

The "annoying" but hard to avoid thing is that on 1.20+ we will build part of the standard library when `go build`ing the builder, then throwing away that standard library because the GOCACHE is temporary, then rebuilding the whole standard library in the //:stdlib rule. But transferring the cache over will be difficult to do.

fmeum (approved):
Thanks for figuring out the Windows build, the change looks good.

We shouldn't merge this yet though as it is only required with Go 1.20 and regresses build times with currently available versions of Bazel.

@matloob Could you add a temporary commit that uses Go 1.20 in CI?

matloob:
> Thanks for figuring out the Windows build, the change looks good.
> 
> We shouldn't merge this yet though as it is only required with Go 1.20 and regresses build times with currently available versions of Bazel.

When do you think we'll be able to get this in? I think it would be nice to make sure that people can easily test with the release candidate once that's issued.
 
> @matloob Could you add a temporary commit that uses Go 1.20 in CI?

Could you give me a pointer to what to modify to do that?

fmeum:
> When do you think we'll be able to get this in? I think it would be nice to make sure that people can easily test with the 1.20 release candidate (which was released yesterday),

That's a good point. At the cost of making this change more complex, we could probably make the new logic conditional on the SDK version. We accept this version [here](https://github.com/bazelbuild/rules_go/blob/1a8fe64877c6e71dbbf51cdc8ceb2eb10c13e521/go/private/sdk.bzl#L203) and would need to wire this through all the way into the sdk properties available at analysis time. @linzhp @achew22 Do you think we should do this?

> 
> > @matloob Could you add a temporary commit that uses Go 1.20 in CI?
> 
> Could you give me a pointer to what to modify to do that?

Are there binaries for the release candidate? If so, you could add [go_download_sdk](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#go-download-sdk) to the `WORKSPACE` file before the `go_register_toolchain` call.

linzhp:
yeah, I think we should add that conditional logic on SDK versions. It may take a while for big repos to fix all issues and upgrade to Go 1.20.

fmeum:
@matloob Do you want to take a stab at this conditional logic? We can also pair on it next week.

matloob:
> @matloob Do you want to take a stab at this conditional logic? We can also pair on it next week.

I'll give it a try

> Are there binaries for the release candidate? If so, you could add [go_download_sdk](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#go-download-sdk) to the `WORKSPACE` file before the `go_register_toolchain` call.

Got it, I'll do that.

matloob:
Added conditional logic to only build a new SDK for Go 1.20+.

Set the version to 1.20rc1 in the WORKSPACE file for CI purposes.

fmeum (approved):
Looks good to me with one comment, we should be able to merge this right away.

@linzhp Could you test at Uber once the comment has been resolved?

go/private/actions/stdlib.bzl
```diff
@@ -59,6 +59,12 @@ def _has_precompiled_stdlib(version_string):
     dot = minor.find(".")
     if dot != -1:
         minor = minor[:dot]
+    rc = minor.find("rc")
+    if rc != -1:
+        minor = minor[:rc]
+    beta = minor.find("beta")
+    if beta != -1:
+        minor = minor[:beta]
```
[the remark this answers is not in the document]
matloob: > ~Looks good to me with one comment, we should be able to merge this right away.~
> 
> ~@linzhp Could you test at Uber once the comment has been resolved?~
> 
> Missed the coverage test failure, will also look into it.

I think the coverage test is the same issue on the boringcrypto cl: that we need to make the path to "Tool" relative on 1.19+.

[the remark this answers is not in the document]
fmeum: I ran it locally with extra output and it complained about `-test.coverprofile` not being supported when built without coverage. Since the CI is passing for the other PR, I rather suspect something changed with respect to coverage instrumentation in Go 1.20.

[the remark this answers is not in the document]
matloob: Ah, got it, that will require a closer look...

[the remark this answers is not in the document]
matloob: Ok so that error was due to the flag-flip setting the GOEXPERIMENT coverageredesign by default. We'll need to eventually update the way coverage works in rules_go, but for now we can set GOEXPERIMENT=nocoverageredesign. I'll try to find a way to set that to get this change moving forward. It might be easiest to plumb something through the same way we do in the boringcrypto cl

matloob:
I've updated the PR with changes to conditionally set GOEXPERIMENT=nocoverageredesign for Go 1.20+ to work around the coverage failures. In the future, on 1.20+ the coverage support should adapt to work with the coverageredesign. See https://github.com/bazelbuild/rules_go/issues/3397

go/private/actions/stdlib.bzl
```diff
@@ -53,19 +57,23 @@ def _should_use_sdk_stdlib(go):
...
-    if beta != -1:
-        minor = minor[:beta]
-    return int(minor) < 20
+    minor_digits = ""
+    for e in minor.elems():
+        if e.isdigit():
+            minor_digits += e
+        else:
+            break
+    return int(minor_digits)
```
fmeum, line 76: Can we reuse the new function in `common.bzl` here?

go/private/sdk.bzl
```diff
@@ -338,6 +338,13 @@ def _sdk_build_file(ctx, platform, version):
     ctx.file("ROOT")
     goos, _, goarch = platform.partition("_")
 
+    coverageredesign = False
+    pv = _parse_version(version)
+    if pv == None or len(pv) < 3:
+        fail("error parsing sdk version: " + version)
+    if pv[1] >= 20:
+        coverageredesign = True
```
fmeum, line 346: Where is this variable used?

fmeum:
@linzhp @achew22 What do you think, should we always build the stdlib with `-c opt` (relevant for the cgo parts)?

linzhp:
What's `-c opt`?

fmeum:
Sorry for the terse question: Now that we are always building the standard library from source, we should decide whether we want to build it with Bazel's current buildmode (`opt`, `fastbuild` or `dbg`) or always with optimizations. The Go part isn't affected by this, but the cgo deps would be.

I personally prefer not meddling with the user-configured build mode, but users that aren't passing `-c opt` to Bazel for their go release builds (because previously it didn't make a difference) may conceivably experience performance regressions in the cgo parts of the stdlib.

go/private/actions/stdlib.bzl
```diff
@@ -44,13 +48,33 @@ def _stdlib_library_to_source(go, attr, source, merge):
         source["stdlib"] = _build_stdlib(go)
 
 def _should_use_sdk_stdlib(go):
-    return (go.mode.goos == go.sdk.goos and
+    return (_has_precompiled_stdlib(go.sdk.version) and
```
linzhp, line 51: Can we instead check the existence of `go.sdk.libs`? So people can download a pre-compile Go SDK even after 1.20 to avoid building it on every machine

go/private/actions/stdlib.bzl
matloob: Done

go/private/sdk.bzl
matloob: removed. was dead code from another earlier approach.

go/private/actions/stdlib.bzl
matloob: Done!

go/private/actions/compilepkg.bzl
```diff
@@ -117,6 +121,10 @@ def emit_compilepkg(
         outputs.append(out_cgo_export_h)
     if testfilter:
         args.add("-testfilter", testfilter)
+    minor = minor_version(go.sdk.version)
+    if minor != None and minor >= 20:
+        # Turn off coverageredesign GOEXPERIMENT
+        args.add("-nocoverageredesign")
```
fmeum, line 127: Do you expect compatibility with older SDKs if we were to always set this flag?

fmeum (commented):
I will do one more `cquery` pass tomorrow to make sure that we are generating the minimal possible number of configured targets.

go/private/actions/compilepkg.bzl
matloob: Do you mean if we always set GOEXPERMENT=nocoverageredesign on older SDKs? That wouldn't be compatible because the Go SDK tools will produce errors if GOEXPERIMENT contains experiment names that are unknown to older versions of the SDK. So setting GOEXPERIMENT=nocoverageredesign on `go install` with Go 1.18, for example, will fail because the coverageredesign experiment was unknown to that version of Go.

go/private/rules/binary.bzl
fmeum: The recent fix brought it down to just target and exec variants. I don't know where the target variant comes from when building all of rules_go and `cquery` doesn't show me the path. It doesn't seem to be built with a normal `go_binary`, so I'm inclined to say it's not an issue.

I also checked the stdlib rebuilds and they all look reasonable.

go/private/sdk.bzl
```diff
@@ -350,6 +350,14 @@ def _sdk_build_file(ctx, platform, version, experiments):
     ctx.file("ROOT")
     goos, _, goarch = platform.partition("_")
 
+    pv = _parse_version(version)
+    if pv != None and pv[1] >= 20:
+        # Turn off coverageredesign GOEXPERIMENT on 1.20+
+        # until rules_go is updated to work with the
+        # coverage redesign.
+        if not "nocoverageredesign" in experiments:
```
fmeum, line 358:
```suggestion
        if not "nocoverageredesign" in experiments and not "coverageredesign" in experiments:
```

Assuming that this is how `GOEXPERIMENTS` work, it would still allow users to enable `coverageredesign` explicitly if they want to use it in some other way (e.g. produce an instrumented binary).

WORKSPACE
```diff
@@ -6,9 +6,7 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 
 go_rules_dependencies()
 
-go_register_toolchains(
-    version = "1.19.4",
-)
+go_register_toolchains(version = "1.19.4")
```
fmeum, line 9: This is back to 1.19.4. Given that 1.20 isn't out yet, maybe we should test it in `go_bazel_test` instead?

matloob: Added a simple test doing a build in 1.20.

fmeum (approved):

fmeum:
Looks good to me, thanks for going through all the iterations!

Since I lean towards not forcing `-c opt` given that we haven't done so in the past and it may still be different from what Go used for release builds, I would can 6 this done.

@linzhp Could you take a final look and/or test it?

linzhp:
OK. I will test it out and let you know

sywhang:
Hey folks, I work at Uber with @linzhp and I'm in charge of upgrading Go versions at Uber. 

I just tested this patch with our testing branch for Go 1.20 rc1, and we're still noticing issues with it. Specifically I'm seeing the stdlib archives are not being picked up by rules_go. 

```
load packages: error building pkg "<some package name>": some/package/file.go:4:2: could not import bytes (opening archive file: open external/go_sdk/pkg/linux_amd64/bytes.a: no such file or directory)

load packages: error building pkg "<another package name>": another/package/file.go:4:2: could not import context (opening archive file: open external/go_sdk/pkg/linux_amd64/context.a: no such file or directory)
```
I'm noticing that rules_go builds the stdlib archive files:
```
[sungyoon@sungyoon]❯ ls -l bazel-bin/external/io_bazel_rules_go/stdlib_/pkg/linux_amd64                                                                                                                go-code/git/arcpatch-D8841493
total 30592
dr-xr-xr-x 2 sungyoon user    4096 Dec 17 21:40 archive
-r-xr-xr-x 1 sungyoon user  368040 Dec 17 21:40 bufio.a
-r-xr-xr-x 1 sungyoon user  419322 Dec 17 21:40 bytes.a
dr-xr-xr-x 2 sungyoon user    4096 Dec 17 21:40 compress
dr-xr-xr-x 2 sungyoon user    4096 Dec 17 21:40 container
-r-xr-xr-x 1 sungyoon user  230702 Dec 17 21:40 context.a
dr-xr-xr-x 4 sungyoon user    4096 Dec 17 21:40 crypto
```
but it seems like rules_go is unable to pick these up and populate them as dependencies.

fmeum:
@sywhang Do you know why CI can't reproduce this when running with `1.20rc1`? I am also unable to reproduce this locally. Which version of Bazel are you using?

matloob:
@sywhang What do the lines in your WORKSPACE file that declares the toolchain/sdk look like?

matloob:
@sywhang Could you also provide details about what action you're running that's failing? Those error messages don't look familiar to me and I'm wondering if something needs to be updated in your repo apart from rules_go to be compatible with Go 1.20. (Just a theory of course)

sywhang:
I think this might have to do with the way we're extracting the binary since we're not using `go_download_sdk` rule. I also have found a workaround for this, so I'm ok with this PR merging as-is. If there are any issues we can just follow-up with a fix afterwards.

matloob:
Ok, thanks @sywhang! @fmeum, Is this CL okay to merge?

sluongng:
It seems like https://github.com/golang/go/issues/55953 is aiming to remove `nocoverageredesign` in 1.22.

Anybody have context on why did we introduce that experiment here and what would it take to adapt to the new design?

cc: @fmeum @matloob

fmeum:
We introduced it because Go changed the way it handles coverage by default. Supporting the new mode is tracked in https://github.com/bazelbuild/rules_go/issues/3513 and it looks like we might need to work on this sooner rather than later. It is not an easy task though.

PR #3390, opened 2022-12-08
Make C++ toolchain optional

illicitonion:
**What type of PR is this?**

Bug fix

Currently, if you transition a go_binary, you must have a C++ toolchain
for it even if it's building in pure mode without cgo.

By making this toolchain optional, we push the error from a
toolchain-resolution-time error to an analysis-time error only if a
toolchain is both _needed_ and not found.

This showed up in testing https://github.com/aspect-build/bazel-lib/pull/289

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -838,7 +838,13 @@ cgo_context_data = rule(
             default = "@bazel_tools//tools/osx:current_xcode_config",
         ),
     },
-    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
+    toolchains = [
+        # In pure mode, a C++ toolchain isn't needed when transitioning.
+        # But if we declare a mandatory toolchain dependency here, a cross-compiling C++ toolchain is required at toolchain resolution time.
+        # So we make this toolchain dependency optional, so that it's only attempted to be looked up if it's actually needed.
+        # Optional toolchain support was added in bazel 6.0.0.
+        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if hasattr(config_common, "toolchain_type") else "@bazel_tools//tools/cpp:toolchain_type",
+    ],
     fragments = ["apple", "cpp"],
     doc = """Collects information about the C/C++ toolchain. The C/C++ toolchain
     is needed to build cgo code, but is generally optional. Rules can't have
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@bazel_features//:features.bzl", "bazel_features")
 load(
     "@bazel_tools//tools/cpp:toolchain_utils.bzl",
     "find_cpp_toolchain",
@@ -643,8 +644,12 @@ def _cgo_context_data_impl(ctx):
     # toolchain (to be inputs into actions that need it).
     # ctx.files._cc_toolchain won't work when cc toolchain resolution
     # is switched on.
-    cc_toolchain = find_cpp_toolchain(ctx)
-    if cc_toolchain.compiler in _UNSUPPORTED_C_COMPILERS:
+    find_cpp_toolchain_kwargs = {}
+    if bazel_features.cc.find_cpp_toolchain_has_mandatory_param:
+        find_cpp_toolchain_kwargs["mandatory"] = False
+
+    cc_toolchain = find_cpp_toolchain(ctx, **find_cpp_toolchain_kwargs)
+    if not cc_toolchain or cc_toolchain.compiler in _UNSUPPORTED_C_COMPILERS:
         return []
 
     feature_configuration = cc_common.configure_features(
@@ -833,7 +838,7 @@ def _cgo_context_data_impl(ctx):
 cgo_context_data = rule(
     implementation = _cgo_context_data_impl,
     attrs = {
-        "_cc_toolchain": attr.label(default = "@bazel_tools//tools/cpp:current_cc_toolchain"),
+        "_cc_toolchain": attr.label(default = "@bazel_tools//tools/cpp:optional_current_cc_toolchain" if bazel_features.cc.find_cpp_toolchain_has_mandatory_param else "@bazel_tools//tools/cpp:current_cc_toolchain"),
         "_xcode_config": attr.label(
             default = "@bazel_tools//tools/osx:current_xcode_config",
         ),
@@ -843,7 +848,7 @@ cgo_context_data = rule(
         # But if we declare a mandatory toolchain dependency here, a cross-compiling C++ toolchain is required at toolchain resolution time.
         # So we make this toolchain dependency optional, so that it's only attempted to be looked up if it's actually needed.
         # Optional toolchain support was added in bazel 6.0.0.
-        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if hasattr(config_common, "toolchain_type") else "@bazel_tools//tools/cpp:toolchain_type",
+        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if bazel_features.cc.find_cpp_toolchain_has_mandatory_param else "@bazel_tools//tools/cpp:toolchain_type",
     ],
     fragments = ["apple", "cpp"],
     doc = """Collects information about the C/C++ toolchain. The C/C++ toolchain
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -644,11 +644,10 @@ def _cgo_context_data_impl(ctx):
     # toolchain (to be inputs into actions that need it).
     # ctx.files._cc_toolchain won't work when cc toolchain resolution
     # is switched on.
-    find_cpp_toolchain_kwargs = {}
     if bazel_features.cc.find_cpp_toolchain_has_mandatory_param:
-        find_cpp_toolchain_kwargs["mandatory"] = False
-
-    cc_toolchain = find_cpp_toolchain(ctx, **find_cpp_toolchain_kwargs)
+        cc_toolchain = find_cpp_toolchain(ctx, mandatory = False)
+    else:
+        cc_toolchain = find_cpp_toolchain(ctx)
     if not cc_toolchain or cc_toolchain.compiler in _UNSUPPORTED_C_COMPILERS:
         return []
 
@@ -848,7 +847,7 @@ cgo_context_data = rule(
         # But if we declare a mandatory toolchain dependency here, a cross-compiling C++ toolchain is required at toolchain resolution time.
         # So we make this toolchain dependency optional, so that it's only attempted to be looked up if it's actually needed.
         # Optional toolchain support was added in bazel 6.0.0.
-        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if bazel_features.cc.find_cpp_toolchain_has_mandatory_param else "@bazel_tools//tools/cpp:toolchain_type",
+        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if hasattr(config_common, "toolchain_type") else "@bazel_tools//tools/cpp:toolchain_type",
     ],
     fragments = ["apple", "cpp"],
     doc = """Collects information about the C/C++ toolchain. The C/C++ toolchain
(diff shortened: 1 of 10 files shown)


achew22:
Can you also include a test to confirm we don't regress on this? Thanks!

go/private/context.bzl
```diff
@@ -785,7 +785,12 @@ cgo_context_data = rule(
             default = "@bazel_tools//tools/osx:current_xcode_config",
         ),
     },
-    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
+    toolchains = [
+        # In pure mode, a C++ toolchain isn't needed when transitioning.
+        # But if we declare a mandatory toolchain dependency here, a cross-compiling C++ toolchain is required at toolchain resolution time.
+        # So we make this toolchain dependency optional, so that it's only attempted to be looked up if it's actually needed.
```
fmeum, line 791: Could you give an example of the error users would see if they need one but it isn't available? I want to make sure that the rest of the code handles the absence of the formerly required toolchain gracefully.

go/private/context.bzl
```diff
@@ -785,7 +785,12 @@ cgo_context_data = rule(
...
+        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if hasattr(config_common, "toolchain_type") else "@bazel_tools//tools/cpp:toolchain_type",
```
fmeum, line 792: Could you add a comment stating which Bazel release first introduced support for optional toolchains so that we can simplify this line at some point in the future?

fmeum:
@illicitonion Is it possible that you based your commit on a rather old version of `master`? Rebasing should fix the CI failure.

go/private/context.bzl
illicitonion: Interesting, I went through a large matrix of:
Bazel version: 5.3.2 vs 6.0.0rc1 (i.e. doesn't have vs does have support for optional toolchains)
`--incompatible_enable_cc_toolchain_resolution` on or off
Target's `pure` on or off
Actually having a cross-compiling C++ toolchain registered or not

I saw the exact same behaviour without and with this PR except for the following cells:

* (The intended change): Bazel 6.0.0rc1 with no cross-compiling C++ toolchain with `--incompatible_enable_cc_toolchain_resolution` off building a `pure="on"` target: Instead of a toolchain error (behaviour on master), we get a successful build
* (Surprising): Bazel 6.0.0rc1 with no cross-compiling C++ toolchain with `--incompatible_enable_cc_toolchain_resolution` off building a `pure="off"` target: Instead of a toolchain error (behaviour on master), we get a build-time error because a C++ toolchain _is_ found, but it's not a cross-compiling one, so an invalid command line is run.

I suspect this change probably isn't worth merging in that case - particularly as I'm a believer in the `--incompatible_enable_cc_toolchain_resolution` future, and this only changes behaviour when that flag is disabled.

fmeum: The surprising case is indeed... surprising. Could you add details about the toolchain that is found in that case? If rules_go is behaving incorrectly in this case, we should fix it. If Bazel does, that would make optional toolchains quite a bit less useful.

illicitonion: Absolutely - I just pushed a new commit

If on a mac (I'm on an x86_64 machine), you run `USE_BAZEL_VERSION=6.0.0rc1 bazel build //tests/core/cross:linux_go_cross_impure_cgo` you should see Bazel hand rules_go a darwin->darwin C++ toolchain, which then fails to build the stdlib for linux like this:

```
ERROR: /Volumes/Source/github.com/bazelbuild/rules_go/BUILD.bazel:42:7: GoStdlib stdlib_/pkg failed: (Exit 1): builder failed: error executing command (from target //:stdlib) bazel-out/darwin-opt-exec-2B5CBBC6-ST-759b62b2e9d9/bin/external/go_sdk/builder stdlib -sdk external/go_sdk -installsuffix linux_amd64 -out bazel-out/darwin-fastbuild-ST-79e94b232e60/bin/stdlib_

# runtime/cgo
linux_syscall.c:67:13: error: implicit declaration of function 'setresgid' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
linux_syscall.c:67:13: note: did you mean 'setregid'?
[redacted]/usr/include/unistd.h:593:6: note: 'setregid' declared here
linux_syscall.c:73:13: error: implicit declaration of function 'setresuid' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
linux_syscall.c:73:13: note: did you mean 'setreuid'?
[redacted]/usr/include/unistd.h:595:6: note: 'setreuid' declared here
Error in child process '/usr/bin/xcrun'. 1
stdlib: error running subcommand external/go_sdk/bin/go: exit status 2
Target //tests/core/cross:linux_go_cross_impure_cgo failed to build
```

WORKSPACE
```diff
@@ -167,3 +167,12 @@ http_archive(
 load("@io_bazel_stardoc//:setup.bzl", "stardoc_repositories")
 
 stardoc_repositories()
+
+local_repository(
+    name = "apple_federation",
+    path = "/Volumes/Source/github.pie.apple.com/bazel/apple_federation",
```
fmeum, line 173: @illicitonion This got committed by accident?

go/private/context.bzl
fmeum: Okay, that actually isn't that surprising: Without toolchain resolution, the transition in `go_cross_binary` is a no-op and it makes sense that the auto-configured host toolchain is used.

Do you know why you didn't see a difference with `--incompatible_enable_cc_toolchain_resolution` *enabled*? I agree that is the interesting case.

fmeum: Tried this myself, I think I have an idea of what is wrong:

1. https://github.com/bazelbuild/rules_go/blob/1a8fe64877c6e71dbbf51cdc8ceb2eb10c13e521/go/private/context.bzl#L779 uses the legacy way to get the C++ toolchain, which always fails if none is available even if it isn't marked as mandatory. 
2. Removing this line changes the error to one in the Starlark `find_cpp_toolchain` helper, which also always fails if the toolchain isn't available. 

I will try what happens if I use a helper that is aware of optional toolchains.

fmeum: @katre Do you see a way for rules to both work without C++ toolchain resolution (using `current_cc_toolchain`) *and* make use of optional toolchains? It seems as if depending on `@bazel_tools//tools/cpp:current_cc_toolchain` is necessary for the former, but breaks the latter.

fmeum: @illicitonion Using `cc_toolchain = ctx.toolchains["@bazel_tools//tools/cpp:toolchain_type"]` directly in `_cgo_context_data_impl` correctly returns `None` in the cross-compilation case with `--incompatible_enable_cc_toolchain_resolution`. Not sure though we can get this behavior without losing support for non-toolchain resolution builds, given the above.

katre: @fmeum You may need to import/copy the latest version of [`find_cc_toolchain.bzl`](https://github.com/bazelbuild/rules_cc/blob/main/cc/find_cc_toolchain.bzl), which correctly handles all these options.

fmeum: I did that, but with the commands provided by @illicitonion and `--incompatible_enable_cc_toolchain_resolution`, I get a Java error in the native `cc_toolchain_alias` rule even before that Starlark function is executed. It only goes away if I remove the `_cc_toolchain` attribute with default `@bazel_tools//tools/cpp:current_cc_toolchain`, but that would break backwards compatibility.

fmeum:
@illicitonion This should be unblocked as soon as Bazel 6.1 is the minimum supported version of rules_go: https://github.com/bazelbuild/bazel/commit/ac47e1085c59654f29b82b0ca3be224447895774

illicitonion:
Amazing, thanks!

LeandroLovisolo:
Bazel 6.1 has been released. Any chance this PR could be merged?

fmeum:
We still support Bazel 5.4.0 though.

@illicitonion Would you be interested in picking this up again and using `bazel_features` for feature detection?

illicitonion:
Sure, happy to pick this up again!

It looks like `bazel_features` requires a two-step init (`http_archive`ing it in, then calling a function from it), but rules_go only has a one-step init (loading `go_rules_dependencies` and calling it) - is there an existing pattern in rules_go for two-step init? (The problem being: You can't `load` the function you need to call at the top of the file because the `http_archive` you're loading it from doesn't exist yet, and you can't `load` it after the `http_archive` has been called because you can't `load` inside a function)

fmeum:
@illicitonion There isn't and adding a second step would be a breaking change. How about adding this feature for Bzlmod only? Then we can look into whether we can support WORKSPACE as a follow-up.

illicitonion:
@fmeum I think we're good to go on this, PTAL!

(Sorry for the horrors of the fake `bazel_features` :))

go/private/context.bzl
```diff
@@ -643,8 +644,12 @@ def _cgo_context_data_impl(ctx):
     # toolchain (to be inputs into actions that need it).
     # ctx.files._cc_toolchain won't work when cc toolchain resolution
     # is switched on.
-    cc_toolchain = find_cpp_toolchain(ctx)
-    if cc_toolchain.compiler in _UNSUPPORTED_C_COMPILERS:
+    find_cpp_toolchain_kwargs = {}
+    if bazel_features.cc.find_cpp_toolchain_has_mandatory_param:
+        find_cpp_toolchain_kwargs["mandatory"] = False
```
fmeum, line 649: I would like this better if we just duplicated the call instead of using `kwargs`.

go/private/context.bzl
```diff
@@ -833,12 +838,18 @@ def _cgo_context_data_impl(ctx):
...
             default = "@bazel_tools//tools/osx:current_xcode_config",
         ),
     },
-    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
+    toolchains = [
+        # In pure mode, a C++ toolchain isn't needed when transitioning.
+        # But if we declare a mandatory toolchain dependency here, a cross-compiling C++ toolchain is required at toolchain resolution time.
+        # So we make this toolchain dependency optional, so that it's only attempted to be looked up if it's actually needed.
+        # Optional toolchain support was added in bazel 6.0.0.
+        config_common.toolchain_type("@bazel_tools//tools/cpp:toolchain_type", mandatory = False) if bazel_features.cc.find_cpp_toolchain_has_mandatory_param else "@bazel_tools//tools/cpp:toolchain_type",
```
fmeum, line 851: This was introduced earlier than that `mandatory` param and can be checked for directly with `hasattr` on `config_common`.

go/private/fake_bazel_features.bzl
```diff
@@ -0,0 +1,35 @@
...
+# We just implement the checks we've seen we actually need, and hope to delete this completely when we are in a pure-bzlmod world.
+
+_FAKE_BAZEL_FEATURES = """bazel_features = struct(
+  cc = struct(
+    find_cpp_toolchain_has_mandatory_param = {find_cpp_toolchain_has_mandatory_param},
+  )
+)
+"""
+
+def _fake_bazel_features_impl(rctx):
```
fmeum, line 14: This looks pretty real to me. Maybe call it a `polyfill` instead? Thanks for going the extra mile!

fmeum (approved):
Looks great, only a few stylistic comments

tests/core/cross/BUILD.bazel
```diff
@@ -24,6 +24,21 @@ go_binary(
     deps = [":platform_lib"],
 )
 
+go_binary(
+    name = "linux_cross_impure",
+    srcs = ["main.go"],
+    goarch = "amd64",
+    goos = "linux",
+    pure = "off",
```
fmeum, line 32: Just having second thoughts about semantics: If someone sets `pure = "off"`, is it okay for us to silently turn it on when no C toolchain is available? I'm wondering whether we should restrict this to unset `pure`. 

@illicitonion What do you think?

illicitonion: It's a slightly weird corner - I _think_ any place that this would have errored before, but wouldn't error now, it's because `pure` doesn't actually need to be `off`, but I agree that ignoring an explicit instruction is probably bad.

I added this check, and a test showing we correctly fail.

tests/core/starlark/cgo/cgo_test.bzl
```diff
@@ -0,0 +1,43 @@
...
+
+    asserts.expect_failure(env, "has pure explicitly set to off, but no C++ toolchain could be found for its platform")
+
+    return analysistest.end(env)
+
+missing_cc_toolchain_explicit_pure_off_test = analysistest.make(
+    _missing_cc_toolchain_explicit_pure_off_test,
+    expect_failure = True,
+    config_settings = {
+        "//command_line_option:extra_toolchains": str(Label("//tests/core/starlark/cgo:fake_go_toolchain")),
```
fmeum, line 15: I should have made the connection ;⁠-⁠)

fmeum (approved):

fmeum:
@linzhp Could you test this?

tyler-french (approved):
Tested these changes at Uber and everything passes

PR #3393, opened 2022-12-13
add support for building with boringcrypto

matloob:
Adds an option to go_register_toolchains and the various go_*_sdk functions to specify boringcrypto = True. Doing so builds the standard library and all library/binary packages using the GOEXPERIMENT=boringcrypto environment variable setting, which switches out the crypto implementation for boringcrypto on Go 1.19+.

**What type of PR is this?**

> Feature

It adds support for building with boringcrypto, which was added in Go 1.19.

Fixes #

**Other notes for review**


matloob:
Abandoned in favor of https://github.com/bazelbuild/rules_go/pull/3398.

PR #3399, opened 2022-12-13
explicitly build relative path for Tool command in lcov_coverage_test

matloob:
filepath.Join(".", "Tool") is just "Tool" because the path is cleaned.

See https://go.dev/play/p/R1KWJ9612fN.

Build the relative path explicitly using "." + string(filepath.Separator) + "Tool".


fmeum (approved):
Shouldn't have tried to be clever here. Thanks!

PR #3400, opened 2022-12-13
Refactoring the extld related logic

linzhp:
**What type of PR is this?**
Refactoring

The original issue that this PR was trying to fix was fixed on zig side. This PR is reduced to a refactoring to:
* avoid adding both `-linkmode internal` and `-linkmode external` in some corner cases
* avoid using `-linkmode external` when external linker is not available
* add C toolchain to go_context as long as it's available, deferring to the link action to decide whether it should be passed down to the linker. This gives us more flexibility later on, for example, to use external linker even in pure mode.

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -95,10 +95,11 @@ def emit_link(
         tool_args.add("-race")
     if go.mode.msan:
         tool_args.add("-msan")
-    if ((go.mode.static and not go.mode.pure) or
-        (go.mode.race and extld) or
+
+    if extld and (go.mode.static or
+        go.mode.race or
         go.mode.link != LINKMODE_NORMAL or
-        go.mode.goos == "windows" and (go.mode.race or go.mode.msan)):
+        go.mode.goos == "windows" and go.mode.msan):
         # Force external linking for the following conditions:
         # * Mode is static but not pure: -static must be passed to the C
         #   linker if the binary contains cgo code. See #2168, #2216.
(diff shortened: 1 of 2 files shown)


motiejus:
> There are situations when the internal linker doesn't work (@motiejus do you have an issue number or example?)

I am still stuck on creating a FOSS test case (I will follow up with you). The error we are getting now is when cross-compiling some arm64 binaries:

```
runtime/cgo(.text): relocation target memset not defined
```

... unless we use an external linker.

go/private/actions/link.bzl
```diff
@@ -89,35 +89,36 @@ def emit_link(
...
+        if extld and (go.mode.static or
+                      go.mode.race or
+                      go.mode.link != LINKMODE_NORMAL or
+                      go.mode.goos == "windows" and go.mode.msan):
+            # Force external linking for the following conditions:
+            # * Mode is static but not pure: -static must be passed to the C
+            #   linker if the binary contains cgo code. See #2168, #2216.
+            # * Non-normal build mode: may not be strictly necessary, especially
+            #   for modes like "pie".
+            # * Race or msan build for Windows: Go linker has pairwise
```
fmeum, line 111: This is outdated now that `go.mode.race` is no longer guarded behind `go.mode.goos == "windows"`.

fmeum (approved):

go/private/actions/link.bzl
linzhp: The reason for using external linker in race mode is different on Windows and Linux. It's more clear to keep them separate in comments.

Issue #3405, opened 2022-12-16
support go:embed of generated files failing


bakjos:
### What version of rules_go are you using?
0.37.0

### What version of gazelle are you using?
0.28.0

### What version of Bazel are you using?
5.4.0

### Does this issue reproduce with the latest releases of all the above?
Yes

### What operating system and processor architecture are you using?
OSX x64 

### What did you do?

When I tried to run the compilation process with the latest release, the embedsrc was failing when using go_path, due a relativize issue

### What did you expect to see?
Compiling correctly

### What did you see instead?
```
 File "/external/io_bazel_rules_go/go/private/tools/path.bzl", line 89, column 51, in _go_path_impl
                dst = pkg.dir + "/" + paths.relativize(embedpath, src_dir)
        File "external/bazel_skylib/lib/paths.bzl", line 182, column 13, in _relativize
                fail("Path '%s' is not beneath '%s'" % (path, start))
Error in fail: Path 'services/test/graphql/schema.graphql' is not beneath 'bazel-out/darwin-fastbuild/bin/services/test/graphql'
```

mrkkrp:
Should this be closed? Seemingly resolved by #3407.

PR #3407, opened 2022-12-16
fix: Relative path of embedsrc

bakjos:
**What type of PR is this?**
 Bug fix

The new relativize introduced by https://github.com/bazelbuild/rules_go/pull/3285 fails if also includes generated code because the `library_to_source` appends the src with the `generated_srcs`, so instead of using the last src_dir, it uses the first

Fixes # [3405](https://github.com/bazelbuild/rules_go/issues/3405)

**Other notes for review**

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -164,7 +164,7 @@ def emit_archive(go, source = None, _recompile_suffix = ""):
         _cxxopts = as_tuple(source.cxxopts),
         _clinkopts = as_tuple(source.clinkopts),
         _cgo_exports = as_tuple(source.cgo_exports),
-        _root_dir = source.root_dir,
+
         # Information on dependencies
         _dep_labels = tuple([d.data.label for d in direct]),
         _dep_importmaps = tuple([d.data.importmap for d in direct]),
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -164,7 +164,7 @@ def emit_archive(go, source = None, _recompile_suffix = ""):
         _cxxopts = as_tuple(source.cxxopts),
         _clinkopts = as_tuple(source.clinkopts),
         _cgo_exports = as_tuple(source.cgo_exports),
-
+        _root_dir = source.root_dir,
         # Information on dependencies
         _dep_labels = tuple([d.data.label for d in direct]),
         _dep_importmaps = tuple([d.data.importmap for d in direct]),
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -77,15 +77,14 @@ def _go_path_impl(ctx):
         src_dir = None
 
         for f in pkg.srcs:
-            src_dir = f.dirname
+            if src_dir == None:
+                src_dir = f.dirname
             dst = pkg.dir + "/" + f.basename
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
         for f in pkg.embedsrcs:
             if src_dir == None:
                 fail("cannot relativize {}: src_dir is unset".format(f.path))
             embedpath = paths.relativize(f.path, f.root.path)
-            if ctx.bin_dir.path == f.root.path:
-                embedpath = f.path
             dst = pkg.dir + "/" + paths.relativize(embedpath, src_dir)
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
     if ctx.attr.include_pkg:
--- a/go/private/tools/path.bzl
+++ b/go/private/tools/path.bzl
@@ -54,6 +54,11 @@ def _go_path_impl(ctx):
             importpath, pkgpath = effective_importpath_pkgpath(archive)
             if importpath == "":
                 continue  # synthetic archive or inferred location
+
+            # Uses an array so it can be updated during the merge package
+            root_dir = []
+            if archive._root_dir:
+                root_dir = [archive._root_dir]
             pkg = struct(
                 importpath = importpath,
                 dir = "src/" + pkgpath,
@@ -61,6 +66,7 @@ def _go_path_impl(ctx):
                 data = as_list(archive.data_files),
                 embedsrcs = as_list(archive._embedsrcs),
                 pkgs = {mode: archive.file},
+                root_dir = root_dir,
             )
             if pkgpath in pkg_map:
                 _merge_pkg(pkg_map[pkgpath], pkg)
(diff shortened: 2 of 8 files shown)


fmeum:
Thanks for looking into this in detail! 

Could you add a test case similar to the one that the PR added that ended up breaking your use case? Path meddling gets tricky and we don't want to break this again.

bakjos:
@fmeum Thanks for the quick response, I added one test and it fails if you discard the change on the `path.bzl` file. There's another scenario of failure when the library only uses generated code and has`embedsrcs`, but not sure how to fix it

go/private/tools/path.bzl
```diff
@@ -77,7 +77,8 @@ def _go_path_impl(ctx):
         src_dir = None
 
         for f in pkg.srcs:
-            src_dir = f.dirname
+            if src_dir == None:
+                src_dir = f.dirname
```
fmeum, line 81: I am not very familiar with `go_path`. Could you explain why using the first rather than the last source directory is the correct choice?

fmeum (commented):
Thanks for adding tests!

Could you expand on the case that still isn't handled correctly?

go/private/tools/path.bzl
bakjos: Sure, as you can see here https://github.com/bazelbuild/rules_go/blob/master/go/private/context.bzl#L237, the `attr.srcs` files are added before the `generated_srcs` which means the relative paths are always first, the unsolved case is when all the files are coming from the `generated_srcs` and `attr_srcs` is empty because none of the files is relative

fmeum: That makes sense, thanks for the pointer.

Could it be that the root of the problem is that we are separately storing all the potential roots (dirnames of source files) and potentially embedded files (embedsrcs) but not the association between the two (which dirnames of sources were present on the target that defined a particular embedsrc). It seems to me that the correct approach could be to store this association in the provider and then make all `embedsrcs` of a particular target relative to all dirnames of source files of that target.

@bakjos What do you think?

bakjos: @fmeum it does makes sense, let me take another pass, because, before https://github.com/bazelbuild/rules_go/pull/3285 was working well if not will try to include the map in the provider, thanks for the guidance

bakjos: @fmeum Took some days off but added a new test case for generated sources only and a root dir per GoSource and seems to work

linzhp (commented):
Can you provide a minimal reproduction of the bug in #3405 to help me understand the situation? The test case in this PR seem to indicate that you are trying to embed a generated text file into a generated Go file?

bakjos:
> Can you provide a minimal reproduction of the bug in #3405 to help me understand the situation? The test case in this PR seem to indicate that you are trying to embed a generated text file into a generated Go file?

@linzhp Thanks for the comments. Yes, the issue is when including an embedded file from a generated source, the reproduction is basically what I did on the test, so if you take the tests I added and try to execute  them in the current version of the library it will fail

go/private/actions/archive.bzl
```diff
@@ -164,7 +164,7 @@ def emit_archive(go, source = None, _recompile_suffix = ""):
         _cxxopts = as_tuple(source.cxxopts),
         _clinkopts = as_tuple(source.clinkopts),
         _cgo_exports = as_tuple(source.cgo_exports),
-
+        _root_dir = source.root_dir,
```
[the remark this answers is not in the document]
bakjos: Yes, is used here https://github.com/bazelbuild/rules_go/blob/e85024cc0caaae026a4273df06d35107ae62557f/go/private/tools/path.bzl#L60-L61

go/private/tools/path.bzl
```diff
@@ -75,15 +81,18 @@ def _go_path_impl(ctx):
...
-            src_dir = f.dirname
             dst = pkg.dir + "/" + f.basename
             _add_manifest_entry(manifest_entries, manifest_entry_map, inputs, f, dst)
         for f in pkg.embedsrcs:
             if src_dir == None:
                 fail("cannot relativize {}: src_dir is unset".format(f.path))
             embedpath = paths.relativize(f.path, f.root.path)
+            if embedpath.startswith(ctx.bin_dir.path + "/"):
+                embedpath = embedpath[len(ctx.bin_dir.path) + 1:]
             dst = pkg.dir + "/" + paths.relativize(embedpath, src_dir)
```
linzhp, line 96:
```suggestion
            dst = pkg.dir + "/" + paths.relativize(embedpath.lstrip(ctx.bin_dir.path + "/"), src_dir.lstrip(ctx.bin_dir.path + "/"))
```
I make this change locally without all other changes you made, and run the tests in this PR, they all passed. Maybe this is all we need?

bakjos: @linzhp Thank you so much for looking into this,  you're right, and that's a better fix. I already applied the changes

tests/core/go_path/pkg/lib/generated_embeded.bzl
```diff
@@ -0,0 +1,107 @@
...
+    ]
+
+_gen_library = rule(
+    _gen_library_impl,
+    attrs = {
+        "importpath": attr.string(mandatory = True),
+        "_go_context_data": attr.label(
+            default = "//:go_context_data",
+        ),
+        "srcs": attr.label_list(
```
linzhp, line 55: Is this attr used in the impl?

tests/core/go_path/pkg/lib/generated_embeded.bzl
```diff
@@ -0,0 +1,107 @@
...
+    embedsrcs = getattr(ctx.attr, "embedsrcs", [])
+
+    embed_declarations = ""
+    i = 0
+    for e in embedsrcs:
+        for f in e.files.to_list():
+            embed_declarations += "//go:embed {0}\nvar embeddedSource{1} string\n".format(f.basename, i)
+            i += 1
+
+    go.actions.write(src, "package " + libname + "\n" +
```
linzhp, line 35: Can you use [expand_template](https://bazel.build/rules/lib/actions#expand_template) instead? It's much harder to read with all the escapes

linzhp (commented):
I am trying to see if we can simplify tests

tests/core/go_path/pkg/lib/generated_embeded.bzl
bakjos: You'll still have the same issues with `expand_template` I mage some changes to improve the readability

tests/core/go_path/pkg/lib/generated_embeded.bzl
```diff
@@ -0,0 +1,115 @@
...
+        ),
+        "embedsrcs": attr.label_list(
+            allow_files = True,
+        ),
+        "libname": attr.string(default = "lib"),
+    },
+    toolchains = ["@io_bazel_rules_go//go:toolchain"],
+)
+
+def _gen_main_src_impl(ctx):
```
linzhp, line 74: It appears that the generation of the main.go is not required to cover the scenario. Can you rewrite the test to create go_path from the `_gen_library` instead?

bakjos: I’m not following, the issue happens when using the `go_path` including a generated library, so it needs to be called outside the _gen_library, and yes I think the main.go could be removed

linzhp: Sorry, I realized what I said is confusing. Yes, go ahead and remove main.go and have `go_path` depend on `_gen_library`. Then you probably don't need the macro and instead rename `_gen_library` to `generated_embeded`.

bakjos: @linzhp  done

linzhp (approved):
Thanks for the contribution and patience.

PR #3414, opened 2023-01-05
Do not use exec paths for rundir

fmeum:
Bazel tests are executed in the runfiles directory, so package relative paths can be looked up directly by prepending the test's repository name - it is not necessary (and not correct with `--nolegacy_external_runfiles`) to use `workspace_root`.


fmeum:
@linzhp Could you test this at Uber?

linzhp:
Looks like some tests are already failing on github

asuffield:
Looks like setting `--nolegacy_external_runfiles` in the test shook out a couple of other things which assume the legacy layout

fmeum:
Sorry, I forgot to remove the TMP commit. I will submit a separate PR to fix the `--nolegacy_external_runfiles` use case.

fmeum:
Tests are passing now.

linzhp:
OK. I will test it in Uber in the coming days

go/private/rules/test.bzl
```diff
@@ -86,13 +85,15 @@ def _go_test_impl(ctx):
...
-    if ctx.attr.rundir:
-        if ctx.attr.rundir.startswith("/"):
-            run_dir = ctx.attr.rundir
-        else:
-            run_dir = pkg_dir(ctx.label.workspace_root, ctx.attr.rundir)
+    repo_relative_rundir = ctx.attr.rundir
+    if not repo_relative_rundir:
+        repo_relative_rundir = ctx.label.package if ctx.label.package else "."
+    if ctx.label.workspace_name:
+        # The test is contained in an external repository. The test runner cd's into the directory
```
linzhp, line 92: If the main workspace has a name, this would walk up and down to the same directory, which is fine, but the comment is misleading.

go/private/rules/test.bzl
```diff
@@ -86,13 +85,15 @@ def _go_test_impl(ctx):
...
 
     # now generate the main function
-    if ctx.attr.rundir:
-        if ctx.attr.rundir.startswith("/"):
-            run_dir = ctx.attr.rundir
-        else:
-            run_dir = pkg_dir(ctx.label.workspace_root, ctx.attr.rundir)
+    repo_relative_rundir = ctx.attr.rundir
+    if not repo_relative_rundir:
+        repo_relative_rundir = ctx.label.package if ctx.label.package else "."
```
linzhp, line 90:
```suggestion
    repo_relative_rundir = ctx.attr.rundir or ctx.label.package or "."
```
slightly simple...

go/private/rules/test.bzl
```diff
@@ -262,14 +263,21 @@ _go_test_kwargs = {
         ),
         "rundir": attr.string(
             doc = """ A directory to cd to before the test is run.
-            This should be a path relative to the execution dir of the test.
+            This should be a path relative to the subdirectory of the runfiles
+            directory corresponding to the test's repository.
```
linzhp, line 267: Would it be more clear to say "This should be a path relative to the root directory of the workspace where the test is defined, which can be the main or an external workspace"?

linzhp (commented):
Most tests in Uber passed, except one that uses `pkg_dir`, which I think we should change on our end.

go/private/rules/test.bzl
fmeum: Very clean!

go/private/rules/test.bzl
fmeum: Good point, I reworded this to avoid mentioning runfiles explicitly (except in the Windows case where knowledge about this is necessary anyway)

fmeum: I am not using the term `workspace` anymore because it means `main repository` to some people and `any repository` to others and is (luckily) going away with Bzlmod.

go/private/rules/test.bzl
fmeum: Even if the main workspace has a name, the value in this line should always be empty. Added a comment explaining that.

linzhp (approved):
LGTM, but some sanity check is failing

fmeum:
Forgot to update the docs, should be fixed now.

PR #3418, opened 2023-01-05
chore: remove experimental warning from bzlmod module

alexeagle:
In practice, now that Bazel 6.0.0 has shipped, rules_go is working about as well as any other supported languages in the ecosystem. Also, if users find bugs with it, those are important and we'll fix them.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,8 +5,6 @@ module(
     compatibility_level = 0,
 )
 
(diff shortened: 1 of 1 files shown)


fmeum:
@linzhp @achew22 I agree with Alex that the `rules_go` and `gazelle` modules have picked up sufficient usage that their (purposefully minimal) API can't easily be changed anymore. I also haven't found any issues with the module API yet. What do you think, should we drop the warnings?

achew22:
Sounds reasonable to me. Thanks for killing the log spam, @alexeagle!

PR #3420, opened 2023-01-06
WIP: Fix tests for `--nolegacy_external_runfiles`

fmeum:


fmeum:
There aren't many failing tests, but most of them seem to fail for slightly different underlying reasons.

@asuffield It seems that you want to see the flag flipped, would you be interested in taking a look at some of the failures as well?

fmeum:
__

PR #3428, opened 2023-01-23
Add `env` attribute to `go_binary`

fmeum:
Using the new `RunEnvironmentInfo` provider in Bazel 5.3.0, `go_binary` targets can now specify environment variables to be set when they are executed with `bazel run`.

Since `RunEnvironmentInfo` generalizes the `testing.TestEnvironment` provider, the existing usage is updated.

Requires raising the minimum required Bazel version to 5.3.0.


fmeum:
Includes the changes of #3427, which are required to get stardoc to recognize `RunEnvironmentInfo`.

linzhp (approved):

PR #3429, opened 2023-01-23
Make the toolchain's `go` binary available as a target

fmeum:
This allows developers to use the `go` command provided by the registered toolchain to e.g. add dependencies to `go.mod`. This ensures that everyone uses the same version of Go and does not require a local installation of Go.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -2,7 +2,7 @@
 tasks:
   ubuntu1804_bazel400:
     platform: ubuntu1804
-    bazel: 5.3.0 # test minimum supported version of bazel
+    bazel: 5.4.0 # test minimum supported version of bazel
     shell_commands:
       - tests/core/cgo/generate_imported_dylib.sh
     build_targets:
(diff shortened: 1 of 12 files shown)


fmeum:
Stacked on #3428, but otherwise ready for review.

fmeum:
@alexeagle This provides the missing hermetic link to the `go` command.

alexeagle:
How is it different from the binary under @go_sdk? This is the one Bazel resolved for the toolchain so the only difference is if the user installed more than one?

fmeum:
There are two differences:
1. The user may have registered their own toolchain and referencing `@go_sdk` would needlessly trigger a fetch of yet another Go SDK.
2. The default `go_sdk` may have been configured to run a remote executor, but may not run on the user's machine (MacBook with Linux RBE).

That's why I turned to toolchain resolution.

achew22:
Did a very light review on this. Seems reasonable at first pass but I'll wait till it's truly ready to review before I do too much more review

achew22 (approved):

fmeum:
Slightly improved the test, but otherwise rebased without changes.

Issue #1986, opened 2019-03-09
Migrate from go_googleapis to com_google_googleapis


vam-google:
`rules_go` depends on [googleapis](https://github.com/googleapis/googleapis) repository by applying gazelle-generated patch. `googleapis` is now Bazel-aware and `rules_go` should depend on googleapis directly without overwriting its BUILD.bazel files. Also the repository name used within rules_go should be migrated from `go_googleapis` to `com_google_googleapis`. 

This is expected to be done in few phases (with deprecation period, when both `go_googleapis` and `com_google_googleaspis` are available).

Without this change there is a circular dependency between `googleapis` and  `rules_go` (because `googleapis` uses rules_go to build go clients).

Currently, the build of googleapis produces the following warnings because of this issue:
```
GoLink: warning: package "google.golang.org/genproto/googleapis/api/annotations" is provided by more than one rule:
    @go_googleapis//google/api:annotations_go_proto
    //google/api:annotations_go_proto
Set "importmap" to different paths in each library.
This will be an error in the future.
```
Using `importmap` should work as a short term solution, but proper migration to `com_google_googleapis` would eliminate these warnings automaticaly.

ashi009:
Is anyone working on this? Looking forward to having a language-agnostic repo to use.

ashi009:
🤷‍♂️ It turns out that go_googleapis disagrees with the proto_library defined in com_google_googleapis.

https://github.com/bazelbuild/rules_go/blob/ef7cca8857f2f3a905b86c264737d0043c6a766e/third_party/go_googleapis-deletebuild.patch#L11-L18

https://github.com/bazelbuild/rules_go/blob/ef7cca8857f2f3a905b86c264737d0043c6a766e/third_party/go_googleapis-gazelle.patch#L512-L520

vam-google:
@ashi009 Yes, those should converge (`go_googleapis` and `com_google_googleapis`), this is what this issue is about. Unfortunately we did not have enough resources to do the proper migration yet, but it is still the plan.

gonzojive:
ping? not sure what's required to implement this... maybe somebody can jot down thoughts that might help the ambitious contributor.

jayconrod:
I don't believe anyone is working on this at the moment. This should eventually done by either me or someone who maintains `com_google_googleapis`. I'd rather not take PRs from anyone else.

I had a conversation with @noahdietz (one of the googleapis maintainers) a couple weeks ago. To summarize what needs to be done:

[...]

PR #3432, opened 2023-01-26
Revert go_googleapis upgrade

linzhp:
**What type of PR is this?**
Other

The go_googleapis upgrade in 0.38.0 is too disruptive. It changes the Go import paths of the proto packages like `@go_googleapis//google/cloud/dialogflow/cx/v3` and `@go_googleapis//google/cloud/dialogflow/v2`, making it necessary to change a large number known imports in Gazelle (https://github.com/bazelbuild/bazel-gazelle/pull/1426), which means that people may not be able to upgrade to rules_go 0.38.0 until we release another version of Gazelle.

In addition, it requires other 3rd party packages to use the new import paths too. For example, it requires people to upgrade cloud.google.com/go/dialogflow/apiv2 to at least v1.19.0, which is less than 3 months old, to use the new import path.

We should explore a better upgrade plan, or at least wait a few more months to upgrade go_googleapis. Reverting the upgrade for now.

**Other notes for review**
I am planning to create a patch release 0.38.1 to include this revert.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -247,13 +247,13 @@ def go_rules_dependencies(force = False):
     wrapper(
         http_archive,
         name = "go_googleapis",
-        # master, as of 2023-01-22
+        # master, as of 2022-12-05
         urls = [
-            "https://mirror.bazel.build/github.com/googleapis/googleapis/archive/9fac84a6942bd3e95fca1ffe87018adbf560c514.zip",
-            "https://github.com/googleapis/googleapis/archive/9fac84a6942bd3e95fca1ffe87018adbf560c514.zip",
+            "https://mirror.bazel.build/github.com/googleapis/googleapis/archive/83c3605afb5a39952bf0a0809875d41cf2a558ca.zip",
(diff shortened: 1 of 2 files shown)


fmeum (approved):

sluongng:
@linzhp @fmeum unfortunately keeping the old version of go_googleapis is blocking folks from using some of the latest dependencies in the ecosystem. This affects the usability of rules_go a lot as some popular libraries such as OpenTelemetry require a new version of these.

Have you guys discussed an upgrade plan? Or do you want to tackle #1986 instead?

linzhp:
We are planning to see if we can remove the [hard-coded import mapping](https://github.com/bazelbuild/bazel-gazelle/blob/master/language/proto/known_go_imports.go) for googleapis and use the pre-generated package managed by Go modules instead. So rules_go and Gazelle no longer need to worry about googleapis.

I am trying to do that for a while, but haven't get there yet. If you have bandwidth, feel free to give it a try.

sluongng:
I think that's the sane approach.

I can't promise anything but I will be taking a closer look at these code paths to see if I could make a nudge.

Issue #2522, opened 2020-05-29
proto: add go_proto_apiv2 and go_grpc_apiv2 compilers


johanbrandhorst:
### What version of rules_go are you using?

v0.23.2

### What version of gazelle are you using?

v0.21.1

### What version of Bazel are you using?

v3.2.0

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

Linux amd64

### What did you do?

Wanted to generate protofiles with the new Go protobuf generator stack, `protoc-gen-go` v1.24.0 and `protoc-gen-go-grpc`.

### What did you expect to see?

I expected to see some `go_proto_compiler` definitions for these compiler options

### What did you see instead?

No such compiler options

### Discussion

I put together some example implementations of these two generators for the grpc-gateway to use, and I think we should consider upstreaming them (after v1 of `protoc-gen-go-grpc` is complete).

Here they are:

`protoc-gen-go` v1.24.0:

```starlark
go_proto_compiler(
    name = "go_proto_apiv2", # up for discussion, of course
    options = [
        "paths=source_relative",
    ],
    plugin = "@org_golang_google_protobuf//cmd/protoc-gen-go",
    suffix = ".pb.go",
    visibility = ["//visibility:public"],
    deps = [
        "@com_github_golang_protobuf//proto:go_default_library",
        "@io_bazel_rules_go//proto/wkt:any_go_proto",
        "@io_bazel_rules_go//proto/wkt:api_go_proto",
        "@io_bazel_rules_go//proto/wkt:compiler_plugin_go_proto",
        "@io_bazel_rules_go//proto/wkt:descriptor_go_proto",
        "@io_bazel_rules_go//proto/wkt:duration_go_proto",
        "@io_bazel_rules_go//proto/wkt:empty_go_proto",
        "@io_bazel_rules_go//proto/wkt:field_mask_go_proto",
        "@io_bazel_rules_go//proto/wkt:source_context_go_proto",
        "@io_bazel_rules_go//proto/wkt:struct_go_proto",
        "@io_bazel_rules_go//proto/wkt:timestamp_go_proto",
        "@io_bazel_rules_go//proto/wkt:type_go_proto",
        "@io_bazel_rules_go//proto/wkt:wrappers_go_proto",
        "@org_golang_google_protobuf//reflect/protoreflect:go_default_library",
        "@org_golang_google_protobuf//runtime/protoimpl:go_default_library",
    ],
)
```

[...]

PR #3437, opened 2023-01-31
Use protoc-gen-go-grpc as gRPC compiler

### LINKED ISSUE #3022: Use protoc-gen-go-grpc for gRPC compilers
The generated gRPC code is different from the result of the [protoc-gen-go-grpc plugin](google.golang.org/grpc/cmd/protoc-gen-go-grpc). By reading the rule definition, it seems that rules_go is using the [deprecated grpc mode](google.golang.org/protobuf) mode in proto-gen-go.

Currently I'm working around it with creating my own rule for the grpc-go proto compiler, but the official bazel rules should provide it.

0.29.0

0.24.0

4.2.1-homebrew

Yes

MacOS 12.0.1 Montery

N/A

### What did you do?

`bazel build` a proto with service defined and compiler using `@io_bazel_rules_go//proto:go_grpc`

The generated code should have `var <ServiceName>_ServiceDesc = ...`

### What did you see instead?

The generated code has `var _<ServiceName>_ServiceDesc = ...`

sessfeld:
**What type of PR is this?**

Feature

Currently we use the deprecated gRPC mode of proto-gen-go. I believe rules_go should use this newer version.

#3022, #2522

**Other notes for review**

From a first look, the change itself looks pretty straightforward. However, because `protoc-gen-go-grpc` no longer generates the Proto message, it requires both `go_proto` and `go_grpc`. That'd require updating Gazelle and `go_googleapis`, and I'm unsure how one would orchestrate that between the two codebases. I also believe that this would require updates to the dependency docs, I'd be great if I could get some pointers where to add that.

### DIFF
--- a/proto/BUILD.bazel
+++ b/proto/BUILD.bazel
@@ -29,13 +29,14 @@ go_proto_compiler(
 
 go_proto_compiler(
     name = "go_grpc",
-    options = ["plugins=grpc"],
+    plugin = "@org_golang_google_grpc_cmd_protoc_gen_go_grpc//:protoc-gen-go-grpc",
+    suffix = "_grpc.pb.go",
+    valid_archive = False,
     visibility = ["//visibility:public"],
-    deps = PROTO_RUNTIME_DEPS + WELL_KNOWN_TYPES_APIV2 + [
+    deps = WELL_KNOWN_TYPES_APIV2 + [
         "@org_golang_google_grpc//:go_default_library",
         "@org_golang_google_grpc//codes:go_default_library",
         "@org_golang_google_grpc//status:go_default_library",
-        "@org_golang_x_net//context:go_default_library",
     ],
 )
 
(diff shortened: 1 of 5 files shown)


mtoader:
Will this make into the new release by any chance ?

proto/BUILD.bazel
```diff
@@ -29,13 +29,14 @@ go_proto_compiler(
 
 go_proto_compiler(
     name = "go_grpc",
-    options = ["plugins=grpc"],
+    plugin = "@org_golang_google_grpc_cmd_protoc_gen_go_grpc//:protoc-gen-go-grpc",
+    suffix = "_grpc.pb.go",
+    valid_archive = False,
```
adam-azarchs, line 34: Defining a new, separate target would avoid issues with things like gazelle and allow incremental migration.

michaelboyd2:
@sessfeld Are you still working on this? I think it would be pretty useful.

DolceTriade:
+1, this is a great fix to start using newer gRPC generated code.

sessfeld:
> @sessfeld Are you still working on this? I think it would be pretty useful.

Unfortunately I don't have the capacity at the moment, I'd love to take this up again some time later this summer.

PR #3439, opened 2023-02-01
Extract generated code from BUILD.toolchains.bazel into a macro

fmeum:
**What type of PR is this?**

Refactoring

As a macro, the code is cleaner and easier to reuse. In a follow-up change, it will be used to simplify toolchain registration with Bzlmod.
    
This commit also removes the public `declare_toolchains` macro. It seems to have been public only because it was accessed by the generated SDK repos. It has never been documented and subject to backwards incompatible changes in recent releases.

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -1,8 +1,8 @@
-load("@{rules_go_repo_name}//go/private/rules:binary.bzl", "go_tool_binary")
-load("@{rules_go_repo_name}//go/private/rules:sdk.bzl", "package_list")
-load("@{rules_go_repo_name}//go/private/rules:transition.bzl", "non_go_reset_target")
-load("@{rules_go_repo_name}//go/private:go_toolchain.bzl", "declare_go_toolchains")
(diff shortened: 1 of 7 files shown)


go/private/BUILD.sdk.bazel
```diff
@@ -50,7 +50,7 @@ go_sdk(
 
 go_tool_binary(
     name = "builder",
-    srcs = ["@{rules_go_repo_name}//go/tools/builders:builder_srcs"],
+    srcs = ["@io_bazel_rules_go//go/tools/builders:builder_srcs"],
```
achew22, line 53: Does this inhibit our ability to rename the repo to `rules_go` in the long run?

tyler-french (approved):
Tested in Uber's Go Monorepo (cc @linzhp)
Before #3439
- Build of a simple target Passes with WORKSPACE
- Build of a simple target Passes with Bzlmod + register toolchains

After #3439
- Build of a simple target Passes with WORKSPACE
- Build of a simple target Passes with Bzlmod

PR #3440, opened 2023-02-01
bzlmod: Prohibit undesirable SDK registrations in non-root modules

fmeum:
This API is still very new and there are no modules in the BCR declaring their own Go SDK.


tyler-french (approved):
Tested in Uber's Go Monorepo with and without Bzlmod

PR #3443, opened 2023-02-01
bzlmod: Improve SDK registration

fmeum:
**What type of PR is this?**

Feature

Improve Bzlmod toolchain registration

Before this commit, users had to do the following in their
`MODULE.bazel` file to register a custom Go SDK:

```
go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
go_sdk.download(name = "my_go", version = "1.19.5")
use_repo(go_sdk, "my_go")
register_toolchains("@my_go_toolchains//:all")
```

This has several drawbacks:
* Names can collide with SDKs defined by transitive dependencies as they
  all live in the same namespace.
* The SDK has to be brought into scope with `use_repo`.
* Since the toolchain definitions are separate from the SDK repo, users
  need to know the naming convention for the separate toolchain repo for
  their `register_toolchains` calls.
* It encourages using the SDK repo directly, which is almost never
  correct: Toolchain resolution should be used instead, possibly in the
  form of the new convenience target `@rules_go//go`.

With this commit, Bazel toolchain definitions are emitted into a single
repo and registered by rules_go itself. SDK repo names are mangled and
kept as an implementation detail of rules_go. Module dependencies are no
longer allowed to register SDKs with a non-mangled name, but an
exception is made for the root module for the sake of backwards
compatibility and to address unforeseen use cases.

As a result of this change, SDK registration now looks as follows:

```
go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
go_sdk.download(version = "1.19.5")
```

**Other notes for review**

Currently stacked on #3438, #3439 and #3440, only the last commit is new.


fmeum:
@Wyverald This is the PR I told you about. Could you review the last commit? Leaving this as draft as it is stacked on other PRs with refactorings that enable this change.

Wyverald (commented):
I can't say I understand everything that's going on, but the basics of generating the SDK hub repo looks good to me. (Only looked at the 4th commit as you asked)

go/private/extensions.bzl
```diff
@@ -15,40 +15,130 @@ _download_tag = tag_class(
...
             if not module.is_root and not download_tag.version:
                 fail("go_sdk.download: version must be specified in non-root module " + module.name)
-            go_download_sdk(
-                name = download_tag.name,
+
+            # SDKs with an explicit name are at risk of colliding with those from other modules.
+            # This is acceptable if brought in by the root module as the user is responsible for any
+            # conflicts that arise. rules_go itself provides "go_default_sdk", which is used by
+            # Gazelle to bootstrap itself.
+            # TODO: Investigate whether Gazelle can use the first user-defined SDK instead to
```
tyler-french, line 49: Might be worth creating an issue in gazelle to track this.

go/private/extensions.bzl
```diff
@@ -15,40 +15,130 @@ _download_tag = tag_class(
...
+    # Keep the version out of the repository name if possible to prevent unnecessary rebuilds when
+    # it changes.
+    return "{name}_{version}_{tag_type}_{index}".format(
+        name = module.name,
+        version = module.version if multi_version else "",
+        tag_type = tag_type,
+        index = index,
+    )
+
+def _toolchain_prefix(index, name):
```
tyler-french, line 134: It might be helpful to have a comment giving an example of this string transformation, as it is a bit difficult to understand.

go/private/sdk.bzl
```diff
@@ -142,73 +142,164 @@ _go_download_sdk = repository_rule(
...
+    prefix = "{prefix}",
+    go_toolchain_repo = "@{sdk_repo}",
+    host_goarch = "{goarch}",
+    host_goos = "{goos}",
+    major = {identifier_prefix}MAJOR_VERSION,
+    minor = {identifier_prefix}MINOR_VERSION,
+    patch = {identifier_prefix}PATCH_VERSION,
+    prerelease = {identifier_prefix}PRERELEASE_SUFFIX,
+    sdk_type = "{sdk_type}",
+)
```
tyler-french, line 208: I think it would be good to either have a test for this function, to prevent changes from causing unintended generated `declare_bazel_toolchains(` rules.

go/private/sdk.bzl
```diff
@@ -142,73 +142,164 @@ _go_download_sdk = repository_rule(
...
+        sdk_types,
+        sdk_versions):
+    if len({len(l): None for l in [
+        prefixes,
+        geese,
+        goarchs,
+        sdk_repos,
+        sdk_types,
+        sdk_versions,
+    ]}) != 1:
```
tyler-french, line 238: optional: might be more clear to extract to a helper `_lengths_are_same( .. )`

tyler-french (approved):
A few stylistic comments. Logic and flow LGTM!

go/private/sdk.bzl
fmeum: Added tests for the two `sdk_version` cases.

go/private/extensions.bzl
fmeum: Created https://github.com/bazelbuild/bazel-gazelle/issues/1469 and referenced it in the comment.

sluongng:
Reading the source today and `goos` plural being `gees` cracked me up. 🤣

PR #3447, opened 2023-02-07
add bazel test rule for excluded packages in popular_repos

### LINKED ISSUE #3446: add bazel tests for excluded packages in popular_repo

0.38.0

0.29.0

6.0.0

Yes

Mac OS
arm64

No

### What did you do?
rules_go has integration tests on popular repos

The integration tests runs the unit tests in those repos. However, they exclude many packages for different reasons.

We should still build those packages as part of integration tests for rules_go. I use [build_test](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/build_test_doc.md#build_test) to catch build issues in popular repos.

`bazel test` works fine with all the test targets

### What did you see instead?
The bazel test works successfully with all the packages that were excluded, except packages with `/internal/` or `/testdata/` in its path.

uberzzr:
**What type of PR is this?**
Feature

rules_go has integration tests on popular repos. The integration tests runs the unit tests in those repos. However, they exclude many packages for different reasons. I add those excluded packages as part of integration tests for rules_go, using [build_test](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/build_test_doc.md#build_test) to catch build issues as early, except packages with /internal/ or /testdata/ in its path.

Fixes #3446 

**Other notes for review**


fmeum (approved):

fmeum:
Thanks, glad to see the tests pass.

tests/integration/popular_repos/popular_repos.py
```diff
@@ -272,6 +276,8 @@ def popular_repos_bzl():
 def build_bazel():
   with open(path.join(path.dirname(__file__), "BUILD.bazel"), "w") as f:
     f.write(BUILD_HEADER)
+    f.write("\n" + LOAD_BAZEL_TEST_RULE)
+    need_test = []
```
linzhp, line 280: The purpose of adding `build_test` target is because we cannot run tests in those packages. So the name `need_test` is confusing. It's better to call it `build_only`

uberzzr: Created another pull request to handle the comment #3449

PR #3454, opened 2023-02-09
Read GOEXPERIMENT env var before setting it

### LINKED ISSUE #3453: setting GOEXPERIMENT env var is ignored
0.38.1

0.26

6.0

yes

linux/amd64 and darwin/arm64

n/a

### What did you do?
Setting GOEXPERIMENT environment variables weren't being respected by rules go. Dug into it, and GOEXPERIMENT environment value is not respected by rules_go, because it's directly resetting it via os.Setenv() in compilepkg.go. 

GOEXPERIMENT env var getting respected.

### What did you see instead?
GOEXPERIMENT env var not getting respected.

sywhang:
**What type of PR is this?**
Bug fix

When GOEXPERIMENT environment variable is set, this gets completely reset by builder to some other value if the build rule specifies goexperiment.

This patch changes that logic so that it appends to the existing GOEXPERIMENT environment variable if one exists.

Fixes #3453

**Other notes for review**

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -134,7 +134,11 @@ func compilePkg(args []string) error {
 	}
 
 	if len(experiments) > 0 {
-		os.Setenv("GOEXPERIMENT", strings.Join(experiments, ","))
+		if fromEnv := os.Getenv("GOEXPERIMENT"); fromEnv != "" {
+			os.Setenv("GOEXPERIMENT", fromEnv+","+strings.Join(experiments, ","))
+		} else {
+			os.Setenv("GOEXPERIMENT", strings.Join(experiments, ","))
+		}
 	}
 
 	return compileArchive(
--- a/go/tools/builders/link.go
+++ b/go/tools/builders/link.go
@@ -149,7 +149,11 @@ func link(args []string) error {
 	goargs = append(goargs, "-o", *outFile)
 
 	if len(experiments) > 0 {
-		os.Setenv("GOEXPERIMENT", strings.Join(experiments, ","))
+		if fromEnv := os.Getenv("GOEXPERIMENT"); fromEnv != "" {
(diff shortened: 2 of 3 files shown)


achew22:
I don't object to this change, but out of curiosity, why do you prefer to set an environment variable for this instead of configuring it in a checked in build file? To my naive eyes, I would expect experimental features to basically be mandatory if you're using them, and checking their configuration in with your build files would be useful. Can you help me understand your use case a little better?

sywhang:
@achew22 sorry, I probably should've clarified that a little bit more in the PR summary. I don't necessarily have a strict use case for it; it's just that I was trying to investigate some build time regression between setting `GOEXPERIMENT=nounified` and not with recent Go 1.20 release, and was trying to do that via env var rather than modifying the build rules. 

Ideally doing either should work, but it took me a while to figure out why setting the environment variable just got ignored.

fmeum (changes requested):
I just noticed that it is missing from the docs on `go_download_sdk`, but all SDK rules accept an [`experiments`](https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#go_register_toolchains) attribute that you can set to a list of `GOEXPERIMENTS` to enable.

Generally speaking, it seems much more in line with Bazel's hermeticity principles to have the experiments enabled on a per-toolchain basis in `WORKSPACE` rather than via environment variables.
 
I am not even sure how the `GOEXPERIMENTS` variable would end up in the individual actions emitted by the ruleset. We aren't specifying `use_default_shell_env` on `ctx.actions.run` as far as I can tell. How are you setting `GOEXPERIMENTS` with this PR?

linzhp:
> I am not even sure how the GOEXPERIMENTS variable would end up in the individual actions emitted by the ruleset.

I am guessing `--action_env`? Otherwise Bazel is not doing its job in hiding env vars.

PR #3455, opened 2023-02-13
go: support "all:" prefix for "go:embed" directives

### LINKED ISSUE #3441: embedsrcs not correctly used when using all: prefixed patterns of go:embed
gazelle [supports now using `go:embed all:` directives](https://github.com/bazelbuild/bazel-gazelle/pull/1432), which isn't yet supported by rules_go.

From [Go docs](https://pkg.go.dev/embed#hdr-Directives):

> If a pattern begins with the prefix ‘all:’, then the rule for walking directories is changed to include those files beginning with ‘.’ or ‘_’. For example, ‘all:image’ embeds both ‘image/.tempfile’ and ‘image/dir/.tempfile’.

v0.37.0

(master)

https://github.com/bazelbuild/bazel-gazelle/archive/af021617bef884e1e40afb01b692092a471d9a48.tar.gz

Build label: 6.0.0
Build time: Mon Dec 19 15:52:35 2022 (1671465155)
Build timestamp: 1671465155
Build timestamp as int: 1671465155

No, I'm using a pre-release version of gazelle.

tested on Ubuntu 22.04 amd64 and NixOS amd64

### What did you do?

assuming

```shell-session
$ tree data
data
└── __hidden.txt

0 directories, 1 file
```
main.go
```go
package "main"

//go:embed data all:*
var Data embed.FS
...
```

Successful build.

### What did you see instead?

```
$ bazel build //cli/...
INFO: Analyzed 16 targets (1 packages loaded, 4949 targets configured).
INFO: Found 16 targets...
ERROR: /home/katexochen/test/constellation/cli/internal/helm/BUILD.bazel:3:11: GoCompilePkg cli/internal/helm/helm.a failed: (Exit 1): builder failed: error executing command (from target //cli/internal/helm:helm) bazel-out/k8-opt-exec-2B5CBBC6-ST-a2b97ed4b8d6/bin/external/go_sdk/builder '-param=bazel-out/k8-fastbuild-ST-a2b97ed4b8d6/bin/cli/internal/helm/helm.a-0.params'

compilepkg: /home/katexochen/.cache/bazel/_bazel_katexochen/f83094fca0f0f2ee3451f9caccc3e2f4/sandbox/linux-sandbox/9977/execroot/__main__/cli/internal/helm/loader.go:37:12: could not embed all:charts/*: no matching files found
INFO: Elapsed time: 1.247s, Critical Path: 0.37s
INFO: 14 processes: 12 internal, 2 linux-sandbox.
```

malt3:
**What type of PR is this?**

Feature

Implements the "all:" prefix for `//go:embed` directives. From [Go docs](https://pkg.go.dev/embed#hdr-Directives):

> If a pattern begins with the prefix ‘all:’, then the rule for walking directories is changed to include those files beginning with ‘.’ or ‘_’. For example, ‘all:image’ embeds both ‘image/.tempfile’ and ‘image/dir/.tempfile’.

Fixes #3441

**Other notes for review**


tests/core/go_library/embedsrcs_test.go
```diff
@@ -119,6 +122,22 @@ func TestFiles(t *testing.T) {
 				"embedsrcs_transitioned",
 			},
 		},
+		{
+			desc: "all",
+			fsys: all,
+			want: []string{
+				".",
+				"embedsrcs_static",
+				"embedsrcs_static/dir",
```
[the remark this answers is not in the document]
malt3: Had some trouble using `.git` as a folder name but I added some alternative examples

fmeum (approved):
Thanks, looks good!

On last thing: I just noticed that the Go docs only ever mention *files* starting with `.` and `_` - do you know whether `some/_hidden/file` would be included also without `all:...` with the standard Go toolchain?

malt3:
> On last thing: I just noticed that the Go docs only ever mention files starting with . and _ - do you know whether some/_hidden/file would be included also without all:... with the standard Go toolchain?

The Go docs seem to be correct! Any directories are always included (including those starting with `.` or `_`). Should I fix the behavior of `rules_go` to follow this more closely?

fmeum:
That would be great, yes. Could you keep that change in a separate commit?

@linzhp @achew22 Do you have any backwards compatibility concerns?

malt3:
A correct implementation should closely follow https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/load/pkg.go;l=2045;drc=2f2c5e41e7728cae99abd14e8bd5ffcfebf9df8e

I do want to get there but I feel like we would need change the implementation more drastically to get it to be correct.

Additional edge cases I found:

- Hidden files can be conditionally in-/excluded based on if the folder name is given with a glob or not
  given the folder `a` and the file `a/.hidden`, the directive `//go:embed a` would exclude the file, while `//go:embed a/*` would include the file.
- There  are some exclude patterns for VCS: https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/load/pkg.go;drc=261fe25c83a94fc3defe064baed3944cd3d16959;bpv=0;bpt=1;l=2203
- Folders named `go.mod` are ignored https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/load/pkg.go;l=2158;drc=261fe25c83a94fc3defe064baed3944cd3d16959;bpv=0;bpt=1

I don't have the time available to implement this behavior correctly.

fmeum:
Thanks for collecting the pointers, that's already very helpful.

Given that you uncovered a number of places in which we (unknowingly) deviate from the native toolchain, let's just keep the scope of this PR limited to `all:` support. I certainly won't ask you to fix all of that.

How do you feel about doing the following as part of this PR?

1. Add excludes for ".bzr", ".hg", ".git", and ".svn", so that users of `all:` don't accidentally pick up their version control metadata when they wouldn't with the native Go toolchain.
2. Add the list in your last comment as a TODO to the code.

Let me know if you think that you won't have the time for it and we can consider merging as is.

linzhp:
> Do you have any backwards compatibility concerns?
No. I don't see a backward-compatibility issue here.

malt3:
I implemented excludes for ".bzr", ".hg", ".git", and ".svn" and added TODOs.
I also noticed that the current implementation is already correct when it comes to the inclusion of hidden leaf nodes that I thought was not correct at first:

> Hidden files can be conditionally in-/excluded based on if the folder name is given with a glob or not
given the folder a and the file a/.hidden, the directive //go:embed a would exclude the file, while //go:embed a/* would include the file.

This works in the same way that it does for the native toolchain.

go/tools/builders/embedcfg.go
```diff
@@ -382,3 +414,23 @@ func fsValidPath(name string) bool {
 		name = name[i+1:]
 	}
 }
+
+// isBadEmbedName reports whether name is the base name of a file that
+// can't or won't be included in modules and therefore shouldn't be treated
+// as existing for embedding.
+//
+// TODO: This should use the equivalent of golang.org/x/mod/module.CheckFilePath instead of fsValidPath.
```
fmeum, line 422:
```suggestion
// TODO: This should use the equivalent of golang.org/x/mod/module.CheckFilePath instead of fsValidPath.
// https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/load/pkg.go;l=2200;drc=261fe25c83a94fc3defe064baed3944cd3d16959
```

go/tools/builders/embedcfg.go
```diff
@@ -314,12 +340,18 @@ func resolveEmbed(embed fileEmbed, root *embedNode) (matchedPaths, matchedFiles
 		}
 
 		// Matching directory. Recursively add all files in subdirectories.
-		// Don't add hidden files or directories (starting with "." or "_").
+		// Don't add hidden files or directories (starting with "." or "_"),
+		// unless "all:" prefix was set.
 		// See golang/go#42328.
 		matchTreeErr := matchNode.walk(func(childRel string, childNode *embedNode) error {
+			// TODO: Should check that directories along path do not begin a new module
```
fmeum, line 347:
```suggestion
			// TODO: Should check that directories along path do not begin a new module
			// https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/load/pkg.go;l=2158;drc=261fe25c83a94fc3defe064baed3944cd3d16959
```

go/tools/builders/embedcfg.go
```diff
@@ -295,17 +295,43 @@ func resolveEmbed(embed fileEmbed, root *embedNode) (matchedPaths, matchedFiles
...
 	// Search for matching files.
 	err = root.walk(func(matchRel string, matchNode *embedNode) error {
-		if ok, _ := path.Match(embed.pattern, matchRel); !ok {
+		if ok, _ := path.Match(pattern, matchRel); !ok {
 			// Non-matching file or directory.
 			return nil
 		}
+
+		// TODO: Should check that directories along path do not begin a new module
+		// (do not contain a go.mod).
```
fmeum, line 320:
```suggestion
		// (do not contain a go.mod).
		// https://cs.opensource.google/go/go/+/master:src/cmd/go/internal/load/pkg.go;l=2158;drc=261fe25c83a94fc3defe064baed3944cd3d16959
```

fmeum (approved):

fmeum:
Thanks! I added the links to the comments. This is good to go now.

PR #3461, opened 2023-02-22
Add missing darwin CPUs to apple matcher

keith:


keith:
I see failures locally with this change, we'll see if CI feels the same

keith:
@fmeum with your fix i think we can merge this one now?

fmeum (approved):
@keith Thanks for the ping

keith:
Thanks

PR #3468, opened 2023-03-01
Revert "Add automatic platform detection from inbound crosstool_top a…

fmeum:
This reverts commit https://github.com/fmeum/rules_go/commit/026db6df5ad760e9002f627bdb4f6b695c04d76a.

Bazel is migrating to platforms and toolchains resolution, but this logic frequently reset platforms back based on the value of the legacy `--cpu` and `--crosstool_top` flags. Instead, users of rules that still rely on these flags should use the `platform_mappings` file.


linzhp:
This PR will partially fix #3535

linzhp (approved):
@steeve We are planning revert your PR to solve a performance issue compiling stdlib on macOS (#3535). To support `--crosstool_top` and `--cpu`, Apple/iOS people should follow [platform mappings](https://bazel.build/concepts/platforms#platform-mappings) instead.

steeve:
Seeing this 2 months later! Sure platform mappings are a cleaner solution.

PR #3471, opened 2023-03-09
feat: provide convenient access to Go binary

cgrindel:
**What type of PR is this?**

Feature

Make the binary target `@rules_go//go` available publicly. This is useful for client projects that define Bazel targets to run maintenance tasks (e.g. `go mod tidy`).

NA

**Other notes for review**

I noticed that the 0.38.1 release archive does not include the `//go/tools/go_bin_runner` package. From the dates, it looks like it should exist. I looked for a list of packages that are used for packaging the release. The best that I can tell the archive is created by the `releaser` tool. That tool seems to just create a git archive of the repo in its current state. 🤷‍♂️

### DIFF
--- a/go/BUILD.bazel
+++ b/go/BUILD.bazel
@@ -1,10 +1,11 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
+# Provides a nice way to execute the Go binary used by rules_go. This can be
+# handy when running certain utility functions like `go mod tidy`.
 alias(
     name = "go",
     actual = "//go/tools/go_bin_runner",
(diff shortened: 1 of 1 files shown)


cgrindel:
@fmeum Will provide a PR instead.

fmeum:
@linzhp Do you happen to know why the release archive doesn't include the `go/tools/go_bin_runner` package?

linzhp:
I don't know :(

fmeum:
Well, turns out the commit just didn't make it into 0.38.0 and 0.38.1 was just a patch release.

PR #3473, opened 2023-03-10
Expand locations in `x_defs` values

fmeum:
**What type of PR is this?**

Feature

Allows embedding rlocationpaths into binaries to allow them to find dependencies at runtime even when run from other tools, meaning that `args` and `env` can't be used for this purpose.


fmeum:
This is preparation for https://github.com/bazelbuild/rules_go/pull/3474, but paired with `$(rlocationpath ...)`, it's quite useful in general. We may be able to use it for `gazelle` eventually.

linzhp (approved):

Issue #2255, opened 2019-10-23
Can't use rules_go toolchain as `toolchains` attribute in genrule


gkousik:
### What version of rules_go are you using?
0.20.1

### What version of gazelle are you using?
0.17.0

### What version of Bazel are you using?
1.1.0

### Does this issue reproduce with the latest releases of all the above?
Yes

### What operating system and processor architecture are you using?
`Linux x86_64`

### Any other potentially useful information about your toolchain?
None

### What did you do?
I have a genrule like:

```
genrule(
    name = "foo",
    srcs = [],
    outs = ["foo.txt"],
    cmd = "touch foo.txt > $@",
    toolchains = [
        "@io_bazel_rules_go//go:toolchain",
    ],
)
```

### What did you expect to see?
That this toolchain is successfully provided to the execution of the genrule.

### What did you see instead?
```
'@io_bazel_rules_go//go:toolchain' does not have mandatory providers: 'TemplateVariableInfo'
ERROR: Analysis of target '//foo:bar' failed; build aborted: Analysis of target '//foo:bar' failed; build aborted
INFO: Elapsed time: 0.662s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (6 packages loaded, 9 targets configured)
```

jayconrod:
Could you explain more about the problem you're trying to solve? The example you gave reproduces the error, but I don't understand why you'd want to use a Go toolchain in a `genrule`.

This is the first time I've heard of the `toolchains` attribute of `genrule`. The documentation says this:

```
The set of toolchains that supply "Make variables" that this target can use in some of its attributes. Some rules have toolchains whose Make variables they can use by default.
```

The Go toolchain does not provide any variables like this. I guess `TemplateVariableInfo` is how they would be provided. What variables do you need? Why not consume the toolchain in a regular rule instead of a `genrule`?

gkousik:
Sure, I'm trying to build a go-plugin based on some C++ libraries (using SWIG). I also build some C++ libraries as part of that, and after the C++ libraries are built, I want to use the same `go_sdk` used by bazel to call `go build -buildmode=plugin <sourcefiles>`.

I need to use the same `go_sdk`, so that the plugin can be built with the same version of go-compiler that all of our other go-tooling uses.

gkousik:
The actual genrule looks something like this:

```
genrule(
name = "buildwithswig",
cmd = "run-build.sh > $@",
srcs = ["foo.go"],
[...]

PR #3474, opened 2023-03-10
Make `//go` usable in scripts run with `bazel run`

fmeum:
**What type of PR is this?**

Feature

Allows developers to build their own local convenience scripts around `go`, e.g. to run `go mod tidy` backed by a hermetic Go toolchain and without the need to hardcode the name of an SDK repository, which we want to get away from with Bzlmod.

This requires getting rid of the `env` attribute as it is not evaluated if the binary is run as a dependency of another target.

Since `//go` select a Go SDK suitable for the host, not the exec platform, we forbid its use in the exec configuration. As remarked in https://github.com/bazelbuild/rules_go/issues/2255, using raw `go` in a genrule is not a good idea to start with.

https://bazelbuild.slack.com/archives/CDBP88Z0D/p1678384340169399

**Other notes for review**

Stacked on https://github.com/bazelbuild/rules_go/pull/3473

### DIFF
--- a/docs/go/core/defines_and_stamping.md
+++ b/docs/go/core/defines_and_stamping.md
@@ -21,6 +21,9 @@ value. You can also override stamp values from libraries using `x_defs`
 on the `go_binary` rule if needed. The `--[no]stamp` option controls whether
 stamping of workspace variables is enabled.
 
+The values of the `x_defs` dictionary are subject to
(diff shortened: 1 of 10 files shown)


fmeum:
@cgrindel

linzhp (approved):

PR #3478, opened 2023-03-13
Avoid repository rule restarts in go_sdk

jfirebaugh:
**What type of PR is this?**

Bug fix

Previously, `go_sdk` would always restart after fetching and extracting the SDK, as seen in Bazel profiles:

Screen Shot 2023-03-13 at 11 50 20 AM

The effectively doubles the amount of time it takes to run the rule implementation. [Repository rule restarts](https://docs.bazel.build/versions/5.0.0/skylark/repository_rules.html#when-is-the-implementation-function-executed) are discussed in https://github.com/bazelbuild/bazel-gazelle/issues/1175 but not resolved for `go_sdk`, or maybe this particular restart is a regression since that issue was resolved.

By adding debug `print`s, I determined that the restart is caused by this line:

https://github.com/bazelbuild/rules_go/blob/89e3296c36b789e8623e1178a59506547a98419e/go/private/sdk.bzl#L440

Moving the label to the default value of an attribute avoids the restarts, as seen here:

Screen Shot 2023-03-13 at 11 55 36 AM

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -38,6 +38,9 @@ go_host_sdk_rule = repository_rule(
         "experiments": attr.string_list(
             doc = "Go experiments to enable via GOEXPERIMENT",
         ),
+        "_sdk_build_file": attr.label(
+            default = Label("//go/private:BUILD.sdk.bazel"),
+        ),
     },
 )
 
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Thank you very much for the really thorough investigation and great fix!

PR #3484, opened 2023-03-18
Canonicalize tags before transitioning

fmeum:
Reduces the number of distinct configurations as the order and multiplicity of tags doesn't matter.

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -49,6 +49,9 @@ TRANSITIONED_GO_SETTING_KEYS = [
     "//go/config:tags",
 ]
 
+def _deduped_and_sorted(strs):
+    return sorted({s: None for s in strs}.keys())
+
 def _original_setting_key(key):
     if not "//go/config:" in key:
         fail("_original_setting_key currently assumes that all Go settings live under //go/config, got: " + key)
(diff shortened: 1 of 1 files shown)


go/private/rules/transition.bzl
```diff
@@ -117,15 +120,13 @@ def _go_transition_impl(settings, attr):
 
     tags = getattr(attr, "gotags", [])
     if tags:
-        tags_label = "//go/config:tags"
-        settings[tags_label] = tags
+        settings["//go/config:tags"] = _deduped_and_sorted(tags)
```
steeve, line 123: you can use `sets.to_list(sets.make(tags))` from skylib: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/new_sets_doc.md

fmeum: That would still require a call to `sorted` though, so I'm not sure it would reduce the overall complexity here.

go/private/rules/transition.bzl
```diff
@@ -117,15 +120,13 @@ def _go_transition_impl(settings, attr):
...
 
     linkmode = getattr(attr, "linkmode", "auto")
     if linkmode != "auto":
         if linkmode not in LINKMODES:
             fail("linkmode: invalid mode {}; want one of {}".format(linkmode, ", ".join(LINKMODES)))
-        linkmode_label = "//go/config:linkmode"
-        settings[linkmode_label] = linkmode
+        settings["//go/config:tags"] = linkmode
```
fmeum, line 129:
```suggestion
        settings["//go/config:linkmode"] = linkmode
```

tyler-french:
Tested against Uber's Go Monorepo

tyler-french (approved):

PR #3485, opened 2023-03-18
Only infer platforms from crosstool/cpu if at default

fmeum:
**What type of PR is this?**

Bug fix

If the current target platform differs from the default value (the host platform), then we don't want to override it with a value inferred from the legacy --cpu and --crosstool_top flags. Otherwise it would become impossible to "platformize" a Go build without having a matching platform mappings file.

This is not expected to break any existing use cases and if it does, it can and should be worked around with a platform mappings file that maps the legacy flag values to `//go/platform` targets.

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -37,6 +37,9 @@ load(
     "platform_from_crosstool",
 )
 
+_DEFAULT_PLATFORMS_VALUE = [Label("@local_config_platform//:host")]
+_PLATFORMS_LABEL = "//command_line_option:platforms"
+
 # A list of rules_go settings that are possibly set by go_transition.
 # Keep their package name in sync with the implementation of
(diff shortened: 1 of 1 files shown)


fmeum:
CC @keith This should unblock https://github.com/bazelbuild/rules_go/pull/3461#issuecomment-1440780828

motiejus:
Do I understand correctly that change is a noop for repositories that do not use `--cpu` and `--crosstool_top`? If yes, then I don't have any concerns with this.

(My stake is platforms, transitions and their interaction in bazel-zig-cc and Uber's Go Monorepo)

fmeum:
@motiejus Yes, this is intended to be a noop if you either don't transition `--platforms` yourself or never change `--cpu` and/or `--crosstool_top` without also changing `--platforms` (e.g. via a platform mappings file).

PR #3488, opened 2023-03-23
Keep build tags that affect the stdlib build

### LINKED ISSUE #3456: gotags in go_binary rules is not handled correctly for the standard library
In rules_go v0.38.1 (specifically after https://github.com/bazelbuild/rules_go/commit/8d309d5af15814b4096d80b60f80fa86128c43f2) building a binary with `gotags=["timetzdata"]` no longer includes the timezone database embedded in the build. Can see that it does trigger a rebuild of the stdlib when adding/removing the tag, but it seems it doesn't link with the right thing at the end.

If an anonymous import (`import _ "time/tzdata"`) is added, or if adding the cmdline flag `--define=gotags=timetzdata` then the build does correctly embed the tzdata.

This is a regression from v0.37.0 where this worked as expected.

v0.38.1

n/a

5.4.0

yes

linux/amd64

n/a

### What did you do?

<details>
<summary>Reproducer</summary>

WORKSPACE
```
workspace(name = "com_example_tztest")

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "dd926a88a564a9246713a9c00b35315f54cbd46b31a26d5d8fb264c07045f05d",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.38.1/rules_go-v0.38.1.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.38.1/rules_go-v0.38.1.zip",
    ],
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies", "go_download_sdk")

go_rules_dependencies()

go_register_toolchains(version = "1.19.5")
```
cmd/tztest/main.go
```
package main

import "time"

func main() {
	_, err := time.LoadLocation("America/New_York")
	if err != nil {
		panic("tzdata is missing")
	}
}
```

cmd/tztest/BUILD
```

go_library(
    name = "go_default_library",
    srcs = ["main.go"],
    importpath = "example.com/cmd/tztest",
)

go_binary(
    name = "tztest",
    gotags = ["timetzdata"],
    pure = "on",
)
```

test.sh
```
#!/bin/sh

bazel build //cmd/tztest:tztest_embed

echo "Checking if result contains embedded tzdata..."
if go tool nm "$(bazel info execution_root 2> /dev/null)/$(bazel cquery --output=files //cmd/tztest:tztest 2> /dev/null)" | grep -q tzdata; then
    echo "Build contains tzdata!"
    exit 0
else
    echo "Build is missing tzdata :("
    exit 1
fi
```

</details>

When the above script is run with v0.37.0 the artefact correctly contains timezone data and the script prints `Build contains tzdata!`

### What did you see instead?

With v0.38.0 or v0.38.1 the artefact does not contain the embedded tzdata and the script prints `Build is missing tzdata :(`

fmeum:
**What type of PR is this?**

Bug fix

Since 8d309d5, all `gotags` were cleared from the configuration when building the standard library. This is not correct as some tags, such as `timetzdata`, do affect the standard library build.

This commit adds a tool that scrapes the list of build tags relevant for the standard library from Go SDK sources and keeps those tags that are relevant for Go 1.18-1.20.

Fixes #3456

**Other notes for review**

### DIFF
--- a/go/tools/internal/stdlib_tags/stdlib_tags.go
+++ b/go/tools/internal/stdlib_tags/stdlib_tags.go
@@ -1,0 +1,174 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"go/build/constraint"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var goVersionRegex = regexp.MustCompile(`^go1.(\d+)$`)
+
+// Used to update the list of tags affecting the standard library kept in
+// transitions.bzl.
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("usage: stdlib_tags <go SDK src directory>...")
+	}
+
+	filteredTags, err := extractBuildTags(os.Args[1:]...)
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	fmt.Printf("_TAG_AFFECTS_STDLIB = {\n")
+	for _, tag := range filteredTags {
+		fmt.Printf("    %q: None,\n", tag)
+	}
+	fmt.Printf("}\n")
+}
+
+func extractBuildTags(sdkPaths ...string) ([]string, error) {
+	tags := make(map[string]struct{})
+	for _, dir := range sdkPaths {
+		err := filepath.WalkDir(dir, func(path string, d os.DirEntry, err error) error {
+			if d.IsDir() {
+				if d.Name() == "testdata" {
+					return filepath.SkipDir
+				}
+				return nil
+			}
+			if filepath.Ext(path) != ".go" {
+				return nil
+			}
+			if strings.HasSuffix(filepath.Base(path), "_test.go") {
+				return nil
+			}
+			return walkFile(path, tags)
+		})
+		if err != nil {
+			return nil, fmt.Errorf("%s: %w", dir, err)
+		}
+	}
+
+	filteredTags := make([]string, 0, len(tags))
+	for tag := range tags {
+		if !shouldExclude(tag) {
+			filteredTags = append(filteredTags, tag)
+		}
+	}
+	sort.Strings(filteredTags)
+
+	return filteredTags, nil
+}
+
+func shouldExclude(tag string) bool {
+	// Set via CGO_ENABLED
+	return tag == "cgo" ||
+		// Set via GOARCH and GOOS
+		knownOS[tag] || knownArch[tag] || tag == "unix" ||
+		// Set via GOEXPERIMENT and GOAMD64
+		strings.HasPrefix(tag, "goexperiment.") || strings.HasPrefix(tag, "amd64.") ||
+		// Set implicitly
+		goVersionRegex.MatchString(tag)
+}
+
+func walkFile(path string, tags map[string]struct{}) error {
+	file, err := os.Open(path)
+	if err != nil {
+		return err
+	}
+
+	scanner := bufio.NewScanner(file)
+	// The Go SDK contains some very long lines in vendored files (minified JS).
+	scanner.Buffer(make([]byte, 0, 128*1024), 1024*1024)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if !isConstraint(line) {
+			continue
+		}
+		c, err := constraint.Parse(line)
+		if err != nil {
+			continue
+		}
+		walkConstraint(c, tags)
+	}
+
+	if err = scanner.Err(); err != nil {
+		return fmt.Errorf("%s: %w", path, err)
+	}
+	return nil
+}
+
+func walkConstraint(c constraint.Expr, tags map[string]struct{}) {
+	switch c.(type) {
+	case *constraint.AndExpr:
+		walkConstraint(c.(*constraint.AndExpr).X, tags)
+		walkConstraint(c.(*constraint.AndExpr).Y, tags)
+	case *constraint.OrExpr:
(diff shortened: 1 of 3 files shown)


fmeum:
Manually verified that this fixes https://github.com/bazelbuild/rules_go/pull/3457.

go/tools/internal/stdlib_tags/stdlib_tags.go
```diff
@@ -0,0 +1,161 @@
...
+		line := scanner.Text()
+		if !isConstraint(line) {
+			continue
+		}
+		c, err := constraint.Parse(line)
+		if err != nil {
+			continue
+		}
+		walkConstraint(c, tags)
+	}
```
tyler-french, line 90:
```suggestion
	}
	if err := scanner.Err(); err != nil {
		return err
	}
```

go/tools/internal/stdlib_tags/stdlib_tags.go
```diff
@@ -0,0 +1,161 @@
...
+	"strings"
+)
+
+// Used to update the list of tags affecting the standard library kept in
+// transitions.bzl.
+func main() {
+	if len(os.Args) < 2 {
+		println("usage: stdlib_tags <go SDK src directory>...")
+		os.Exit(1)
+	}
```
tyler-french, line 20: optional nit: Generally prefer to only exit once in `main()`, and also prefer to parameterize the Output as a writer.

This way, we can write a test for this function with some other `io.Writer`

```suggestion
// Used to update the list of tags affecting the standard library kept in
// transitions.bzl.
func main() {
	if err := run(os.Stdout); err != nil {
		log.Fatal(err)
	}
}
func run(output io.Writer) error {
	if len(os.Args) < 2 {
		return errors.New("usage: stdlib_tags <go SDK src directory>...")
	}
	...
	_, err := fmt.Fprintf(output, "_TAG_AFFECTS_STDLIB = {\n")
	...
```

go/tools/internal/stdlib_tags/stdlib_tags.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			return walkFile(path, tags)
+		})
+		if err != nil {
+			println(dir + ": " + err.Error())
+			os.Exit(1)
+		}
+	}
+
+	var filteredTags []string
```
tyler-french, line 45: We can pre-allocate the capacity of this for performance.
```suggestion
	filteredTags := make([]string, 0, len(tags))
```

go/tools/internal/stdlib_tags/stdlib_tags.go
```diff
@@ -0,0 +1,161 @@
...
+	sort.Strings(filteredTags)
+
+	fmt.Printf("_TAG_AFFECTS_STDLIB = {\n")
+	for _, tag := range filteredTags {
+		fmt.Printf("    %q: None,\n", tag)
+	}
+	fmt.Printf("}\n")
+}
+
+var goVersionRegex = regexp.MustCompile(`^go1.(\d+)$`)
```
tyler-french, line 60: optional nit: I generally prefer to keep top-level variable declarations at the top. This includes the `knownOS` and `knownArch`

go/tools/internal/stdlib_tags/stdlib_tags.go
```diff
@@ -0,0 +1,161 @@
...
+	// Set via CGO_ENABLED
+	return tag == "cgo" ||
+		// Set via GOARCH and GOOS
+		knownOS[tag] || knownArch[tag] || tag == "unix" ||
+		// Set implicitly
+		goVersionRegex.MatchString(tag) ||
+		// Set via GOEXPERIMENT and GOAMD64
+		strings.HasPrefix(tag, "goexperiment.") || strings.HasPrefix(tag, "amd64.")
+}
+
```
tyler-french, line 72: optional: For performance, would it be faster to join the first 4 conditions into a single map at startup?

This way, we only need to do once map lookup.

Also, would it make sense to check the regex last since this is likely the slowest operation?

go/tools/internal/stdlib_tags/stdlib_tags.go
```diff
@@ -0,0 +1,161 @@
...
+		if !isConstraint(line) {
+			continue
+		}
+		c, err := constraint.Parse(line)
+		if err != nil {
+			continue
+		}
+		walkConstraint(c, tags)
+	}
+
```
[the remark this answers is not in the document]
fmeum: Good catch, there actually was a scanner error on a very long line. I increased the buffer size.

go/tools/internal/stdlib_tags/stdlib_tags.go
fmeum: I refactored out a testable function, but left in the prints as we would have to distinguish between `stdout` and `stderr` and the remainder of the `main` function can't really be effectively unittested anyway.

go/tools/internal/stdlib_tags/stdlib_tags.go
fmeum: Done

go/tools/internal/stdlib_tags/stdlib_tags.go
fmeum: Moved up the `goVersionRegex`, but I ended up keeping the maps at the end. They are essentially generated code that I think would distract from this otherwise very short tool.

PR #3491, opened 2023-03-26
Quote `CC` if it contains whitespace

fmeum:
**What type of PR is this?**

Bug fix

The `cgo` tool splits `CC` on unquoted whitespace, which can cause failures on Windows when `CC` points to an absolute path such as `C:\Program Files\some\compiler.exe`.

### DIFF
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -436,6 +436,23 @@ func passLongArgsInResponseFiles(cmd *exec.Cmd) (cleanup func()) {
 	return cleanup
 }
 
+// quotePathIfNeeded quotes path if it contains whitespace and isn't already quoted.
+// Use this for paths that will be passed through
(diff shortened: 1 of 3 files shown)


achew22 (approved):

tyler-french (approved):

ash2k:
This is a bit weird. I'd think the consumer of an environment variable needs to quote it IF it uses it in a shell script or some such. What if a program uses contents of `CC` to open/copy/etc the file? It will fail because filename is incorrect - it contains quotes.

ash2k:
Unrelated to this PR, but something I noticed: it's weird to mutate OWN environment before calling an external program to pass the environment vars to it. Why not just pass them? Current approach will not work if code needs to run multiple programs concurrently.

fmeum:
@ash2k No need to worry, we are only quoting `CC` when calling the `go` tool itself to *build* Go binaries. The code changed in this PR also doesn't run in parallel, so while `Setenv` is certainly not good practice, it should be perfectly fine in this case.

PR #3492, opened 2023-03-26
Only install `runtime/cgo` with `pure = False`

### LINKED ISSUE #1885: GoStdLib: don't install runtime/cgo in pure mode
I don't think this is necessary, and I'm mildly surprised it works.

fmeum:
**What type of PR is this?**

Other

We don't know why installing `runtime/cgo` even if cgo is disabled doesn't seem to break anything, so we shouldn't do it.

Fixes #1885


go/private/actions/stdlib.bzl
```diff
@@ -82,6 +82,9 @@ def _build_stdlib(go):
     if go.mode.race:
         args.add("-race")
     args.add_all(go.sdk.experiments, before_each = "-experiment")
+    args.add("-package", "std")
+    if not go.mode.pure:
+        args.add("-package", "runtime/cgo")
```
achew22, line 87: Does this unconditionally include `runtime/cgo` in the binary? If that isn't used do you get an unnecessary import/link directive error?

achew22 (approved):

tyler-french (approved):

go/private/actions/stdlib.bzl
fmeum: No, this just affects the `stdlib` install action, it won't link anything extra into binaries. The term `package` is slightly confusing here, but it is what the arguments to `go install` are (with `std` being a special value).

PR #3493, opened 2023-03-26
Disable cgo by default with unsupported compilers

fmeum:
**What type of PR is this?**

Bug fix

MSVC-style compilers are explicitly not supported by cgo. If the current C/C++ toolchain is detected to be of that type, disable cgo by default unless explicitly requested.

Without this commit, building the standard library on Windows fails even with #3491 without forcing use of mingw-gcc, which we don't want to ask of our Go-only users.


go/private/context.bzl
```diff
@@ -73,6 +73,12 @@ load(
     "request_nogo_transition",
 )
 
+# cgo requires a gcc/clang style compiler.
+_UNSUPPORTED_C_COMPILERS = {
```
achew22, line 77: I would probably go the other direction and enumerate that `gcc` and possibly `musl-gcc` are supported. It's unlikely that `tinycc` (or any of the other thousands of esoteric compilers) is a supported compiler but it's not on this list.

achew22 (approved):

go/private/context.bzl
fmeum: I agree that would be conceptually more correct, but Bazel hasn't always been doing a good job at returning something meaningful in the `compiler` field: Until I believe 6.1.0, most compilers just have `"compiler"` in there. I don't think that there is any further unsupported compiler for which Bazel ever returned a distinctive compiler name.

achew22: Could we leave a comment requesting the next person who edits this list update it to be a inclusion list instead of an exclusion list assuming version 6.1 to spend released and is generally in use by that point?

go/private/context.bzl
```diff
@@ -783,15 +792,18 @@ cgo_context_data = rule(
...
     optional toolchains, so instead, we have an optional dependency on this
     rule.""",
 )
 
 def _cgo_context_data_proxy_impl(ctx):
-    return [ctx.attr.actual[CgoContextInfo]] if ctx.attr.actual else []
+    if not ctx.attr.actual:
+        return []
+    if not CgoContextInfo in ctx.attr.actual:
+        return []
```
tyler-french, line 805:
```suggestion
    if not ctx.attr.actual or not CgoContextInfo in ctx.attr.actual:
        return []
```
Can we combine these conditions?

tyler-french (approved):

PR #3494, opened 2023-03-27
prepare release 0.39

tyler-french:
Prepare lease 0.39

This diff updates all dependencies in go/private/repositories.bzl **_except_** for `@go_googleapis`

This is because of issues relating to https://github.com/bazelbuild/bazel-gazelle/issues/1422

When tested downstream, many build failures arise from this upgrade. Therefore, we are keeping the version of this dependency stagnant while a better solution to these dependencies is explored. 

This PR was accomplished by running:
```
bazel run //go/tools/releaser -- upgrade-dep -mirror=false all      
```
And then reverting the changes to `@go_googleapis` and its patches.

This change was tested downstream in Uber's Go Monorepo

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,13 @@ Slack: `#go on Bazel Slack`_, `#bazel on Go Slack`_
 
 Announcements
 -------------
+2023-03-27
+  Release
+  `v0.39.0 <https://github.com/bazelbuild/rules_go/releases/tag/v0.39.0>`_
+  is now available. This release includes a simpler interface for Bzlmod
+  `go_sdk` registration, makes the `//go` tool available to users, and
+  fixes various bugs.
+
 2022-12-06
   Release
   `v0.37.0 <https://github.com/bazelbuild/rules_go/releases/tag/v0.37.0>`_
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "rules_go",
-    version = "0.38.1",
+    version = "0.39.0",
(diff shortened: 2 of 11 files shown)


README.rst
```diff
@@ -72,6 +72,12 @@ Slack: `#go on Bazel Slack`_, `#bazel on Go Slack`_
 
 Announcements
 -------------
+2023-03-27
+  Release
+  `v0.39.0 <https://github.com/bazelbuild/rules_go/releases/tag/v0.39.0>`_
+  is now available. This release includes support for Bzlmod and various
```
fmeum, line 78: We had bzlmod before this release. Maybe just say that SDK registration has been simplified? We could also point out the new `//go` target.

fmeum:
@sluongng While I do think that these patches would overall be extremely helpful, they would still be somewhat backwards incompatible. This release includes some important fixes for Bzlmod, so maybe we'll have it as is and follow up with a mildly breaking `0.40.0` that fixes the proto mess. What do you think?

sluongng:
I have no objection to that. We definitely want to treat the breaking change with care 🤗

fmeum (approved):

linzhp:
I agree. The fix for googleapis proto is quite a significant effort. We can create a separate rules_go/Gazelle release as soon as it's fixed.

wyattanderson:
> @wyattanderson are you still working on porting https://github.com/wyattanderson/googleapis-reproducer changes into rules_go?

Sorry, I've sort of lost steam on this effort. Might get to it eventually, but I'm not actively working on it.

PR #3499, opened 2023-03-28
Revert "Add new darwin CC toolchain for tests (#3460)"

### LINKED ISSUE #3498: //go/tools/releaser doesn't build on macOS

515b5d03429098e93ae461d37ee5b55bdd42edb9

6.1.1

Yes

macOS arm64

I have XCode command line tools but not full XCode

### What did you do?

```
$ bazel build //go/tools/releaser
```

build succeeds

### What did you see instead?

```
ERROR: /Users/zplin/go/src/github.com/bazelbuild/rules_go/WORKSPACE:181:27: fetching _apple_cc_autoconf rule //external:local_config_apple_cc: Traceback (most recent call last):
        File "/private/var/tmp/_bazel_zplin/5b0cce1975e01e8748828c47e46c38ba/external/build_bazel_apple_support/crosstool/setup.bzl", line 40, column 17, in _apple_cc_autoconf_impl
                fail("Failed to configure Apple CC toolchain, if you only have the command line tools installed and not Xcode, you cannot use this toolchain. You should either remove it or temporarily set '{}=1' in the environment".format(_DISABLE_ENV_VAR))
Error in fail: Failed to configure Apple CC toolchain, if you only have the command line tools installed and not Xcode, you cannot use this toolchain. You should either remove it or temporarily set 'BAZEL_NO_APPLE_CPP_TOOLCHAIN=1' in the environment
```

### Other info
I tried `BAZEL_NO_APPLE_CPP_TOOLCHAIN=1 bazel build //go/tools/releaser` too, but it failed with:

```
ERROR: /private/var/tmp/_bazel_zplin/5b0cce1975e01e8748828c47e46c38ba/external/bazel_tools/tools/cpp/BUILD:58:19: every rule of type cc_toolchain_alias implicitly depends upon the target '@local_config_apple_cc//:toolchain', but this target could not be found because of: no such target '@local_config_apple_cc//:toolchain': target 'toolchain' not declared in package '' defined by /private/var/tmp/_bazel_zplin/5b0cce1975e01e8748828c47e46c38ba/external/local_config_apple_cc/BUILD (Tip: use `query "@local_config_apple_cc//:*"` to see all the targets in that package)
```

linzhp:
This reverts #3460.

**What type of PR is this?**

Bug fix

Revert cd9817098856f6b993937c2dd794a69850cd90ea to unblock rules_go release

Fixes #3498

**Other notes for review**
@keith FYI

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,8 +1,3 @@
-build --enable_platform_specific_config
-build:macos --apple_crosstool_top=@local_config_apple_cc//:toolchain
-build:macos --crosstool_top=@local_config_apple_cc//:toolchain
-build:macos --host_crosstool_top=@local_config_apple_cc//:toolchain
-
 build:windows --cpu=x64_windows
 build:windows --compiler=mingw-gcc
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -165,17 +165,3 @@ http_archive(
 load("@io_bazel_stardoc//:setup.bzl", "stardoc_repositories")
 
 stardoc_repositories()
-
-# For testing objc_library interop, users should not need to install it
-http_archive(
-    name = "build_bazel_apple_support",
-    sha256 = "77a121a0f5d4cd88824429464ad2bfb54bdc8a3bccdb4d31a6c846003a3f5e44",
-    url = "https://github.com/bazelbuild/apple_support/releases/download/1.4.1/apple_support.1.4.1.tar.gz",
-)
-
-load(
-    "@build_bazel_apple_support//lib:repositories.bzl",
-    "apple_support_dependencies",
-)
-
-apple_support_dependencies()


fmeum (approved):

keith:
if you have only the command line tools you won't be able to run those tests starting with bazel 7.x. So this isn't a long term fix.

tyler-french (approved):

linzhp:
> if you have only the command line tools you won't be able to run those tests starting with bazel 7.x

Can you give more context on that? Does it only affect those tests or Go packages in general? `//go/tools/releaser` is a pretty simple Go package, so I assume if affects all Go packages. It would be a pretty big migration effort to install full XCode on all dev machines in a big corporation.

linzhp:
Also, we should move towards a direction of hermetic C++ compilers (e.g., using https://github.com/uber/bazel-zig-cc) . Overriding a generic `@bazel_tools//tools/cpp:toolchain` with a local toolchain further breaks the hermeticity, right?

keith:
The toolchain you're removing here is only needed for the test that uses an `objc_library`, so otherwise you should be able to disable it. The problem is the env var isn't enough because of this config:

https://github.com/bazelbuild/rules_go/blob/13aa4706394734625e441c9334d88dc3bd566e40/.bazelrc#L1-L4

so we also need a way to opt out of that in this case 🤔

linzhp:
We can have a different buildkite step to add those toolchains and run the `objc_library`.

keith:
sounds good, note that this change might break rules_go in the bazel downstream pipeline

linzhp:
Should we add those flags to bazel downstream pipeline too?

keith:
i think it gets them from your .bazelci config

PR #3501, opened 2023-03-29
Use `go_default_sdk` in `.bcr/presubmit.yml`

fmeum:
Context: https://github.com/bazelbuild/bazel-central-registry/pull/547


fmeum:
FYI @tyler-french @linzhp

PR #3509, opened 2023-04-01
Fix use of `//go` with Bzlmod by removing dependency on `@org_golang_x_sys`

bricedp:
**What type of PR is this?**

Bug fix

Running `bazel run //go` (or `bazel run @rules_go//go` from another module) fails when using Bzlmod because the `@org_golang_x_sys` repository is not visible. Without Bzlmod, the repository is loaded through a WORKSPACE dependency and the error does not happen.

Related to these recent changes:

- https://github.com/bazelbuild/rules_go/pull/3429
- https://github.com/bazelbuild/rules_go/pull/3474

Two-step minimal reproduction from root of this repo:

1. `touch WORKSPACE.bzlmod`
2. `bazel run --enable_bzlmod //go`

Here's the error that produces:

```shell
$ bazel run --enable_bzlmod //go
DEBUG: gazelle@0.27.0/MODULE.bazel:7:6: WARNING: The bazel_gazelle Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
WARNING: For repository 'com_google_protobuf', the root module requires module version protobuf@3.19.2, but got protobuf@3.19.6 in the resolved dependency graph.
ERROR: /tmp/rules_go/go/tools/go_bin_runner/BUILD.bazel:11:11: no such package '@[unknown repo 'org_golang_x_sys' requested from @]//unix': The repository '@[unknown repo 'org_golang_x_sys' requested from @]' could not be resolved: No repository visible as '@org_golang_x_sys' from main repository and referenced by '//go/tools/go_bin_runner:go_bin_runner_lib'
ERROR: Analysis of target '//go:go' failed; build aborted: 
INFO: Elapsed time: 3.043s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (45 packages loaded, 865 targets configured)
    currently loading: go/platform
ERROR: Build failed. Not running target
```

Here's the point in the BUILD file where this repository is being referenced:

https://github.com/bazelbuild/rules_go/blob/e558f56df9995f571e49de336c3831d407d9ff70/go/tools/go_bin_runner/BUILD.bazel#L25

None

**Other notes for review**

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -43,8 +43,16 @@ tasks:
   macos:
     shell_commands:
       - tests/core/cgo/generate_imported_dylib.sh
+    build_flags:
+      - "--apple_crosstool_top=@local_config_apple_cc//:toolchain"
+      - "--crosstool_top=@local_config_apple_cc//:toolchain"
+      - "--host_crosstool_top=@local_config_apple_cc//:toolchain"
     build_targets:
     - "//..."
+    test_flags:
+      - "--apple_crosstool_top=@local_config_apple_cc//:toolchain"
+      - "--crosstool_top=@local_config_apple_cc//:toolchain"
+      - "--host_crosstool_top=@local_config_apple_cc//:toolchain"
     test_targets:
     - "//..."
   rbe_ubuntu1604:
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -165,3 +165,17 @@ http_archive(
 load("@io_bazel_stardoc//:setup.bzl", "stardoc_repositories")
 
 stardoc_repositories()
+
+# For testing objc_library interop, users should not need to install it
+http_archive(
+    name = "build_bazel_apple_support",
+    sha256 = "77a121a0f5d4cd88824429464ad2bfb54bdc8a3bccdb4d31a6c846003a3f5e44",
+    url = "https://github.com/bazelbuild/apple_support/releases/download/1.4.1/apple_support.1.4.1.tar.gz",
+)
+
+load(
+    "@build_bazel_apple_support//lib:repositories.bzl",
+    "apple_support_dependencies",
+)
+
+apple_support_dependencies()


fmeum:
Thanks for the report and fix. 

This dependency is indeed missing. In fact, even with your fix, we only pick it up indirectly from Gazelle as it's not declared via `go_deps` in rules_go's MODULE.bazel file.

Given that `golang.org/x/sys` isn't promising backwards compatibility, I think that we may just be better of getting rid of this direct dependency - it isn't even clear that it is better than the pure Go alternative.

@bricedp Could you try deleting `process_unix.go` and the references to it and `golang.org/x/sys` in https://github.com/bazelbuild/rules_go/blob/e558f56df9995f571e49de336c3831d407d9ff70/go/tools/go_bin_runner/BUILD.bazel#L17? If that works, I think I would prefer that fix.

bricedp:
Thanks @fmeum, I agree that's much cleaner. I've removed `process_unix.go` and updated the BUILD file — I'm able to run `//go` without issue.

If ever needed for the future, I think the `@org_golang_x_sys` I had added was being brought in by the explicit call to `go_rules_dependencies`, which appears to be a newer version than the indirect dependency in `go.mod`. 

https://github.com/bazelbuild/rules_go/blob/e558f56df9995f571e49de336c3831d407d9ff70/go/private/extensions.bzl#L157

https://github.com/bazelbuild/rules_go/blob/e558f56df9995f571e49de336c3831d407d9ff70/go/private/repositories.bzl#L86-L101

fmeum (approved):
Looks good, thanks!

I will try to add a simple `sh_test` to `tests/bcr` that runs `go version` so that we don't regress this functionality.

fmeum:
@bricedp Sorry, I wanted to add a commit to your  branch, but unfortunately pushed the `master` branch of rules_go rather than my checkout of your fork to it, thus closing the PR. I can no longer push it to clean up the mess. Could you restore the commits from https://github.com/fmeum/rules_go/tree/bricedp/master and open a new PR?

bricedp:
Thanks @fmeum for the tests, opened https://github.com/bazelbuild/rules_go/pull/3512 with our changes.

PR #3514, opened 2023-04-02
WIP - added command to generate json

### LINKED ISSUE #3503: gopls - generated go files - tree artifacts

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "dd926a88a564a9246713a9c00b35315f54cbd46b31a26d5d8fb264c07045f05d",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.38.1/rules_go-v0.38.1.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.38.1/rules_go-v0.38.1.zip",
    ],
)

Not using it

6.0.0

Yes

Linux, NixOs

Local build, nothing special

### What did you do?

I wrote a special rule to generate golang files within bazel. The set of generated files is unknown except at runtime. 
I am aware of `actions.declare_directory` as a solution which indeed works. The only problem is that it seems the gopackagedriver doesn't seem to be able to pickup that the returned depset is actually a folder hence (I guess) it doesn't resolve to the right location.

By manually enumerating the files and defining them during action.run_shell everything works fine!

To gopls find the generated package.

### What did you see instead?

An error!

lromor:
**What type of PR is this?**

Feature

Adds support to tree artifacts for the gopackagedriver.

Fixes #3503 

**Other notes for review**

Mostly added the aspect code. What is left is to write the golang code that performs json encoding.


lromor:
@fmeum , I'm almost ready to open the PR for review. If you have some wishes to steer the development let me know! It's not extensively tested for now, but I can definitely add more features and checks. Do you have any remarks for now?

go/tools/gopackagesdriver/archive_to_json.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+		}).
+		WithArg("--data-srcs", func(value string) {
+			if strings.HasSuffix(value, ".go") {
+ 				archive.CompiledGoFiles = append(archive.CompiledGoFiles, value)
+ 			}
+		}).
+		WithArg("--output-file", func(value string) {
+			outputPath = value
+		})
```
fmeum, line 160: The argument parser implementation looks pretty good and there is no need to change it if it works, but I'm wondering whether you considered using https://pkg.go.dev/flag#FlagSet.Parse at some point.

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -48,31 +48,29 @@ def file_path(f):
...
+    args.add("--pkg-path", archive.data.importpath)
+    args.add("--export-file", archive.data.export_file)
+
+    # Will expand into GoFiles and OtherFiles.
+    args.add_all("--orig-srcs", archive.data.orig_srcs, map_each=file_path)
+
+    # Expands in CompiledGoFiles
+    args.add_all("--data-srcs", archive.data.srcs, map_each=file_path)
+    args.add("--output-file", pkg_json_file)
+    args.use_param_file("@%s.args.txt", use_always = True)
```
fmeum, line 66: I could see the param file adding overhead here (may be negligible though). Is `use_always = True` necessary here or could we let bazel decide?

fmeum (commented):
This looks pretty good already, I like where this is going.

go/tools/gopackagesdriver/aspect.bzl
lromor: Yes, my bad, should have left this out of the commit.

go/tools/gopackagesdriver/archive_to_json.go
lromor: Yes, I used it initially! The problem I had with it is that I couldn't find a neat way to handle the multi valued args. I think bazel allows you to customize the output formats quite a lot, yet it felt easier to match what I wrote in the bazel aspect without having to format stuff to match go flags requirements.

lromor: Anyways, I'm now testing it to check for potential issues and if it fixes my problem.

lromor:
I noticed that in order to make bazel expand the file direcotory, all files need to be set as inputs. I'm updating the action declaring all the sources as inputs.

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -48,31 +48,31 @@ def file_path(f):
...
+    args.add_all("--orig-srcs", archive.data.orig_srcs, map_each=file_path)
+
+    # Expands in CompiledGoFiles
+    args.add_all("--data-srcs", archive.data.srcs, map_each=file_path)
+    args.add("--output-file", pkg_json_file)
+    args.use_param_file("@%s", use_always = True)
+
+    inputs = archive.data.orig_srcs + archive.data.srcs
+    ctx.actions.run(
+        inputs = inputs,
```
fmeum, line 70: Adding an execution_requirement `local` may improve performance here. This action never actually reads its inputs, so any kind of sandboxing or remote execution would be wasted effort.

lromor: (solved)

lromor:
@fmeum there's still the test not really included as target in bazel. Shall I remove it?

lromor:
@fmeum If you are ok with the PR I won't add any more changes. Let me know if you need me to do anything else!

fmeum (approved):
Looks good, thanks!

@JamyDev Could you test this at Uber, especially with respect to its performance?

JamyDev:
Will test when I have some time. We actually ran into this, but fixed our rule to statically define the outputs, I'll have to revert that to test it.

fmeum:
Sorry, I forgot about this for a long time. 

@lromor Could you resolve the merge conflicts?
@JamyDev Did you get to test this?

lromor:
@fmeum , I should have some cycles this weekend!

JamyDev:
@fmeum I'll give it a shot after rebase. Albeit we don't have any declare_directory rules anymore, and I'm concerned that this adds additional complexity to something that could be fixed by just declaring the files directly.

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -49,32 +49,53 @@ def file_path(f):
...
+    return "{}{}{}".format(escaped_key, delimiter, escaped_value)
 
-def make_pkg_json(ctx, name, pkg_info):
+def _go_archive_to_pkg_json(ctx, name, archive):
     pkg_json_file = ctx.actions.declare_file(name + ".pkg.json")
-    ctx.actions.write(pkg_json_file, content = json.encode(pkg_info))
+
+    # Build the args in a config file.
+    args = ctx.actions.args()
+    args.add("--id", archive.data.label)
```
ChrisVandoo, line 70: Hi @lromor,

Thanks for tackling this issue as this is something I would also love to see resolved!

I was doing some testing with this branch and ran into an issue where the package ID set in the `.pkg.json` file that is created didn't match what `bazel query` ([here](https://github.com/bazel-contrib/rules_go/blob/0433213eb9474a592de87fd4dd5f49e8aef11432/go/tools/gopackagesdriver/bazel_json_builder.go#L170)) was returning since the `--consistent_labels` flag is set (packages returned from the query were formatted like `@//some/thing` vs. the ID in `.pkg.json` was formatted `//something/thing` which caused issues when trying to match packages to root later on.

I think this also matches what is happening in L59 that was removed.

go/tools/gopackagesdriver/archive_to_json.go
```diff
@@ -0,0 +1,233 @@
...
+		}).
+		WithArg("--pkg-path", func(value string) {
+			archive.PkgPath = value
+		}).
+		WithArg("--export-file", func(value string) {
+			archive.ExportFile = value
+		}).
+		WithArg("--data-srcs", func(value string) {
+			if strings.HasSuffix(value, ".go") {
+				archive.GoFiles = append(archive.GoFiles, value)
```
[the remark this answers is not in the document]
lromor: This I really didn't notice. Thank you for pointing that out! I'll try to pass that to the archive jsonifier.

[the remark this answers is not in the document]
ChrisVandoo: no problem, thanks for the work you've done to get a fix in for this!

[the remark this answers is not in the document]
lromor: I'm now hunting why there's a nil map running: `//go/tools/gopackagesdriver:gopackagesdriver_test`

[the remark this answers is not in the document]
lromor: Alright, fixed, the `Imports` field must be not omitted even if empty, to match how bazel generates the json.

lromor:
@fmeum ?

fmeum:
@jayconrod This looks good to me, but the packagesdriver is definitely a weak spot of mine. Could you review this as well?

fmeum (approved):

jayconrod:
Oh, this is a big one! I'll block out some time tomorrow to review.

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -49,32 +49,53 @@ def file_path(f):
...
+    args.use_param_file("@%s")
+
+    ctx.actions.run(
+        inputs = archive.data.srcs,
+        outputs = [pkg_json_file],
+        executable = ctx.executable._archive_to_json,
+        execution_requirements = {
+            "local": "1",
+        },
+        mnemonic = "ArchiveToJSON",
```
jayconrod, line 96: Should have `Go` in the name, maybe `GoArchiveJSON`?

go/tools/gopackagesdriver/BUILD.bazel
jayconrod, line 1: I think this needs a test as part of `gopackagesdriver_test.go`.

The smallest test case I can think of would be to add a `.bzl` file in the workspace created in `TestMain` with a trivial rule that declares a directory artifact, then runs a shell command to write a .go file there. That directory artifact could then be used in a `go_library`'s `srcs`.

go/tools/gopackagesdriver/archive_to_json.go
```diff
@@ -0,0 +1,233 @@
...
+	"bufio"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+type scanner interface {
```
jayconrod, line 13: If you can squash this with `//go/tools/builder`, then you can reuse `expandParamsFiles`, which already does this processing for most actions.

If not, this can be simplified quite a bit: it's fine to expand to a big slice of arguments rather than buffering and parsing incrementally.

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -49,32 +49,53 @@ def file_path(f):
...
-    )
+def _escape_special_chars(value, delimiter):
+    # We only have "\" and "=" as special characters.
+    # Double escape the escaper char.
+    escaped = value.replace("\\", "\\\\")
+
+    # Escape delimiter.
+    return escaped.replace(delimiter, "\\" + delimiter)
+
+def format_key_value_arg(key, value, delimiter = "="):
```
jayconrod, line 60: This doesn't seem necessary: it's fine to pass a `-arg=key=value` without escaping as long as `key` does not contain `=`. This is only used for `importpath`, and [Go import paths cannot contain =`](https://pkg.go.dev/golang.org/x/mod/module#CheckImportPath).

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -49,32 +49,53 @@ def file_path(f):
...
+    escaped_key = _escape_special_chars(key, delimiter)
+    escaped_value = _escape_special_chars(value, delimiter)
+    return "{}{}{}".format(escaped_key, delimiter, escaped_value)
 
-def make_pkg_json(ctx, name, pkg_info):
+def _go_archive_to_pkg_json(ctx, name, archive):
```
jayconrod, line 65: As an optimization, let's keep the old code that builds the JSON at analysis time and only run this action if we know there's a tree artifact. Running an action is going to be slower, and this is latency-sensitive, so let's only do it when needed.

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -49,32 +49,53 @@ def file_path(f):
...
+
+    args.add("--output-file", pkg_json_file)
+    args.use_param_file("@%s")
+
+    ctx.actions.run(
+        inputs = archive.data.srcs,
+        outputs = [pkg_json_file],
+        executable = ctx.executable._archive_to_json,
+        execution_requirements = {
+            "local": "1",
```
jayconrod, line 94: Let's drop this or document if there's a strong reason to keep it.

In most cases, people won't have remote execution enable when using `gopackagesdriver`. If remote execution is enabled, let's respect it. Mixing local and remote actions causes a lot of additional file copying, which can ultimately be slower.

go/tools/gopackagesdriver/archive_to_json.go
```diff
@@ -0,0 +1,233 @@
...
+
+// The bazel args will be parsed either from the program args or
+// from a file if its path prepended with "@" is provided.
+func parseArchiveAndOutputPath(arguments []string, archive *archive) (string, error) {
+	var outputPath string
+	parser := NewBazelArgsParser().
+		WithArg("--id", func(value string) {
+			archive.ID = value
+		}).
+		WithArg("--pkg-path", func(value string) {
```
jayconrod, line 183: Let's use the `flag` package for argument parsing. I'd also suggest using `_` instead of `-` for consistency with other commands.

jayconrod (commented):
Thanks for working on this. I have some comments, which I know is going to be frustrating since this PR has been open for nearly two years, and I'm sorry for that. I think we'll land in a good place though.

Issue #3515, opened 2023-04-03
`//go/tools/gopackagesdriver` fails on generated code when used with `@rules_go` as repo name


bricedp:
### What version of rules_go are you using?

0.39.0

### What version of gazelle are you using?

0.30.0

### What version of Bazel are you using?

7.0.0-pre.20230316.2 (rolling)

### Does this issue reproduce with the latest releases of all the above?

Yes: https://github.com/bricedp/rules_go_repro

### What operating system and processor architecture are you using?

macOS 11.7.4 (Intel)

### Any other potentially useful information about your toolchain?

Using Bzlmod without any content in `WORKSPACE.bzlmod` (fully migrated to `MODULE.bazel`)

### What did you do?

Currently, the implementation of `//go/tools/gopackagesdriver` includes a hard-coded reference to the `@io_bazel_rules_go` repo name. When users import this module using the `@rules_go` repo name instead (without a `repo_name="io_bazel_rules_go"` override), the driver fails in some cases. For instance, I observed issues in VS Code when importing generated code built from protocol buffers. The package could not be found at all.

The GitHub repository below contains a reproduction of the issue with potential fix to verify:

https://github.com/bricedp/rules_go_repro

See [`README.md`](https://github.com/bricedp/rules_go_repro/blob/main/README.md) for instructions on use.

Note: My experience is with VS Code but this likely affects other editors as they all use the same driver.

### What did you expect to see?

There is already a variable for overriding repo name with the `GOPACKAGESDRIVER_RULES_GO_REPOSITORY_NAME` (in this case it needs to be set to `"@rules_go"`), but the code in `bazel_json_builder.go` does not make use of this variable.

https://github.com/bazelbuild/rules_go/blob/d756ad91feb9ca43800e781ab29c117623abed90/go/tools/gopackagesdriver/bazel_json_builder.go#L35

https://github.com/bazelbuild/rules_go/blob/d756ad91feb9ca43800e781ab29c117623abed90/go/tools/gopackagesdriver/main.go#L54

I propose we apply the following commit to `bazel_json_builder.go` to use the `rulesGoRepositoryName` variable, sourced from the `GOPACKAGESDRIVER_RULES_GO_REPOSITORY_NAME` environment variable. That environment variable can either be added to one's driver shell script or to VS Code settings under `go.toolsEnvVars`.

**Fix in commit:** https://github.com/bricedp/rules_go/commit/53a8f5ae44f4055cdfa962e153a85be9a8888b11

[...]

PR #3516, opened 2023-04-03
bzlmod: Fix repo name used by gopackagesdriver

fmeum:
**What type of PR is this?**

Bug fix

Ensure that we always specify the correct repo name when invoking the aspect by injecting it via `x_defs`. This works even with Bzlmod (where users usually refer to rules_go as `@rules_go`, not `@io_bazel_rules_go`, but could choose any apparent repository name for it).

Also replace one instance of a hardcoded repo name (thanks @bricedp).

#3515 

Fixes #

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/BUILD.bazel
+++ b/go/tools/gopackagesdriver/BUILD.bazel
@@ -1,4 +1,5 @@
 load("//go:def.bzl", "go_binary", "go_library")
+load(":aspect.bzl", "bazel_supports_canonical_label_literals")
 
 go_library(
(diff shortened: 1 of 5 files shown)


fmeum:
@tyler-french Could you review and check that this works at Uber (ideally with and without Bzlmod)?

@bricedp Could you check that this works for you even without the environment variable set?

bricedp:
Thanks @fmeum! Confirming that this works without the `GOPACKAGESDRIVER_RULES_GO_REPOSITORY_NAME` environment variable.

Issue #3472, opened 2023-03-10
switching from bazel test to coverage causes GoStdlib to be rebuilt with go1.20


com6056:
### What version of rules_go are you using?

v0.38.1 with go1.20

### What version of gazelle are you using?

v0.29.0

### What version of Bazel are you using?

5.4.0

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

macOS 13.2, M1 Max (but reproduces on our Linux systems as well)

### Any other potentially useful information about your toolchain?

Related thread in Bazel Slack: https://bazelbuild.slack.com/archives/CDBP88Z0D/p1677506103789169
Potentially related issue: https://github.com/golang/go/issues/58848

### What did you do?

Ran `bazel test //foo/bar:bar_test --nocache_test_results`, see actions like this show up
```
    GoStdlib external/io_bazel_rules_go/stdlib_/pkg; 10s darwin-sandbox
```

Ran `bazel coverage //foo/bar:bar_test --nocache_test_results`, see actions like this show up again:
```
    GoStdlib external/io_bazel_rules_go/stdlib_/pkg; 10s darwin-sandbox
```

### What did you expect to see?

The `GoStdlib` actions to not be run again

### What did you see instead?

The `GoStdlib` actions ran again

fmeum:
Go 1.20 stopped shipping the standard library in precompiled form and current version of Bazel aren't good at preventing rebuilds. The situation should improve with https://github.com/bazelbuild/bazel/pull/16910, so if you can, please try `last_green` after that has been merged.

com6056:
Bringing in the info from the Slack thread, seems to be a separate issue from https://github.com/bazelbuild/bazel/pull/16910.

When diffing the 2 configs, we don't see anything standing out:
```
diff test_config coverage_config
1,2c1,2
< BuildConfigurationValue c3cee25049c1cec9cc8257e5ee797b1ced86dd67427ffa666cb2917c41b58d0d:
< Skyframe Key: BuildConfigurationKey[c3cee25049c1cec9cc8257e5ee797b1ced86dd67427ffa666cb2917c41b58d0d]
---
> BuildConfigurationValue d9700ee9a063bff991db657d797c04995659d0648b00bf8228d2071c73cc65ea:
> Skyframe Key: BuildConfigurationKey[d9700ee9a063bff991db657d797c04995659d0648b00bf8228d2071c73cc65ea]
34c34
[...]

PR #3521, opened 2023-04-06
Remove LLVM coverage flags for CGo

fmeum:
**What type of PR is this?**

Bug fix

**What does this PR do? Why is it needed?**

We already remove the GCC variants of these flags.

PR #3522, opened 2023-04-06
Drop coverage linker flags from stdlib build

fmeum:
**What type of PR is this?**

Bug fix

We never compile CGo with coverage instrumentation and the stdlib contains no user-provided C/C++ code, so linking coverage runtimes is never needed.

Work towards #3472 

(fixing the issue requires a change to Bazel)

**Other notes for review**


fmeum:
Stacked on #3521

achew22 (approved):

PR #3524, opened 2023-04-07
gopackagesdriver: fix interface to work with golangci-lint (#3523)

### LINKED ISSUE #3523: `gopackagesdriver` doesn't work with golangci-lint

v0.39.0

v0.30.0

6.1.1

Yes

linux-amd64

### What did you do?

Exported `GOPACKAGESDRIVER` and `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE`, ran `golangci-lint`.

A successful run of `golangci-lint` with no `typecheck` errors.

### What did you see instead?

Silent failures because `golangci-lint` wasn't able to load the types from source and gave up on non-local libraries.

grampelberg:
Unlike gopls which uses `file=` queries exclusively, golangci-lint uses the more generic `go list` query syntax. This change allow for file paths not prefixed with `file=` to be queried correctly.

Additionally, some bugs have been fixed up:
- Packages with different IDs and identical import paths work now.
- WKT for protobufs are no longer skipped when the proto wrapper is used.`

**What type of PR is this?**

Bug fix

This PR makes it possible to run `golangci-lint` and use the bazel source tree, not `go list`.

- Expands the supported queries to most of what `go list` supports. `golangci-lint` naively passes the patterns through to `go/packages` which normally delegates to `go list`. Some examples would be:
    - `./foo`
    - `/abs/foo`
    - `./foo/bar.go`
    - `./foo/...`
- Packages with the same import path but separate IDs now work. This allows for multiple targets per package and aliases (rules_go's proto vs google's).
- WKT for protobufs from `rules_go` are now included in the result.

Fixes #3523 

**Other notes for review**

See #3552  for the companion PR.


JamyDev:
Had a quick scan over it and didn't immediately see anything concerning. I'm going to give this change a try on Uber's monorepo to see if all our scenarios still work and get back to you.

grampelberg:
@JamyDev thank you! If you run into any issues I'd love to get it updated. I added some debugging statements to `golangci-lint` for my testing (happy to push a branch if you'd like to see) because that was the easiest way to see the source compilation errors that arise. The easier, but somewhat less illuminating way is to set `GL_DEBUG` for `goanalysis`.

JamyDev:
Hey, so I tried applying this as a patch, but was unable to do so. Would you be able to rebase?

grampelberg:
@JamyDev sorry about that! Should be fixed now =)

JamyDev:
Some notes:

- In the current implementation setting `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE` to a bazel target pattern, will limit queries like `./...` to that query scope. It seems that in your implementation this will not happen and query the entire module (which in a monorepo is a lot 😅 ) (Check `bazelJsonBuilder.localQuery`)
- Not quite yet sure what happened, but we hit a crash with CGo stuff, may be a zig-cc issue, I'll discuss this internally with our team:
```
INFO: Running command line: bazel-bin/external/io_bazel_rules_go/go/tools/gopackagesdriver/gopackagesdriver_/gopackagesdriver 'file=src/code.uber.internal/devexp/green-keeper/main.go'
Running: [bazel --output_base /home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747 info --tool_tag=gopackagesdriver --ui_actions_shown=0]
INFO: Invocation ID: f8c77bc2-378b-461e-acf3-8f5f4f87004e
Running: [bazel --output_base /home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747 query --tool_tag=gopackagesdriver --ui_actions_shown=0 --ui_event_filters=-info,-stderr --noshow_progress --order_output=no --output=label --nodep_deps --noimplicit_deps --notool_deps kind("go_library|go_test|go_binary|cff", same_pkg_direct_rdeps("src/path/to/pkg/main.go"))]
Running: [bazel --output_base /home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747 build --tool_tag=gopackagesdriver --ui_actions_shown=0 --show_result=0 --build_event_json_file=/tmp/gopackagesdriver_bep_1360650674 --build_event_json_file_path_conversion=no --experimental_convenience_symlinks=ignore --ui_event_filters=-info,-stderr --noshow_progress --aspects=//tools/ide/packagesdriver:aspect.bzl%cff_pkg_info_aspect,@io_bazel_rules_go//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect --output_groups=go_pkg_driver_json_file,go_pkg_driver_stdlib_json_file,go_pkg_driver_srcs --keep_going //src/path/to/pkg:go_default_library]
ERROR: /home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747/external/io_bazel_rules_go/BUILD.bazel:42:7: GoStdlibList external/io_bazel_rules_go/stdlib_/stdlib.pkg.json failed: (Exit 1): builder failed: error executing command (from target @io_bazel_rules_go//:stdlib) bazel-out/k8-opt-exec-2B5CBBC6-ST-1d3589a9f1f7/bin/external/go_sdk/builder_reset/builder stdliblist -sdk external/go_sdk -installsuffix linux_amd64 -sdk external/go_sdk -out ... (remaining 1 argument skipped)

# runtime/cgo
cgo: C compiler "/home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747/sandbox/processwrapper-sandbox/4256/execroot/__main__/external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++" not found: exec: "/home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747/sandbox/processwrapper-sandbox/4256/execroot/__main__/external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++": stat /home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747/sandbox/processwrapper-sandbox/4256/execroot/__main__/external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++: no such file or directory
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x54b575]
```

grampelberg:
> In the current implementation setting GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE to a bazel target pattern, will limit queries like ./... to that query scope. It seems that in your implementation this will not happen and query the entire module (which in a monorepo is a lot 😅 ) (Check bazelJsonBuilder.localQuery)

I *think* that I made that considerably more efficient than it had been with `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE`. Previously, we would do a wildcard match that would be pretty expensive. The intent with `localQuery` is that it will always be a specific path (obviously `...` would be the entire sub-tree). Is that kind of specific `kind()` query still expensive in a large monorepo? My goal was to try and simplify the number of potential knobs you need.

All that said, I'm happy to add it back in.

> Not quite yet sure what happened, but we hit a crash with CGo stuff, may be a zig-cc issue

I'd be happy to poke around if you have any way for me to try out an approximation locally.

grampelberg:
> The query itself is fine, but the intent of GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE was to limit the view of the packagesdriver to a specific scope, since building everything would be near impossible. 

Should this be a blocker for all queries then? eg, before that if/else tree in `queryFromRequests` we check it against the scope? That'd block file + package queries though and wouldn't respect the original behavior where query scope was ignored for file queries.

JamyDev:
Hmm that's a good question. `file=` is a very pointed query, so skipping there may make sense, but in that case you would just not have the QUERY_FILTER. Now let's say you've got a dep outside of your filter, that is used by a package in your filter, it should still resolve (since you have a chance of opening that file in your IDE, or needing to resolve it); so build query may be a bit risky in that case.

I would say let's keep it as much to the original behavior as possible.

JamyDev:
@linzhp any idea why zig-cc wouldn't be included in the execution sandbox when building with the updated aspect in this diff active?

```
▶ bazel build --sandbox_debug --aspects=@io_bazel_rules_go//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect --output_groups=go_pkg_driver_json_file,go_pkg_driver_stdlib_json_file,go_pkg_driver_srcs //src/some/pkg:go_default_library

INFO: Writing tracer profile to '/tmp/bazel_20230426231308_f5kju'
INFO: Invocation ID: b884be6c-38b4-43e6-9587-d05e97f25767
INFO: Analyzed target //src/some/pkg:go_default_library (1150 packages loaded, 17273 targets configured).
INFO: Found 1 target...
ERROR: /home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f/external/io_bazel_rules_go/BUILD.bazel:42:7: GoStdlibList external/io_bazel_rules_go/stdlib_/stdlib.pkg.json failed: (Exit 1): process-wrapper failed: error executing command 
  (cd /home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f/sandbox/processwrapper-sandbox/23/execroot/__main__ && \
  exec env - \
    CC=external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++ \
    CGO_CFLAGS='-no-canonical-prefixes -Wno-builtin-macro-redefined -D__DATE__="redacted" -D__TIMESTAMP__="redacted" -D__TIME__="redacted" -include external/zig_sdk/glibc-hacks/fcntl.h -include external/zig_sdk/glibc-hacks/res_search-amd64.h -fno-sanitize=undefined -fno-lto -Wl,-S' \
    CGO_ENABLED=1 \
    CGO_LDFLAGS='-fno-sanitize=undefined -fno-lto -Wl,-S' \
    GOARCH=amd64 \
    GOOS=linux \
    GOPATH='' \
    GOROOT=external/go_sdk \
    GOROOT_FINAL=GOROOT \
    PATH=external/zig_sdk/tools/x86_64-linux-gnu.2.19:/bin:/usr/bin \
    TMPDIR=/tmp \
  /home/user/.cache/bazel/_bazel_jamy/install/e63ada1a9406bff70c6efc9d97c3aeba/process-wrapper '--timeout=0' '--kill_delay=15' '--stats=/home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f/sandbox/processwrapper-sandbox/23/stats.out' bazel-out/k8-opt-exec-2B5CBBC6-ST-1d3589a9f1f7/bin/external/go_sdk/builder_reset/builder stdliblist -sdk external/go_sdk -installsuffix linux_amd64 -sdk external/go_sdk -out bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_/stdlib.pkg.json)
# runtime/cgo
cgo: C compiler "/home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f/sandbox/processwrapper-sandbox/23/execroot/__main__/external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++" not found: exec: "/home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f/sandbox/processwrapper-sandbox/23/execroot/__main__/external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++": stat /home/user/.cache/bazel/_bazel_jamy/b97476d719d716accead0f2d5b93104f/sandbox/processwrapper-sandbox/23/execroot/__main__/external/zig_sdk/tools/x86_64-linux-gnu.2.19/c++: no such file or directory
Aspect @io_bazel_rules_go//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect of //src/code.uber.internal/devexp/green-keeper:go_default_library failed to build
INFO: Elapsed time: 6.595s, Critical Path: 1.89s
INFO: 4 processes: 4 internal.

▶ ls -al /home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747/sandbox/processwrapper-sandbox/766/execroot/__main__/external/
total 12
drwxr-xr-x 3 jamy user 4096 Apr 26 22:51 .
drwxr-xr-x 4 jamy user 4096 Apr 26 23:08 ..
drwxr-xr-x 5 jamy user 4096 Apr 26 23:08 go_sdk
```

Here's the [diff](https://gist.github.com/JamyDev/038ba7a6e7863d423de66b314b099846) I applied to the monorepo.

grampelberg:
@JamyDev alright, makes sense. I'll add the query scope into `localQuery` then. Lemme see what I can do with the docs as well, it was a little challenging for me to figure out what bits and pieces needed to be enabled for this use case.

linzhp:
My impression is that the action `GoStdlibList` uses the C compiler without declaring it in its `tools` attribute. However, it's unclear to me why the action `GoStdlib` is able to use it without declaring it. Maybe aspect enforces more strict sandbox?

If that's the case, I am not sure how to pass the C compiler to the `tools` attribute. What we get from `go.cgo_tools.cc_toolchain` is a `CcToolchainInfo` provider, which provides the compiler as a string, not a `File` object.

@fmeum Any idea?

fmeum:
`GoStdlib` adds `go.crosstool` to inputs, which is just `go.cgo_tools.cc_toolchain.all_files.to_list()`. I am not entirely sure why `GoStdlibList` needs the compiler, but if it does, that's the set of inputs we should add.

The reason we need this now and didn't before is probably that we pass `_cgo_env` into `_build_stdlib_list_json` unconditionally with this PR [here](https://github.com/bazelbuild/rules_go/pull/3524/files#diff-fd05bb9d040364753a7c9bf458bea610d632fdba2ced6db62151898571305d59R105) - do we really want to do this? It seems that it should at least be gated on the `pure` setting.

linzhp:
`GoStdLibList` runs `go list -json builtin std runtime/cgo`, which would preprocess some C code when cgo is enabled.

@grampelberg @JamyDev `GoStdLibList` needs to have the same input as `GoStdLib` after cgo is enabled, so the action has access to C compiler and libc headers.

Like Fabian said, do we really need to enable cgo for Go pkg driver? I ran `go list -json builtin std runtime/cgo` with `CGO_ENABLED=1` and `CGO_ENABLED=0`, the output is not too different. In fact, `CGO_ENABLED=0` produces more entries in `GoFiles` section for some packages, which is fine.

@grampelberg You may consider splitting the part of the PR that is not related to cgo to a different PR, so you can be unblocked for that part.

JamyDev:
> do we really need to enable cgo for Go pkg driver? 

Maybe we don't need it for stdlib, but we do need it for any of the queries, albeit that may change so it may make sense to support it from the getgo.

grampelberg:
@linzhp you shouldn't see any significant changes in `GoFiles`, you want to look at `CompiledGoFiles`. That's where `cc` preprocesses the go files and compiles them into something that is more usable. Because we don't have that today, `golangci-lint` is unable to work with anything that has imports from stdlib + cgo. The `net` library's a great example of this.

Separately, I'll get the changes split into stdlib + cgo and driver changes. That's a great suggestion.

grampelberg:
@linzhp yup! You also need `-compiled=true`.

linzhp:
I see the difference now. Thanks for pointing out. OK, let's split the PR, make `GoStdlibList` take the same list of input as `GoStdlib` and iterate from there. There are some rough edges that I need to get help from other people to make `GoStdlibList` work with hermetic cc toolchain.

> does it make sense to disable cgo entirely for packages driver?

Worst case, we can disable cgo for stdlib but we should still be able to enable cgo for other packages.

linzhp:
I mean something like this on top of your change:

```
diff --git a/go/private/actions/stdlib.bzl b/go/private/actions/stdlib.bzl
index 1cb953d4..fedb317d 100644
--- a/go/private/actions/stdlib.bzl
+++ b/go/private/actions/stdlib.bzl
@@ -70,7 +70,7 @@ def _build_stdlib_list_json(go, env):
     args.add("-out", out)
 
     go.actions.run(
-        inputs = go.sdk_files,
+        inputs = go.sdk_files + go.crosstool,
         outputs = [out],
         mnemonic = "GoStdlibList",
```

So `go list` can access the hermetic C compiler in Bazel sandbox.

grampelberg:
@linzhp I see! If you can point me in the direction of a way that I can test hermetic C compilers locally, I'd be happy to iterate towards something that works there.

In the meantime, this has been split into two PRs, see #3552 for the companion PR.

JamyDev:
Sorry for the delay here, I was able to get it functioning with your latest changes, so this is good to go

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -84,7 +89,13 @@ def _go_pkg_info_aspect_impl(target, ctx):
     transitive_compiled_go_files = []
 
     for attr in DEPS_ATTRS + PROTO_COMPILER_ATTRS:
-        for dep in getattr(ctx.rule.attr, attr, []) or []:
+        deps = getattr(ctx.rule.attr, attr, []) or []
+
+        # Some attrs are not iterable, ensure that deps is always iterable.
+        if type(deps) != "list":
```
fmeum, line 95: Could you replace that with a constant for `type([])`? The returned strings aren't guaranteed to be stable.

fmeum (approved):

fmeum:
Looks good to me, but I don't understand the details. @linzhp Could you merge this?

go/tools/gopackagesdriver/aspect.bzl
grampelberg: Well that makes sense and is way more readable!

grampelberg:
@JamyDev back to our conversation about `GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE`, the only way I can look to match the behavior with `localQuery` is to do something like `filter(//foo/bar/baz, kind(go_library, //my_other_thing/...))`. That feels like it is moving in the wrong direction, especially because `filter` doesn't work quite the same as `deps(//foo/bar/baz)` in `packageQuery`.

kev-the-dev:
Your PR summary mentioned my issue #3554... did you mean #3552?

grampelberg:
@kev-the-dev I did, sorry about that!

PR #3526, opened 2023-04-11
config: make strip a bool_flag

### LINKED ISSUE #3325: Unrecognized build option errors when trying to use @io_bazel_rules_go//go/config:strip

0.35.0

0.26.0

5.3.1

Yes

Linux x86-64

### What did you do?

```
❯ bazel build //... --@io_bazel_rules_go//go/config:strip
INFO: Invocation ID: 548294b5-5b52-465c-ad99-b6361c85c0af
INFO: Invocation ID: 548294b5-5b52-465c-ad99-b6361c85c0af
ERROR: Unrecognized option: @io_bazel_rules_go//go/config:strip=true
```

### What did you see instead?

`Unrecognized option` error. Other options appear to work that are documented

malt3:
**What type of PR is this?**

>
Bug fix
> Feature
> Documentation
> Other

Allows specifying the `strip` config setting via a flag. This allows configuration of the strip setting via a bazel command line flag and makes it easier to switch between stripped and non-stripped builds.

Fixes #3325

**Other notes for review**

As far as I can tell, changing this setting to be a bool_flag just works. It does perform stripping as expected[^1]. I couldn't find any reason as to why the strip setting is handled differently from the rest. If there was a reason, please state it and we can close this.

Edit: This is one possible resolution. See also #3527 for an alternative solution that respects the global bazel strip setting.

[^1]: running `file` on a resulting binary when using the flag shows `ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=redacted, not stripped`, while it shows `ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=redacted, with debug_info, not stripped` without the flag. This seems to be consistent with what the docs say about this flag (it is equivalent to `-w` on the `go build` options). I was expecting this to also set `-s` for the go build options which would change the last part of the output to say `stripped`, but this is a separate issue.

### DIFF
--- a/go/config/BUILD.bazel
+++ b/go/config/BUILD.bazel
@@ -1,7 +1,6 @@
 load(
     "@bazel_skylib//rules:common_settings.bzl",
     "bool_flag",
-    "bool_setting",
     "string_flag",
     "string_list_flag",
 )
@@ -34,7 +33,7 @@ bool_flag(
     visibility = ["//visibility:public"],
 )
 
-bool_setting(
+bool_flag(
     name = "strip",
     build_setting_default = False,
     visibility = ["//visibility:public"],


fmeum:
I do see one reason: Bazel's `--strip` flag should also control Go stripping and, compared to the Go-specific flag, ensures that stripping is consistent across languages.

Did you check whether this works? Do you have a use case which for stripping Go but not other languages?

malt3:
> I do see one reason: Bazel's `--strip` flag should also control Go stripping [...].
> Did you check whether this works? Do you have a use case which for stripping Go but not other languages?

I added this to my .bazelrc: `build --strip=always` and the file command still says `with debug_info` if I use rules_go from main. Is this maybe a bug or an incorrect use on my side?

malt3:
Superseded by #3527

PR #3528, opened 2023-04-11
Add defines for assembly build

### LINKED ISSUE #1894: Define GOOS_<goos> and GOARCH_<goarch> for assembly
`go build` does this. We should, too.

Example:

```
/opt/go1.12beta1/pkg/tool/linux_amd64/asm -trimpath $WORK/b001 -I $WORK/b001/ -I /opt/go1.12beta1/pkg/include -D GOOS_linux -D GOARCH_amd64 -o $WORK/b001/foo.o ./foo.s
```

patrickmscott:
Fixes https://github.com/bazelbuild/rules_go/issues/1894 by adding GOOS_ GOARCH_ and GOOS_GOARCH_ defines for building assembly.

**What type of PR is this?**

Bug fix

This adds a few definitions to the assembly build that are used by several projects.

Fixes #1894 

**Other notes for review**


go/tools/builders/asm.go
```diff
@@ -107,13 +106,22 @@ func asmFile(goenv *env, srcPath, packagePath string, asmFlags []string, outPath
...
 	}
+	asmDefines(args)
 	args = append(args, "-trimpath", ".")
 	args = append(args, "-o", outPath)
 	args = append(args, "--", srcPath)
 	absArgs(args, []string{"-I", "-o", "-trimpath"})
 	return goenv.runCommand(args)
 }
 
+func asmDefines(args []string) {
```
fmeum, line 117: You can't (reliably) mutate a slice in a function by passing it in as a parameter, you need to pass it back. Here it would be simpler to just return the three-element slice and append it at the call site.

patrickmscott: Updated to return a copy

go/tools/builders/asm.go
```diff
@@ -107,13 +106,22 @@ func asmFile(goenv *env, srcPath, packagePath string, asmFlags []string, outPath
...
 	args = append(args, "-trimpath", ".")
 	args = append(args, "-o", outPath)
 	args = append(args, "--", srcPath)
 	absArgs(args, []string{"-I", "-o", "-trimpath"})
 	return goenv.runCommand(args)
 }
 
+func asmDefines(args []string) []string {
+	return append(args,
+		"-D", "GOOS_"+runtime.GOOS,
```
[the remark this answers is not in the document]
patrickmscott: You are right. I dug more into how the go tool does this and they use `build.Default.GOOS` and friends. Updated and added a compilation test to verify.

fmeum (approved):
Looks great, thanks for adding a test!

patrickmscott:
Looks like the build continues to fail on xcode. Any ideas?

fmeum:
Pretty sure that failure is caused by https://github.com/bazelbuild/continuous-integration/commit/a95a916098d3015bb4ea20b7e33bc7d27d00bffc, I don't think it has anything to do with your PR. Will try to fix it tomorrow.

PR #3530, opened 2023-04-13
Fix macOS CI after `--remote_download_minimal` flip

fmeum:
This was flipped in bazelbuild/continuous-integration#1579, but breaks `//tests/core/cgo:versioned_dylib_test`.


fmeum:
FYI @achew22 @linzhp I am going to merge this now to get CI unstuck.

PR #3539, opened 2023-04-18
Re-enable go_path_test on RBE

sluongng:
**What type of PR is this?**

>
Bug fix

Re-enable GoPath action to test against RBE now that symlinks are handled correctly.

Fixes #

**Other notes for review**

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -90,7 +90,6 @@ tasks:
     test_targets:
     - "--"
     - "//..."
-    - "-//tests/core/go_path:go_path_test"
     - "-//tests/core/stdlib:buildid_test"
   windows:
     build_flags:


sluongng:
@fmeum this change is realtively straight forward. Since you touched symlink code on Bazel side, could you please give it a quick look?

fmeum:
@sluongng We could add a feature for `declare_symlink` being available by default to `bazel_features` and start using it if it is available. What do you think?

fmeum:
@sluongng That warning should definitely go away, the feature is stable - toggling it off is what may not work as expected at this point. The issue you linked isn't directly related, it's about `ctx.actions.symlink`, not `ctx.actions.declare_symlink`.

sluongng:
Ah I was confusing the 2 😵‍💫 

Yup, happy to depend on https://github.com/bazel-contrib/bazel_features and switch to `ctx.actions.declare_symlink` when it's available.

sluongng:
Sent https://github.com/bazelbuild/bazel/pull/18313

PR #3543, opened 2023-04-24
go_sdk extension: create `go_host_compatible_sdk` repository

ylecornec:
**What type of PR is this?**
Feature

**What does this PR do? Why is it needed?**

As discussed in [this issue](https://github.com/bazelbuild/bazel-gazelle/issues/1469), this PR makes the `go_sdk` module extension define a new repository called `go_host_compatible_sdk`.

This repository exposes the label of the first host compatible sdk declared by a module via the `download` or `host` tag (according to bazel's iteration order over modules).

The gazelle module could then rely on this toolchain for bootstrapping instead of always using `go_default_sdk`.

**Which issues(s) does this PR fix?**

This PR is a first step toward closing https://github.com/bazelbuild/bazel-gazelle/issues/1469

**Other notes for review**

- As discussed in the issue, this PR does not yet provide a way do use toolchains declared outside of rules_go. This is left for another PR.

PR #3546, opened 2023-04-25
Add `custom` tag to `go_sdk`

ylecornec:
Depends on #3543

**What type of PR is this?**

Feature

This PR adds a new `custom` tag to the `go_sdk` module extension.
This tag is used to declare custom go toolchains to rules_go, to take them into consideration when creating the `go_host_compatible_sdk` repository introduced in #3543.

See https://github.com/bazelbuild/bazel-gazelle/issues/1469

**Other notes for review**

- The attribute of the `custom` tag is not directly the label of a toolchain, but the label of a file containing a `toolchain` variable. This way the file can be generated by the user in a repository rule (and the `toolchain` variable can be set to `None` to be deactivated if some conditions are not met). 

- Custom toolchains have precedence over other ones (defined in the same module) because the user can deactivate them if needed.

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -1,5 +1,23 @@
-load("//go/private:sdk.bzl", "go_download_sdk_rule", "go_host_sdk_rule", "go_multiple_toolchains")
+load("//go/private:sdk.bzl", "detect_host_platform", "go_download_sdk_rule", "go_host_sdk_rule", "go_multiple_toolchains")
 load("//go/private:repositories.bzl", "go_rules_dependencies")
+
+def host_compatible_toolchain_impl(ctx):
+    ctx.file("BUILD.bazel", content = "")
+    host_compatible_sdk = ctx.attr.toolchains[0]
+    ctx.file("defs.bzl", content = """
+host_compatible_sdk = Label("{}")
+""".format(host_compatible_sdk))
+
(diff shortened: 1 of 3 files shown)


fmeum:
Thanks for performing the split.

This adds sufficient complexity that I would first want to get back to the formulating the exact use case(s) we are trying to address with it. Non-Bzlmod rules_go offers `go_wrap_sdk`, which I think may help us here as well. 

@ylecornec If we had e.g. `go_sdk.wrap(root_files = {"linux_amd64": "@my_nix_go_sdk//:ROOT")`, would that adress your use case? If not, could you provide more details including perhaps some code snippets to explain how you fetch your SDKs?

ylecornec:
Ah, something like `go_sdk.wrap` is close, but to replicate the setup of non bzlmod [rules_haskell](https://github.com/tweag/rules_haskell), the `@my_nix_go_sdk//:ROOT` value would need to depend on some environment variable (to test whether we are in a nix-shell or not).

[Here](https://github.com/ylecornec/gazelle_nixos_repro) is an example use-case that replicates this with bzlmod (using the `custom` tag) and has more details in the README.
(The nix toolchains are defined in `non_module_deps.bzl` using [nixpkgs_go_configure](https://github.com/tweag/rules_nixpkgs/blob/master/toolchains/go/README.md#nixpkgs_go_configure)).

PR #3551, opened 2023-04-28
Move CI flags to .bazelrc and fix BwoB build

fmeum:
This simplifies the CI config and also ensures that the recent BwoB flip in CI is disabled in all runs.


fmeum:
I expect this to fix the failures in the Bazel@HEAD downstream pipeline.

linzhp (approved):

achew22:
Thank you for doing this! This is a big win that we should have done a long time ago

PR #3552, opened 2023-05-02
stdliblist: return generated files

### LINKED ISSUE #3523: `gopackagesdriver` doesn't work with golangci-lint

grampelberg:
**What type of PR is this?**

Bug fix

The stdlib json currently is not populated with generated files. When CGO is being used, this means that the file list returned does not actually compile. This change adds `-compiled=true` to go list's output which results in the full set required to compile stdlib.

Fixes #3523

**Other notes for review**

See #3524 for the companion PR.

### DIFF
--- a/go/private/actions/stdlib.bzl
+++ b/go/private/actions/stdlib.bzl
@@ -68,15 +68,45 @@ def _build_stdlib_list_json(go):
     args = go.builder_args(go, "stdliblist")
     args.add("-sdk", go.sdk.root_file.dirname)
     args.add("-out", out)
+
+    inputs = go.sdk_files
+    if not go.mode.pure:
+        inputs += go.crosstool
+
     go.actions.run(
-        inputs = go.sdk_files,
+        inputs = inputs,
         outputs = [out],
         mnemonic = "GoStdlibList",
         executable = go.toolchain._builder,
         arguments = [args],
-        env = go.env,
+        env = _build_env(go),
     )
     return out
+
+def _build_env(go):
+    env = go.env
+
+    if go.mode.pure:
+        env.update({"CGO_ENABLED": "0"})
+        return env
+
+    # NOTE(#2545): avoid unnecessary dynamic link
+    # go std library doesn't use C++, so should not have -lstdc++
+    # Also drop coverage flags as nothing in the stdlib is compiled with
+    # coverage - we disable it for all CGo code anyway.
+    ldflags = [
+        option
+        for option in extldflags_from_cc_toolchain(go)
+        if option not in ("-lstdc++", "-lc++") and option not in COVERAGE_OPTIONS_DENYLIST
+    ]
+    env.update({
+        "CGO_ENABLED": "1",
+        "CC": go.cgo_tools.c_compiler_path,
+        "CGO_CFLAGS": " ".join(go.cgo_tools.c_compile_options),
+        "CGO_LDFLAGS": " ".join(ldflags),
+    })
+
+    return env
 
 def _sdk_stdlib(go):
     return GoStdLib(
@@ -96,26 +126,9 @@ def _build_stdlib(go):
     if not go.mode.pure:
         args.add("-package", "runtime/cgo")
     args.add_all(link_mode_args(go.mode))
-    env = go.env
-    if go.mode.pure:
-        env.update({"CGO_ENABLED": "0"})
-    else:
-        # NOTE(#2545): avoid unnecessary dynamic link
-        # go std library doesn't use C++, so should not have -lstdc++
-        # Also drop coverage flags as nothing in the stdlib is compiled with
-        # coverage - we disable it for all CGo code anyway.
-        ldflags = [
-            option
-            for option in extldflags_from_cc_toolchain(go)
-            if option not in ("-lstdc++", "-lc++") and option not in COVERAGE_OPTIONS_DENYLIST
-        ]
-        env.update({
-            "CGO_ENABLED": "1",
-            "CC": go.cgo_tools.c_compiler_path,
-            "CGO_CFLAGS": " ".join(go.cgo_tools.c_compile_options),
-            "CGO_LDFLAGS": " ".join(ldflags),
-        })
+
     args.add("-gcflags", quote_opts(go.mode.gc_goopts))
+
     inputs = (go.sdk.srcs +
               go.sdk.headers +
               go.sdk.tools +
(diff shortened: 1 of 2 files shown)


go/private/actions/stdlib.bzl
```diff
@@ -63,24 +63,46 @@ def _should_use_sdk_stdlib(go):
...
+        "CGO_CFLAGS": " ".join(go.cgo_tools.c_compile_options),
+        "CGO_LDFLAGS": " ".join(ldflags),
+    })
+
+    return env
+
 def _sdk_stdlib(go):
     return GoStdLib(
-        _list_json = _build_stdlib_list_json(go),
+        _list_json = _build_stdlib_list_json(go, _cgo_env(go)),
```
linzhp, line 105: This means we would enable cgo unconditionally here, even for targets built with cgo disabled. Can you gate on `go.mode.pure`?

go/private/actions/stdlib.bzl
```diff
@@ -63,24 +63,46 @@ def _should_use_sdk_stdlib(go):
...
 
-def _build_stdlib_list_json(go):
+def _build_stdlib_list_json(go, env):
     out = go.declare_file(go, "stdlib.pkg.json")
     args = go.builder_args(go, "stdliblist")
     args.add("-sdk", go.sdk.root_file.dirname)
     args.add("-out", out)
+
     go.actions.run(
         inputs = go.sdk_files,
```
linzhp, line 73: If cgo is enabled, you also need to pass `go.crosstool` as input

go/tools/builders/stdliblist.go
```diff
@@ -116,20 +116,45 @@ func outputBasePath(cloneBase, p string) string {
...
+		ret = append(ret, absPath)
 	}
 	return ret
 }
 
+// filterGoFiles keeps only files either ending in .go or those without an
+// extension (which are from the cache). This is a work around for
+// https://golang.org/issue/28749: cmd/go puts assembly, C, and C++ files in
+// CompiledGoFiles.
+func filterGoFiles(srcs []string) []string {
```
linzhp, line 142: Compiled Go files don't have extension either. So this essentially removed all compiled Go files. If we are going to do this, do we still need to enable cgo and `-compiled=true` flag?

go/private/actions/stdlib.bzl
linzhp: I was going to help set up a hermetic cc toolchain for you to reproduce the error we see internally, but found that the CI of this PR is alway failing with the same issue

JamyDev:
Here's a repro that repro the c++ not found, although this one isn't crashing with an NPE: https://github.com/JamyDev/go-bazel-sample/tree/jamy/zig-cc

go/private/actions/stdlib.bzl
JamyDev: This suggestion fixes the c++ issue!

JamyDev:
We found the root cause and solution to the NPE. Will have to discuss a plan to fix this in the grander scheme, but this diff is okay from a functional perspective (assuming the c++ issue gets fixed)

go/private/actions/stdlib.bzl
grampelberg: Isn't the SDK always built with CGO?

grampelberg:
@JamyDev `go.crosstool` is all it took! Thank you =)

go/tools/builders/stdliblist.go
grampelberg: This keeps all the go files in `CompiledGoFiles` that either end in `.go` or have no extension. I actually copypasta'd this from the `go list` codebase.

go/private/actions/stdlib.bzl
linzhp: Also, `_build_stdlib()` sets `CGO_ENABLED=0` when `go.mode.pure`. To be consistent, this function should also set that.

go/private/actions/stdlib.bzl
```diff
@@ -63,24 +63,46 @@ def _should_use_sdk_stdlib(go):
...
-def _build_stdlib_list_json(go):
+def _build_stdlib_list_json(go, env):
     out = go.declare_file(go, "stdlib.pkg.json")
     args = go.builder_args(go, "stdliblist")
     args.add("-sdk", go.sdk.root_file.dirname)
     args.add("-out", out)
+
     go.actions.run(
-        inputs = go.sdk_files,
+        inputs = go.sdk_files + go.crosstool,
```
linzhp, line 73: `go.crosstool` is only necessary when cgo is enabled.

go/private/actions/stdlib.bzl
grampelberg: Ahhha! That explains a lot for me. Thank you =) I'll get the `go.mod.pure` gate in place!

go/tools/builders/stdliblist.go
```diff
@@ -139,13 +164,31 @@ func flatPackageForStd(cloneBase string, pkg *goListPackage) *flatPackage {
...
 		GoFiles:         goFiles,
-		CompiledGoFiles: goFiles,
+		CompiledGoFiles: filterGoFiles(compiledGoFiles),
+	}
+
+	// imports
+	//
+	// Imports contains the IDs of all imported packages.
+	// ImportsMap records (path, ID) only where they differ.
+	ids := make(map[string]bool)
```
linzhp, line 174:
```suggestion
	ids := make(map[string]struct{})
```
so people don't need to speculate what true or false means

go/tools/builders/stdliblist.go
```diff
@@ -116,20 +116,45 @@ func outputBasePath(cloneBase, p string) string {
...
+// https://golang.org/issue/28749: cmd/go puts assembly, C, and C++ files in
+// CompiledGoFiles.
+func filterGoFiles(srcs []string) []string {
+	ret := make([]string, 0, len(srcs))
+	for _, f := range srcs {
+		if ext := filepath.Ext(f); ext != ".go" && ext != "" {
+			continue
+		}
+
+		ret = append(ret, f)
```
linzhp, line 149:
```suggestion
		if ext := filepath.Ext(f); ext == ".go" || ext == "" {
			ret = append(ret, f)
		}
```

slightly simpler

linzhp (approved):
a few nit comments. looks good to me otherwise

grampelberg:
@linzhp great suggestions, thank you!

PR #3559, opened 2023-05-16
Revert "Re-enable go_path_test on RBE (#3539)"

sluongng:
This reverts commit f6be5e045c95a681989749a6e054c0c2fa91746e.

**What type of PR is this?**

>
Bug fix
> Feature
> Documentation
> Other

Disable go_path test on CI.

It seems that current way Bazel CI setup make it hard for symlink tests to work

**Other notes for review**

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -77,6 +77,7 @@ tasks:
     test_targets:
     - "--"
     - "//..."
+    - "-//tests/core/go_path:go_path_test"
     - "-//tests/core/stdlib:buildid_test"
   windows:
     build_flags:


sluongng:
@fmeum it seems like the issue was not caused by PR #3559.

Instead, it is caused by Bazel CI enabling `--remote_download_minimal` unconditionally even when RBE is not used.

The build failure seems to be `20.04 LTS` that are supposed to execute locally, not remotely. The fix in 841933051 was not effective.

Looking a bit deeper into this, the error seems to be applicable for `download = toplevel` as well. See https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java;l=1330-1332;drc=a5dde12b0d945b54967969fafa4d5e40692f9a44

and toplevel config seems to be set in https://github.com/bazelbuild/continuous-integration/blob/a7c6d0a1387e6a0d6a8eaafbb14ccf78e7973da2/buildkite/bazelci.py#L1899

fmeum:
@sluongng The PR you linked sets `--remote_download_outputs=all`, which I would have assumed fixes this. Did I apply it to the wrong pipeline? We could also add this flag to rules_go's `.bazelrc` to ensure it's always picked up.

sluongng:
The build log is like this

```
bazel build ... --remote_download_toplevel ... -- //...
...
(10:44:38) INFO: Reading rc options for 'build' from /workdir/.bazelrc:
  'build' options: --remote_download_outputs=all --build_runfile_links --experimental_remote_download_regex=.*tests/core/go_path/.*
```

So the flags given at command line override the flag written to `.bazelrc`.
In particular, `--remote_download_toplevel` get higher priority here and override your setting.

The fix is to either exclude go_path from all CI builds, or to put an option into https://github.com/bazelbuild/continuous-integration/ setup to allow picking remote download mode.

PR #3562, opened 2023-05-17
Implement //nolint parsing similar to golangci-lint

### LINKED ISSUE #3561: Parse nolint comments like golangci-lint within nogo
nogo has configuration that allows filtering files but has no way to annotate specific errors with comments like `//nolint`. golangci-lint can collect those comments and filters out the issues.

I think nogo should do the same, following the same format to filter out specific errors where requested. I already have a PR ready to implement the change.

patrickmscott:
https://github.com/bazelbuild/rules_go/issues/3561

Look for nolint comments and collect the ranges of where they apply. If a node immediately follows a range with the same column, expand the range to include the node.

Add tests that verify errors are filtered out.

**What type of PR is this?**

Feature

Parses `//nolint` comments and filters out errors from nogo.

Fixes #3561 

**Other notes for review**

### DIFF
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -202,6 +202,55 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
 		act.pkg = pkg
 	}
 
+	// Process nolint directives similar to golangci-lint.
+	for _, f := range pkg.syntax {
+		// First, gather all comments and apply a range to the entire comment block.
+		// This will cover inline comments and allow us to detect comments above a
+		// node later.
+		for _, group := range f.Comments {
+			for _, comm := range group.List {
+				linters, ok := parseNolint(comm.Text)
+				if !ok {
+					continue
+				}
+				nl := &Range{
+					from: pkg.fset.Position(group.Pos()),
+					to:   pkg.fset.Position(group.End()).Line,
+				}
+				for analyzer, act := range actions {
+					if linters.Contains(analyzer.Name) {
+						act.nolint = append(act.nolint, nl)
+					}
+				}
+			}
+		}
+
+		// For each node in the ast, check if the previous line is covered by a
+		// range for the linter and expand it.
+		ast.Inspect(f, func(n ast.Node) bool {
+			if n == nil {
+				return true
+			}
+
+			nodeStart := pkg.fset.Position(n.Pos())
+			nodeEnd := pkg.fset.Position(n.End())
+			for _, act := range actions {
+				for _, rng := range act.nolint {
+					if rng.from.Filename != nodeStart.Filename {
+						continue
+					}
+					if rng.to != nodeStart.Line-1 || rng.from.Column != nodeStart.Column {
+						continue
+					}
+					// Expand the range to cover this node
+					rng.to = nodeEnd.Line
+				}
+			}
+
+			return true
+		})
+	}
+
 	// Execute the analyzers.
 	execAll(roots)
 
@@ -209,6 +258,11 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
 	diagnostics := checkAnalysisResults(roots, pkg)
 	facts := pkg.facts.Encode()
 	return diagnostics, facts, nil
+}
+
+type Range struct {
+	from token.Position
+	to   int
 }
 
 // An action represents one unit of analysis work: the application of
@@ -226,6 +280,7 @@ type action struct {
 	diagnostics []analysis.Diagnostic
 	usesFacts   bool
 	err         error
+	nolint      []*Range
 }
 
 func (act *action) String() string {
@@ -279,13 +334,26 @@ func (act *action) execOnce() {
 		factFilter[reflect.TypeOf(f)] = true
 	}
 	pass := &analysis.Pass{
-		Analyzer:          act.a,
-		Fset:              act.pkg.fset,
-		Files:             act.pkg.syntax,
-		Pkg:               act.pkg.types,
-		TypesInfo:         act.pkg.typesInfo,
-		ResultOf:          inputs,
-		Report:            func(d analysis.Diagnostic) { act.diagnostics = append(act.diagnostics, d) },
+		Analyzer:  act.a,
+		Fset:      act.pkg.fset,
+		Files:     act.pkg.syntax,
+		Pkg:       act.pkg.types,
+		TypesInfo: act.pkg.typesInfo,
+		ResultOf:  inputs,
+		Report: func(d analysis.Diagnostic) {
+			pos := act.pkg.fset.Position(d.Pos)
+			for _, rng := range act.nolint {
+				if pos.Filename != rng.from.Filename {
+					continue
+				}
+				if pos.Line < rng.from.Line || pos.Line > rng.to {
+					continue
+				}
+				// Found a nolint range. Ignore the issue.
+				return
+			}
+			act.diagnostics = append(act.diagnostics, d)
+		},
 		ImportPackageFact: act.pkg.facts.ImportPackageFact,
 		ExportPackageFact: act.pkg.facts.ExportPackageFact,
 		ImportObjectFact:  act.pkg.facts.ImportObjectFact,
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -204,61 +204,29 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
 
 	// Process nolint directives similar to golangci-lint.
 	for _, f := range pkg.syntax {
-		// First, gather all comments and apply a range to the entire comment block.
-		// This will cover inline comments and allow us to detect comments above a
-		// node later.
-		for _, group := range f.Comments {
-			for _, comm := range group.List {
-				linters, ok := parseNolint(comm.Text)
-				if !ok {
-					continue
-				}
-				nl := &Range{
-					from: pkg.fset.Position(group.Pos()),
-					to:   pkg.fset.Position(group.End()).Line,
-				}
-				for analyzer, act := range actions {
-					if linters == nil || linters[analyzer.Name] {
-						act.nolint = append(act.nolint, nl)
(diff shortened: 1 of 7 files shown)


patrickmscott:
Test failures appear unrelated to the change

go/tools/builders/nolint.go
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import "strings"
+
+type Linters []string
+
+func (l Linters) Contains(s string) bool {
```
fmeum, line 21: How many different linters does a typical project use? I am wondering whether a `map[string]bool` would be more efficient (and whether that difference actually matters).

fmeum (commented):
This looks good to me at first glance, but I am not familiar enough with nogo to review this more deeply.

@sluongng Could you take a look?

go/tools/builders/nolint.go
patrickmscott: I use about 40 in my project and did not see a noticeable performance issue.

go/tools/builders/nogo_main.go
```diff
@@ -202,6 +202,55 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
...
+				linters, ok := parseNolint(comm.Text)
+				if !ok {
+					continue
+				}
+				nl := &Range{
+					from: pkg.fset.Position(group.Pos()),
+					to:   pkg.fset.Position(group.End()).Line,
+				}
+				for analyzer, act := range actions {
+					if linters.Contains(analyzer.Name) {
```
sluongng, line 221: could `linters` be nil here?

go/tools/builders/nogo_main.go
```diff
@@ -202,6 +202,55 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
...
+						act.nolint = append(act.nolint, nl)
+					}
+				}
+			}
+		}
+
+		// For each node in the ast, check if the previous line is covered by a
+		// range for the linter and expand it.
+		ast.Inspect(f, func(n ast.Node) bool {
```
sluongng, line 230: I am not sure why we are checking the previous line of an AST node and not the first line. I think future maintainers might be confused as well.

It's best if you could provide a separate test(s) that would help demonstrate the range of what would be covered and NOT covered when a `//nolint` comment is made.

go/tools/builders/nogo_main.go
```diff
@@ -279,13 +334,26 @@ func (act *action) execOnce() {
...
+					continue
+				}
+				if pos.Line < rng.from.Line || pos.Line > rng.to {
+					continue
+				}
+				// Found a nolint range. Ignore the issue.
+				return
+			}
+			act.diagnostics = append(act.diagnostics, d)
+		},
```
sluongng, line 356: Nit, now that it's no longer an one-lined, could you please move the func out to above struct?

```
ignoreNoLintReporter := func(...) {
  ...
}
pass := &analysis.Pass{
  ...
  Report: ignoreNoLintReporter,
  ...
}
```

go/tools/builders/nolint.go
sluongng: This is not scale against the analyzers inside a project, it scales against the `//nolint:<comma separated list of string>` so we don't expect too bad of a search time here, as a typical user should not be ignoring a billion linters in a directive comment.

With that said, I think it would be cleaner/better if `Linters` is constructed as a Set: `map[string]struct{}` because we will be performing this lookup inside 4-5 levels of nested loops for "each compilation action". It would also save us from needing a type alias as it's very straight forward to do a key lookup in a map.

tests/core/nogo/nolint/nolint_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+func InlineComment() {
+	s := "hello" // nolint
+	fmt.Printf("%d", s)
+}
+
+func LinterMatch() bool {
+	return true || true //nolint:printf
+}
+`,
```
sluongng, line 75: I would much prefer if you could split this test into smaller tests, each have 1-2 `//nolint` directive demonstrating different behaviors.

go/tools/builders/nogo_main.go
```diff
@@ -202,6 +202,55 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
...
+				return true
+			}
+
+			nodeStart := pkg.fset.Position(n.Pos())
+			nodeEnd := pkg.fset.Position(n.End())
+			for _, act := range actions {
+				for _, rng := range act.nolint {
+					if rng.from.Filename != nodeStart.Filename {
+						continue
+					}
```
sluongng, line 241: In which scenarios does the File name differ here?

I thought we are still in the same `for _, f := range pkg.syntax {` loop where `f == file`? 🤔
So these should all have the same file name no?

go/tools/builders/nogo_main.go
```diff
@@ -211,6 +260,11 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
 	return diagnostics, facts, nil
 }
 
+type Range struct {
+	from token.Position
+	to   int
+}
```
sluongng, line 266: I thought we only care about line numbers here?

It's unclear to me in what scenarios would `from.Filename` and `from.Column` would be helpful. It would be nice to have those scenarios documented as a test.

sluongng (commented):
This is great work @patrickmscott!! Damn impressive.

I have a small brain problem where I don't understand all of these though.
I think splitting these into 2 smaller PRs might be easier for reviewers to consume:
1. Introduce `//nolint` + relevant tests
2. Allow `//nolint: linterA,linterB` + relevant tests

You would also want to update `nogo` documentation to highlight what do we support.
Keep in mind that down in each analyzer, they could also have there own `ignore lint error` directive, so we want to highlight that for future users of this feature to be aware of.

go/tools/builders/nogo_main.go
patrickmscott: Yes, to indicate _all_ linters. `Contains` handles that case and always returns true.

go/tools/builders/nogo_main.go
patrickmscott: I'm doing a lot of same logic as golangci-lint but did not provide a lot of context here.

The way I reasoned about this is there are 2 cases we are trying to handle:
1. Inline comments - these apply to the same line as the ast node
2. Comments above the block they apply to - this is the case handled here

The idea is that something like
```
//nolint
complex(stmt.Func())
```
Will apply the `nolint` directive to everything within the following line

A more complex example:
```
//nolint
return Struct{
  foo: bar,
}, nil
```

The complex example should cover any issues that arise in the _entire_ return block. I can add more tests to show these examples.

tests/core/nogo/nolint/nolint_test.go
patrickmscott: Will do

go/tools/builders/nogo_main.go
patrickmscott: The ranges are built for the entire package across multiple files. I didn't want a range for one file to apply to an error in a different file with a similar line number.

go/tools/builders/nogo_main.go
patrickmscott: The Filename is because the ranges are built for the entire package. The column helps me associate a comment that applies to the immediately following block. I suppose, in theory, an inline comment could have the same column as a node on the next line, causing the range expand. I think golangci-lint as the same issue though.

patrickmscott:
@sluongng any update? The test failure appears unrelated to my change.

go/tools/builders/nolint_test.go
```diff
@@ -0,0 +1,47 @@
...
+	assert(
+		"// nolint:foo,bar,baz",
+		map[string]bool{
+			"foo": true,
+			"bar": true,
+			"baz": true,
+		},
+		true,
+	)
+}
```
sluongng, line 47: Nit

This should be rewritten as Go's test table convention. https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

In particular, I think it should be written as

```
tests := struct{ ...<fields>... }{ <entries> }
for _, tc := tests {
  t.Run(tc.Name, func(t *testing.T) { ... })
}
```

See https://github.com/bazelbuild/rules_go/blob/f5aa81cecbaf966fb9f749d4cd653ff16c9a7003/go/tools/bzltestutil/lcov_test.go#L57 for example

tests/core/nogo/nolint/nolint_test.go
```diff
@@ -0,0 +1,202 @@
...
+	// Execute tests that should filter out errors
+	runTest(t, "//:inline", "")
+	runTest(t, "//:inline_filter", "")
+	runTest(t, "//:block", "")
+	runTest(t, "//:block_multiline", "")
+
+	// Execute tests that should return unfiltered errors
+	runTest(t, "//:inline_errors", "inline_errors.go:9:15: nil dereference in load (nilness)")
+	runTest(t, "//:inline_column", "inline_column.go:8:27: redundant or: true || true (bools)")
+	runTest(t, "//:large_block", "large_block.go:10:5: redundant or: true || true (bools)")
```
sluongng, line 192: nit: use test table convention.

tests/core/nogo/nolint/nolint_test.go
```diff
@@ -0,0 +1,202 @@
...
+	})
+}
+
+func runTest(t *testing.T, target string, expected string) {
+	t.Helper()
+
+	cmd := bazel_testing.BazelCmd("build", target)
+	stderr := &bytes.Buffer{}
+	cmd.Stderr = stderr
+	err := cmd.Run()
```
sluongng, line 164: nit: `cmd.Output()` or `cmd.CombinedOutput()` could be more suitable.

sluongng (approved):
This looks great! I have a few style nits that I think should reduce future churns, but overall this looks good. We should be able to iterate from here based on the nolint_test setup.

Passing this PR to @fmeum and @linzhp for further review.

patrickmscott:
@fmeum or @linzhp can you merge?

PR #3566, opened 2023-05-19
Create empty `.go` file in a random location 

### LINKED ISSUE #3558: Repeated issues with unable to create directory for _empty.go on Windows
v0.39.0

v0.29.0

bazel 6.1.0

Yes. tried 0.39.1 of rules, bazel 6.2.0, 0.30.0 of gazelle.

Windows x64

### What did you do?
Just trying to build golang

That it works

### What did you see instead?
Many golang targets (almost always tests, don't know whether that is significant) fail with errors like:
```
compilepkg: could not create directory for _empty.go: mkdir C:\users\me\_bazel_me\dotzqxa5\execroot\altr_sw_hub\bazel-out\x64_windows-fastbuild\bin\golang\pkg\auth\auth\go_altair_com_slchub_pkg_auth_auth: Cannot create a file when that file already exists.
```

There is an issue elsewhere that reports that this only happens when there are multiple `go_test` targets in the BUILD.bazel file. That isn't the case here, or at least not in all cases (it is in a couple).

It also happens when running with `--jobs=1`, though I haven't run a completely clean build like that. I ran bazel with `--keep_going` to try and run all targets that would succeed and then tried running with `--jobs=1` to see whether the rest would succeed. 
I could get it to run using `--jobs=1` and then each time it failed, deleting the directory it was trying to create. After a few rounds of that the build succeeded.

I have repeated the test by first manually deleting the "execroot" output directory to force a completely clean build and I get the same thing.

The curious thing is that looking at the source code for the rules, I don't really understand why it's even trying to make an `_empty.go` file. The appears to be that it is only created if `len(srcs.goSrcs) == 0`. But all of my rules have lists of go files, and none of the files have golang build conditions in them, so it ought to have real code to build.

fmeum:
**What type of PR is this?**

Bug fix

Isolating the empty `.go` files generated for targets without sources
and removing them after a build turned out not to solve issues with
concurrent unsandboxed builds causing races. Instead, just generate the
file in a temporary location, which for a truly empty file does not
result in (non-hermetic) source file paths being included in the
archive.

Along the way fix a potential source of non-hermeticity in
`go_bazel_test`.

Fixes #3558

**Other notes for review**

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -206,33 +206,31 @@ func compileArchive(
 	nogoSrcsOrigin := make(map[string]string)
 
 	if len(srcs.goSrcs) == 0 {
-		// We need to run the compiler to create a valid archive, even if there's
-		// nothing in it. GoPack will complain if we try to add assembly or cgo
-		// objects.
-		//
-		// _empty.go needs to be in a deterministic location (not tmpdir) in order
-		// to ensure deterministic output. The location also needs to be unique
-		// otherwise platforms without sandbox support may race to create/remove
-		// the file during parallel compilation.
-		emptyDir := filepath.Join(filepath.Dir(outPath), sanitizePathForIdentifier(importPath))
-		if err := os.Mkdir(emptyDir, 0o700); err != nil {
-			return fmt.Errorf("could not create directory for _empty.go: %v", err)
+		// We need to run the compiler to create a valid archive, even if there's nothing in it.
+		// Otherwise, GoPack will complain if we try to add assembly or cgo objects.
+		// A truly empty archive does not include any references to source file paths, which
+		// ensures hermeticity even though the temp file path is random.
+		emptyGoFile, err := os.CreateTemp(filepath.Dir(outPath), "*.go")
+		if err != nil {
+			return err
 		}
-		defer os.RemoveAll(emptyDir)
-
-		emptyPath := filepath.Join(emptyDir, "_empty.go")
-		if err := os.WriteFile(emptyPath, []byte("package empty\n"), 0o666); err != nil {
+		defer os.Remove(emptyGoFile.Name())
+		defer emptyGoFile.Close()
+		if _, err := emptyGoFile.WriteString("package empty\n"); err != nil {
+			return err
+		}
+		if err := emptyGoFile.Close(); err != nil {
 			return err
 		}
 
 		srcs.goSrcs = append(srcs.goSrcs, fileInfo{
-			filename: emptyPath,
+			filename: emptyGoFile.Name(),
 			ext:      goExt,
 			matched:  true,
 			pkg:      "empty",
 		})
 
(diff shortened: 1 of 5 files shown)


fmeum:
cc @sluongng

linzhp (approved):
Tests passed in Uber

go/tools/builders/compilepkg.go
```diff
@@ -206,33 +206,31 @@ func compileArchive(
...
+		emptyGoFile, err := os.CreateTemp(filepath.Dir(outPath), "*.go")
+		if err != nil {
+			return err
 		}
-		defer os.RemoveAll(emptyDir)
-
-		emptyPath := filepath.Join(emptyDir, "_empty.go")
-		if err := os.WriteFile(emptyPath, []byte("package empty\n"), 0o666); err != nil {
+		defer os.Remove(emptyGoFile.Name())
+		defer emptyGoFile.Close()
```
sluongng, line 218: Nit: it would be a bit easier to read if we combine these 2 defer into 1 single `defer func() {}`.
It's a bit hard to read here whether we are intentionally rm-ing an opened file handle or not.

Then we could also add err check for the `os.Remove()` call.

go/tools/builders/compilepkg.go
```diff
@@ -206,33 +206,31 @@ func compileArchive(
...
+		if _, err := emptyGoFile.WriteString("package empty\n"); err != nil {
+			return err
+		}
+		if err := emptyGoFile.Close(); err != nil {
```
sluongng, line 222: So we are closing the same file... twice. And we ignore the error happening when we close the second time inside the defer.

tests/core/go_library/no_srcs_test.go
```diff
@@ -0,0 +1,78 @@
...
+    for i in range(1000)
+]
+`,
+	})
+}
+
+func Test(t *testing.T) {
+	commonArgs := []string{
+		"--spawn_strategy=local",
+		"--compilation_mode=dbg",
```
sluongng, line 47: any particular reason why `dbg` was needed here?

sluongng (commented):
Thanks a ton for fixing this. I did not know that using a random path would still ensure the hermeticity for the action the last time I touch this code path.

I have some minor nits but love the overall solution 👍

tests/core/go_library/no_srcs_test.go
fmeum: It's not needed, but I wanted to ensure that the output does include all debug symbols, even if we move to not generate them in other compilation modes in the future.

go/tools/builders/compilepkg.go
fmeum: `defer func() {}` would definitely be more readable. But what would we do if `os.Remove()` or `Close()` return an error? This is part of best-effort cleanup, so I don't think we would want to fail compilation over it.

PR #3569, opened 2023-05-23
Pass `-trimpath` to `cgo` in `stdlib` action

fmeum:
When `go install` invokes `cgo`, we need to pass `-trimpath`, but there doesn't seem to be a way to influence the command line. Instead, use `filterbuildid` to inject the argument.

This resolves one source of non-hermeticity when building the stdlib with the hermetic C++ toolchain, but another one remains.

### DIFF
--- a/go/tools/builders/filter_buildid.go
+++ b/go/tools/builders/filter_buildid.go
@@ -17,6 +17,7 @@ package main
 import (
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


fmeum:
Seems hard to get right: https://github.com/golang/go/issues/27751
Maybe we also have to pass `-trimpath` to other uses of `cgo` outside `stdlib`?

PR #3570, opened 2023-05-25
Fix typo in load statement for bzlmod docs

Sovietaced:
**What type of PR is this?**
- Documentation

- This PR fixes a typo that made the docs incorrect. There is no defs.bzl, only [def.bzl](https://github.com/bazelbuild/bazel-gazelle/blob/master/def.bzl).


fmeum (approved):
Thanks!

PR #3577, opened 2023-05-31
docs: Mention that `gazelle:prefix` can be replaced by `go.mod`

fmeum:


fmeum:
@tyler-french

docs/go/core/bzlmod.md
```diff
@@ -75,11 +75,12 @@ If you have a use case that would require this, please explain it in an issue.
 
 ## Generating BUILD files
 
-Add the following to your top-level BUILD file, including your Go module's path in a [Gazelle directive](https://github.com/bazelbuild/bazel-gazelle#directives):
+Add the following to your top-level BUILD file, including your Go module's path in a [Gazelle directive](https://github.com/bazelbuild/bazel-gazelle#directives) if there is no `go.mod` file with that information in the same directory:
```
seh, line 78: Though it was easier to qualify the existing sentence, it would be clearer to the reader if we flip the sentence around. Consider:
```suggestion
If there is no `go.mod` file in the same directory as your top-level BUILD file, add the following [Gazelle directive](https://github.com/bazelbuild/bazel-gazelle#directives) to that BUILD file to supply Gazelle with your Go module's path:
```

seh: I wasn't sure that I captured the advice exactly right; I was mostly trying to convey the shape of the sentence.

fmeum: I applied your suggestion, but added another paragraph with the `go.mod` case.

seh (approved):
I'm happy to be moving past needing to know about this directive. Thank you.

PR #3586, opened 2023-06-14
enable nogo tag in bzlmod

tyler-french:
**What type of PR is this?**

Feature

This PR enables the ability to use `nogo` when running Bzlmod. This is done using the `.nogo` tag from `go_sdk` extension.

Fixes #

**Other notes for review**


go/private/extensions.bzl
```diff
@@ -39,6 +40,12 @@ _host_tag = tag_class(
     },
 )
 
+_nogo_tag = tag_class(
+    attrs = {
+        "nogo": attr.string(mandatory = True),
```
seh, line 45: The "nogo" attribute stutters with the tag class name. Could we either use something different for the attribute name here such as "rule", or perhaps generalize the tag class name to something like "tools"?

PR #3588, opened 2023-06-16
cleanup: remove usage of deprecated ioutil

sluongng:
**What type of PR is this?**

Other

Replace usage of ioutil with newer API in "os" and "io".

Fixes #

**Other notes for review**


sluongng:
too much churn and not worth the effort

PR #3602, opened 2023-06-24
cgo: Control static linking through rules_go

### LINKED ISSUE #3601: Allow the static on/off attribute of go_binary to control static linking with cgo
This is a follow up from #3590 where we noticed that client cgo could also lead to static linking.

0.40

0.30.0

6.2.1

yes

Linux, Debian, AARCH64

### What did you do?

```
$ cat cat .bazelrc 
build --features=fully_static_link --dynamic_mode=off

$ cat WORKSPACE
workspace(name = "rules_go_static")

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "6b65cb7917b4d1709f9410ffe00ecf3e160edf674b78c54a894471320862184f",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.39.0/rules_go-v0.39.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.39.0/rules_go-v0.39.0.zip",
    ],
)

http_archive(
    name = "bazel_gazelle",
    sha256 = "727f3e4edd96ea20c29e8c2ca9e8d2af724d8c7778e7923a854b2c80952bc405",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.30.0/bazel-gazelle-v0.30.0.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.30.0/bazel-gazelle-v0.30.0.tar.gz",
    ],
)

load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")

go_rules_dependencies()

go_register_toolchains(version = "1.19.3")
#go_register_toolchains(version = "1.20.3")

$ cat BUILD.file
load("@io_bazel_rules_go//go:def.bzl", "go_binary")

go_binary(
    name = "c_hello",
    srcs = ["c_hello.go"],
    cgo = True,
    static = "off", # "on"
)

$ cat c_hello.go
package main

// typedef int (*intFunc) ();
//
// int
// bridge_int_func(intFunc f)
// {
//		return f();
// }
//
// int fortytwo()
// {
//	    return 42;
// }
import "C"
import "fmt"

func main() {
	f := C.intFunc(C.fortytwo)
	fmt.Println(int(C.bridge_int_func(f)))
	// Output: 42
}

$ bazel build //:c_hello && file bazel-bin/c_hello_/c_hello
```

Given the `static = "off"`, I expected this binary to not be linked statically (regardless of my toolchain config).

### What did you see instead?

```
bazel-bin/c_hello_/c_hello: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, for GNU/Linux 3.7.0, Go BuildID=redacted, stripped
```

zecke:
Let the link action of rules_go determine how to link the binary. This aligns the behavior of the stdlib with other cgo code.

A binary can be linked statically by either using `static = "on"` in the go_binary rule or using `--@io_bazel_rules_go//go/config:static`.

Fixes #3601

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -58,7 +58,10 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
     cxxopts = go.cgo_tools.cxx_compile_options + cxxopts
     objcopts = go.cgo_tools.objc_compile_options + copts
     objcxxopts = go.cgo_tools.objcxx_compile_options + cxxopts
-    clinkopts = extldflags_from_cc_toolchain(go) + clinkopts
+    clinkopts = [
+        option for option in extldflags_from_cc_toolchain(go)
+        if option not in ("-static")
+    ] + clinkopts
 
     # NOTE(#2545): avoid unnecessary dynamic link
     if "-static-libstdc++" in clinkopts:


go/private/rules/cgo.bzl
```diff
@@ -58,7 +58,10 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
     cxxopts = go.cgo_tools.cxx_compile_options + cxxopts
     objcopts = go.cgo_tools.objc_compile_options + copts
     objcxxopts = go.cgo_tools.objcxx_compile_options + cxxopts
-    clinkopts = extldflags_from_cc_toolchain(go) + clinkopts
+    clinkopts = [
+        option for option in extldflags_from_cc_toolchain(go)
+        if option not in ("-static")
```
fmeum, line 63: This could be:
```suggestion
        if option != "-static"
```
for now. We can always generalize it later.

go/private/rules/cgo.bzl
```diff
@@ -58,7 +58,10 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
     cxxopts = go.cgo_tools.cxx_compile_options + cxxopts
     objcopts = go.cgo_tools.objc_compile_options + copts
     objcxxopts = go.cgo_tools.objcxx_compile_options + cxxopts
-    clinkopts = extldflags_from_cc_toolchain(go) + clinkopts
+    clinkopts = [
```
fmeum, line 61: Could you add a comment explaining why we exclude `-static`?

fmeum (commented):
Thanks, this generally makes sense.

Given the subtle effects passing/filtering this flag can have, I would really like to see an end-to-end test that e.g. runs `file` or some other tool on a `go_binary` with and without `static` to verify that it is built as expected.

Would you be okay with adding such a test? It probably doesn't have to be a full-blown `go_bazel_test`, just `go_binary` targets that are `data` dependencies of a regular `go_test`.

zecke:
> Thanks, this generally makes sense.
> 
> Given the subtle effects passing/filtering this flag can have, I would really like to see an end-to-end test that e.g. runs `file` or some other tool on a `go_binary` with and without `static` to verify that it is built as expected.
> 
> Would you be okay with adding such a test? It probably doesn't have to be a full-blown `go_bazel_test`, just `go_binary` targets that are `data` dependencies of a regular `go_test`.

Happy to do that. The best test would need to pass --features=fully_static_link to the invocation of bazel. How would this be done with the framework used? Should I add a new task to .bazelci/presubmit.yml?

fmeum:
You can define a `go_bazel_test` target, which allows you to run bazel in bazel with any kind of flags. There are many examples in rules_go that you could use as a base. Let me know if you need more help.

PR #3603, opened 2023-06-25
Remove references to go_googleapis

### LINKED ISSUE #2845: remove dependency on go-genproto

v0.26.0

v0.23.0

4.0.0

Yes

Linux / MacOS, AMD64.

### What did you do?

Used rules_go with [cloud.google.com/go](https://github.com/googleapis/google-cloud-go).

A build that compiled.

### What did you see instead?

The build failed due to rules_go declaring an older version of [go-genproto](https://github.com/googleapis/go-genproto). It is possible to get everything working by using the override instructions but that is not super intuitive. Unless there is a need for the genproto to be bundled I suggest removing the dependency on it so users of rules_go can more easily use the latest version of cloud.google.com/go without the need for overriding dependencies.

linzhp:
**What type of PR is this?**
Other

* Removing `go_googleapis` from `go_rules_dependencies`, and adding it to WORKSPACE of rules_go for tests only.
* Update integration tests on Google APIs to cover the new usage.
* Update docs

**Other notes for review**


WORKSPACE
```diff
@@ -119,6 +119,30 @@ go_repository(
     version = "v0.6.0",
 )
 
+go_repository(
+    name = "googleapis",
+    build_directives = [
+        "gazelle:go_generate_proto false",
```
fmeum, line 125: If we expect users to add these directives, we may want to add them to https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/bzlmod/default_gazelle_overrides.bzl so that they don't have to. This would also soften the migration.

WORKSPACE
```diff
@@ -119,6 +119,30 @@ go_repository(
     version = "v0.6.0",
 )
 
+go_repository(
+    name = "googleapis",
```
fmeum, line 123: Shouldn't we name this `com_github_googleapis_googleapis` now that it isn't subject to any special handling anymore?

linzhp: This repo will be treated as a pure proto repo. All references to it will have to be resolved manually (possibly with the new `resolve_regexp` directive). Because of that, I am trying to save some typing...

Do you see any benefit for using `com_github_googleapis_googleapis`?

fmeum: I see. I was just wondering what this would look like with Bzlmod: If the repo is managed by `go_deps`, it would naturally be assigned this name unless we special case it. Having WORKSPACE and Bzlmod builds align on the directives users need to add could have sufficient value that the longer repo name becomes acceptable.

linzhp: googleapis is not going to be managed by `go_deps`, because the Go modules are in `org_golang_google_genproto`. The only reason to download it as a `go_repository` is for Gazelle to generate the `proto_library` rules, hence `gazelle:go_generate_proto false` and `build_file_generation = "on"`.

fmeum: I see thanks for the explanation!

WORKSPACE
linzhp: Since `googleapis` is not `go_deps`, [default_gazelle_overrides.bzl](https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/bzlmod/default_gazelle_overrides.bzl) won't make a difference, right?

fmeum: That's true.

Do we happen to have an example for how depending on this would look like with Bzlmod? I am a bit worried that strict repo visibility would make it difficult for users to specify the correct resolve directives.

linzhp: Is it possible to load a `go_repository` in bzlmod? Worse case, we can pre-generate all the `proto_library` targets and upload it to BCR.

fmeum: It is, not if we do this in the most direct way possible, users won't easily be able to update the version, bringing us back to the problem we are trying to solve.

I will think more about this and get back to you.

linzhp: We can also commit all missing `proto_library` rules to googleapis repo, which is already a bazel repo. It gives us a bit more protection because their CI may catch some issues when the proto files and build rules are out of sync.

fmeum: I think that the best model would be a `googleapi` Bazel module that only contains the `proto_library` targets and `.proto` files and separate `googleapi-lang` repos with only the `lang_proto_library` targets.

linzhp: The original googleapis repo contains both `proto_library` (although not for every proto file, which is why we need to generate more) and `lang_proto_library` targets. I think it's a good idea to use it as a pure proto Bazel module.

I am not sure whether we need another googleapi-lang repos though. For Go, all generated packages have import prefix like google.golang.org/genproto/googleapis. They have a checked-in version published to that path. The plan is to use the go_deps `org_golang_google_genproto` instead. WDYT?

linzhp: With https://github.com/googleapis/googleapis/pull/831, we can use http_archive for googleapis

fmeum (approved):
Great work! 🎉

PR #3609, opened 2023-06-28
Adding Uber's hermetic cc toolchain for tests

linzhp:
**What type of PR is this?**
Other

Enable [uber/hermetic_cc_toolchain](https://github.com/uber/hermetic_cc_toolchain) to run tests on CI

**Other notes for review**
@fmeum Is there a better way to detect the current C toolchain in the test, so I can skip some test cases?


WORKSPACE
```diff
@@ -179,3 +179,24 @@ load(
...
 
 apple_support_dependencies()
+
+# For testing the compatibility with a hermetic cc toolchain. Users should not have to enable it.
+
+http_archive(
+    name = "hermetic_cc_toolchain",
+    sha256 = "bd2234acd0837251361be3270d7d3ce599b418be123d902d84762302e31a3014",
+    strip_prefix = "hermetic_cc_toolchain-13c904dce0cb9b6d07f0d557e6ce3cf7013a562e",
+    urls = ["https://github.com/uber/hermetic_cc_toolchain/archive/13c904dce0cb9b6d07f0d557e6ce3cf7013a562e.zip"],
```
motiejus, line 189: curious, why not the released version? Any known blockers?

.bazelci/presubmit.yml
```diff
@@ -12,21 +12,37 @@ tasks:
...
-    - "--config=incompatible"
+      - "--config=incompatible"
+    test_flags:
+      - "--config=incompatible"
+    build_targets:
+      - "//..."
+    test_targets:
+      - "//..."
+  ubuntu2004_uber_hermetic_cc:
+    platform: ubuntu2004
```
motiejus, line 31: not sure where to get the list of open platforms, but is there one that doesn't have a C/C++ compiler installed?

motiejus (commented):
It would be great to split `.bazelci/presubmit.yaml` lint checks to a separate commit or even PR for easier review.

Also, please squash the changes, there are extra unneeded merges.

WORKSPACE
linzhp: Yeah, I need https://github.com/uber/hermetic_cc_toolchain/commit/a9d87b21a5dddd691336c6c0004fa5dcfe5b9b48 to fix some tests.

.bazelci/presubmit.yml
linzhp: This is the list: https://github.com/bazelbuild/continuous-integration/blob/8c5e4b1322ea5751842bfd03d2e03d31dbe1f8f2/buildkite/bazelci.py#L465-L615

I am planning to add macOS cross-compiling into Linux, which is a common scenario

linzhp:
> It would be great to split `.bazelci/presubmit.yaml` lint checks to a separate commit or even PR for easier review.
Split to #3611 

> Also, please squash the changes, there are extra unneeded merges.
Will be squashed before merge

.bazelrc
```diff
@@ -1,11 +1,17 @@
 common --enable_platform_specific_config
+test --test_output=errors
 
 # Go requires a C toolchain that accepts options and emits errors like
+
```
fmeum, line 5:
```suggestion
```

fmeum (approved):

PR #3612, opened 2023-07-01
Fix `go_sdk` extension failure for unnamed root module

fmeum:
Fixes failures such as the following if the root module never calls `module`:

```
                 go_download_sdk_rule(
 Error in repository_rule: invalid user-provided repo name '__download_0': valid names may contain only A-Z, a-z, 0-9, '-', '_', '.', and must start with a letter
```

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -154,7 +154,8 @@ def _default_go_sdk_name(*, module, multi_version, tag_type, index):
     # Keep the version out of the repository name if possible to prevent unnecessary rebuilds when
     # it changes.
     return "{name}_{version}_{tag_type}_{index}".format(
-        name = module.name,
+        # "main_" is not a valid module name and thus can't collide.
+        name = module.name or "main_",
         version = module.version if multi_version else "",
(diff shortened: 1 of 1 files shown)


go/private/extensions.bzl
```diff
@@ -154,7 +154,8 @@ def _default_go_sdk_name(*, module, multi_version, tag_type, index):
     # Keep the version out of the repository name if possible to prevent unnecessary rebuilds when
     # it changes.
     return "{name}_{version}_{tag_type}_{index}".format(
-        name = module.name,
+        # "main_" is not a valid module name and thus can't collide.
+        name = module.name or "main_",
```
linzhp, line 158: Should we default it `__main__` so it's the same as the default workspace name when it's not given?

fmeum: That's not possible since a repository name (which we construct here) can't start with `_`. The name of the main repository is always the empty string, `__main__` is only used for the runfiles directory without Bzlmod and will go away when WORKSPACE is removed.

linzhp (approved):

PR #3617, opened 2023-07-05
fix: don't cause Bazel to use system patch(1)

alexeagle:
Resolves https://github.com/bazelbuild/rules_go/pull/1954#issuecomment-1622215160
in a way that requires BSD users to pass an explicit boolean argument.

**What type of PR is this?**

Bug fix

Prevent rules_go users always needing a system `patch(1)` installed


fmeum:
I think that a plain revert would suffice, see https://github.com/bazelbuild/rules_go/pull/1954#issuecomment-1622324085

alexeagle:
Let's just wait on #3603

PR #3618, opened 2023-07-06
Implement functionality to skip tests via TESTBRIDGE_TEST_ONLY env

### LINKED ISSUE #2785: go_test: skipping tests with --test_filter
We have a system that monitors the health (e.g., slow, flaky) of tests so we can dynamically skip unhealthy test cases in some environments. Using `t.Skip` is not very flexible, as it involves frequently adding and removing code snippets to a very large number of test cases. 

Can we support a minus (`-`) sign before the regex in `--test_filter` to indicate the pattern we want to skip? For example, `--test_filter=-^TestCase1$` means run all test cases except `TestCase1`.

Currently `--test_filter` is passed to `test.run`, which doesn't support the minus sign. Maybe we can filter the `allTests` slice before this line?

https://github.com/bazelbuild/rules_go/blob/642ff3d80c3454e8d9758c54ce6841cc089507be/go/tools/builders/generate_test_main.go#L141

yushan26:
**What type of PR is this?**

Feature

This PR implements the functionality to skip tests via TESTBRIDGE_TEST_ONLY env field. TESTBRIDGE_TEST_ONLY will expect a comma-separated list of regex expressions. A prefix '-' to the regex specifies to excluded test case matching the filter (similar to [instrumentaion_filter](https://bazel.build/reference/command-line-reference#flag--instrumentation_filter).
For example:
To indicate tests to run all the tests except for TestFQNTaskFiler and TestTargetTaskFiler, provide the environment in go_test:
`env = {"TESTBRIDGE_TEST_ONLY":"-^TestFQNTaskFiler$,-^TestTargetTaskFiler$"}`

This is the opposite behavior of the existing behavior `go test -run`.
This is needed in order to skip flaky test cases in CI.

Fixes #2785 

**Other notes for review**

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -103,6 +103,7 @@ import (
 	"reflect"
 {{end}}
 	"strconv"
+	"strings"
 	"testing"
 	"testing/internal/testdeps"
 
@@ -190,10 +191,21 @@ func main() {
   {{end}}
 
 	if filter := os.Getenv("TESTBRIDGE_TEST_ONLY"); filter != "" {
-		if filter[0:1] == "-" {
-			flag.Lookup("test.skip").Value.Set(filter[1:])
-		} else {
-			flag.Lookup("test.run").Value.Set(filter)
+		filters := strings.Split(filter, ",")
+		var runTests []string
+		var skipTests []string
+		for _, f := range filters {
+			if strings.HasPrefix(f, "-") {
+				skipTests = append(skipTests, f[1:])
+			} else {
+				runTests = append(runTests, f)
+			}
+		}
+		if len(runTests) > 0 {
+			flag.Lookup("test.run").Value.Set(strings.Join(runTests, "|"))
+		}
+		if len(skipTests) > 0 {
+			flag.Lookup("test.skip").Value.Set(strings.Join(skipTests, "|"))
 		}
 	}
 
@@ -401,7 +413,6 @@ func genTestMain(args []string) error {
 			}
 		}
 	}
-
 	for name := range importMap {
 		// Set the names for all unused imports to "_"
 		if !pkgs[name] {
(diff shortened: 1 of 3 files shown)


tests/core/go_test/filter_test_cases_test.go
```diff
@@ -0,0 +1,144 @@
...
+func Test(t *testing.T) {
+	tests := []struct {
+		name                  string
+		args                  []string
+		expectedRunTestCases  map[string]bool
+		expectedSkipTestCases map[string]struct{}
+	}{
+		{
+			name:                  "skip_tests",
+			args:                  []string{"test", "//:skip_tests", "--test_env=GO_TEST_WRAP_TESTV=1"},
```
linzhp, line 90: Can you pass `--test_filter` here instead of setting `TESTBRIDGE_TEST_ONLY`? That way, you only need 1 `go_test` target to cover 3 cases.

go/tools/builders/generate_test_main.go
```diff
@@ -397,7 +413,6 @@ func genTestMain(args []string) error {
 			}
 		}
 	}
-
```
linzhp, line 400: nit: keep this blank line

tests/core/go_test/filter_test_cases_test.go
```diff
@@ -0,0 +1,144 @@
...
+				t.Fatalf("could not unmarshall generated xml: %s", err)
+			}
+
+			for _, suite := range suites.Suites {
+				for _, tc := range suite.TestCases {
+					if _, ok := tt.expectedRunTestCases[tc.Name]; ok {
+						tt.expectedRunTestCases[tc.Name] = true
+					}
+					if _, ok := tt.expectedSkipTestCases[tc.Name]; ok {
+						t.Fatalf("unexpected test case ran %s", tc.Name)
```
linzhp, line 133:
```suggestion
						t.Errorf("unexpected test case ran %s", tc.Name)
```

The error here is not fatal. Using `Errorf` allow us to report all test failures at once.

linzhp (approved):
Thank your for working on this. I am really excited to see this implemented finally!

PR #3620, opened 2023-07-10
Preparing 0.41.0 release

linzhp:
Updating version strings for 0.41.0 release. For this release, I intentionally avoid upgrading dependencies, because we just updated them 2 weeks ago and it might bring additional disruption to an already disruptive release.

This release requires next release of Gazelle; and next release of Gazelle requires this release. @fmeum How to release new versions of two Bazel modules that depends on each other at the same time?

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -25,7 +25,7 @@ use_repo(go_sdk, "go_toolchains")
 
 register_toolchains("@go_toolchains//:all")
 
-bazel_dep(name = "gazelle", version = "0.32.0")
+bazel_dep(name = "gazelle", version = "0.30.0")
 
 go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
 go_deps.from_file(go_mod = "//:go.mod")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "rules_go",
-    version = "0.40.0",
+    version = "0.41.0",
     compatibility_level = 0,
     repo_name = "io_bazel_rules_go",
 )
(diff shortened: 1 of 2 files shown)


linzhp:
It seems OK for rules_go to depend on a future release of Gazelle, but not OK for Gazelle to depend on a future release of rules_go. So we should release rules_go first

MODULE.bazel
```diff
@@ -19,13 +19,13 @@ use_repo(
...
     name = "go_default_sdk",
-    version = "1.19.8",
+    version = "1.20.2",
 )
 use_repo(go_sdk, "go_toolchains")
 
 register_toolchains("@go_toolchains//:all")
 
-bazel_dep(name = "gazelle", version = "0.30.0")
+bazel_dep(name = "gazelle", version = "0.32.0")
```
fmeum, line 28: Do we need to raise this? In general, with Bzlmod, we should keep all deps at the lowest version possible.

fmeum (approved):
It is possible to release two versions simultaneously, but only manually and with some tricks. Let me know if this is necessary and I can take care of it, but if it's not, separate releases are far easier.

MODULE.bazel
linzhp: hmm, maybe not, if rules_go only needs Gazelle for testing itself. Would we mark this is as dev dependency?

fmeum: I think that we need the dep, but we don't need to update the version: The tests passed even without the bump. MVS will take care of it when used by end users.

If you just want to test against the new version, we could use `single_version_override`, which is ignored if `rules_go` isn't the root module.

PR #3622, opened 2023-07-12
fix indeterminism with cgo during coverage collection

joeljeske:
**What type of PR is this?**

>
> Bug fix

Archives including cgo with coverage instrumentation are indeterministic. 

When compiling c files with --collect_code_coverage, the intermediate .o files contain absolute paths  that include the temp dir prefix which leads to indeterminism. Tell the compiler to strip this temp dir prefix. 

**Other notes for review**

I am using https://github.com/grailbio/bazel-toolchain as my underlying toolchain, which may change this behavior, but I am not certain.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -318,6 +318,10 @@ func defaultCFlags(workDir string) []string {
 	flags := []string{
 		"-fdebug-prefix-map=" + abs(".") + "=.",
 		"-fdebug-prefix-map=" + workDir + "=.",
+		"-fcoverage-prefix-map=" + abs(".") + "=.",
+		"-fcoverage-prefix-map=" + workDir + "=.",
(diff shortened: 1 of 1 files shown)


fmeum (approved):
@linzhp Could you test at Uber?

go/tools/builders/cgo2.go
```diff
@@ -318,6 +318,10 @@ func defaultCFlags(workDir string) []string {
 	flags := []string{
 		"-fdebug-prefix-map=" + abs(".") + "=.",
 		"-fdebug-prefix-map=" + workDir + "=.",
+		"-fcoverage-prefix-map=" + abs(".") + "=.",
```
fmeum, line 321: Could you check whether `-ffile-prefix-map` has better support with old GCC versions? Otherwise we would need some kind of feature detection.

fmeum (changes requested):

PR #3631, opened 2023-07-19
[proto] Add extra_outputs to go_proto_compile

### LINKED ISSUE #3630: Allow extra_outputs on go_proto_compile

0.40.0

0.28.0

 6.3.0rc1

Yes

### What did you do?

See below. 

We have a custom plugin to generate enums .pb.go files from .proto files. For `example.proto`, we expect both
`example_dbenum.pb.go` and `example.pb.go` to be generated as Bazel provider outputs. Currently, only files with the same name and `.pb.go` file extensions are allowed. I'd like to propose adding attributes to allow outputting more w/ custom proto plugins.

tingilee:
**What type of PR is this?**

> Feature

Fixes #3630

**Other notes for review**

### DIFF
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -35,7 +35,7 @@ GoProtoCompiler = provider(
     fields = {
         "compile": """A function with the signature:
 
-    def compile(go, compiler, protos, imports, importpath)
+    def compile(go, compiler, protos, extra_outputs, imports, importpath)
 
 where go is the go_context object, compiler is this GoProtoCompiler, protos
 is a list of ProtoInfo providers for protos to compile, imports is a depset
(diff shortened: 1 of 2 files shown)


proto/compiler.bzl
```diff
@@ -93,6 +98,17 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
                 ext = compiler.internal.suffix,
             )
             go_srcs.append(out)
+
+            if extra_outputs.get(src.basename):
```
fmeum, line 102: You can assign this to a variable so that you don't need to `get` it again in the next line.

proto/def.bzl
```diff
@@ -165,6 +166,9 @@ go_proto_library = rule(
             providers = [GoLibrary],
             aspects = [_go_proto_aspect],
         ),
+        "extra_outputs": attr.string_dict(
+            doc = "Extra output files to be included in the expected generated output",
```
fmeum, line 170: Same as for the other doc, please explain the role of keys and values.

fmeum (commented):
Tests in `tests/core/go_proto_library` would be highly appreciated. There are already a number of tests there that you can use as a reference, let me know if you need further support.

tingilee:
Abandoning this in favor of https://github.com/bazelbuild/rules_go/pull/3650

Issue #3402, opened 2022-12-15
`go_download_sdk` should use the exec OS and arch instead of the host


jpsim:
My goal is to compile something using rules_go on a macOS arm64 host, using a macOS amd64 remote executor.

From my understanding of reading the [bazel docs on platforms](https://bazel.build/extending/platforms), it seems like instead of downloading the toolchain for the host platform, it should download it for the exec platform.

https://github.com/bazelbuild/rules_go/blob/1a8fe64877c6e71dbbf51cdc8ceb2eb10c13e521/go/private/sdk.bzl#L359-L372

### What version of rules_go are you using?

https://github.com/bazelbuild/rules_go/releases/tag/v0.36.0

### What version of gazelle are you using?

https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.26.0

### What version of Bazel are you using?

`6.0.0rc4`

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Host: macOS 13 (arm64)
Exec: macOS 13 (amd64)

### Any other potentially useful information about your toolchain?

Using remote execution.

### What did you do?

```console
$ git clone https://github.com/envoyproxy/envoy.git
$ cd envoy/mobile
# ./bazelw build \
  --tls_client_certificate=<redacted> \
  --tls_client_key=<redacted> \
  --config remote-ci-macos \
  //library/common:envoy_main_interface_lib
```

### What did you expect to see?

I expected `go_download_sdk` to download the darwin/arm64 version of the SDK.

### What did you see instead?

I see a toolchain resolution error:

> ERROR: /private/var/tmp/_bazel_jsimard/e99b8d924fe95f277e909ecdb8681535/external/com_envoyproxy_protoc_gen_validate/BUILD:11:10: While resolving toolchains for target @com_envoyproxy_protoc_gen_validate//:protoc-gen-validate: No matching toolchains found for types @io_bazel_rules_go//go:toolchain.

Here's a link to the invocation on EngFlow: https://envoy.cluster.engflow.com/invocation/7c14f5f1-1765-4589-8562-b1c5fc86318a#console

---

With https://github.com/envoyproxy/envoy/pull/24501 I can get past the toolchain resolution error when using remote exec on a macOS amd64 host and a macOS arm64 remote executor, but then the toolchain resolution fails on Linux ([CI logs](https://github.com/envoyproxy/envoy/actions/runs/3679971288/jobs/6225402653)).

[...]

Issue #3007, opened 2021-11-16
Can't easily configure differing host and remote execution SDKs


brentleyjones:
### What version of rules_go are you using?

0.29.0

### What version of gazelle are you using?

0.24.0

### What version of Bazel are you using?

5.0.0rc1

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

macOS 11.6.1, arm64

### What did you do?

I'm running a bazel build on an arm64 Mac, but targeting an x86 remote execution platform. I need the SDK to be the `darwin-amd64` version, but since the repository rule detects the architecture with `uname -m`,`go_download_sdk` downloads the `darwin-arm64` version. If I use `go_download_sdk` to download the `darwin-amd64` version, I also have to call it explicitly for every other platform I support (e.g. linux). In addition, because I call it multiple times, it downloads all of the SDKs, even if they aren't used.

nacl:
Also seeing a variant of this with HEAD rules_go and Bazel 5.2.0.  When I provide an explicit URL for `go_download_sdk`, rules_go seems to ignore the URL and download the SDK for the host architecture.

The OS autodetection feature can be worked around by explicitly providing `goos` and `goarch` to `go_download_sdk`, assuming that's what you are using, of course.

And as for everything being downloaded, you're probably seeing #3196.

nacl:
I've also seen cases where `goos` and `goarch` end up being unset or empty in the generated BUILD files for SDKs in our configuration -- it might be an issue with `native.existing_rules`, as we have our golang toolchains configured in a separate repository from the one bazel is invoked from.

achew22:
Strongly agree. Having our toolchains defined via flags is almost certainly the correct way for us to go.

PR #3634, opened 2023-07-24
Automatically register SDKs for common execution platforms

fmeum:
**What type of PR is this?**

Feature

The `go_sdk` module extension now registers unnamed SDKs for all major execution platforms in addition to the host platform. This ensures that Go builds should "just work" if the exec platform is on this list, even if it differs from the host platform.

Fixes https://github.com/bazelbuild/rules_go/issues/3007
Fixes https://github.com/bazelbuild/rules_go/issues/3402

**Other notes for review**

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -67,6 +67,18 @@ _host_tag = tag_class(
     },
 )
 
+# A list of (goos, goarch) pairs that are commonly used for remote executors in cross-platform
+# builds (where host != exec platform). By default, we register toolchains for all of these
+# platforms in addition to the host platform.
+_COMMON_EXEC_PLATFORMS = [
+    ("darwin", "amd64"),
+    ("darwin", "arm64"),
+    ("linux", "amd64"),
+    ("linux", "arm64"),
+    ("windows", "amd64"),
+    ("windows", "arm64"),
+]
+
 # This limit can be increased essentially arbitrarily, but doing so will cause a rebuild of all
 # targets using any of these toolchains due to the changed repository name.
 _MAX_NUM_TOOLCHAINS = 9999
@@ -96,10 +108,9 @@ def _go_sdk_impl(ctx):
 
             # SDKs with an explicit name are at risk of colliding with those from other modules.
             # This is acceptable if brought in by the root module as the user is responsible for any
-            # conflicts that arise. rules_go itself provides "go_default_sdk", which is used by
-            # Gazelle to bootstrap itself.
-            # TODO(https://github.com/bazelbuild/bazel-gazelle/issues/1469): Investigate whether
-            #  Gazelle can use the first user-defined SDK instead to prevent unnecessary downloads.
+            # conflicts that arise. rules_go itself provides "go_default_sdk".
+            # TODO: Now that Gazelle relies on the go_host_compatible_sdk_label repo, remove the
+            #       special case for "go_default_sdk". Users should migrate to @rules_go//go.
             if (not module.is_root and not module.name == "rules_go") and download_tag.name:
                 fail("go_sdk.download: name must not be specified in non-root module " + module.name)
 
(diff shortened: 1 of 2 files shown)


go/private/extensions.bzl
```diff
@@ -101,6 +112,42 @@ def _go_sdk_impl(ctx):
...
+                    )
+
+                    toolchains.append(struct(
+                        goos = goos,
+                        goarch = goarch,
+                        sdk_repo = default_name,
+                        sdk_type = "remote",
+                        sdk_version = download_tag.version,
+                    ))
+
```
sluongng, line 150: would this still work if you move the `go_download_sdk_rule` above this `if` clause into an `else` clause?

i.e.
```
            if download_tag.version and not download_tag.goos and not download_tag.goarch:
                for goos, goarch in _COMMON_EXEC_PLATFORMS:
                    ...
                    go_download_sdk_rule(...)
                    ...
            else:
                ...
                go_download_sdk_rule(...)
                ...
```

fmeum: It makes a difference if the host SDK is not contained in the list of common exec platforms. We always want to register an SDK for the host, even if it is not in the list.

sluongng: So with the current approach, there's potentially be an SDK registered twice (with different names).

I don't think it will be a big deal as the final Merkle tree hash would still be the same regardless. But it could be a confusing UX for the end users if both ended up being used somehow. 🤔

fmeum: Toolchain registration order definitely matters. I would think that "all 1.19 SDKs with varying exec platforms" coming before "all 1.20 SDKs with varying exec platforms" would be the most reasonable behavior, but of course this PR could end up changing which SDK is selected for existing declarations of the form:

```starlark
go_sdk.download(version = "1.19")
go_sdk.download(version = "1.20", goos = "linux", goarch = "arm64")
```

On a Linux amd64 host with Linux arm64 executor, with this PR 1.19 would be selected instead of the previous selection of 1.20. I don't know how common this setup is in practice, it does look error-prone as is.

Re duplicate SDKs: Unfortunately the Merkle tree hash includes the repo name, meaning that currently cache hits can fail for a variety of reasons, including just bumping the rules_go version (which is part of the repo name) or registering an additional toolchain earlier in the module even if it doesn't end up being selected. The ongoing path mapping work may help with that some day, but it's a difficult problem to solve in general.

sluongng:
Since our repo is not net on bzlmod, we need to make similar modifications to `sdk.bzl` for it to work.

There was an error on our end on the initial attempt:

```
Error in fail: gazelle found more than one suitable Go SDK (go_sdk, go_sdk_darwin_amd64). Specify which one to use with gazelle_dependencies(go_sdk = "go_sdk").
```

Work around is to add the if clause to skip on duplicated declaration (or to change gazelle's declaration).

```
diff --git a/go/private/sdk.bzl b/go/private/sdk.bzl
index 08c6e74c..7cea13bc 100644
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -303,6 +303,18 @@ def _go_toolchains(name, sdk_repo, sdk_type, sdk_version = None, goos = None, go
         sdk_versions = [sdk_version or ""],
     )

+# A list of (goos, goarch) pairs that are commonly used for remote executors in cross-platform
+# builds (where host != exec platform). By default, we register toolchains for all of these
+# platforms in addition to the host platform.
+_COMMON_EXEC_PLATFORMS = [
+    ("darwin", "amd64"),
+    ("darwin", "arm64"),
+    ("linux", "amd64"),
+    ("linux", "arm64"),
+    ("windows", "amd64"),
+    ("windows", "arm64"),
+]
+
 def go_download_sdk(name, register_toolchains = True, **kwargs):
     go_download_sdk_rule(name = name, **kwargs)
     _go_toolchains(
@@ -316,6 +328,28 @@ def go_download_sdk(name, register_toolchains = True, **kwargs):
     if register_toolchains:
         _register_toolchains(name)

+    for goos, goarch in _COMMON_EXEC_PLATFORMS:
+        if goos == kwargs.get("goos") and goarch == kwargs.get("goarch"):
+            continue
+
+        n = "{}_{}_{}".format(name, goos, goarch)
+        go_download_sdk_rule(
+            name = n,
+            goos = goos,
+            goarch = goarch,
+            **kwargs
+        )
+        _go_toolchains(
+            name = n + "_toolchains",
+            sdk_repo = n,
+            sdk_type = "remote",
+            sdk_version = kwargs.get("version"),
+            goos = goos,
+            goarch = goarch,
+        )
+        if register_toolchains:
+            _register_toolchains(n)
+
 def _go_local_sdk_impl(ctx):
     goroot = ctx.attr.path
     platform = _detect_sdk_platform(ctx, goroot)
```

Could you please include the sdk.bzl change in this PR as well? (please feel free to modify my monkey patch above)

darkrift:
@fmeum I was just hit by this problem this week, any way we can get this move forward ?

go/private/extensions.bzl
fmeum: I would like to revive this PR as I think it's still very relevant.

> I think as long as we provide a knob for people to opt-out of the default automatic toolchain registering to use their own, then they could do w/e they want. So what do you think we just add a flag (rules / extension attribute) in to disable the auto toolchain registration for known platforms?

I would like to do away with the "registering toolchains is costly" point of view that IMO has historically plagued rulesets. As part of this, I would prefer not viewing registering additional toolchains as a breaking change. How about we assume that the current mechanisms for controlling resolution (e.g. `go_cross_binary`) are sufficient for now and if really turns out that they aren't, we improve user control over resolution instead of making registration more complicated? I can see that being more maintainable down the road and also more useful for users.

fmeum:
Given the limited time all of us get to spend on maintaining this ruleset, I would like to restrict this to Bzlmod. It's less likely to break anything there with better way for us to contain any problems that may arise (WORKSPACE macros offer limited options here).

sluongng (approved):
Agree with the reasoning above.

tyler-french (approved):
Tested this, looks good!

Issue #3411, opened 2022-12-21
builds with cgo appear to miss .S assembly files


tals:
### What version of rules_go are you using?

0.35.0

### What version of gazelle are you using?

0.27.0

### What version of Bazel are you using?

Build label: 5.4.0

### Does this issue reproduce with the latest releases of all the above?

Yes

### What operating system and processor architecture are you using?

```
$ lsb_release -a
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
```

### What did you do?

I am attempting to [build Go program](https://github.com/tals/bazel-rulesgo-broken-asm-repro) using the popular [zstd package from DataDog](https://github.com/DataDog/zstd)
This library vendors zstd, which includes accelerated operations in the .S file.

This works well with the vanilla go toolchain, however building this via rules_go fails to pick up an .S file for amd64, causing it to fail during link time

```
$ git clone https://github.com/tals/bazel-rulesgo-broken-asm-repro
$ cd bazel-rulesgo-broken-asm-repro
$ go run main.go
Hello, World!

$ bazel run  //:hello-go

INFO: Analyzed target //:hello-go (26 packages loaded, 195 targets configured).
INFO: Found 1 target...
ERROR: /home/tal/.cache/bazel/_bazel_tal/d6d42962320318bd35977430f4d78b15/external/com_github_datadog_zstd/BUILD.bazel:3:11: GoCompilePkg external/com_github_datadog_zstd/zstd.a failed: (Exit 1): builder failed: error executing command bazel-out/k8-opt-exec-2B5CBBC6/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix linux_amd64 -src external/com_github_datadog_zstd/errors.go -src ... (remaining 211 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
/tmp/rules_go_work-585792782/cgo/github.com/DataDog/zstd/_x16.o:huf_decompress.c:function HUF_decompress4X1_usingDTable_internal_bmi2_asm: error: undefined reference to 'HUF_decompress4X1_usingDTable_internal_bmi2_asm_loop'
[...]

Issue #2006, opened 2019-03-26
Assembly files not classified correctly, based on cgo mode


jayconrod:
### What version of rules_go are you using?

0.18.1

### What version of Bazel are you using?

0.23.2

### Does this issue reproduce with the latest releases of all the above?

Yes

### What did you do?

`//tests/legacy/build_constraints:go_default_test` is incorrect. It includes Go assembly files in a package with cgo.

### What did you expect to see?

Go assembly sources should be rejected with a message like `package using cgo has Go assembly file asm_linux.s`. Both .s and .S files should be included in the build and should be assembled with the C compiler. This should only happen for packages with `cgo = True` when cgo is enabled.

### What did you see instead?

Go assembly sources are assembled with the Go assembler. .S files should not be included in the build unless `cgo = True` and cgo is enabled.

jsharpe:
This happens for datadog/zstd - I hit this when using gazelle to generate deps for ethereum-go : this is the workaround in that case: https://github.com/aspect-build/rules_sol/pull/8/files#diff-1f2c31b701f066691b4a4024644f86e092ab731eb81cc5d6282017fe485082ae

fmeum:
If we can't fix the issue in Gazelle, we could add the affected Go repos to the BCR with the fixing patches applied, as we did for [circl](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/circl).

PR #3648, opened 2023-08-10
compilepkg: cgo assembly uses the C compiler

evanj:
**What type of PR is this?**

>
Bug fix

This changes compilepkg to use the C compiler for .S and .s files to use the C compiler, like go build does. Previously it would use the Go assembler, which is used for pure Go packages. This should help fix issue: https://github.com/bazelbuild/rules_go/issues/3411

I have added a cgo test for this issue that is intended to work with both arm64 and amd64 machines. I have only tested it on Linux amd64 and Darwin arm64. Without this change it fails to build with the following output. This fails to parse the assembly file because it uses the Go assembler.

```
ERROR: rules_go/tests/core/cgo/asm/BUILD.bazel:3:11: GoCompilePkg tests/core/cgo/asm/asm.a failed: (Exit 1): builder failed: error executing command (from target //tests/core/cgo/asm:asm) bazel-out/k8-opt-exec-2B5CBBC6/bin/external/go_sdk/builder_reset/builder compilepkg -sdk external/go_sdk -installsuffix linux_amd64 -src tests/core/cgo/asm/cgoasm.go -src tests/core/cgo/asm/asm_amd64.S ...

tests/core/cgo/asm/asm_amd64.S:4: expected identifier, found "."
asm: assembly of tests/core/cgo/asm/asm_amd64.S failed
compilepkg: error running subcommand external/go_sdk/pkg/tool/linux_amd64/asm: exit status 1
```

Fixes # https://github.com/bazelbuild/rules_go/issues/3411

**Other notes for review**

This seems to pass `bazelisk test //tests/core/...` on my machine.

### DIFF
--- a/tests/core/cgo/asm/cgoasm_test.go
+++ b/tests/core/cgo/asm/cgoasm_test.go
@@ -1,4 +1,4 @@
-//go:build unix && (amd64 || arm64)
+//go:build amd64 || arm64
 
 package asm
 
--- a/tests/core/cgo/asm/cgoasm_test.go
+++ b/tests/core/cgo/asm/cgoasm_test.go
@@ -1,8 +1,10 @@
+//go:build unix && (amd64 || arm64)
+
 package asm
 
 import (
-	"testing"
 	"runtime"
+	"testing"
 )
 
 func TestNativeAssembly(t *testing.T) {
--- a/tests/core/cgo/asm/cgoasm_test.go
+++ b/tests/core/cgo/asm/cgoasm_test.go
@@ -1,0 +1,21 @@
+package asm
+
+import (
+	"testing"
+	"runtime"
+)
+
+func TestNativeAssembly(t *testing.T) {
+	expectedGOARCH := map[string]int{
+		"amd64": 42,
+		"arm64": 44,
+	}
+	expected := expectedGOARCH[runtime.GOARCH]
+	if expected == 0 {
+		t.Fatalf("expected=0 for GOARCH=%s; missing value?", runtime.GOARCH)
+	}
+	actual := callAssembly()
+	if actual != expected {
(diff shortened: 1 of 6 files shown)


evanj:
I used this example that uses the DataDog/zstd library to test this: https://github.com/evanj/ddzstdbazel

evanj:
This might also fix https://github.com/bazelbuild/rules_go/issues/2006

tests/core/cgo/asm/cgoasm_test.go
```diff
@@ -0,0 +1,21 @@
+package asm
```
fmeum, line 1: Could you add build constraints here so that this test doesn't break if we start testing it on non-amd64, non-am64 platforms or e.g. Windows?

fmeum (approved):
Thanks, this is a very helpful contribution!

@tyler-french Could you test this at Uber (assuming you have some Assembler CGo deps)?

CC @jsharpe

evanj:
Hmm looks like I didn't run the correct set of tests on my local branch; As noted in #2006: 

> `//tests/legacy/build_constraints:go_default_test` is incorrect. It includes Go assembly files in a package with cgo.

I may need to modify that test. Let me look into that, and the build constraints are a great idea. Thanks.

evanj:
I will attempt to fix the failing tests in a separate PR. It is related, but does not depend on this. Once that PR is merged, I'll update this to the latest main branch. See https://github.com/bazelbuild/rules_go/pull/3652

I still need to add the build constraints. I'll do that in the next couple of days.

evanj:
okay some progress: I'll get back to this in a bit; looks like I need to use some more build constraints to get the test to work on Windows

evanj:
Did you intend to enable auto-merge on this PR? It was enabled then disabled. Thanks.

PR #3650, opened 2023-08-11
[proto] Allow multiple outputs from a proto compiler

tingilee:
**What type of PR is this?**
 Feature

This PR allows a proto compiler to output to multiple files. For example, in the test, the protoc compilation of the enum.proto will output to enum.pb.go and enum_dbenum.pb.go.

This PR adds a new field `suffixes` which takes precedent over `suffix` if set. `suffix`'s behavior remains the same for backwards compatibility for code that already uses this field. However, long term, `suffix` can be replaced by `suffixes`.

Fixes https://github.com/bazelbuild/rules_go/issues/3630

**Other notes for review**


proto/compiler.bzl
```diff
@@ -87,14 +87,24 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
...
+                    )
+                    go_srcs.append(out)
+            else:
+                out = go.declare_file(
+                    go,
+                    path = importpath + "/" + src.basename[:-len(".proto")],
+                    ext = compiler.internal.suffix,
+                )
+                go_srcs.append(out)
+
```
fmeum, line 105: I would prefer unifying these code paths, e.g. like this (untested):

```suggestion
            suffixes = compiler.internal.suffixes
            if not suffixes:
                suffixes = [compiler.internal.suffix]
            for suffix in suffixes:
                out = go.declare_file(
                    go,
                    path = importpath + "/" + src.basename[:-len(".proto")],
                    ext = suffix,
                )
                go_srcs.append(out)

```

fmeum (approved):
I really like the comprehensive and realistic test coverage, great job!

The diff to production code is also pleasantly small now, thanks for iterating over https://github.com/bazelbuild/rules_go/pull/3631 in this way.

proto/compiler.bzl
tingilee: Addressed

PR #3653, opened 2023-08-12
Allow generating export files for the stdlib.

### LINKED ISSUE #3651: Generate Exports for stdlib for golangci-lint
0.41.0 and master

0.32.0

```
Build label: 6.0.0- (@non-git)
Build time: Tue Jan 1 00:00:00 1980 (315532800)
Build timestamp: 315532800
Build timestamp as int: 315532800
```

ya

Linux amd64

No

### What did you do?

```
export GOPACKAGEDRIVER=$PWD/tools/gopackagedriver.sh
golangci-lint run -v ./src/... --timeout=1h
```
`golangci-lint` complete

### What did you see instead?
`golangci-lint` hangs.

After debugging this, this seems to be an issue with gopackagesdriver not providing the correct Export files for the stdlib. `golanglint-ci` read the export files for all packages it receives from gopackagedriver, including packages in the stdlib. However, since Bazel sets the ExportFile for every stdlib module to the `__BAZEL_OUTPUTDIR__`, it fails to open. This can be thousands of imports across all the files which leads to 
 1. Incorrect results where stdlib modules fail typecheck
 2. Hanging for extremely long times (I lost patience at like 10min) as `golangci-lint` continously fails to read the export data here: https://github.com/golangci/golangci-lint/blob/9fc1e2075330561fb31888c4f2aff12b652466c2/pkg/golinters/goanalysis/runner_loadingpackage.go#L175

However, we do have the ability to generate the export files for the stdlib, at the cost of some extra build time on the first build:
```
diff --git a/go/tools/builders/stdliblist.go b/go/tools/builders/stdliblist.go
index 32f217f3..4c5eb295 100644
--- a/go/tools/builders/stdliblist.go
+++ b/go/tools/builders/stdliblist.go
@@ -158,7 +158,7 @@ func flatPackageForStd(cloneBase string, pkg *goListPackage, pathReplaceFn func(
 		ID:              stdlibPackageID(pkg.ImportPath),
 		Name:            pkg.Name,
 		PkgPath:         pkg.ImportPath,
-		ExportFile:      outputBasePath(cloneBase, pkg.Target),
+		ExportFile:      pathReplaceFn(pkg.Export),
 		Imports:         map[string]string{},
 		Standard:        pkg.Standard,
 		GoFiles:         goFiles,
@@ -265,7 +265,7 @@ func stdliblist(args []string) error {
 		listArgs = append(listArgs, "-compiled=true")
 	}

-	listArgs = append(listArgs, "-json", "builtin", "std", "runtime/cgo")
+	listArgs = append(listArgs, "-export", "-json", "builtin", "std", "runtime/cgo")

 	jsonFile, err := os.Create(*out)
 	if err != nil {
```

And this lets `golangci-lint` finish quickly and provide more accurate results.

nogo works great for analyzer based linters but golangci-lint offers many more linters that aren't written as analyzers which makes moving to bazel mean that you lose a lot of the lints you were accustomed to when using native Go tooling.

It would be great to support this!

DolceTriade:
**What type of PR is this?**

Feature

By default, rules_go only generates Export files for the code it compiles using bazel, but not the stdlib. This is problematic for tools like golangci-lint that has analyzers that utilize the export files for their checks, including the stdlib.

It's possible to generate export files for the stdlib by passing the `-export` flag to the `go list` command, however, it comes with a time penalty (about 20s with the flag vs 2s without the flag).

Therefore, add it as an option for go_register_toolchains.

Now, you can do something like:
```
go_register_toolchains(
    exports_for_stdlib = True,
    version = "1.20.1",
)
```

And it will generate the export files in the stdlib.pkg.json. Note that this time penalty only seems to be when using gopackagedriver not for regular builds.

Fixes #3651


DolceTriade:
ps: Will add tests if this approach seems sensible.

fmeum:
@JamyDev Could you take a look at this?

If we add this new functionality, it would be great if we could also expose it to Bzlmod users. For example, we could add a `config` tag to the `go_sdk` extension to configure this globally for all SDKs.

DolceTriade:
@fmeum @JamyDev Any update here? If the approach is viable, I can go ahead and add/fix tests.

JamyDev:
I totally lost track of this. If it's optional and can be enabled on the fly I think this is a fine approach.

tokongs:
Anything I can do to help this along? I'd love to get golangci-lint working with rules_go!

dbernadett:
I have also found this fix to be useful for getting golangci-lint working with the bazel package manager. I arrived at the exact same patch outlined here: https://github.com/bazelbuild/rules_go/issues/3651

DolceTriade:
Sorry I've been occupied. Let me rebase and clean up this PR some time next week. The main missing piece is bzlmod support. Our project hasn't migrated to it yet so it wasn't a priority and the hacky patch is much smaller than this one so it wasn't too burdensome to carry for us.

lbcjbb:
If you want @DolceTriade, I can help you to add the missing bzlmod support

DolceTriade:
@lbcjbb Sorry, that would be appreciated. I haven't found the time to do this yet :(

DolceTriade:
I rebased the changes. Hopes this helps. Let me know if I can help. Feel free to open a new PR and I can close this one as well.

fmeum:
One test is still failing, but thanks for rebasing the PR. If anyone figures out how to fix the test, let me know and I can apply the fix.

DolceTriade:
I should have fixed the test and I think I added bzlmod support, but I would appreciate feedback to see if I did it correctly.

I'm a little confused by the two `go_download_sdk_rule` calls in the bzlmod path, where the 2nd call doesn't pass in all the sdk configuration flags.

fmeum:
@DolceTriade Sorry for noticing this so late, but I would expect the need for export files to be independent of the particular SDK and instead determined by per-build needs that could change between local usage and CI. This would probably be better addressed by a Starlark flag in `//go/config` (which also requires much less boilerplate in various places) instead of a repo rule attribute.

lbcjbb:
Yes, it's a better solution. This would make it possible to build the stdllib in export mode only when it's needed, when the `packages.NeedExportFile` is given in driver request.

albertocavalcante:
@DolceTriade @fmeum should this PR be closed and consider implemented via https://github.com/bazel-contrib/rules_go/pull/4244?

PR #3654, opened 2023-08-12
update x/tools and use x/tools/go/vcs module

### LINKED ISSUE #3640: nogo_main fail to build when to upgrade golang.org/x/tools

Build label: 6.3.1-homebrew

if you update the tools like this

```
    go_repository(
        name = "org_golang_x_tools",
        importpath = "golang.org/x/tools",
        sum = "h1:ojD5zOW8+7dOGzdnNgersm8aPfcDjhMp12UfG93NIMc=",
        version = "v0.11.1",
    )
```

you will get the error

```
        have ()
        want (bool)

```

malt3:
**What type of PR is this?**

Bug fix

- Update golang.org/x/tools to v0.12.0
- Use deprecated module golang.org/x/tools/go/vcs (the package was removed from golang.org/x/tools)
- Update usage of x/tools

Fixes #3640

**Other notes for review**

After implementing this, I'm a little bit worried of the consequences this might have for downstream users of rules_go / gazelle.
It seems like we may somehow need to update rules_go and gazelle together, since both depend on golang.org/x/tools.
See also: https://github.com/bazelbuild/bazel-gazelle/pull/1603
Maybe we should copy the functionality of x/tools into rules_go as needed and not depend on it at all?


fmeum:
Could we break this up into smaller pieces by starting to use `golang.org/x/tools/go/vcs` instead of `golang.org/x/tools` now, but deferring the update and/or removal of `golang.org/x/tools`?

malt3:
> Could we break this up into smaller pieces by starting to use golang.org/x/tools/go/vcs instead of golang.org/x/tools now [...]

I will give this another try but I think this may be hard to do in practice, since gazelle will still use the `org_golang_x_tools` repo if it exists. Maybe this is only broken in the other PR.. I'll report back.

PR #3655, opened 2023-08-14
bzltestutil: document init() behavior change in Go 1.21

sluongng:
Also adjust test_chdir accordingly.

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**

### DIFF
--- a/go/tools/bzltestutil/init.go
+++ b/go/tools/bzltestutil/init.go
@@ -34,7 +34,7 @@ var (
 //
 // Since Go 1.21, the order of `init()` calls is determined based
 // on how Go runtime sort the package names. So it's not guaranteed
-// that this initializer runs before any user packages' initializer.
+// that this initializer runs before any user package's initializer.
 func init() {
 	var err error
 	testExecDir, err = os.Getwd()
--- a/go/tools/bzltestutil/init.go
+++ b/go/tools/bzltestutil/init.go
@@ -32,10 +32,7 @@ var (
 
 // Before Go 1.21, this initializer runs before any user's package.
 //
-// Since Go 1.21, the order of `init()` calls is determined based
-// on how Go runtime sort the package names. So it's not guaranteed
-// that this initializer runs before any user package's initializer.
-//
+// Since Go 1.21, the order of this initializer calls is not guarenteed to be first.
 // See https://go.dev/doc/go1.21#language for more details.
 func init() {
 	var err error
--- a/go/tools/bzltestutil/init.go
+++ b/go/tools/bzltestutil/init.go
@@ -30,7 +30,11 @@ var (
 	testExecDir string
 )
 
-// This initializer runs before any user packages.
+// Before Go 1.21, this initializer runs before any user packages.
+//
+// Since Go 1.21, the order of `init()` calls is determined based
+// on how Go runtime sort the package names. So it's not guaranteed
+// that this initializer runs before any user packages' initializer.
 func init() {
 	var err error
 	testExecDir, err = os.Getwd()
--- a/go/tools/bzltestutil/init.go
+++ b/go/tools/bzltestutil/init.go
@@ -30,11 +30,13 @@ var (
 	testExecDir string
 )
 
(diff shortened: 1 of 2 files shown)


go/tools/bzltestutil/init.go
```diff
@@ -30,7 +30,11 @@ var (
 	testExecDir string
 )
 
-// This initializer runs before any user packages.
+// Before Go 1.21, this initializer runs before any user packages.
+//
+// Since Go 1.21, the order of `init()` calls is determined based
+// on how Go runtime sort the package names. So it's not guaranteed
+// that this initializer runs before any user packages' initializer.
```
fmeum, line 37:
```suggestion
// that this initializer runs before any user package's initializer.
```

go/tools/bzltestutil/init.go
```diff
@@ -30,7 +30,11 @@ var (
 	testExecDir string
 )
 
-// This initializer runs before any user packages.
+// Before Go 1.21, this initializer runs before any user packages.
+//
+// Since Go 1.21, the order of `init()` calls is determined based
```
fmeum, line 35: Based on my understanding of the process, the sorting isn't determined by the runtime but fixed by the spec. Instead of trying to describe it here, we could just link to the authoritative source.

sluongng: fixed

achew22:
I'm not aware of any legitimate use-cases of this in my own codebase. That's not to say that there aren't any, but more that this is a thing changing in go 1.21 by the language. I think we should track the language and the raw go version of this would exhibit the exact same behavior

fmeum:
But wouldn't the raw Go equivalent of this be the test runner that changes directories before any test code is run? This isn't something we easily do here without a launcher, but maybe we should for compatibility with native Go tests.

sluongng:
```
--- a/go/tools/bzltestutil/init.go
+++ b/go/tools/bzltestutil/init.go
@@ -41,6 +41,10 @@ func init() {
                panic(err)
        }

+       ChangeTestDir()
+}
+
+func ChangeTestDir() {
        // Check if we're being run by Bazel and change directories if so.
        // TEST_SRCDIR and TEST_WORKSPACE are set by the Bazel test runner, so that makes a decent proxy.
        testSrcDir, hasSrcDir := os.LookupEnv("TEST_SRCDIR")
```

I was thinking of doing something like this, so the logic of switching dir is available for others to re-use.
But given that:
a. It's only reading 2 environment variables
b. It's only affecting test execution that uses the relative path in `init()`

I think the potential breakage here is small-to-none. Whoever is facing the incompatibility should either:
a. Switch to use an absolute path, OR reuse the current logic
b. Refactor out the usage of `init()`

Normal file I/O in tests is not affected as it would happen _after_ `bzltestutil.init()` runs.

sluongng:
Another problem with calling `ChangeTestDir()` in the example above, before `bzltestutil.init()` is run, is that it could potentially affect `testExecDir` value. So I think folks should be looking at this case-by-case basis.

And if anyone found a better generic solution, we should welcome that contribution to rules_go 🚀

linzhp:
In Uber's the [style guide](https://github.com/uber-go/guide/blob/master/style.md#avoid-init), we recommend against any I/O in `init()`. We fixed all the places that are affected by this behavior change as part of Go 1.21 upgrade.

achew22:
I guess I'm interested in the circumstances where changing the order of `init` execution would matter. Could we get some links to examples of people doing this?

If you're performing IO in `init` that's going to be used in a test case you already, by definition, need to be able to handle being invoked from any directory in the repository because each of those tests would have a different launch directory. Our launching with the root as the cwd is equivalent to having a test file in the root of the repo. I guess it's technically possible that someone could have a package that's only ever imported by a single test in another package, but it seems like you've gone out of your way to build something brittle. I think documenting this behavior change is totally fine, but I am not sure it motivates writing a wrapper program, unless you have bigger plans for it.

PR #3666, opened 2023-08-17
sdk.bzl: include go.env since 1.21.0

### LINKED ISSUE #3665: Build failures when using a local go 1.21.0 SDK

v0.41.0

v0.32.0

6.3.2

Yes

Linux/amd64

Using `go_host_sdk`, with go 1.21.0.

### What did you do?

Attempt to build with external repository dependencies (not bzlmod, though I doubt that matters here)

Working build.

### What did you see instead?

```
ERROR: An error occurred during the fetch of repository 'org_golang_google_grpc_cmd_protoc_gen_go_grpc':
        File "/.../external/bazel_gazelle/internal/go_repository.bzl", line 256, column 17, in _go_repository_impl
                fail("failed to fetch %s: %s" % (ctx.name, result.stderr))
Error in fail: failed to fetch org_golang_google_grpc_cmd_protoc_gen_go_grpc: fetch_repo: google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0: GOPROXY list is not the empty string, but contains no entries
```
Building with `go build` works fine.

A little bit of Googling around turned up https://github.com/golang/go/issues/61928.  I suspect the issue is that
https://github.com/bazelbuild/rules_go/blob/139ec3c5ebf86682527aa1decc09b01d17235b97/go/private/BUILD.sdk.bazel#L77-L84
is not pulling in the new `go.env` file from the root of the SDK directory.

Honestly I'm not sure whether I'd count this as a bug in `rules_go` (for not including the file) or `go` itself (for failing this way if the file is missing, rather than just falling back on sensible defaults like it would if the file was present but empty).

sluongng:
Include go.env file in the root of the sdk repo starting from Go 1.21.0.

Fixes #3665

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -80,8 +80,10 @@ filegroup(
         "bin/go*",
         "src/**",
         "pkg/**",
-        "go.env*",
-    ]),
+    ]) + glob(
+        ["go.env*"],
+        allow_empty = True,
+    ),
 )
 
 exports_files(
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -80,7 +80,7 @@ filegroup(
         "bin/go*",
         "src/**",
         "pkg/**",
-        {goenv}
+        "go.env*",
     ]),
 )
 
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -80,6 +80,7 @@ filegroup(
         "bin/go*",
         "src/**",
         "pkg/**",
+        {goenv}
     ]),
 )
 
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -80,7 +80,7 @@ filegroup(
         "bin/go*",
         "src/**",
         "pkg/**",
-        "go.env*",
+        {goenv}
     ]),
 )
 
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -478,6 +478,9 @@ def _sdk_build_file(ctx, platform, version, experiments):
         if not "nocoverageredesign" in experiments and not "coverageredesign" in experiments:
             experiments = experiments + ["nocoverageredesign"]
 
+    # TODO(sluongng): Remove this once we drop support for Go 1.20
+    include_go_env = pv[1] >= 21
+
     ctx.template(
         "BUILD.bazel",
         ctx.path(ctx.attr._sdk_build_file),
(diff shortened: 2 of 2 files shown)


sluongng:
```
!master ~/work/bazelbuild/rules_go> cat /private/var/tmp/_bazel_sluongng/5c02d8a9b82454292bdfef7b6f9ac04e/external/go_sdk/go.env
# This file contains the initial defaults for go command configuration.
# Values set by 'go env -w' and written to the user's go/env file override these.
# The environment overrides everything else.

# Use the Go module mirror and checksum database by default.
# See https://proxy.golang.org for details.
GOPROXY=https://proxy.golang.org,direct
GOSUMDB=sum.golang.org

# Automatically download newer toolchains as directed by go.mod files.
# See https://go.dev/doc/toolchain for details.
GOTOOLCHAIN=auto
```

for reference, this is the content of the file.

go/private/BUILD.sdk.bazel
```diff
@@ -80,6 +80,7 @@ filegroup(
         "bin/go*",
         "src/**",
         "pkg/**",
+        {goenv}
```
fmeum, line 83: `go.env*` should match `go.env` if it exists and nothing otherwise. Just slightly hacky, but in my opinion better than version detection logic. I could see a point for `*.env` instead.

sluongng: Nice call. I switched to using `go.env*` instead, which is much cleaner.

Let's start with the more conservative solution and change it later when needed.

sluongng:
```
(09:55:02) ERROR: Traceback (most recent call last):
--
  | File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/go_sdk/BUILD.bazel", line 79, column 16, in <toplevel>
  | srcs = glob([
  | Error in glob: glob pattern 'go.env*' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
```

I guess it does not work if empty glob is not allowed.

fmeum: You can set `allow_empty = True`

fmeum: You can also extract out a separate `glob` call so that we don't lose the error in case the SDK directory is empty.

fmeum (approved):

fmeum:
@sluongng Could you submit the update to Go SDK 1.21.0 combined with the revert of the earlier approach to fix the GO_PROXY issue? The work should be done and that way we can test that's actually the case.

Thank you very much for driving these fixes!

sluongng:
I don't think we need a revert on Gazelle side just *yet*.

Instead, let's proceed with the Go SDK upgrade in rules_go.
Once rules_go is updated, we shall bump rules_go + GoSDK version in Gazelle and revert the change.

WDYT?

sluongng:
Here is the revert on Gazelle side anyway ☝️ 

You already have [your PR](https://github.com/bazelbuild/rules_go/pull/3643) for bumping Go SDK in rules_go. Let's try merging that @fmeum

PR #3667, opened 2023-08-22
[go_proto_library] File granularity for output groups in OutputGroupInfo 

### LINKED ISSUE #3664: Enable generate pb.go outputs as Bazel targets

0.40.0

0.28.0

6.3.0

Yes

We have `mockgen` rules that we need to refer to the generated pb.go files, sometimes only for one specific file service.pb.go among the many generated pb.go files from `go_proto_library`. I'd like to request for adding a list of `output_list` to expose //package:service.pb.go file from other Bazel targets. For files in this list, go_proto_library will declare targets for us to reference the generated file.  

This will allow us to declare:
```
mockgen(
    name = "retriever_service_mock",
    package = "clients",
    source = "//retriever/v1:service.pb.go",
)
```

### What did you see instead?

Today, this feature is not supported.

tingilee:
**What type of PR is this?**

> Feature

Make OutputGroupInfo more modularized in go_proto_library. This allows us to create filegroups referring to subset of the generated Go files. See tests for examples of only referring to one of the generated files. 

Fixes #3664

**Other notes for review**

### DIFF
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -134,9 +134,9 @@ def _go_proto_library_impl(ctx):
     )
     source = go.library_to_source(go, ctx.attr, library, False)
     providers = [library, source]
-    output_groups = {
-        "go_generated_srcs": go_srcs,
-    }
+    output_groups = {}
+    for go_src in go_srcs:
+        output_groups[go_src.basename] = depset([go_src])
     if valid_archive:
         archive = go.archive(go, source)
         output_groups["compilation_outputs"] = [archive.data.file]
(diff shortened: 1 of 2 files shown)


proto/def.bzl
```diff
@@ -134,9 +134,9 @@ def _go_proto_library_impl(ctx):
     )
     source = go.library_to_source(go, ctx.attr, library, False)
     providers = [library, source]
-    output_groups = {
-        "go_generated_srcs": go_srcs,
```
fmeum, line 138: We need to keep the old group for backwards compatibility.

fmeum (changes requested):

fmeum (approved):
@linzhp Looks good to me, do you have any concerns about this?

tingilee:
Closing this with comment:
https://github.com/bazelbuild/rules_go/issues/3664#issuecomment-1690984461

PR #3671, opened 2023-08-23
use carved out module for x/tools/go/vcs

### LINKED ISSUE #3640: nogo_main fail to build when to upgrade golang.org/x/tools

malt3:
This is a part of #3654 that was carved out to make it smaller and easier to understand. 
The vcs part of golang.org/x/tools was removed from tools. The vcs code is still available in a separate module with a version marked as "deprecated". Let's use it for now.

**What type of PR is this?**

Bug fix

- Use deprecated module golang.org/x/tools/go/vcs (the package was removed from golang.org/x/tools)

You can now upgrade golang.org/x/tools and rules_go will no longer try to use the vcs package from that module.
Note however, that updating golang.org/x/tools still triggers #3640.
To fix this, we need to update `golang.org/x/tools` in rules_go and use the new function signature (as shown in https://github.com/bazelbuild/rules_go/pull/3654/commits/d09bc19a2a4b8d726d2ddc5f097ac7b1e4129de3).

**Other notes for review**

We should probably vendor this code in the long term.

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -82,6 +82,25 @@ def go_rules_dependencies(force = False):
         patch_args = ["-p1"],
     )
 
+    # Needed for go/tools/fetch_repo
+    # releaser:upgrade-dep golang tools
+    wrapper(
+        http_archive,
+        name = "org_golang_x_tools_go_vcs",
+        # v0.12.0, latest as of 2023-08-12
+        urls = [
+            "https://mirror.bazel.build/github.com/golang/tools/archive/refs/tags/go/vcs/v0.1.0-deprecated.zip",
+            "https://github.com/golang/tools/archive/refs/tags/go/vcs/v0.1.0-deprecated.zip",
+        ],
+        sha256 = "1b389268d126467105305ae4482df0189cc80a13aaab28d0946192b4ad0737a8",
+        strip_prefix = "tools-go-vcs-v0.1.0-deprecated/go/vcs",
+        patches = [
+            # releaser:patch-cmd gazelle -repo_root . -go_prefix golang.org/x/tools/go/vcs -go_naming_convention import_alias
+            Label("//third_party:org_golang_x_tools_go_vcs-gazelle.patch"),
+        ],
+        patch_args = ["-p1"],
+    )
+
     # releaser:upgrade-dep golang sys
     wrapper(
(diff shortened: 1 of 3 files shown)


hawkingrei:
@fmeum @sluongng PTAL

sluongng (approved):
LGTM. I am a bit concerned about the potential conflict of `importpath` for downstream users, but I think the usage of this package is rare enough and we could always advise users to upgrade/patch instead.

hawkingrei:
> LGTM. I am a bit concerned about the potential conflict of `importpath` for downstream users, but I think the usage of this package is rare enough and we could always advise users to upgrade/patch instead.

Yes, I agree with you. TiDB use this package in some linters. we will upgrade rules_go with upgrading all linters to the latest one.

go/private/repositories.bzl
```diff
@@ -82,6 +82,25 @@ def go_rules_dependencies(force = False):
...
     )
 
+    # Needed for go/tools/fetch_repo
+    # releaser:upgrade-dep golang tools
+    wrapper(
+        http_archive,
+        name = "org_golang_x_tools_go_vcs",
+        # v0.12.0, latest as of 2023-08-12
+        urls = [
+            #"https://mirror.bazel.build/github.com/golang/tools/archive/refs/tags/go/vcs/v0.1.0-deprecated.zip",
```
fmeum, line 92:
```suggestion
          "https://mirror.bazel.build/github.com/golang/tools/archive/refs/tags/go/vcs/v0.1.0-deprecated.zip",
```

after https://github.com/bazelbuild/bazel/issues/19335 has been resolved.

malt3: Thanks for noticing and requesting the file to be mirrored. Totally forgot about that!

sluongng:
There are a few references in tests/integration/popular_repos/... that were missed by this PR.

But I will attempt to fix it when upgrade x/tools

PR #3672, opened 2023-08-25
bug fix: parse embeds in files that contain the double quote rune

andyscott:
**What type of PR is this?**

Bug fix

Updates the embed parser to parse between single quotes. As it stands, the double quote rune `'"'` effectively halts embed parsing for a given file.

**Other notes for review**

In the first commit, I've added a failing test using a fixture file with the double quote rune. The second commit resolves the failure and fixes the bug.

### DIFF
--- a/go/tools/builders/read.go
+++ b/go/tools/builders/read.go
@@ -198,6 +198,28 @@ func (r *importReader) findEmbed(first bool) bool {
 		case ' ', '\t':
 			// leave startLine alone
 
+		case '\'':
+			startLine = false
+			for r.err == nil {
+				if r.eof {
+					r.syntaxError()
+				}
+				c = r.readByteNoBuf()
+				if c == '\\' {
+					_ = r.readByteNoBuf()
+					if r.err != nil {
+						r.syntaxError()
+						return false
+					}
+					continue
+				}
+				if c == '\'' {
+					c = r.readByteNoBuf()
+					goto Reswitch
+				}
+			}
+			goto Reswitch
+
 		case '"':
 			startLine = false
 			for r.err == nil {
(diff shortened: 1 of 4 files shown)


fmeum (approved):

go/tools/builders/read.go
```diff
@@ -198,6 +198,28 @@ func (r *importReader) findEmbed(first bool) bool {
...
 
+		case '\'':
+			startLine = false
+			for r.err == nil {
+				if r.eof {
+					r.syntaxError()
+				}
+				c = r.readByteNoBuf()
+				if c == '\\' {
+					r.readByteNoBuf()
```
tyler-french, line 209:
```suggestion
					_ = r.readByteNoBuf()
```

Let's show that we're suppressing

go/tools/builders/read.go
```diff
@@ -198,6 +198,28 @@ func (r *importReader) findEmbed(first bool) bool {
 		case ' ', '\t':
 			// leave startLine alone
 
+		case '\'':
```
tyler-french, line 201: optional: since we use `\'` more than once, consider extracting it to a constant

tyler-french (approved):
LGTM! Small optional comments

andyscott:
Hello again! What can do I do to help get this merged?

fmeum:
@andyscott Looks like the CLA bot got confused. Could you squash all the commits into a single one with just you as the Git author?

andyscott:
Can do!

andyscott:
Squashing to a single commit didn't do the trick, so I opted to revert prior to the merged suggestion and then manually apply the change. Following that, I rebased on master and force pushed.

PR #3677, opened 2023-09-05
releaser: turn off upload to mirror by default

sluongng:
Most rules_go maintainers do not have direct access to Bazel team's
GSC bucket to upload this mirror. Let's disable it by default for
eventually removal.

Mirroring should be done by filing a requests in Bazel.git's Github
issues.

**What type of PR is this?**

> Uncomment one line below and remove others.
>
> Bug fix
> Feature
> Documentation
> Other

**What does this PR do? Why is it needed?**

**Which issues(s) does this PR fix?**

Fixes #

**Other notes for review**

PR #3678, opened 2023-09-05
upgrade x/tools to v0.12.0

sluongng:
The upgrade was done with

```bash
> bazel run go/tools/releaser -- upgrade-dep org_golang_x_tools
```

after the patch in #3677.

---

A few additional updates are needed for this to work:

- Update to `tests/integration/popular_repos` to remove unused packages
- Update to `nogo` to use the new interface of x/tools

Because this upgrade is NOT backward compatible (by design of x/tools
repo), downstream users of rules_go are advised to upgrade `x/tools`
version in their repo to v0.12.0 or later to avoid breakages.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -128,8 +128,11 @@ def _go_test_impl(ctx):
         # Disable symbol table and DWARF generation for test binaries.
         test_gc_linkopts.extend(["-s", "-w"])
 
-    # Link in the run_dir global for bzltestutil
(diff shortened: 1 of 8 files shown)


sluongng:
@fmeum it looks like Bazel CI is having an issue today. Could you please retry this PR later as well when you have a chance?

sluongng:
may require gazelle to be upgraded to a version that includes https://github.com/bazelbuild/bazel-gazelle/pull/1603 so that `//tests/integration/gazelle:gazelle_test` would pass.

dzbarsky:
We're past this version

PR #3679, opened 2023-09-05
bzltestutil: set importmap to fix run_dir

### LINKED ISSUE #3675: go_test: I/O in package init function fails after Go 1.21 due to initialization order change

v0.41.0 but also reproduced at `5206498b4f67ff3e6a9222a923cf67ff5191754e` (master as of 2023-09-01).

n/a

7.0.0-pre.20230530.3

yes

darwin arm64

n/a

### What did you do?

We're in the process of upgrading to Go 1.21, running all our tests including those from vendored packages.

All tests pass, as before.

### What did you see instead?

A couple vendored tests are failing for us as part of that upgrade, for example `github.com/klauspost/compress/huff0`. This fails in [`compress_test.go`](https://github.com/klauspost/compress/blob/master/huff0/compress_test.go#L64) because the test reads a testdata file in its init function.

I think we can all agree that doing I/O in an init function is bad and anti-social. Regrettably there are no test-I/O cops, so people do this somewhat frequently.

This is caused by the change documented in the [Go 1.21 release notes](https://tip.golang.org/doc/go1.21):

> Package initialization order is now specified more precisely. The new algorithm is:
>
> - Sort all packages by import path.
> - Repeat until the list of packages is empty:
>   - Find the first package in the list for which all imports are already initialized.
>   - Initialize that package and remove it from the list.
> 
> This may change the behavior of some programs that rely on a specific initialization ordering that was not expressed by explicit imports. The behavior of such programs was not well defined by the spec in past releases. The new rule provides an unambiguous definition.

`go_test` links tests with `go/tools/bzltestutil`. The `init` function in that package is responsible for changing into the test directory from the repository root directory where Bazel starts the test process. I added some print statements and confirmed that `bzltestutil`'s init function does not run in the test above before it panics.

I see there's some discussion already in #3643. #3655 also adds comments documenting the change.

It would be great to fix the regression here. I'm happy to work on a fix, but this is definitely the kind of change that needs some discussion. Some ideas:

1. We could inject a source file into the internal and external test packages for each test that imports `bzltestutil`. That would add a dependency edge, forcing `bzltestutil` to be initialized first. It might still be initialized before other transitively imported packages though, so this may not be a sufficient fix.
2. We could set the package path for `bzltestutil` to something that sorts before anything else. Of course this is an egregious hack, but it's lightweight and would work reasonably well under the new semantics.
3. We could introduce a separate test wrapper executable that calls the main executable without linking in any user code. We already have a test wrapper *process*, but it runs code linked into the test executable (`bzltestutil/wrap.go`). Although it would be cleaner to separate that, I think it would mean an extra link action per `go_test`.

cc @fmeum @sluongng

jayconrod:
**What type of PR is this?**

Bug fix

Go 1.21 clarified package initialization order and changed behavior: packages with lexicographically lower paths are now initialized earlier.

We need bzltestutil to initialize before user packages so it can change to the correct directory. To accomplish this, we set an importmap prefix that starts with '+', the lowest allowed character.

Fixes #3675

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -128,8 +128,11 @@ def _go_test_impl(ctx):
         # Disable symbol table and DWARF generation for test binaries.
         test_gc_linkopts.extend(["-s", "-w"])
 
-    # Link in the run_dir global for bzltestutil
-    test_gc_linkopts.extend(["-X", "github.com/bazelbuild/rules_go/go/tools/bzltestutil.RunDir=" + run_dir])
+    # Link in the run_dir global for bzltestutil.
+    # We add "+initfirst/" to the package path so the package is initialized
+    # before user code. See comment above the init function
+    # in bzltestutil/init.go.
+    test_gc_linkopts.extend(["-X", "+initfirst/github.com/bazelbuild/rules_go/go/tools/bzltestutil.RunDir=" + run_dir])
 
     # Now compile the test binary itself
     test_library = GoLibrary(
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -87,11 +87,10 @@ func (c *Cases) Version(v string) bool {
 const testMainTpl = `
 package main
 
-// This package must be initialized before packages being tested.
-// NOTE: this relies on the order of package initialization, which is the spec
-// is somewhat unclear about-- it only clearly guarantees that imported packages
-// are initialized before their importers, though in practice (and implied) it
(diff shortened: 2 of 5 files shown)


fmeum (approved):
Thanks, very neat!

fmeum:
@jayconrod I rebased https://github.com/bazelbuild/rules_go/pull/3643 onto your fix, but `test_chdir` fails (https://buildkite.com/bazel/rules-go-golang/builds/5510). Could you look into that?

PR #3681, opened 2023-09-06
bzltestutil: move os.Chdir call into new package

### LINKED ISSUE #3675: go_test: I/O in package init function fails after Go 1.21 due to initialization order change

jayconrod:
**What type of PR is this?**

Bug fix

Sorry for missing this the first time in #3679. The package init algorithm was a little different than I thought, and I imagined that Go 1.21 was already in use, so I didn't look into it deeply.

The problem was that even though "+initfirst/.../bzlutil" sorts lower than other packages, it imports "sync" and a number of other std packages that sort higher than "github.com/..." user packages. If a user package has no dependencies, it's eligible to be initialized before bzltestutil, even with the lower name.

I moved the init function with the os.Chdir call into a separate tiny package that only imports "os" and "runtime". It should get initialized much earlier.

Fixes #3675


jayconrod:
@fmeum I think this will fix the issue in #3679. I've verified at least that my company's affected vendored tests now pass when using 1.21.

go/tools/bzltestutil/BUILD.bazel
```diff
@@ -3,17 +3,15 @@ load("//go:def.bzl", "go_test", "go_tool_library")
...
     srcs = [
-        "init.go",
         "lcov.go",
         "test2json.go",
         "wrap.go",
         "xml.go",
     ],
-    # We add "+initfirst/" to the package path so this package is initialized
-    # before user code. See comment above the init function in init.go.
     importmap = "+initfirst/github.com/bazelbuild/rules_go/go/tools/bzltestutil",
```
fmeum, line 11: Do we still need the funky importpath on this package?

go/tools/bzltestutil/chdir/init.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil && runtime.GOOS != "windows" {
+			panic("could not change to test directory: " + err.Error())
+		}
+		if err == nil {
+			os.Setenv("PWD", abs)
+		}
+	}
+}
+
+func filepathIsAbs(path string) bool {
```
fmeum, line 106: Is this copied from the Go stdlib sources? If so, please add a link.

go/tools/bzltestutil/chdir/init.go
```diff
@@ -0,0 +1,138 @@
...
+//    several other std packages may be get initialized later. For example,
+//    the "sync" package is initialized after a user package named
+//    "github.com/a/b" that only imports "os", and because bzltestutil imports
+//    "sync", it would get initialized even later. A user package that imports
+//    nothing may still be initialized before "os", but we assume "os"
+//    is needed to observe the current directory.
+package chdir
+
+// This package should import nothing other than "os"
+// and packages imported by "os".
```
fmeum, line 64: But it imports `"runtime"` :-)

How about replacing `runtime.GOOS == "windows"` with a check on `os.PathSeparator`?

sluongng:
If you want to future proof this, you could add `github.com/klauspost/compress` into https://github.com/bazelbuild/rules_go/blob/master/tests/integration/popular_repos/popular_repos.py

I think that repo is quite relevant for the Bazel and Go ecosystem in general (we use it as well).

go/tools/bzltestutil/chdir/init.go
sluongng: I don't think it's worth it. `os` imports `runtime` anyway, so IIUC, the init order would include `runtime` with/without it being imported here.

Just modifying the comment would suffice.

go/tools/bzltestutil/BUILD.bazel
jayconrod: Nope, deleted.

go/tools/bzltestutil/chdir/init.go
jayconrod: Since we only need to know whether we're on Windows, `os.PathSeparator` is a good enough check, I've added a global `isWindows` for that.

It is true though that `"runtime"` needs to be initialized first anyway. I've expanded the comment with a `go list` command to show what's okay.

go/tools/bzltestutil/chdir/init.go
jayconrod: No, this is a more minimal hack, same with `filepathJoin`. I've expanded the comment to explain its limitations.

jayconrod:
> If you want to future proof this, you could add github.com/klauspost/compress into https://github.com/bazelbuild/rules_go/blob/master/tests/integration/popular_repos/popular_repos.py
>
> I think that repo is quite relevant for the Bazel and Go ecosystem in general (we use it as well).

I tried, but it's not as easy as I hoped. Most of the tests in that repo read testdata files from a top-level testdata directory, so some hand-edited `filegroup`s and `data` attributes would be needed: Gazelle / `go_repository` can't do it automatically. We've excluded tests in `org_golang_x_tools` for similar reasons.

Some tests are slow: `zstd` takes 99s on my machine.

There's also some desync between `popular_repos.py` and rules_go's dependencies.

fmeum (approved):

PR #3688, opened 2023-09-11
Fix bazel query scope not being used

### LINKED ISSUE #3687: Go packages driver does not use bazelQueryScope

0.41.0

0.32.0

6.3.1 

Yes

Linux Amd64

### What did you do?

Command:
`echo {} | GOPACKAGESDRIVER_BAZEL_QUERY_SCOPE=//path/to/target ./gopackagesdriver ./...  `

Deps query with bazelQueryScope ignored, resulting in a very large query: 
```
Running: [bazel --output_base X query --tool_tag=gopackagesdriver --ui_actions_shown=0 --ui_event_filters=-info,-stderr --noshow_progress --order_output=no --output=label --nodep_deps --noimplicit_deps --notool_deps kind("go_library", ...)]
```

### What did you see instead?

Deps query scoped correctly: 
```
Running: [bazel --output_base /home/user/xx query --tool_tag=gopackagesdriver --ui_actions_shown=0 --ui_event_filters=-info,-stderr --noshow_progress --order_output=no --output=label --nodep_deps --noimplicit_deps --notool_deps kind("go_library", attr(importpath, ".", deps(//path/to/target)))]
```

rastenis:
**What type of PR is this?**

Bug fix

Addresses an issue with Gopackagesdriver not using bazelQueryScope.

The packageQuery currently returns a bazel query formatted to find dependencies by import path (line 116, `kind("%s", attr(importpath, "%s", deps(%s)))`), so even if localQuery was restricted to the bazelQueryScope, it would not produce the same required output. It seems this behavior change happened after commit b84cd75.

Fixes #3687 

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/bazel_json_builder.go
+++ b/go/tools/gopackagesdriver/bazel_json_builder.go
@@ -125,12 +125,12 @@ func (b *BazelJSONBuilder) queryFromRequests(requests ...string) string {
 		if strings.HasSuffix(request, ".go") {
 			f := strings.TrimPrefix(request, "file=")
 			result = b.fileQuery(f)
+		} else if bazelQueryScope != "" {
+			result = b.packageQuery(request)
 		} else if isLocalPattern(request) {
 			result = b.localQuery(request)
 		} else if request == "builtin" || request == "std" {
 			result = fmt.Sprintf(RulesGoStdlibLabel)
-		} else if bazelQueryScope != "" {
-			result = b.packageQuery(request)
 		}
 
 		if result != "" {


go/tools/gopackagesdriver/bazel_json_builder.go
```diff
@@ -125,12 +125,12 @@ func (b *BazelJSONBuilder) queryFromRequests(requests ...string) string {
 		if strings.HasSuffix(request, ".go") {
 			f := strings.TrimPrefix(request, "file=")
 			result = b.fileQuery(f)
+		} else if bazelQueryScope != "" {
+			result = b.packageQuery(request)
```
fmeum, line 129: Just to understand this fix better: Would it make sense to restrict the `localQuery` to the `packageQuery` (e.g. via `intersect`) instead of always using the `packageQuery`?

rastenis: Hi, the packageQuery currently returns a bazel query formatted to find dependencies by import path (line 116, `kind("%s", attr(importpath, "%s", deps(%s)))`), so even if localQuery was restricted to the bazelQueryScope, it would not produce the same required output. It seems this behavior change happened after commit b84cd75919a40ce7d1409a6effbdd9a343cdaf72.

fmeum (approved):

JamyDev:
Now that we have a test suite for the driver, @rastenis would you be able to write a vallidation for this scenario?

go/tools/gopackagesdriver/bazel_json_builder.go
waltercacau: just an FYI but I believe this PR may have introduced a bit of a subtle bug when gopls sends "./..." and "bultin" args to the driver: https://github.com/bazelbuild/rules_go/issues/3999

PR #3696, opened 2023-09-15
Symlink all SDK files with `go_local_sdk`

fmeum:
**What type of PR is this?**

Bug fix

This is required with Go 1.21.1 as the `go` tool reads the `go.env` at the root of the SDK. Instead of explicitly symlinking that file, symlink all files except for the ones that already exist, e.g. because they were added by another SDK rule wrapped with `go_wrap_sdk`.


linzhp (approved):

tyler-french (approved):
Tested at Uber

PR #3697, opened 2023-09-15
prepare release 0.42.0

tyler-french:
This PR does not upgrade `x/tools` because there is a backward incompatible change that will force downstream users to upgrade.
```
        have ()
        want (bool)
```
We should wait to upgrade `x/tools` as it is pretty invasive: https://github.com/bazelbuild/rules_go/pull/3702


tyler-french:
Just noticed: https://github.com/bazelbuild/rules_go/issues/3640#issuecomment-1665406010
We should upgrade these.

fmeum (approved):

PR #3684, opened 2023-09-07
[go_sdk download] allow patches to standard library

tyler-french:
**What type of PR is this?**

Feature

**What does this PR do? Why is it needed?**

Allows users to provide patches to the Go SDK when using the `go_download_sdk` rule or the `go_sdk.download` Bzlmod tag.

This is helpful when some modification is needed to a package in the go standard library, but the user does not want to manage a fork of the standard library itself.

Since we build the standard library from source, the `patches` should apply appropriately across multiple OS and Architectures. However, we should not expect patches to apply correctly across multiple `version`s of the Go toolchain. For that reason, we only allow patches if the version is specified.

**Which issues(s) does this PR fix?**

Fixes #3683 

**Testing**
Added a test to the BCR which patches the `os` package and then runs a test to assert the patch properly applied.

Show that `patches` cannot apply when `version` is not specified:
```
% bazel-6.3.2 test //...
...
ERROR: An error occurred during the fetch of repository 'rules_go~override~go_sdk~rules_go_bcr_tests__download_0':
   Traceback (most recent call last):
        File "/home/user/.cache/bazel/_bazel_tfrench/565e6f63e863d6a7567b33b30501a80c/external/rules_go~override/go/private/sdk.bzl", line 93, column 21, in _go_download_sdk_impl
                fail("version must be specified when patches are specified")
Error in fail: version must be specified when patches are specified
...
ERROR: Couldn't start the build. Unable to run tests
```

PR #3698, opened 2023-09-16
go_download_sdk: allow patching to stdlib

tyler-french:
This commit merges the features from https://github.com/bazelbuild/rules_go/pull/3670 and https://github.com/bazelbuild/rules_go/pull/3684

We should be as restrictive as possible since we will need to maintain all uses in the long term.

The signature of the allowed patch arguments allows for `patch_strip` rather than `patch_args` to prevent misuse of the patch args. This design decision was also taken by Bzlmod.

`go_local_sdk` and `go_wrap_sdk` also do not allow patching, since these directories/repositories can be modified with patches before being registered as a Go SDK. Allowing these features again is risky as it gives users unnecessary additional control.

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -17,6 +17,34 @@ load("//go/private:nogo.bzl", "go_register_nogo")
 load("//go/private/skylib/lib:versions.bzl", "versions")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "patch")
 
+patch_attrs = {
+    "patches": attr.label_list(
+        default = [],
+        doc =
+            "A list of files that are to be applied as patches to go sdk. " +
+            "By default, it uses the Bazel-native patch implementation " +
+            "which doesn't support fuzz match and binary patch, but Bazel will fall back to use " +
+            "patch command line tool if `patch_tool` attribute is specified or there are " +
+            "arguments other than `-p` in `patch_args` attribute.",
+    ),
+    "remote_patches": attr.string_dict(
+        default = {},
+        doc =
+            "A map of patch file URL to its integrity value, they are applied before applying patch" +
+            "files from the `patches` attribute. It uses the Bazel-native patch implementation, " +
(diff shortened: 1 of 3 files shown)


go/private/sdk.bzl
```diff
@@ -113,9 +141,9 @@ def _go_download_sdk_impl(ctx):
     filename, sha256 = sdks[platform]
 
     _remote_sdk(ctx, [url.format(filename) for url in ctx.attr.urls], ctx.attr.strip_prefix, sha256)
-    patch(ctx, patch_args = _get_patch_args(ctx.attr.patch_strip))
 
     detected_version = _detect_sdk_version(ctx, ".")
+    patch(ctx, patch_args = _get_patch_args(ctx.attr.patch_strip))
```
fmeum, line 146: How are we enforcing that `version` is set here?

go/private/sdk.bzl
```diff
@@ -17,6 +17,34 @@ load("//go/private:nogo.bzl", "go_register_nogo")
...
+    "patches": attr.label_list(
+        default = [],
+        doc =
+            "A list of files that are to be applied as patches to go sdk. " +
+            "By default, it uses the Bazel-native patch implementation " +
+            "which doesn't support fuzz match and binary patch, but Bazel will fall back to use " +
+            "patch command line tool if `patch_tool` attribute is specified or there are " +
+            "arguments other than `-p` in `patch_args` attribute.",
+    ),
+    "remote_patches": attr.string_dict(
```
fmeum, line 30: I would vote for dropping the `remote_*` attributes until we see a clear use case for them.

fmeum (approved):

go/private/sdk.bzl
tyler-french: I see no need to separate out the `patch_attrs` since they're only used once.

I update the docs in https://github.com/bazelbuild/rules_go/pull/3716

Closing this PR

PR #3702, opened 2023-09-18
[BREAKING CHANGE] upgrade x/tools to 0.13.0

tyler-french:
This PR introduces the breaking changes of upgrading `x/tools` to `0.13.0`

Downstream users who are not able to upgrade will need to apply the following patch: [x_tools_old.patch](https://github.com/bazelbuild/rules_go/files/12653580/x_tools_old.patch)

Related comment: https://github.com/bazelbuild/rules_go/issues/3640#issuecomment-1665406010

Resolves https://github.com/bazelbuild/rules_go/issues/3640

1. This PR needs to be merged. 
2. We need to release https://github.com/bazelbuild/rules_go/pull/3697 `0.42.0`
3. Then, we need to upgrade `x/tools` and `rules_go` simultaneously in https://github.com/bazelbuild/bazel-gazelle
4. Then, we need to release a new version of Gazelle.

We will need to send out communication that downstream users will need to upgrade all 3 simultaneously, or patch `rules_go` with [x_tools_old.patch](https://github.com/bazelbuild/rules_go/files/12653580/x_tools_old.patch)


sluongng:
Including the patch into the repo might be a brilliant move. We could add into the release note an instruction on how to consume the patch in some scenarios.

tyler-french:
> Including the patch into the repo might be a brilliant move. We could add into the release note an instruction on how to consume the patch in some scenarios.

Great idea. I haven't had a chance to get back and fix the builds, but we should probably at least include the patch download link with the release

hawkingrei:
In fact, ```x/tools``` is important for the linter such as golangci-linter. We haven't updated to the latest version of the linter for a long time, and we are also unable to update to the latest version of grpc.

PR #3703, opened 2023-09-19
Mark `go_sdk` module extension as `{os,arch}_dependent`

fmeum:
**What type of PR is this?**

Bug fix

This ensures that the lockfile maintains separate instances of this extension for each OS/arch combination.

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -214,7 +214,6 @@ go_sdk_extra_kwargs = {
     "os_dependent": True,
     "arch_dependent": True,
 } if bazel_features.external_deps.module_extension_has_os_arch_dependent else {}
-print(go_sdk_extra_kwargs)
 
 go_sdk = module_extension(
     implementation = _go_sdk_impl,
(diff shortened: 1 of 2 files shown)


fmeum:
CC @tyler-french

go/private/extensions.bzl
```diff
@@ -207,12 +208,21 @@ def _left_pad_zero(index, length):
         fail("index must be non-negative")
     return ("0" * length + str(index))[-length:]
 
+go_sdk_extra_kwargs = {
+    # The choice of a host-compatible SDK is expressed in repository rule attribute values and
+    # depends on host OS and architecture.
+    "os_dependent": True,
+    "arch_dependent": True,
+} if bazel_features.external_deps.module_extension_has_os_arch_dependent else {}
+print(go_sdk_extra_kwargs)
```
linzhp, line 217: debug statement :D

fmeum: Whoops :-)

fmeum, line 217:
```suggestion
```

linzhp (approved):

fmeum:
@tpasternak There isn't and given that a module extension can run essentially arbitrary binaries, this can't be done. I'm looking into ways to resolve the situation for `go_sdk` and similar extensions.

PR #3705, opened 2023-09-21
Infer importpath if not set explicitly

mering:
**What type of PR is this?**

> Feature

This infers the `importpath` automatically if it is not set. If set explicitly it continues to use this one.

**Other notes for review**


go/private/context.bzl
```diff
@@ -357,17 +357,19 @@ def _check_importpaths(ctx):
         if ":" in p:
             fail("import path '%s' contains invalid character :" % p)
 
-def _infer_importpath(ctx):
+def _infer_importpath(ctx, attr = None):
```
fmeum, line 360: I think that this is only called in one place and that call site provides `attr`, so we can make this attribute required.

fmeum (approved):
LGTM, thanks!

@linzhp Do you have any concerns?

go/private/context.bzl
mering: Good idea. Done.

linzhp (approved):
LGTM

tests/core/go_proto_library_importpath/bar.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package tests.core.go_proto_library_importpath.bar;
```
linzhp, line 3: Actually, the mapping from "tests.core.go_proto_library_importpath.bar" to "tests/core/go_proto_library_importpath/bar_go_proto" sounds very opinionated. Can we only infer the import path when `go_package` is set?

mering: The logic to infer the importpath is defined in https://github.com/bazelbuild/rules_go/blob/69d0fc88d0815bd11867ed3f26f36244faa3c06c/go/private/context.bzl#L384-L393 which I do not touch in this PR. It is basically `package/label` which seems logical within the Bazel world. This is also what we are currently using internally via a macro on top of the `go_proto_library` rule. To me, it sounds reasonable to keep the current behavior while at the same time maintaining the possibility of overwriting the importpath as before. So for existing users for which the explicit setting of importpath was obligatory, the behavior doesn't change as well.

linzhp: I see. Thanks for the explanation.

PR #3706, opened 2023-09-21
Use aspect to infer proto deps automatically

### LINKED ISSUE #3668: Remove `deps` attribute from `go_proto_library` by automatically inferring them

0.37.0

0.31.0

6.2.0

Checking the release notes, the requested feature is not there yet

NA

NA

### What did you do?
I wrote a `go_proto_library` rule for a corresponding `proto_library` rule.

For example,
```
proto_library(
    name = "foo_proto",
    deps = [":bar_proto"],
)

go_proto_library(
    name = "foo_go_proto",
    protos = [":foo_proto"],
    deps = [":bar_go_proto"],
)
```

The `deps` attribute needs the corresponding `bar_go_proto` to be included. Since the `protos` attribute has the corresponding proto rules, it would be nice to see the `bar_go_proto` to be inferred by the rule automatically. In other words, we'd like to see the `deps` attribute removed altogether in the `go_proto_library` rule.  This is how it works in blaze, and it is implemented using `aspects` there. 

`cc_proto_library` from `rules_cc` and `py_proto_library` from `rules_python` already does this too.

Expected:
```
proto_library(
    name = "foo_proto",
    deps = [":bar_proto"],
)

go_proto_library(
    name = "foo_go_proto",
    protos = [":foo_proto"],
)
```

### What did you see instead?

This feature is not part of the releases yet. I can also support in developing this but I'm not sure about the work needed and where.

mering:
**What type of PR is this?**

> Feature

This automatically infers transitive proto deps with an aspect to avoid manually maintaining two parallel dependency trees (proto deps and go_proto deps).

Fixes #3668

**Other notes for review**


mering:
@fmeum Could I have your input on this? I am quite new to Bazel and therefore any input is highly appreciated!

The approach in this PR is to generate the go proto bindings within an aspect on the `proto_library` rules instead of on a `go_proto_library` rule. One problem I am facing because of this is that it is now no longer possible to pass down attributes to the aspect. So one could either overwrite defaults (like `importpath`) on the `proto_library` rule instead of the `go_proto_library` one and access it somehow from there. Another alternative could be to use something like an `importpath_prefix` on a `package` level or similar. Same problem for compilers which are essential (e.g. for gRPC). I have seen some `TODO` comments regarding gRPC, so it probably makes sense to align with your plans as this will probably be a breaking change anyways.

So any input to my general approach or ideas on how to move forward is highly appreciated!

mering:
One idea which I just had is to use separate aspects for compilers (might still share the same implementation but need to set the `_compiler` attribute differently). Then the `go_proto_library` rule would just attach the _proto_ aspect while the `go_grpc_library` would attach both the _proto_ and the _grpc_ aspect. Could this work?

mering:
@fmeum I would really appreciate your help here. Thanks in advance!

fmeum:
> @fmeum I would really appreciate your help here. Thanks in advance!

Sorry for the delay, I do really appreciate your work on this. I can't promise that I will get to this before BazelCon as I'm somewhat busy preparing for it. I am not that knowledgeable about gRPC, so I will have to delve into that a bit before I can provide meaningful input.

CC @linzhp just in case you already have some ideas in mind

mering:
> > @fmeum I would really appreciate your help here. Thanks in advance!
> 
> Sorry for the delay, I do really appreciate your work on this. I can't promise that I will get to this before BazelCon as I'm somewhat busy preparing for it. I am not that knowledgeable about gRPC, so I will have to delve into that a bit before I can provide meaningful input.

Thanks for your answer. Take your time. Looking forward for your comments.

linzhp:
> The approach in this PR is to generate the go proto bindings within an aspect on the proto_library rules instead of on a go_proto_library rule

`proto_library` rule should be language agnostic. You should pass any information related to Go in `go_proto_library` and its aspect. One implementation you can refer to is the `py_proto_library` in rules_python.

proto/def.bzl
```diff
@@ -31,161 +28,109 @@ load(
...
-        output_groups["compilation_outputs"] = [archive.data.file]
-        providers.extend([
-            archive,
-            DefaultInfo(
-                files = depset([archive.data.file]),
-                runfiles = archive.runfiles,
-            ),
-        ])
-    return providers + [OutputGroupInfo(**output_groups)]
+    archive = go.archive(go, source)
```
mering, line 79: Return `GoProtoInfo` provider in the aspect and convert in the rule.

proto/def.bzl
```diff
@@ -31,161 +28,109 @@ load(
...
+    return providers
+
+go_proto_library = rule(
+    implementation = _go_proto_library_impl,
+    attrs = {
+        "deps": attr.label_list(
+            providers = [ProtoInfo],
+            aspects = [_go_proto_aspect],
+        ),
+    },
```
mering, line 133: Specify `provides = [GoLibrary, GoSource, GoArchive]` explicitly.

mering:
Note to self: When rebasing, make sure to migrate the breaking changes in [v0.51.0](https://github.com/bazel-contrib/rules_go/releases/tag/v0.51.0).

Issue #3063, opened 2022-01-26
nogo: support the analysis.Analyzer repair feature


ztstewart:
### What version of rules_go are you using?
v0.29.0

### What version of gazelle are you using?
Commit: 6bbfc47f1b0a27ee1efeddcc6671f3e4e03235dc
This is newer than v0.24.0, which was released in October 2021. The commit is dated January 11th, 2022

### What version of Bazel are you using?
bazel 4.2.2

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Mac OS X, x86-64, but is applicable to all OSes / architectures given this is an enhancement request.

### What did you do?

Write an `analysis.Analyzer` that supports `SuggestedFixes` ([docs](https://pkg.go.dev/golang.org/x/tools/go/analysis#SuggestedFix)) with `TextEdit` ([docs](https://pkg.go.dev/golang.org/x/tools/go/analysis#TextEdit)) values. Using `analysistest.RunWithSuggestedFixes()` ([docs](https://pkg.go.dev/golang.org/x/tools@v0.1.8/go/analysis/analysistest#RunWithSuggestedFixes)) shows the suggested changes make sense and are getting applied (or otherwise the test would fail).

### What did you expect to see?

`golang.org/x/tools/go/analysis/multichecker` and the `singlechecker` version both support a `-fix` flag ([relevant source](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.1.8:go/analysis/internal/checker/checker.go;l=54;drc=d88ec18c69c70d818214bd2c98f3ce431ea4fd62)). This instructs the tool to make use of the `SuggestedFixes` output to modify the source code being analyzed. This allows teams to not only find and prevent issues at scale, but also repair them at scale.

Given that `analysis.Analyzer` is becoming the foundation of more and more Go tools, the `-fix` flag is becoming increasingly useful as many third party developers have come up with some pretty nifty code repair tools.

### What did you see instead?

`nogo` and related machinery do not appear to have any equivalent to the `-fix` flag that I can see in the `nogo` binary. To a degree, this is expected; `nogo` as designed runs static analysis tools and fails the build if any report an error.

That said, the `-fix` flag (and / or the `go vet` equivalents) are a pretty big deal; repairing problems across tens of thousands of files is seemingly a design goal of `analysis.Analyzer`, and this limitation prevents realizing this full potential when using `bazel` (without coming up with an alternative strategy). 

[...]

PR #3707, opened 2023-09-22
nogo: use bazel validations and don't fail compile

### LINKED ISSUE #3695: Add validation actions for nogo outputs

0.40.0

0.28.0

6.3.0rc1

Yes

### What did you do?

Set up nogo and ran `bazel build //...` across the monorepo which contains many static analysis errors. 

Should see an exhaustive list of all static analysis errors. This is very important as we're enabling static analysis for the first time across our monorepo. 

### What did you see instead?

With compilePkg and nogo being coupled, when nogo failed, the compilation also fails. Once compilePkg action fails, the targets that depend on the failed build target never run action of compilePkg. Therefore, `bazel build //...` only show a subset of errors. 

I'd like to propose adding https://bazel.build/extending/rules#validation_actions in the current nogo implementation, and would be happy to contribute upstream. High level proposal is to add validation output here https://github.com/bazelbuild/rules_go/blob/f03a723c49d5ec8afa0f9c74c4c3a705b37e8f40/go/private/actions/compilepkg.bzl#L160 and add validation action runs in compilepkg.bzl.

joeljeske:
**What type of PR is this?**

Feature

This PR allows failing nogo findings to be captured as a [Bazel validation action](https://bazel.build/extending/rules#validation_actions) and allow the compilepkg actions to exit 0 even when there are nogo failures. 

This is needed for a couple reasons: 

1. When you flip on a new validation rule, there may be lots of breakages and you would like to see the entire set of violations, as opposed to the only those appearing in the leaf go_libraries in the DAG. This allows go_libraries to propagate their results as to see downstream nogo violations. 
2. When iterating locally, it is _expensive_ to turn off nogo; you currently have to change the nogo registration for the sdk which then requires recompiling the entire chain of dependencies without the nogo check. With this change, you can simply pass --norun_validations and all the nogo output is silenced. 

Fixes #3695

**Other notes for review**

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -62,9 +62,17 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
     out_export = go.declare_file(go, name = source.library.name, ext = pre_ext + ".x")
     out_cgo_export_h = None  # set if cgo used in c-shared or c-archive mode
 
+    # capture the nogo_log output but don't fail the compile, later we assert that the log is empty
+    out_nogo_log = None
+    out_nogo_validation = None
+    if go.nogo:
+        out_nogo_log = go.declare_file(go, name = source.library.name, ext = pre_ext + ".nogo.log")
+        out_nogo_validation = go.declare_file(go, name = source.library.name, ext = pre_ext + ".nogo")
+
     direct = [get_archive(dep) for dep in source.deps]
     runfiles = source.runfiles
     data_files = runfiles.files
+    validations = []
 
     files = []
     for a in direct:
@@ -106,6 +114,7 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
             out_lib = out_lib,
             out_export = out_export,
             out_cgo_export_h = out_cgo_export_h,
+            out_nogo_log = out_nogo_log,
             gc_goopts = source.gc_goopts,
             cgo = True,
             cgo_inputs = cgo.inputs,
@@ -129,10 +138,26 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
             archives = direct,
             out_lib = out_lib,
             out_export = out_export,
+            out_nogo_log = out_nogo_log,
             gc_goopts = source.gc_goopts,
             cgo = False,
             testfilter = testfilter,
             recompile_internal_deps = recompile_internal_deps,
+        )
+
+    if go.nogo:
+        validations.append(out_nogo_validation)
+        go.actions.run_shell(
+            inputs = [out_nogo_log],
+            outputs = [out_nogo_validation],
+            # Create the expected output file if there are no errors, otherwise print the errors to stderr.
+            command = '[ -s "$1" ] && cat <(echo) "$1" <(echo) 1>&2 || touch $2',
+            arguments = [
+                out_nogo_log.path,
+                out_nogo_validation.path,
+            ],
+            progress_message = "Nogo Static Analysis",
+            mnemonic = "GoNogo",
         )
 
     data = GoArchiveData(
@@ -198,4 +223,5 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         cgo_exports = cgo_exports,
         runfiles = runfiles,
         mode = go.mode,
+        validations = validations,
     )
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -54,6 +54,7 @@ def emit_compilepkg(
         clinkopts = [],
         out_lib = None,
         out_export = None,
+        out_nogo_log = None,
         out_cgo_export_h = None,
         gc_goopts = [],
         testfilter = None,  # TODO: remove when test action compiles packages
@@ -112,6 +113,9 @@ def emit_compilepkg(
     if go.nogo:
         args.add("-nogo", go.nogo)
         inputs.append(go.nogo)
+        if out_nogo_log:
+            args.add("-nogo_log", out_nogo_log.path)
+            outputs.append(out_nogo_log)
     if out_cgo_export_h:
         args.add("-cgoexport", out_cgo_export_h)
         outputs.append(out_cgo_export_h)
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -51,7 +51,7 @@ func compilePkg(args []string) error {
 	goenv := envFlags(fs)
 	var unfilteredSrcs, coverSrcs, embedSrcs, embedLookupDirs, embedRoots, recompileInternalDeps multiFlag
 	var deps archiveMultiFlag
-	var importPath, packagePath, nogoPath, packageListPath, coverMode string
+	var importPath, packagePath, nogoPath, nogoLogPath, packageListPath, coverMode string
 	var outPath, outFactsPath, cgoExportHPath string
 	var testFilter string
 	var gcFlags, asmFlags, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags quoteMultiFlag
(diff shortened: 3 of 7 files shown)


fmeum:
Cc @sluongng

go/private/actions/archive.bzl
```diff
@@ -129,12 +138,28 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
             recompile_internal_deps = recompile_internal_deps,
         )
 
+    if go.nogo:
+        validations.append(out_nogo_validation)
+        go.actions.run_shell(
+            inputs = [out_nogo_log],
+            outputs = [out_nogo_validation],
+            # Create the expected output file if there are no errors, otherwise print the errors to stderr.
+            command = '[ -s "$1" ] && cat <(echo) "$1" <(echo) 1>&2 || touch $2',
```
sluongng, line 154: I don't like this approach. The actual nogo is still running as part of GoCompilePkg action, producing an output that is consumed by this no-ops command.

I would prefer if you start with separating `nogo` part of the go code into a separate `builder` sub-command, or ever a separate binary. Then in the validation action, we instrument that new subcommand/binary instead.

Not quite sure if it's a thing, but ideally user should be able to request for a `--output_group=_validation` and only trigger the validation actions to run linter, without much dependency on the compilation action.

go/tools/builders/compilepkg.go
```diff
@@ -449,18 +452,25 @@ func compileArchive(
...
+		nogoLogFile, err := os.OpenFile(nogoLogPath, os.O_CREATE|os.O_RDWR, 0644)
+		if err != nil {
+			return fmt.Errorf("error opening nogo log file: %v", err)
+		}
+		defer nogoLogFile.Close()
+		if len(nogoSrcs) > 0 {
+			ctx, cancel := context.WithCancel(context.Background())
+			nogoChan = make(chan error)
+			go func() {
+				nogoChan <- runNogo(ctx, workDir, nogoPath, nogoSrcs, deps, packagePath, importcfgPath, outFactsPath, nogoLogFile)
```
fmeum, line 465: Since `runNogo` effectively just runs a separate binary, have you considered moving that invocation into the validation action? That would allow us to fully remove nogo from the critical path. We may also not need the shell post-processing anymore.

sluongng (changes requested):
Another downside of the current approach will increase the maintenance complexity over time, by adding more features into compilepkg.go. I think it would also help you refactor CI tests much easier if you separate `nogo` logic to a new subcommand.

The idea of writing out an output file is good. However, I don't think it should be solely used to output errors. Instead, we should consider formats such as https://github.com/reviewdog/reviewdog?tab=readme-ov-file#reviewdog-diagnostic-format-rdformat so that we could let other tools consume these validation outputs and help the user fix their listing mistakes automatically (where possible).

joeljeske:
Hey @sluongng thanks for the review! I considered moving this out of the compilepkg action altogether, and I do think your are right; that is the best strategy long term. One issue right now is it seems though that we combine the nogo facts and the export data into a single .x archive that is used by downstream compilepkg actions, and thus the actions are still a bit coupled together.

I do not have the bandwidth right now to fully split these actions into its ideal long-term state. However, I would argue that this is still a net benefit over the current state, and has the proper abstractions in place we could improve behind in the future: a CompilePkg action that only fails on compile errors, and a separate Nogo validation action that fails on static analysis errors. 

Would you be willing to accept this patch that still provides many improvements over the current state and allow it to be improved later in the future?

sluongng:
> One issue right now is it seems though that we combine the nogo facts and the export data into a single .x archive that is used by downstream compilepkg actions, and thus the actions are still a bit coupled together.

Ah good point. There are a few other clean-ups to the .x and .a dependencies that folks do not have the bandwidth to act on right now.

> I do not have the bandwidth right now to fully split these actions into its ideal long-term state. However, I would argue that this is still a net benefit over the current state, and has the proper abstractions in place we could improve behind in the future: a CompilePkg action that only fails on compile errors, and a separate Nogo validation action that fails on static analysis errors but skipped with --norun_validations

I think there is a tradeoff being made here.
It's a strict feature improvement, but I think it's a downgrade for open-source maintenance overall.
Similar to adding a new feature without tests, we are simply delegating the verification and refactoring burden to future maintainers.

> Would you be willing to accept this patch that still provides many improvements over the current state and allow it to be improved later in the future?

I thought about accepting this PR if we could somehow "feature flag" this new behavior and make it opt-in instead of opt-out. However, in cases such as https://github.com/bazelbuild/rules_go/issues/3063#issuecomment-1151363481 and some private conversations I have had, folks to maintain downstream patch(es) to `nogo`.

So this PR will be a `permanent until replaced` kind of change and will have a downstream impact, which causes churn for users. So it's a no for me until the concerns I have stated above are addressed (or until we have more users demanding this change). If other maintainers are willing to accept this PR, I won't object 🤗.

fmeum:
> I thought about accepting this PR if we could somehow "feature flag" this new behavior and make it opt-in instead of opt-out. However, in cases such as [#3063 (comment)](https://github.com/bazelbuild/rules_go/issues/3063#issuecomment-1151363481) and some private conversations I have had, folks to maintain downstream patch(es) to `nogo`.

@sluongng Could you elaborate on this? The linked patch mostly touches `nogo_main.go`, not the Starlark or `compilepkg` logic for `nogo`. The current PR only makes a tiny change to `nogo_main.go`, so I wouldn't expect that to cause much churn.

sluongng:
It was simply an example of folks patching `nogo` code paths in their own system for different purposes.

I don't have a public example, but this is more common in enterprises that use Go and rules_go and want to have tight integration between the linter and code review system to make automatic changes suggestions from CI runs.

fmeum:
@joeljeske I want to get back to this, but won't be able to look into the foundations (`.x` file generation) in more detail before BazelCon.

joeljeske:
Sure! Is there anything I could do for you in the meantime?

fmeum:
> Sure! Is there anything I could do for you in the meantime?

It would definitely help if anyone could solve:

> One issue right now is it seems though that we combine the nogo facts and the export data into a single .x archive that is used by downstream compilepkg actions, and thus the actions are still a bit coupled together.

But I don't want to make that a requirement for clear incremental improvements.

go/private/actions/archive.bzl
```diff
@@ -129,12 +138,28 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
+            out_nogo_log = out_nogo_log,
             gc_goopts = source.gc_goopts,
             cgo = False,
             testfilter = testfilter,
             recompile_internal_deps = recompile_internal_deps,
         )
 
+    if go.nogo:
+        validations.append(out_nogo_validation)
+        go.actions.run_shell(
```
fmeum, line 150: We try to avoid relying on `run_shell` on Windows. Can you think of a way to replace this with a `run`? 

For example, could this logic be moved into https://github.com/bazelbuild/rules_go/blob/master/go/tools/builders/generate_nogo_main.go?

go/private/actions/archive.bzl
```diff
@@ -129,12 +138,28 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
+            inputs = [out_nogo_log],
+            outputs = [out_nogo_validation],
+            # Create the expected output file if there are no errors, otherwise print the errors to stderr.
+            command = '[ -s "$1" ] && cat <(echo) "$1" <(echo) 1>&2 || touch $2',
+            arguments = [
+                out_nogo_log.path,
+                out_nogo_validation.path,
+            ],
+            progress_message = "Nogo Static Analysis",
```
fmeum, line 159: This is slightly misleading as the action just processes the results, it doesn't analyze anything. 

Also, purely from a stylistic point of view, progress messages typically start with a verb in progressive form, e.g. `Post-processing nogo findings`.

go/private/actions/archive.bzl
```diff
@@ -129,12 +138,28 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
+            mnemonic = "GoNogo",
```
fmeum, line 160: In the future in which we have moved out the nogo analysis into its own action, this is exactly the mnemonic we would like to use for that. Since mnemonics are public API, please use something else here.

go/private/actions/compilepkg.bzl
```diff
@@ -112,6 +113,9 @@ def emit_compilepkg(
     if go.nogo:
         args.add("-nogo", go.nogo)
         inputs.append(go.nogo)
+        if out_nogo_log:
```
fmeum, line 116: Can this ever validly be `None`? If not, let's `fail` if it is.

go/tools/builders/compilepkg.go
fmeum: I will look into this as a follow-up.

go/tools/builders/compilepkg.go
```diff
@@ -449,18 +452,25 @@ func compileArchive(
...
+			}()
```
fmeum, line 466: This looks suspicious to me - don't we need explicit synchronisation to wait on the `runNogo` call here?

fmeum (commented):
Sorry for the radio silence. Now that we have figured out how to make nogo work with Bzlmod, I have the capacity to look into this.

fmeum:
Update: I submitted https://github.com/bazelbuild/rules_go/pull/3789 to extract nogo facts into separate files. I will look into clearing further blockers for the separation of nogo into a separate action.

PR #3711, opened 2023-09-26
use dedicated exec group for link actions

sluongng:
When remote execution is used, a user may want to use different
exec properties for different actions under the same target.

For example:

A go_test could include 3 different actions:
- GoCompilePkg
- GoLink
- GoTest

If users wish to use a special exec_properties for GoLink actions,
after this PR, they could do so with:

```python
go_test(
  ...
  exec_properties = {
    # Default
    "container-image": "docker://alpine",
    "EstimatedMemory": "500MB",

    # Only applied to GoLink exec_group
    "GoLink.container-image": "docker://ubuntu",
    "GoLink.EstimatedMemory": "2GB",
  }
  ...
)
```

Reference: https://bazel.build/extending/exec-groups

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/actions/BUILD.bazel
+++ b/go/private/actions/BUILD.bazel
@@ -55,6 +55,7 @@ bzl_library(
         "//go/private:common",
         "//go/private:mode",
         "//go/private:rpath",
+        "//go/private/rules:execgroup",
         "@bazel_skylib//lib:collections",
     ],
 )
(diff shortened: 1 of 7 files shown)


sluongng:
This seems to be a breaking change for downstream rules that consume rules_go action directly (i.e. gazelle_binary.bzl).

The solution is those rules must declare exec_groups in the rules declaration in order to consume the action 🤔 

I don't really see a way around it.

PR #3722, opened 2023-10-10
Include `go.env` in `//go` runfiles

fmeum:
Speculative fix for user reports of this error message showing up when using `//go`, which indicates that `go.env` is missing:

```
GOPROXY list is not the empty string, but contains no entries
```

Adding `go.env` to the `tools` files of the SDK rather than creating a new field on `GoSDK` for it may end up fixing latent bugs.

https://bazelbuild.slack.com/archives/CDBP88Z0D/p1696892815685419

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -31,7 +31,17 @@ filegroup(
     srcs = glob([
         "pkg/tool/**",
         "bin/gofmt*",
-    ]),
+    ]) + [
+        ":config",
+    ],
+)
+
+filegroup(
+    name = "config",
+    srcs = glob(
+        ["go.env*"],
+        allow_empty = True,
+    ),
 )
 
 go_sdk(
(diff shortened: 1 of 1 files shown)


fmeum:
Has been confirmed to fix the issue in https://bazelbuild.slack.com/archives/CDBP88Z0D/p1697043722156949?thread_ts=1696892815.685419&cid=CDBP88Z0D.

@tyler-french I will merge since this seems trivial enough and has been tested, please take a look when you have the time.

sluongng:
Ran into this issue today... digged around a bunch and ended up finding your PR here 🙃 

For future curious readers: we initially included the `go.env` file to `filegroup(name = "files")` under a downloaded Go SDK's BUILD file. However, the action `GoToolchainBinaryBuild` does not use this filegroup. Instead, it uses the `libs`, `headers`, `srcs` and `tools` filegroup in the same BUILD file.

This PR fixes it so that we include `go.env` to both `files` and `tools` filegroup, which fixes `GoToolchainBinaryBuild` action.

---

@fmeum this might be worth a patch release. WDYT?

fmeum:
Patch release not sure, but minor release without obvious breaking changes probably yes. Which means that we might want to hold onto the tools update. Sigh.

cgrindel:
@fmeum It would be great to get this released. I just ran into it trying to upgrade my rulesets. Because they are rulesets, I cannot rely on an override.

Issue #3597, opened 2023-06-20
ERROR FAILED: //tests/core/transition:hermeticity_test  with Bazel@HEAD in Downstream CI


sgowroji:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3109#0188d690-a57f-42b0-891f-8b2986d9db52

Platform: Windows, MacOs, Ububtu

Logs : 
```
FAIL: //tests/core/transition:hermeticity_test 
```

Steps: 
```
git clone https://github.com/bazelbuild/rules_go.git 
git reset 9d5416bcc9d48da3735b73fb5540b1cdbd470b85 --hard
export USE_BAZEL_VERSION=211fd6c14bef2f656265b3d3eb98a476f02ebf88
bazel build //...
```

Reference issue, which is causing error : https://github.com/bazelbuild/bazel/commit/cb558978820b1fb9b579d8e9578cbb9ca590fe60

CC Green Team @ fweikert

fmeum:
@sgowroji How certain are you that the given commit is the culprit? Was there some kind of bisection that I could look into the results of?

meteorcloudy:
It got the result from https://buildkite.com/bazel/culprit-finder/builds/5736#0188d70d-c9dc-438e-8119-76650a30b54c, which seems to be the same issue reported in https://github.com/bazelbuild/bazel/issues/18666

fmeum:
@meteorcloudy I think that there may be two different issues at play that confused the culprit finder. The log linked in the issue has a different kind of error: It looks like the output of `cquery --output=jsonproto` changed to new-line delimited JSON.

fmeum:
Maybe this is caused by https://github.com/bazelbuild/bazel/commit/607d0f7335f95aa0ee236ba3c18ce2a232370cdb?

meteorcloudy:
Oh, you are probably right, I forgot to check the actual error in the downstream pipeline.

zhengwei143:
Sorry I just saw this, let me take a quick look.

zhengwei143:
I've just rolled back the change in https://github.com/bazelbuild/bazel/commit/dba9e43a26d41c2559568e4abfdece57e79de37d which should fix this.

sgowroji:
We are also seeing this test failiure in the latest CI https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3160#01894849-482e-4d46-9522-beab4a40754e

```
[...]

PR #3729, opened 2023-10-17
Work around Bazel@HEAD bug breaking `hermeticity_test`

fmeum:
`bazel config` doesn't work correctly with Bzlmod, which is enabled by default with Bazel@HEAD.

Fixes https://github.com/bazelbuild/rules_go/issues/3597
See https://github.com/bazelbuild/bazel/issues/19823


fmeum:
FYI @linzhp @tyler-french

fmeum:
I have no idea why, but the test is failing when I unflip Bzlmod, which used to be disabled. Very weird.

fmeum:
I will close this PR as I found a fix for the upstream issue: https://github.com/bazelbuild/bazel/pull/19842.

Issue #3732, opened 2023-10-18
FEATURE: Include data in go_proto_compiler


louisbuchbinder:
## Feature Request:
Include data in the go_proto_compiler rule.

## Motivation:
Currently the go_proto_compiler rule accepts custom proto compiler executables, but not the ability to pass custom input data in to the compiler runtime.

### What version of rules_go are you using?
v0.42.0

### What version of gazelle are you using?
v0.33.0

### What version of Bazel are you using?
Bazelisk version: development
Build label: 6.3.2

PR #3733, opened 2023-10-18
allow data input to go_proto_compiler

louisbuchbinder:
**What type of PR is this?**
Feature

Currently the go_proto_compiler rule accepts custom proto compiler executables, but not the ability to pass custom input data in to the compiler runtime.

This change adds a `data` attribute to the `go_proto_compiler` rule. The data files provided are then appended to the input file list in the `go.actions.run` where the proto compiler is executed.

Fixes https://github.com/bazelbuild/rules_go/issues/3732

**Other notes for review**
Let me know what you are looking for regarding tests and I would be happy to add.

### DIFF
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -117,13 +117,14 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
     args.add_all(imports, before_each = "-import")
     args.add_all(proto_paths.keys())
     args.use_param_file("-param=%s")
+    data = [file for target in compiler.internal.data for file in target.files.to_list()]
     go.actions.run(
         inputs = depset(
             direct = [
                 compiler.internal.go_protoc,
                 compiler.internal.protoc,
                 compiler.internal.plugin,
-            ],
+            ] + data,
             transitive = [transitive_descriptor_sets],
         ),
         outputs = go_srcs,
@@ -176,6 +177,7 @@ def _go_proto_compiler_impl(ctx):
             compile = go_proto_compile,
             valid_archive = ctx.attr.valid_archive,
             internal = struct(
+                data = ctx.attr.data,
                 options = ctx.attr.options,
                 suffix = ctx.attr.suffix,
                 suffixes = ctx.attr.suffixes,
@@ -192,6 +194,7 @@ def _go_proto_compiler_impl(ctx):
 _go_proto_compiler = rule(
     implementation = _go_proto_compiler_impl,
     attrs = {
+        "data": attr.label_list(),
         "deps": attr.label_list(providers = [GoLibrary]),
         "options": attr.string_list(),
         "suffix": attr.string(default = ".pb.go"),


proto/compiler.bzl
```diff
@@ -117,13 +117,14 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
     args.add_all(imports, before_each = "-import")
     args.add_all(proto_paths.keys())
     args.use_param_file("-param=%s")
+    data = [file for target in compiler.internal.data for file in target.files.to_list()]
```
fmeum, line 120: This is problematic as it flattens a `depset` with `to_list()`, which should always be avoided. 

Instead, could you collect the individual `target.files` in a list and pass them to the `transitive` parameter of `depset` below?

louisbuchbinder: @fmeum Thanks for the tip. Updated as you described.

proto/compiler.bzl
```diff
@@ -118,14 +118,15 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
     args.add_all(imports, before_each = "-import")
     args.add_all(proto_paths.keys())
     args.use_param_file("-param=%s")
+    data_depsets = [target.files for target in compiler.internal.data]
```
fmeum, line 121: How does your compiler learn of the paths of these extra files? This will add them to the sandbox, but it doesn't tell the compiler or plugin where to look for them. Paths of generated files have the `bazel-out/k8-exec-ST-some-hash` bit in them which makes it impossible to hardcode them.

I'm wondering whether it would be better to add them to the runfiles of the `plugin` executable instead, which would allow the plugin to find them using their runfiles path via the Go runfiles library.

louisbuchbinder: The idea of this change is that the data targets become available during the runtime of the compiler. In my case the compiler knows where to look for a file based on the name of the proto file being processed. The file paths of the data targets are available as relative paths from the working directory when the compiler is executed.

fmeum: I see, so the data files required depend on the input files and aren't fixed for this particular plugin?

louisbuchbinder: That is correct. In my case I am doing two protoc passes. Both use the same `.proto` input files, but the second uses the output of the first as the data

louisbuchbinder:
Bumping this one. Please let me know if this change is acceptable

proto/compiler.bzl
fmeum: Sorry for the delay, I forgot to reply.

I'm still not sure I understand how this can work, but I guess it does :⁠-⁠) I have some reservations about merging this as the `data` attributes of executable rules almost universally end up adding *runfiles* to actions using the executable, not *input files*. I would feel much more comfortable supporting adding runfiles to the compiler.

Have you been able to try using runfiles for your purpose instead? That would also resolve the questions of how the file paths are known at execution time: runfiles paths don't contain the `bazel-out` path segments.

PR #3740, opened 2023-10-30
Add toolchain param to affected actions

kotlaja:
**What type of PR is this?**

Other - Migration

It sets go toolchain type inside affected actions. This is part of the migration of Automatic Exec Groups inside Bazel downstream.

Please check https://github.com/bazelbuild/bazel/issues/19981 for additional info.

**Other notes for review**

### DIFF
--- a/go/private/actions/stdlib.bzl
+++ b/go/private/actions/stdlib.bzl
@@ -82,6 +82,7 @@ def _build_stdlib_list_json(go):
         executable = go.toolchain._builder,
         arguments = [args],
         env = _build_env(go),
+        toolchain = "@io_bazel_rules_go//go:toolchain",
     )
     return out
 
@@ -149,6 +150,7 @@ def _build_stdlib(go):
         executable = go.toolchain._builder,
         arguments = [args],
         env = _build_env(go),
+        toolchain = "@io_bazel_rules_go//go:toolchain",
     )
     return GoStdLib(
         _list_json = _build_stdlib_list_json(go),
--- a/go/private/actions/stdlib.bzl
+++ b/go/private/actions/stdlib.bzl
@@ -15,6 +15,7 @@
 load(
     "//go/private:common.bzl",
     "COVERAGE_OPTIONS_DENYLIST",
+    "GO_TOOLCHAIN",
 )
 load(
     "//go/private:providers.bzl",
@@ -82,7 +83,7 @@ def _build_stdlib_list_json(go):
         executable = go.toolchain._builder,
         arguments = [args],
         env = _build_env(go),
-        toolchain = "@io_bazel_rules_go//go:toolchain",
+        toolchain = GO_TOOLCHAIN,
     )
     return out
 
@@ -150,7 +151,7 @@ def _build_stdlib(go):
         executable = go.toolchain._builder,
         arguments = [args],
         env = _build_env(go),
-        toolchain = "@io_bazel_rules_go//go:toolchain",
+        toolchain = GO_TOOLCHAIN,
     )
     return GoStdLib(
         _list_json = _build_stdlib_list_json(go),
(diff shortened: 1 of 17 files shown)


go/private/actions/stdlib.bzl
```diff
@@ -82,6 +82,7 @@ def _build_stdlib_list_json(go):
         executable = go.toolchain._builder,
         arguments = [args],
         env = _build_env(go),
+        toolchain = "@io_bazel_rules_go//go:toolchain",
```
kotlaja, line 85: I tried loading and using https://github.com/bazelbuild/rules_go/blob/master/go/private/go_toolchain.bzl#L26 instead of hardcoded string, but that creates a cycle:
 `//go/private:go_toolchain.bzl` ->  `//go/private/actions:stdlib.bzl` -> `//go/private:go_toolchain.bzl`
 I can break the cycle by moving `GO_TOOLCHAIN` to another file. WDYT?

fmeum: Sounds reasonable, please go ahead!

kotlaja: I've moved it inside `go/private:common.bzl` and updated it inside files where its been used. Can you have another look? Thanks!

fmeum (approved):

go/private/actions/stdlib.bzl
fmeum: Thanks!

kotlaja:
Hey @fmeum, thanks for helping out here! Also, do you maybe know when will be the next release? I'm interested in when I'll be able to use this change inside rules_cc. Thanks! :)

fmeum:
We can probably do one this or next week. 

Cc @tyler-french

fmeum:
@kotlaja rules_go 0.43.0 is out now

anguslees:
For visibility, I think this PR (and the followup #3760) are incomplete.  Golang actions need _both_ go and cpp toolchains available - when cgo is enabled.  The AEG [docs](https://bazel.build/extending/auto-exec-groups) say this needs to be done by declaring a new exec_group that mentions both toolchains, so platform resolution is able to take both into account:

> If you need to use multiple toolchains on a single execution platform (an action uses executable or tools from two or more toolchains), you need to manually define [exec_groups](https://bazel.build/extending/exec-groups) (check [When should I use a custom exec_group?](https://bazel.build/extending/auto-exec-groups#when-should-use-exec-groups) section).

kotlaja:
Thanks, @anguslees, for catching this! I will take a look next week.

Yes, you're right, when we're using more than 1 toolchain in an action, we must define a custom exec group for it. Ironically, I wrote that documentation you're referring to, used it in other rules, but missed this case with cgo in go rules. Will take another look. Thanks again!

anguslees:
> Thanks, @anguslees, for catching this! I will take a look next week.
> 
> Yes, you're right, when we're using more than 1 toolchain in an action, we must define a custom exec group for it. Ironically, I wrote that documentation you're referring to, used it in other rules, but missed this case with cgo in go rules. Will take another look. Thanks again!

@kotlaja : Was this followup fix done?  From a quick skim of the code, I don't think so?

kotlaja:
anguslees@ I haven't tried fixing it since changing subteams inside Bazel steered my focus to some other stuff. Sorry for not letting you know.

Fabian did some work for rules_go recently (https://github.com/bazel-contrib/rules_go/commit/b17435b816b36e9776e73879e973a148d15f82f5). @fmeum did you maybe take a look at `Golang actions need both go and cpp toolchains available - when cgo is enabled`?

fmeum:
I'm working on this now. It's not just the `toolchain` attribute that is wrong though, I need to fundamentally change how rules_go accesses the C++ toolchain.

PR #3743, opened 2023-11-01
feat(gopackagesdriver): add base test case for go packages driver

JamyDev:
**What type of PR is this?**

Other - Test cases

Validates the basic functionality for the Go Packages Driver, and sets up a framework to add additional tests in the future.

Two major changes were needed to support all testing platforms:
- For bazel <6.0 we need to use the old style labels. Currently it would always prefix with @ and fail the lookup on bazel 5.
- For Windows we need to adjust the paths returned from Bazel. I think because we're running inside MINGW the paths are unix like where everything starts with `\`, but `since` Bazel runs in Java it uses regular Windows paths (eg `C:\<some>\<path>`). So if we see a path like `\C:\<some>\<path>` we have to remove the leading `\` for go to be able tom resolve this path.

This should help contributors validate that changes they make to this functionality and hopefully means less regressions overall.

**Other notes for review**

The goal is to test more scenarios, but wanted to send in the initial approach for review.

### DIFF
--- a/go/tools/bazel_testing/bazel_testing.go
+++ b/go/tools/bazel_testing/bazel_testing.go
@@ -219,6 +219,28 @@ func BazelOutput(args ...string) ([]byte, error) {
 	return stdout.Bytes(), err
 }
 
+// BazelOutputWithInput invokes a bazel command with a list of arguments and
+// an input stream and returns the content of stdout.
+//
+// If the command starts but exits with a non-zero status, a *StderrExitError
+// will be returned which wraps the original *exec.ExitError.
+func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, error) {
+	cmd := BazelCmd(args...)
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+	if stdin != nil {
+		cmd.Stdin = stdin
+	}
+	err := cmd.Run()
+	if eErr, ok := err.(*exec.ExitError); ok {
+		eErr.Stderr = stderr.Bytes()
+		err = &StderrExitError{Err: eErr}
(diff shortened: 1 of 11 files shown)


fmeum:
Thank you very much, this was sorely lacking!

Just mention me when you think this is ready. CI is unfortunately still failing.

JamyDev:
Thanks yeah, it was passing locally so I'll investigate today!

JamyDev:
Ready to be reviewed @fmeum

go/tools/bazel_testing/bazel_testing.go
```diff
@@ -219,6 +219,28 @@ func BazelOutput(args ...string) ([]byte, error) {
 	return stdout.Bytes(), err
 }
 
+// BazelOutputWithInput invokes a bazel command with a list of arguments and
+// an input stream and returns the content of stdout.
+//
+// If the command starts but exits with a non-zero status, a *StderrExitError
+// will be returned which wraps the original *exec.ExitError.
+func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, error) {
```
fmeum, line 227: Can we implement `BazelOutput` in terms of this new method?

fmeum (approved):
This is great, thank you very much! 
The bug fixes may even end up resolving some open issues.

go/tools/bazel_testing/bazel_testing.go
JamyDev: Do you want me to just modify BazelOutput so it takes stdin by default and then nil it out in all the calls to it?

fmeum: Ah, no, I just wanted to suggest implementing `BazelOutput` as `BazelOutputWithInput(nil, ...)` so that the two implementations don't deviate over time.

PR #3749, opened 2023-11-09
Panic when a test times out

### LINKED ISSUE #3746: Improvement/feature request: consider allowing process under test to timeout BEFORE Bazel-level `--test_timeout`
Under normal circumstances it is highly desirable to let the binary under test in a `go_test` rule time out *before Bazel allows the test to timeout*. This is because when the binary under test times out (e.g. using `-test.timeout`), it dumps stacktraces and some other potentially useful info.

In `cockroach` we use [this commit](https://github.com/cockroachdb/rules_go/commit/1360c0a3a4a1b3ff12ed685013e34b37f99fdf01) which checks the `TEST_TIMEOUT` and, when spawning the test subprocess, passes `-test.timeout N` where `N` is 3 seconds *less* than the `TEST_TIMEOUT`. (The 3 second figure is selected arbitrarily and is meant to give "enough time" for the binary to clean up.) This means that if you run `bazel test :test_target --test_timeout 60`, then the *actual* timeout will be 57 seconds (as 60 - 3 = 57).

While the disparity between the `go test`-level timeout and the Bazel-level timeout may be confusing at first, it creates a hugely superior UX. Compare the `test.log` output from a timed-out test without that patch:

```
Executing tests from //pkg/cmd/dev:dev_test
=== RUN   TestDataDriven
-- Test timed out at 2023-11-07 17:49:38 UTC --
```

This says almost nothing except that a timeout occurred, and the test is not given any time to clean up before being unceremoniously killed. Compare that to *with* the patch:

```
Executing tests from //pkg/cmd/dev:dev_test
=== RUN   TestDataDriven
panic: test timed out after 2s
running tests:
	TestDataDriven (2s)

goroutine 7 [running]:
testing.(*M).startAlarm.func1()
	GOROOT/src/testing/testing.go:2259 +0x320
created by time.goFunc
	GOROOT/src/time/sleep.go:176 +0x38

goroutine 1 [chan receive]:
testing.(*T).Run(0x140000a7380, {0x1051300c0?, 0x66f0a118049?}, 0x10531a668)
...
```

Now we get stack traces.

This change carries two "regressions" in the UX:
1. The timeout you pass to `--timeout` is not the real timeout, and the "real" timeout is a few seconds shorter. In `cockroach` we have gotten used to this and factor it in, though the behavior could be confusing for others.
2. Timed-out tests show as `FAILED` instead of `TIMEOUT` in the Bazel-produced test summary which may be technically inaccurate.

For my taste, neither of these "regressions" outweighs the benefit gained from allowing the test process time to gracefully clean up, report stack traces, etc.

`release-0.42`

`v0.33.0`

6.2.1

Yes

MacOS/ARM64

No

fmeum:
**What type of PR is this?**

Feature

By calling `panic` when a Bazel test times out, which is indicated by Bazel sending a SIGTERM, stack traces of all Go routines are printed, similar to the behavior of native `go test`.

Fixes #3746

**Other notes for review**

### DIFF
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -23,10 +23,12 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/bazelbuild/rules_go/go/tools/bzltestutil/chdir"
 )
@@ -67,9 +69,9 @@ func shouldAddTestV() bool {
 // ensure that one line at a time is written to the inner writer.
 type streamMerger struct {
 	OutW, ErrW *io.PipeWriter
-	mutex sync.Mutex
-	inner io.Writer
-	wg sync.WaitGroup
+	mutex      sync.Mutex
+	inner      io.Writer
+	wg         sync.WaitGroup
 	outR, errR *bufio.Reader
 }
 
@@ -78,10 +80,10 @@ func NewStreamMerger(w io.Writer) *streamMerger {
 	errR, errW := io.Pipe()
 	return &streamMerger{
 		inner: w,
-		OutW: outW,
-		ErrW: errW,
-		outR: bufio.NewReader(outR),
-		errR: bufio.NewReader(errR),
+		OutW:  outW,
+		ErrW:  errW,
+		outR:  bufio.NewReader(outR),
+		errR:  bufio.NewReader(errR),
 	}
 }
 
@@ -126,6 +128,17 @@ func Wrap(pkg string) error {
 	cmd.Env = append(os.Environ(), "GO_TEST_WRAP=0")
 	cmd.Stderr = io.MultiWriter(os.Stderr, streamMerger.ErrW)
 	cmd.Stdout = io.MultiWriter(os.Stdout, streamMerger.OutW)
+	go func() {
+		// When the Bazel test timeout is reached, Bazel sends a SIGTERM that
+		// we need to forward to the inner process.
+		// TODO: This never triggers on Windows, even though Go should simulate
+		//  a SIGTERM when Windows asks the process to close. It is not clear
+		//  whether Bazel uses the required graceful shutdown mechanism.
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, syscall.SIGTERM)
+		<-c
+		cmd.Process.Signal(syscall.SIGTERM)
+	}()
 	streamMerger.Start()
 	err := cmd.Run()
 	streamMerger.ErrW.Close()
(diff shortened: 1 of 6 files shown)


fmeum:
CC @rickystewart

rickystewart:
Ping on this :) IMO this is a colossal UX improvement over the status quo.

go/tools/bzltestutil/wrap.go
```diff
@@ -126,6 +128,17 @@ func Wrap(pkg string) error {
 	cmd.Env = append(os.Environ(), "GO_TEST_WRAP=0")
 	cmd.Stderr = io.MultiWriter(os.Stderr, streamMerger.ErrW)
 	cmd.Stdout = io.MultiWriter(os.Stdout, streamMerger.OutW)
+	go func() {
```
tyler-french, line 131: optional: we could abstract this to a function:
```
go signalForward(syscall.SIGTERM)
```
but maybe this would hurt readibility

tyler-french (approved):
LGTM. Nice work making the signal channel have a buffer of 1. I do wonder if there is a more idiomatic way to handle a stack trace with [runtime.Stack](https://pkg.go.dev/runtime#Stack) and `os.Exit()`, but `panic` is probably good for simplicity.

fmeum:
> I do wonder if there is a more idiomatic way to handle a stack trace with [runtime.Stack](https://pkg.go.dev/runtime#Stack) and `os.Exit()`, but `panic` is probably good for simplicity.

Since the `testing` package itself uses `panic`, I see value in using the same technique. Unfortunately we can't get the list of running tests as it does appear to be internal to that package.

timonwong:
FYI: I came across goleak problem after rules_go v0.44.0.

Since bzltestutil.RegisterTimeoutHandler() is registered before TestMain, I have to use `goleak.IgnoreCurrent()` to workaround that, for example:

```go
func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m, goleak.IgnoreCurrent())
}
```

hawkingrei:
> ```go
> IgnoreCurrent
> ```

I meet this problem, too.

@fmeum PTAL

hawkingrei:
> FYI: I came across goleak problem after rules_go v0.44.0.
> 
> Since bzltestutil.RegisterTimeoutHandler() is registered before TestMain, I have to use `goleak.IgnoreCurrent()` to workaround that, for example:
> 
> ```go
> func TestMain(m *testing.M) {
> 	goleak.VerifyTestMain(m, goleak.IgnoreCurrent())
> }
> ```

I meet this problem, too.

@fmeum PTAL

fmeum:
This will be addressed via a default ignore rule in a future goleak release - rules_go can't cancel the go routine in time for the leak check. 

Cc @linzhp

linzhp:
You can call [IgnoreTopFunction](https://pkg.go.dev/go.uber.org/goleak#IgnoreTopFunction) too, which is less dependent on lifetime of the goroutine that rules_go creates

PR #3750, opened 2023-11-10
Add support for `xtest` packages in Go Packages Driver

JamyDev:
**What type of PR is this?**

Feature

xtests, also named external tests, are test cases that validate the public facing API of a package, even though the files are still located in the same package.

A sample folder structure:

> code.go -> `package code` -> has both public and private methods/interfaces of things
> code_test.go -> `package code` -> tests the private logic
> code_ext_test.go -> `package code_test` -> tests the public API

From what I gather, the go compiler is happy to handle this since the package names are different, but when the Go packages driver runs its aspect, we'll get all of this in one FlatPackage (since they're part of the `go_test` rule) with the same package ID. This causes Go internally to mark the `package code_test` as inconsistent with the `package code` of the other test files.

The solution:

- Read the package id field from any file that has an `_test.go` suffix, if it doesn't match the package id provided by the FlatPackage, separate it out into a list of external test files.
- If we see that we have more than 0 xtestfiles, we create a new xtest package to add to the registry.
- Finally, we need to ensure that this xtest package gets added to the roots if we created one. We do this unconditionally, since if you're trying to resolve the tests of a package, you'd likely also want the external tests.

### DIFF
--- a/go/tools/bazel_testing/bazel_testing.go
+++ b/go/tools/bazel_testing/bazel_testing.go
@@ -206,7 +206,8 @@ func RunBazel(args ...string) error {
 // If the command starts but exits with a non-zero status, a *StderrExitError
 // will be returned which wraps the original *exec.ExitError.
 func BazelOutput(args ...string) ([]byte, error) {
-	return BazelOutputWithInput(nil, args...)
+	stdout, _, err := BazelOutputWithInput(nil, args...)
+	return stdout, err
 }
 
 // BazelOutputWithInput invokes a bazel command with a list of arguments and
@@ -214,7 +215,7 @@ func BazelOutput(args ...string) ([]byte, error) {
 //
 // If the command starts but exits with a non-zero status, a *StderrExitError
 // will be returned which wraps the original *exec.ExitError.
-func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, error) {
+func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, []byte, error) {
 	cmd := BazelCmd(args...)
 	stdout := &bytes.Buffer{}
 	stderr := &bytes.Buffer{}
@@ -228,7 +229,7 @@ func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, error) {
 		eErr.Stderr = stderr.Bytes()
 		err = &StderrExitError{Err: eErr}
 	}
-	return stdout.Bytes(), err
+	return stdout.Bytes(), stderr.Bytes(), err
 }
 
 // StderrExitError wraps *exec.ExitError and prints the complete stderr output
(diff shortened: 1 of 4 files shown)


go/tools/bazel_testing/bazel_testing.go
```diff
@@ -206,15 +206,16 @@ func RunBazel(args ...string) error {
...
+	return stdout, err
 }
 
 // BazelOutputWithInput invokes a bazel command with a list of arguments and
 // an input stream and returns the content of stdout.
 //
 // If the command starts but exits with a non-zero status, a *StderrExitError
 // will be returned which wraps the original *exec.ExitError.
-func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, error) {
+func BazelOutputWithInput(stdin io.Reader, args ...string) ([]byte, []byte, error) {
```
fmeum, line 218: I can see this being a useful change, but it looks like it is unused by this PR. Is that correct?

tests/integration/gopackagesdriver/gopackagesdriver_test.go
```diff
@@ -124,3 +147,37 @@ func TestBaseFileLookup(t *testing.T) {
...
+		if strings.HasSuffix(id, "_xtest") {
+			xTestId = id
+		} else {
+			testId = id
+		}
+	}
+
+	for _, p := range resp.Packages {
+		if p.ID == xTestId && !strings.HasSuffix(p.GoFiles[0], "external_test.go") {
+			t.Errorf("Expected only one file in _xtest pkg, got %+v", p.GoFiles)
```
fmeum, line 178: It looks like we aren't testing the cardinality here.

tests/integration/gopackagesdriver/gopackagesdriver_test.go
```diff
@@ -124,3 +147,37 @@ func TestBaseFileLookup(t *testing.T) {
...
+		} else if p.ID == testId && len(p.GoFiles) != 2 {
+			t.Errorf("Expected only 2 files in _test pkg, got %+v", p.GoFiles)
```
fmeum, line 180: Can we also check what these two files are?

tests/integration/gopackagesdriver/gopackagesdriver_test.go
JamyDev: We are only supposed to have one file in here, not sure about the necessity to check cardinality. Will add a len() == 1 in the check as well.

go/tools/bazel_testing/bazel_testing.go
JamyDev: Correct, I did use it while debugging why the test was failing, but ended up not needing to use it to actually validate the change. I may add an "after" handler that outputs this in case of a failure.

fmeum (approved):

PR #3753, opened 2023-11-14
fix `aux_files` relative paths for gomock source mocks (fix #3752)

### LINKED ISSUE #3752: `aux_files` incorrect behavior in source gomocks

0.42.0

0.32.0

6.4.0

yup!

darwin/arm64, but I can reproduce it on linux/amd64 too 

the issue can be reproduced on the go toolchain with cgo/cross-compilation disabled

### What did you do?

If we add the following code to the [gomock rule test suite](https://github.com/bazelbuild/rules_go/tree/master/tests/extras/gomock):

```go
# client_wrapper.go

package client

type ClientWrapper interface {
	Client
	Close() error
}
```

```go
# client_test.go

package client

var _ Client = (*MockClient)(nil)
var _ ClientWrapper = (*MockClientWrapper)(nil)
```

```starlark
# BUILD.bazel

gomock(
    name = "wrapper_mocks",
    out = "wrapper_mock.go",
    library = ":client",
    package = "client",
    self_package = "github.com/bazelbuild/rules_go/gomock/client",
    source = "client_wrapper.go",
)

# + use the wrapper_mock.go in the test
```

The `bazel build //tests/extras/gomock:wrapper_mocks` will legitimately fail with 
```
2023/11/14 13:43:23 Loading input failed: bazel-out/darwin_arm64-fastbuild/bin/tests/extras/gomock/gopath/src/github.com/bazelbuild/rules_go/gomock/client/client_wrapper.go:4:2: could not parse package github.com/bazelbuild/rules_go/gomock/client: go/build: go list github.com/bazelbuild/rules_go/gomock/client: fork/exec GOROOT/bin/go: no such file or directory
```

`mockgen` needs the `client.go` in the bazel's `gopath` and the only method to do it hermetically right now is to use `aux_files` like this:
```starlark
# BUILD.bazel

gomock(
    name = "wrapper_mocks",
    out = "wrapper_mock.go",
    aux_files = {
        "client.go": "github.com/bazelbuild/rules_go/gomock/client",
    },
    library = ":client",
    package = "client",
    self_package = "github.com/bazelbuild/rules_go/gomock/client",
    source = "client_wrapper.go",
)
```

Which is what I'm currently trying to do.

`bazel build //tests/extras/gomock:wrapper_mocks` works and `bazel test //tests/extras/gomock:client_test` passes, when the changes above are applied. 

### What did you see instead?

```go
2023/11/14 13:44:48 Loading input failed: open bazel-out/darwin_arm64-fastbuild/bin/tests/extras/gomock/bazel-out/darwin_arm64-fastbuild/bin/tests/extras/gomock/gopath/src/github.com/bazelbuild/rules_go/gomock/client/client.go: no such file or directory
```

### Further details

I think the following patch to the `gomock` rule fixes the issue:

```go
diff --git a/extras/gomock.bzl b/extras/gomock.bzl
index 3046c9f1..21149c6a 100644
--- a/extras/gomock.bzl
+++ b/extras/gomock.bzl
@@ -55,7 +55,7 @@ def _gomock_source_impl(ctx):
         aux_files = []
         for target, pkg in ctx.attr.aux_files.items():
             f = target.files.to_list()[0]
-            aux = ctx.actions.declare_file(paths.join(gopath, "src", pkg, f.basename))
+            aux = ctx.actions.declare_file(paths.join("gopath", "src", pkg, f.basename))
             ctx.actions.run_shell(
                 outputs = [aux],
                 inputs = [f],
@@ -63,6 +63,7 @@ def _gomock_source_impl(ctx):
             )
             aux_files.append("{0}={1}".format(pkg, aux.path))
             needed_files.append(f)
+            needed_files.append(aux)
         args += ["-aux_files", ",".join(aux_files)]
 
     inputs = (

```

I can convert it to the PR if it looks alright.

ikavalio:
**What type of PR is this?**

Bug fix

The PR tries to fix the behavior documented in #3752 

Fixes #3752


extras/gomock.bzl
```diff
@@ -55,14 +55,15 @@ def _gomock_source_impl(ctx):
...
             f = target.files.to_list()[0]
-            aux = ctx.actions.declare_file(paths.join(gopath, "src", pkg, f.basename))
+            aux = ctx.actions.declare_file(paths.join("gopath", "src", pkg, f.basename))
             ctx.actions.run_shell(
                 outputs = [aux],
                 inputs = [f],
                 command = "mkdir -p {0} && cp -L {1} {0}".format(aux.dirname, f.path),
             )
             aux_files.append("{0}={1}".format(pkg, aux.path))
             needed_files.append(f)
```
fmeum, line 65: Do you know whether this line is needed? Maybe adding `f` instead of `aux` was just a typo back then?

ikavalio: Hm, yeah, you could be right. Let me test if swapping `s/f/aux/` will be enough here. Thanks for the suggestion!

ikavalio: Looks like the `needed_files.append(f)` is redundant here, indeed.

fmeum (approved):
@linzhp Could you also take a look at this?

extras/gomock.bzl
```diff
@@ -55,14 +55,14 @@ def _gomock_source_impl(ctx):
         aux_files = []
         for target, pkg in ctx.attr.aux_files.items():
             f = target.files.to_list()[0]
-            aux = ctx.actions.declare_file(paths.join(gopath, "src", pkg, f.basename))
+            aux = ctx.actions.declare_file(paths.join("gopath", "src", pkg, f.basename))
```
linzhp, line 58: Good catch! Can you extract the string "gopath" to a variable like `go_path_prefix` and use that in both Line 39 and here? So future readers know whether the string "gopath" is from.

ikavalio: Sure, that's a great idea! Updated the PR

fmeum (approved):

ikavalio:
@fmeum @linzhp thank you very much for the reviews! would it be possible to merge the PR at your convenience (as I don't have write access to the repo), thank you!

fmeum:
@tyler-french will run some tests and then we can merge

tyler-french (approved):
> @tyler-french will run some tests and then we can merge

Tested, LGTM

PR #3761, opened 2023-11-21
Update to modern protoc plugins

### LINKED ISSUE #3022: Use protoc-gen-go-grpc for gRPC compilers

mering:
**What type of PR is this?**

Update

The module [`github.com/golang/protobuf`](https://pkg.go.dev/mod/github.com/golang/protobuf) has been superseded by the [`google.golang.org/protobuf`](https://pkg.go.dev/mod/google.golang.org/protobuf) and [`google.golang.org/grpc`](https://pkg.go.dev/mod/google.golang.org/grpc) modules, which contain an updated and simplified API, support for protobuf reflection, and many other improvements.

The `go_grpc_library()` macro maintains the current behavior of generating both the proto and the service definitions.

Fixes #3022

Supercedes #3437, #3783

**Other notes for review**

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -37,6 +37,7 @@ use_repo(
     "com_github_golang_protobuf",
     "org_golang_google_genproto",
     "org_golang_google_grpc",
+    "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
     "org_golang_google_protobuf",
     "org_golang_x_net",
 )
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	golang.org/x/tools v0.15.0
 	google.golang.org/genproto v0.0.0-20231127180814-3a041ad873d4
 	google.golang.org/grpc v1.59.0
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.3.0
 	google.golang.org/protobuf v1.31.0
 )
 
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -174,19 +174,19 @@ def go_rules_dependencies(force = False):
     )
 
     # gRPC protoc plugin
-    # releaser:upgrade-dep grpc grpc-go
+    # releaser:upgrade-dep grpc grpc-go cmd/protoc-gen-go-grpc
     wrapper(
         http_archive,
         name = "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
         sha256 = "1e84df03c94d1cded8e94da7a2df162463f3be4c7a94289d85c0871f14c7b8e3",
-        # v1.3.0, latest as of 2023-12-07
+        # cmd/protoc-gen-go-grpc/v1.3.0, latest as of 2023-12-13
         urls = [
             "https://mirror.bazel.build/github.com/grpc/grpc-go/archive/refs/tags/cmd/protoc-gen-go-grpc/v1.3.0.zip",
             "https://github.com/grpc/grpc-go/archive/refs/tags/cmd/protoc-gen-go-grpc/v1.3.0.zip",
         ],
         strip_prefix = "grpc-go-cmd-protoc-gen-go-grpc-v1.3.0/cmd/protoc-gen-go-grpc",
         patches = [
-            # releaser:patch-cmd gazelle -repo_root ./cmd/protoc-gen-go-grpc -go_prefix google.golang.org/grpc/cmd/protoc-gen-go-grpc -go_naming_convention import_alias -proto disable_global ./cmd/protoc-gen-go-grpc
+            # releaser:patch-cmd gazelle -repo_root . -go_prefix google.golang.org/grpc/cmd/protoc-gen-go-grpc -go_naming_convention import_alias -proto disable_global
             Label("//third_party:org_golang_google_grpc_cmd_protoc_gen_go_grpc.patch"),
         ],
         patch_args = ["-p1"],
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -173,9 +173,26 @@ def go_rules_dependencies(force = False):
         patch_args = ["-p1"],
     )
 
+    # gRPC protoc plugin
+    # releaser:upgrade-dep grpc grpc-go
+    wrapper(
+        http_archive,
+        name = "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
+        sha256 = "1e84df03c94d1cded8e94da7a2df162463f3be4c7a94289d85c0871f14c7b8e3",
+        # v1.3.0, latest as of 2023-12-07
+        urls = [
+            "https://mirror.bazel.build/github.com/grpc/grpc-go/archive/refs/tags/cmd/protoc-gen-go-grpc/v1.3.0.zip",
+            "https://github.com/grpc/grpc-go/archive/refs/tags/cmd/protoc-gen-go-grpc/v1.3.0.zip",
+        ],
+        strip_prefix = "grpc-go-cmd-protoc-gen-go-grpc-v1.3.0/cmd/protoc-gen-go-grpc",
+        patches = [
+            # releaser:patch-cmd gazelle -repo_root ./cmd/protoc-gen-go-grpc -go_prefix google.golang.org/grpc/cmd/protoc-gen-go-grpc -go_naming_convention import_alias -proto disable_global ./cmd/protoc-gen-go-grpc
+            Label("//third_party:org_golang_google_grpc_cmd_protoc_gen_go_grpc.patch"),
+        ],
+        patch_args = ["-p1"],
+    )
+
     # Legacy protobuf compiler, runtime, and utilities.
-    # We still use protoc-gen-go because the new one doesn't support gRPC, and
-    # the gRPC compiler doesn't exist yet.
     # We need to apply a patch to enable both go_proto_library and
     # go_library with pre-generated sources.
     # releaser:upgrade-dep golang protobuf
--- a/proto/BUILD.bazel
+++ b/proto/BUILD.bazel
@@ -23,13 +23,18 @@ go_proto_compiler(
 
 go_proto_compiler(
     name = "go_proto",
+    plugin = "@org_golang_google_protobuf//cmd/protoc-gen-go",
     visibility = ["//visibility:public"],
     deps = PROTO_RUNTIME_DEPS + WELL_KNOWN_TYPES_APIV2,
 )
 
 go_proto_compiler(
     name = "go_grpc",
-    options = ["plugins=grpc"],
+    # Do not break existing code with this upgrade
+    # See https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/README.md#future-proofing-services
+    options = ["require_unimplemented_servers=false"],
+    plugin = "@org_golang_google_grpc_cmd_protoc_gen_go_grpc//:protoc-gen-go-grpc",
+    suffix = "_grpc.pb.go",
     visibility = ["//visibility:public"],
     deps = PROTO_RUNTIME_DEPS + WELL_KNOWN_TYPES_APIV2 + [
         "@org_golang_google_grpc//:go_default_library",
--- a/proto/BUILD.bazel
+++ b/proto/BUILD.bazel
@@ -30,6 +30,20 @@ go_proto_compiler(
 
 go_proto_compiler(
     name = "go_grpc",
+    deprecation = "Migrate to //proto:go_grpc_v2 compiler via go_grpc_library() rule.",
+    options = ["plugins=grpc"],
+    plugin = "@com_github_golang_protobuf//protoc-gen-go",
+    visibility = ["//visibility:public"],
+    deps = PROTO_RUNTIME_DEPS + WELL_KNOWN_TYPES_APIV2 + [
+        "@org_golang_google_grpc//:go_default_library",
+        "@org_golang_google_grpc//codes:go_default_library",
+        "@org_golang_google_grpc//status:go_default_library",
+        "@org_golang_x_net//context:go_default_library",
+    ],
+)
+
+go_proto_compiler(
+    name = "go_grpc_v2",
     # Do not break existing code with this upgrade
     # See https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/README.md#future-proofing-services
     options = ["require_unimplemented_servers=false"],
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -183,9 +183,15 @@ go_proto_library = rule(
 # go_proto_library is a rule that takes a proto_library (in the proto
 # attribute) and produces a go library for it.
 
-def go_grpc_library(**kwargs):
-    # TODO: Deprecate once gazelle generates just go_proto_library
-    go_proto_library(compilers = [Label("//proto:go_grpc")], **kwargs)
+def go_grpc_library(name, **kwargs):
+    go_proto_library(
+        name = name,
+        compilers = [
+            Label("//proto:go_proto"),
+            Label("//proto:go_grpc_v2"),
+        ],
+        **kwargs
+    )
 
 def proto_register_toolchains():
     print("You no longer need to call proto_register_toolchains(), it does nothing")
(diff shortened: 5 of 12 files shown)


mering:
@fmeum There is a BCR build failure:
```
ERROR: C:/b/bxkbc7u6/external/rules_go~override/proto/BUILD.bazel:31:18: no such package '@gazelle~0.34.0~go_deps~org_golang_google_grpc//cmd/protoc-gen-go-grpc': BUILD file not found in directory 'cmd/protoc-gen-go-grpc' of external repository @gazelle~0.34.0~go_deps~org_golang_google_grpc. Add a BUILD file to a directory to mark it as a package. and referenced by '@rules_go~override//proto:go_grpc_reset_plugin_'
```

Do I see it correctly that this is because the bzlmod dependency doesn't have `build_file_generation = "on"` set which can only be corrected by an upstream PR to Gazelle like bazelbuild/bazel-gazelle#1672?

mering:
@fmeum like what I did in my last commit? Tests still seem to be failing...

mering:
@fmeum Any ideas what else I could try?

mering:
@fmeum as the other tests succeed, I expect the `WORKSPACE` build to work fine. See the generated patch checked into the `third_party` folder.

mering:
@fmeum as this is my first time working with bzlmod and the WORKSPACE version works fine, do you have any ideas what might be wrong with the bzlmod one? The generated BUILD file for WORKSPACE is included in this PR.

mering:
@fmeum thank you so much! Tests are green now!

go/private/repositories.bzl
```diff
@@ -173,9 +173,25 @@ def go_rules_dependencies(force = False):
         patch_args = ["-p1"],
     )
 
+    # releaser:upgrade-dep grpc grpc-go
+    wrapper(
```
fmeum, line 177: Do we even need this repo now? It doesn't seem to be referenced in our BUILD files.

WORKSPACE
```diff
@@ -136,6 +136,13 @@ go_repository(
     version = "v0.0.0-20221024183307-1bc688fe9f3e",
 )
 
+go_repository(
+    name = "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
```
fmeum, line 140: This is referenced in BUILD files, so we would either need to set up the release updater for it or explicitly document that users have to provide it themselves.

Cc @linzhp as I'm not as firm on the state of our proto deps.

go/private/repositories.bzl
mering: Indeed, I removed it!

WORKSPACE
mering: @linzhp friendly ping for your input

linzhp: Yes, this is used by `go_grpc` compiler and will be depended by users of rules_go. We need to move this to `go/private/repositories.bzl` so it can be updated by releaser and loaded by users automatically, together with `org_golang_google_protobuf` etc.

mering: @linzhp I just removed it from there after the https://github.com/bazelbuild/rules_go/pull/3761#discussion_r1410484447 from @fmeum ...

mering: Done.

go/private/repositories.bzl
```diff
@@ -173,9 +172,26 @@ def go_rules_dependencies(force = False):
         patch_args = ["-p1"],
     )
 
+    # gRPC protoc plugin
+    # releaser:upgrade-dep grpc grpc-go
```
mering, line 176: @fmeum how can I tell the releaser that the `go.mod` is rooted in the `cmd/protoc-gen-go-grpc` subfolder?

MODULE.bazel
```diff
@@ -15,6 +15,7 @@ non_module_dependencies = use_extension("//go/private:extensions.bzl", "non_modu
 use_repo(
     non_module_dependencies,
     "io_bazel_rules_nogo",
+    "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
```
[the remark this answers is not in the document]
mering: @linzhp told me to move it from `WORKSPACE` to `repositories.bzl` (see  https://github.com/bazelbuild/rules_go/pull/3761#discussion_r1417577923). As this is also loaded by bzlmod, the name would conflict.

proto/def.bzl
```diff
@@ -183,9 +183,24 @@ go_proto_library = rule(
...
+    # This is roughly equivalent to _infer_importpath() from rules_go/go/private/context.bzl
+    if not importpath:
+        importpath = "%s/%s" % (native.package_name(), name)
+
+    go_proto_library(
+        name = name + "_proto",
+        importpath = importpath,
+        **kwargs
+    )
+    go_proto_library(
```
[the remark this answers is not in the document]
mering: Done.

go/private/repositories.bzl
```diff
@@ -148,7 +148,6 @@ def go_rules_dependencies(force = False):
     #   the macro here.
     # * rules_proto also has a dependency macro. It's only needed by tests and
     #   by gogo_special_proto. Users will need to declare it anyway.
-    # * org_golang_google_grpc has too many dependencies for us to maintain.
```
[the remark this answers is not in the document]
mering: Good point. I re-added this comment.

MODULE.bazel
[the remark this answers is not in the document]
mering: Ah I see. This seems to work.

go.mod
```diff
@@ -18,4 +18,5 @@ require (
 	golang.org/x/sys v0.14.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20231120223509-83a465c0220f // indirect
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.3.0 // indirect
```
fmeum, line 21: We now have a `tools.go` file that we add pseudo-imports to to keep the corresponding `go.mod` entries even after running `go mod tidy`. Could you add an import there so that this becomes a direct dep?

mering: Good idea. Done.

proto/BUILD.bazel
```diff
@@ -23,13 +23,16 @@ go_proto_compiler(
...
     name = "go_proto",
+    plugin = "@org_golang_google_protobuf//cmd/protoc-gen-go",
     visibility = ["//visibility:public"],
     deps = PROTO_RUNTIME_DEPS + WELL_KNOWN_TYPES_APIV2,
 )
 
 go_proto_compiler(
     name = "go_grpc",
-    options = ["plugins=grpc"],
+    options = ["require_unimplemented_servers=false"],
```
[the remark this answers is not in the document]
mering: Done.

fmeum (approved):
One of us will sort out the releaser issue. Thanks for sticking with us through the review!

mering:
I noticed some issues while trying to apply this change to our code base. I will investigate and report back.

mering:
I added a `go_proto_library()` macro to migrate users from `go_proto_library(compiler="//proto:go_grpc")` go `go_grpc_library()`. Our monorepo including all dependencies compile with the latest version in this PR.

proto/def.bzl
```diff
@@ -180,12 +180,59 @@ go_proto_library = rule(
...
-# go_proto_library is a rule that takes a proto_library (in the proto
-# attribute) and produces a go library for it.
 
-def go_grpc_library(**kwargs):
-    # TODO: Deprecate once gazelle generates just go_proto_library
-    go_proto_library(compilers = [Label("//proto:go_grpc")], **kwargs)
+# Add macro for compatibility mode to ease upgrades.
+# See go_grpc_library() for more information.
+def go_proto_library(name, **kwargs):
+    def is_grpc_compiler(label):
```
fmeum, line 187: It looks like this could be a private top-level function instead.

proto/def.bzl
```diff
@@ -180,12 +180,59 @@ go_proto_library = rule(
...
+        if type(label) == "string":
+            return label.endswith("//proto:go_grpc")
+        return label.package == "proto" and label.name == "go_grpc"
+
+    is_grpc_library = False
+    if kwargs.get("compiler"):
```
fmeum, line 193:
```suggestion
    if "compiler" in kwargs:
```
Same below.

fmeum (commented):
You mentioned restoring a test, did you perhaps forget to commit it?

mering:
> You mentioned restoring a test, did you perhaps forget to commit it?

In the second commit I migrated `tests/integration/googleapis/BUILD.bazel` to use `go_grpc_library()` instead of `go_proto_library()`. In the third commit I restored the original file that's why one doesn't see it in the PR diff any more.

proto/def.bzl
mering: Done.

proto/def.bzl
```diff
@@ -180,12 +180,59 @@ go_proto_library = rule(
...
-# attribute) and produces a go library for it.
 
-def go_grpc_library(**kwargs):
-    # TODO: Deprecate once gazelle generates just go_proto_library
-    go_proto_library(compilers = [Label("//proto:go_grpc")], **kwargs)
+# Add macro for compatibility mode to ease upgrades.
+# See go_grpc_library() for more information.
+def go_proto_library(name, **kwargs):
+    def is_grpc_compiler(label):
+        if type(label) == "string":
```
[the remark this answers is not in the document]
mering: Done.

proto/def.bzl
mering: Done

proto/def.bzl
```diff
@@ -180,12 +180,59 @@ go_proto_library = rule(
...
+    if kwargs.get("compiler"):
+        if is_grpc_compiler(kwargs["compiler"]):
+            kwargs.pop("compiler")
+            is_grpc_library = True
+    elif kwargs.get("compilers"):
+        if any([is_grpc_compiler(compiler) for compiler in kwargs["compilers"]]):
+            kwargs.pop("compilers")
+            is_grpc_library = True
+    if is_grpc_library:
+        print("Compatibility mode for %s//%s:%s. Please use go_grpc_library() for compiling gRPC services." % (native.repository_name(), native.package_name(), name))
```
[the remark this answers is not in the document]
mering: While I would like to make this a `fail()` as Bazel will use a single version of a dependency, all dependencies also using `rules_go` would need to be migrated before one is able to update the `rules_go` itself in the WORKSPACE, potentially blocking upgrades for some time. Do you have any idea to make it more visible than the release notes which people rarely read?

proto/def.bzl
```diff
@@ -145,7 +145,7 @@ def _go_proto_library_impl(ctx):
         ])
     return providers + [OutputGroupInfo(**output_groups)]
 
-go_proto_library = rule(
+_go_proto_library = rule(
```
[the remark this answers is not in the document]
mering: I found another solution not requiring the macro in the first place.

fmeum (approved):
I really like the new approach!

go/private/repositories.bzl
tyler-french: I do not know off the top of my head, but maybe we'll need to modify the releaser

tyler-french (approved):
Tested this PR at Uber with no build issues ✅

proto/BUILD.bazel
```diff
@@ -23,13 +23,32 @@ go_proto_compiler(
...
+        "@org_golang_google_grpc//status:go_default_library",
+        "@org_golang_x_net//context:go_default_library",
+    ],
+)
+
+go_proto_compiler(
+    name = "go_grpc_v2",
+    # Do not break existing code with this upgrade
+    # See https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/README.md#future-proofing-services
+    options = ["require_unimplemented_servers=false"],
```
fmeum, line 49: Could we get rid of this option now that the new compiler requires explicit opt-in?

mering:
Based on the inspiration from @ryanpbrewster in #3783 I could simplify the `go_grpc_library()` macro to use an array of both compilers instead of embedding these targets manually.

mering:
> This PR LGTM, though I do have a mild preference for using the array-of-two-compilers approach rather than embedding the proto library into the grpc library.

Rephrasing, I did use the array-of-two-compilers approach in the latest push and tested it with our code base successfully.

ryanpbrewster:
Ah, I missed this, my apologies (I'm out of practice reading PRs with force-pushes)

ryanpbrewster (approved):

fmeum (approved):
Nice!

@tyler-french Could you test once more?

fmeum:
I pushed a commit with the required releaser changes.

mering:
> I pushed a commit with the required releaser changes.

Awesome, thanks a lot!

tyler-french (approved):

ryanpbrewster:
@fmeum do you think we can get a fresh release out with this protoc plugin? Would be nice to give folks a chance to start trying it out.

fmeum:
https://github.com/bazelbuild/rules_go/pull/3788

proto/def.bzl
```diff
@@ -183,9 +183,15 @@ go_proto_library = rule(
 # go_proto_library is a rule that takes a proto_library (in the proto
 # attribute) and produces a go library for it.
 
-def go_grpc_library(**kwargs):
-    # TODO: Deprecate once gazelle generates just go_proto_library
-    go_proto_library(compilers = [Label("//proto:go_grpc")], **kwargs)
+def go_grpc_library(name, **kwargs):
```
com6056, line 186: Should this support passing in additional values for `compilers` for things like `@com_github_grpc_ecosystem_grpc_gateway_v2//protoc-gen-grpc-gateway:go_gen_grpc_gateway`?

ryanpbrewster: For custom scenarios I think you can always "drop down" to the lower-level `go_proto_library` and pass in whatever compilers you want. IMO this method is mostly a matter of convenience for the 90% of use-cases that want these two specific compilers.

com6056: Ya totally fair, I only bring it up because the deprecation notice says `via go_grpc_library() rule` which didn't end up working for us since we use grpc-gateway, so maybe the deprecation message could mention both options to clear up any confusion

ryanpbrewster: That seems entirely reasonable

ryanpbrewster:
I think you need two compilers:
```
[
  "@io_bazel_rules_go//proto:go_proto",
  "@io_bazel_rules_go//proto:go_grpc_v2",
]
```

That's what the `go_grpc_library` helper does (I believe it's convenience wrapper around `go_proto_library` that automatically configures those two compilers.

mering:
@sluongng https://github.com/golang/protobuf has been superceded by the new compilers 4 years ago. So I would argue the deprecation warning in this repo was actually rather late than early.

ryanpbrewster:
For anyone else that finds there way to this PR because they are fighting with gazelle: you can add this directive to a BUILD file and it will affect all subpackages:
```
# gazelle:go_grpc_compilers	@io_bazel_rules_go//proto:go_proto,@io_bazel_rules_go//proto:go_grpc_v2
```

This instructs gazelle to install the appropriate compilers

sluongng:
@mering I am not opposed to upgrading to the new compiler in rules_go and gazelle. However, we should not mark the compiler in rules_go as deprecated until the new compiler is on feature parity with the old compiler (i.e. Gazelle support).

It reminds me of this Goomics https://goomics.net/50/ that folks often share with me when adopting Google's APIs/tools. And it would cost us little to make it happen (deplaying the rules_go release a bit more until the Gazelle change was merged).

@ryanpbrewster that's a useful workaround. Thanks! I hope https://github.com/bazelbuild/bazel-gazelle/pull/1711 will make it obsolete. 🤞

PR #3762, opened 2023-11-22
Update GO_TOOLCHAIN string to Label and add toolchain param to one more action

kotlaja:
What type of PR is this?
Other - Migration

Because of the relative paths, we'll need GO_TOOLCHAIN as a Label, not a String. I saw this error by building `@go_default_sdk//...` from tests/bcr workspace.

Since it's not easy to see which actions are affected by these changes (sometimes it depends on projects which are using them), I'm updating these actions until I see green Bazel downstream projects. Before that I build the whole repo manually with the incompatible flag, but in some projects that is not enough (combinations of flags can affect this, executable/tools as variables from other projects, etc).

Please check [official documentation for Automatic Exec Groups](https://bazel.build/extending/auto-exec-groups) for additional info.


kotlaja:
@fmeum Can you please check this one too? :)

fmeum:
@kotlaja Looks good to me, but there is a failure with Bazel 5.4.0. Could you take a look?

kotlaja:
Ahh, Bazel 5.4.0 supports toolchain registration on a rule only as a String, not as a Label ([see String parsing here](https://github.com/bazelbuild/bazel/blob/release-5.4.0/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java#L627)).

I've used `str` func for this. PTAL. :)

go/private/common.bzl
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-GO_TOOLCHAIN = "@io_bazel_rules_go//go:toolchain"
+GO_TOOLCHAIN = Label("@io_bazel_rules_go//go:toolchain")
```
fmeum, line 15: How about applying `str` here instead?

kotlaja: I also thought of adding one more constant as a string, but let's go with your suggestion. Thanks! :)

fmeum (approved):

fmeum:
CI is still failing with a very obscure failure.

kotlaja:
Seems that Bazel 5.4. has issues with `str(Label(toolchain_string))` when registering toolchains on rules. The error is related to modified string of a toolchain which must be mapped 1-1 (changing it to label and then to a string changes that) or otherwise toolchain resolution doesn't find it as a registered one. I've separated it now into a Label and a String and updated toolchain parameters.

fmeum (approved):

kotlaja:
Thanks @fmeum for the quick responses and all the help! :)
By checking rules_go with multiple rules today, it seems that these changes are sufficient for the AEGs migration. 

I know we had a release 4 days ago, but I'm interested when can we expect the next one? 
I'll CC also @tyler-french here - Hi Tyler! :)

PR #3764, opened 2023-11-24
Do not choose prereleases as highest versions

mering:
**What type of PR is this?**

Bug fix

This skips prerelease versions when determining the latest version.

**Other notes for review**


fmeum (approved):
Cc @tyler-french

PR #3766, opened 2023-11-28
Only use workaround for darwin deadlock on darwin

mbeards:
**What type of PR is this?**
Bug fix

Only use the workaround from https://github.com/bazelbuild/rules_go/pull/3536 on darwin.

(The lack of parallelism here makes this step heavy in the critical path for some of our builds.)

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -459,12 +459,21 @@ def _go_tool_binary_impl(ctx):
         )
     else:
         # Note: GOPATH is needed for Go 1.16.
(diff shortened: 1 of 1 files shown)


fmeum:
As this is only used for certain special binaries that are typically built very early in a build, I am slightly surprised to see this contribute to your critical path. Are you able to share a profile (`--profile`) that we can use to understand this better?

Generally speaking, an action should really only use up a single CPU core, so this setting is correct on all platforms. But of course we don't want it to negatively affect performance.

PR #3767, opened 2023-11-28
Remove deprecated gogo/validator

mering:
**What type of PR is this?**

Cleanup

Cleanup deprecated and unmaintained dependencies in order to make changes to this project easier:

- gogo is [deprecated](https://github.com/gogo/protobuf#readme)
- go-proto-validator is [unmaintained](https://github.com/mwitkow/go-proto-validators/issues/118)

**Other notes for review**


fmeum:
@linzhp Do you see any reason to keep these around?

linzhp:
No objection removing the validator, but can we please keep gogo for a bit longer. Uber still needs more time moving away from it.

mering:
I split out the validator removal to #3768 and rebased this PR on top.

@linzhp do you have a rough timeline about Uber gogo proto migration?

linzhp:
We need 2-3 years to migrate away from gogo. However, we only use "gogo", "gogofast" and "gogoslick" variants. So you can remove other variants if that helps.

mering:
@linzhp what is the state of Uber migrating away from gogo?

mering:
> gogo proto is still faster than go proto, due to the way it serializes and deserializes messages. It also has some features that go proto doesn't have, such as annotations to disallow a field to have zero value. The migration is expensive due to different type names, while people would lose some features and performance. I suspect Uber is not the only one using gogo proto right now.

Are you using the [Opaque API](https://go.dev/blog/protobuf-opaque)?
Are you using [Protobuf Editions](https://protobuf.dev/editions/overview/)?
IIUC these could solve at least some of the feature and performance issues you are describing.

> When you said "in order to make changes to rules_go easier.", is there any change you have in mind that would benefit from this removal?

This came up when working on #3706.

PR #3770, opened 2023-11-28
nogo: use original source files instead of coverage-instrumented

### LINKED ISSUE #3769: Nogo should use original source file instead of coverage instrumented ones for static analysis

0.39.1

0.28.0

6.3.0

MacOS M1 Ventura and Linux AMD64

### What did you do?

With nogo setup in monorepo, I'm trying to enable a `gochecknoinits` analyer, which check that no inits functions are present in Go code. When I ran `bazel coverage //...`, nogo reports errors for `init()` functions it found on non-existing lines in some source files. 

nogo should not report errors on files where there is no init() function in the original source file. 

### What did you see instead?

When coverage mode is set, the source file is modified with an `init()` function added to register files to coverdata, which is where nogo static analysis reported the error.

https://github.com/bazelbuild/rules_go/blob/91e35309c55253d323e072b7a85f33188835903b/go/tools/builders/cover.go#L96-L105

Ideally nogo static analysis should not run against coverage instrumented source files, but the original ones. Under GoCompilePkg actions, the nogoSrcsOrigin map was introduced in #3216 for this purpose. But this mechanism doesn't work when cgo is enabled, as all source files path would change after cgo compilation: https://github.com/bazelbuild/rules_go/blob/91e35309c55253d323e072b7a85f33188835903b/go/tools/builders/compilepkg.go#L342

hence no file would match the nogoSrcsOrigin map, resulting in nogo uses the coverage instrumented source files instead: https://github.com/bazelbuild/rules_go/blob/91e35309c55253d323e072b7a85f33188835903b/go/tools/builders/compilepkg.go#L434-L444

I'd like to propose a fix to this issue by having cgo compile source files that are not coverage instrumented for nogo's use.

emmaxy:
**What type of PR is this?**

Bug fix

Nogo static analysis should not run against coverage instrumented source files, but the original ones.

The current mechanism with `nogoSrcsOrigin` map doesn't work when cgo is enabled, because source files paths change after cgo compilation, resulting in no match to the `nogoSrcsOrigin` map, hence nogo checks against modified source files, reporting unexpected errors.

This PR allows nogo to run static analysis against the original source files, instead of coverage-instrumented ones, by making a hard copy of the source file slice if coverage mode is set and before they are modified.

Fixes #3769 

**Other notes for review**


fmeum:
CC @sluongng

sluongng (commented):
Oh hey that’s _my_ TODO 😂
Great job on this PR!

Could you help provide us with a test that demonstrate the broken behavior before this change and it’s now fixed? There has been several discussions about reworking `nogo` to enable it with bzlmod and validation action. A test could help us avoid repeating the same mistake.

sluongng:
On second thought, I am not quite sure if this is the right thing to do since we do have analyzers that could operate on cgo code. https://cs.opensource.google/go/x/tools/+/master:go/analysis/passes/cgocall/cgocall.go

It would be nice if we have a test to demonstrate how this PR would affect such an analyzer.

go/tools/builders/compilepkg.go
```diff
@@ -343,6 +351,15 @@ func compileArchive(
 		if err != nil {
 			return err
 		}
+		if coverMode != "" {
+			// Compile original source files, not coverage instrumented, for nogo
+			_, goSrcsNogo, _, err = cgo2(goenv, goSrcsNogo, cgoSrcsNogo, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSrcs, hSrcs, packagePath, packageName, cc, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags, cgoExportHPath)
```
emmaxy, line 356: This PR shouldn't affect how cgo analyzers operate, because the source files for nogo are also cgo compiled, they are just not coverage instrumented.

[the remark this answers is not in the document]
emmaxy: `goSrcsNogo` and `cgoSrcsNogo` need to be cgo compiled, otherwise cgo enabled targets cannot run static analysis succesfully. Because the source files with `import "C"` are treated as cgoSrcs, and nogo will report type-checking errors on those source files if not cgo compiled.

sluongng (commented):
Great job. The PR looks much cleaner now.

Is it possible to add some tests demonstrating the new behavior?

sluongng:
This goes without saying, but might be worth highlighting in the changelog if we were to merge this in the future:

This change will double the cgo/cc part in a `compilepkg` action when coverage is enabled, leading to slower action, depending on how much cgo/cc is in your package.

emmaxy:
> Is it possible to add some tests demonstrating the new behavior?

Could you help provide some example/reference for such tests? I'd really appreciate it. I'm not very familiar with how tests are done in rules_go, as this is my first time working in this repo.

go/tools/builders/compilepkg.go
```diff
@@ -429,24 +437,10 @@ func compileArchive(
...
-		// Need to determine whether we want to feed these CGO-generated files into static analyzers.
-		//
-		// Add unknown origin source files into the mix.
+	nogoSrcs := make([]string, 0, len(goSrcsNogo))
+	// goSrcsNogo contains original source files that are not coverage-instrumented in case coverage mode is set
+	// to avoid nogo reporting errors introduced by the modification.
+	for _, goSrc := range goSrcsNogo {
 		nogoSrcs = append(nogoSrcs, goSrc)
 	}
 	if nogoPath != "" && len(nogoSrcs) > 0 {
```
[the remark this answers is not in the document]
emmaxy: good call, I've updated to use `goSrcsNogo` directly.

sluongng:
https://github.com/bazelbuild/rules_go/blob/master/tests/core/nogo/coverage/README.rst should be a good starting point for this case since it tests for a combination of nogo+coverage.

emmaxy:
Unit test added! Sorry for the delay.
This unit test passes with the fix in this PR, and fails on current master, with error message:
```
foo.go:22:1: don't use init function (gochecknoinits)
Target //:foo_test failed to build
```
This shows that the nogo was running static analysis on coverage instrumented source files, resulting in error non-existing in the original source file.

But I'm not sure why the [build](https://buildkite.com/bazel/rules-go-golang/builds/5966#018c5704-47d8-4616-bfde-122443629a39) is failing for windows (OpenJDK 11, VS2022), any suggestions?

tests/core/nogo/coverage/coverage_cgo_test.go
```diff
@@ -0,0 +1,124 @@
...
+		Main: `
+-- BUILD.bazel --
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test", "nogo")
+
+go_library(
+    name = "foo",
+	cgo = True,
+    srcs = ["foo.go"],
+    importpath = "foo"
+)
```
sluongng, line 34:
```suggestion
go_library(
    name = "foo",
	  cgo = True,
    srcs = ["foo.go"],
    importpath = "foo"
)
```

Nit: indentation

tests/core/nogo/coverage/coverage_cgo_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return nil, nil
+}
+`,
+		Nogo: `@//:nogo`,
+	})
+}
+
+func TestCoverageWithNogo(t *testing.T) {
```
sluongng, line 119: Let's rename this func to be specific to nogo + coverage.

sluongng: I suspect this is why the Windows test was failing btw. There is no Bazel sandbox on Windows so targets with similar name could collide.

emmaxy:
Took a while before I realized the build failure was because a known issue #2518 that coverage not able to run on windows platform, so I updated the tests accordingly. Also changed it so that tests are skipped with `target_compatible_with` in BUILD.bazel instead of CI configuration. 
Now that build is green,  @sluongng could you help review again, and if looks good to you I could merge this PR

go/tools/builders/compilepkg.go
```diff
@@ -343,6 +342,15 @@ func compileArchive(
...
 		}
+		if coverMode != "" {
+			// Compile original source files, not coverage instrumented, for nogo
+			_, goSrcsNogo, _, err = cgo2(goenv, goSrcsNogo, cgoSrcsNogo, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSrcs, hSrcs, packagePath, packageName, cc, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags, cgoExportHPath)
+			if err != nil {
+				return err
+			}
+		} else {
+			goSrcsNogo = goSrcs
+		}
```
sluongng, line 353: Nit: do we need this else clause? Seems redundant.

```suggestion
		}
```

.bazelci/presubmit.yml
```diff
@@ -293,8 +293,6 @@ tasks:
       - "-//tests/core/go_proto_library_importmap:importmap_test"
       - "-//tests/core/go_test:data_test"
       - "-//tests/core/go_test:pwd_test"
-      - "-//tests/core/nogo/coverage:coverage_test"
-      - "-//tests/core/nogo/coverage:gen_code_test"
```
sluongng, line 297: I will let @fmeum weigh in here, but I think a benefit of having a list like this is documenting what is supported and not supported. If we move things out of this list by leveraging `target_compatible_with`, then there is no longer a single place acting as the maintainers' TODO list anymore.

go/tools/builders/compilepkg.go
```diff
@@ -280,8 +270,18 @@ func compileArchive(
...
+	// over original source files and not the modified ones.
+	// goSrcsNogo and cgoSrcsNogo are copies of the original source files for nogo to run static analysis.
+	goSrcsNogo := goSrcs
+	cgoSrcsNogo := cgoSrcs
+
 	// Instrument source files for coverage.
 	if coverMode != "" {
+		// deep copy original source files for nogo static analysis, avoid being modified by coverage.
+		goSrcsNogo = append([]string{}, goSrcs...)
+		cgoSrcsNogo = append([]string{}, cgoSrcs...)
```
sluongng, line 283: Nit: would it make more sense to always use a copy of the slice instead of only copying during coverage?

```suggestion
	// goSrcsNogo and cgoSrcsNogo are copies of the original source files for nogo to run static analysis.
	// deep copy original source files for nogo static analysis, avoid being modified by coverage.
	goSrcsNogo := append([]string{}, goSrcs...)
	cgoSrcsNogo := append([]string{}, cgoSrcs...)

	// Instrument source files for coverage.
	if coverMode != "" {
```

sluongng (approved):
Functionally LGTM. I have a few optional nits/questions from a second reading of the diff.

Will let the active maintainers discuss the style concerns of moving targets out of presubmit.yml. cc: @fmeum @linzhp

.bazelci/presubmit.yml
fmeum: I agree with @sluongng here. We could go with `target_compatible_with` instead, but since we already have this central list, we should stick to it and maintain it as the single source of truth.

go/tools/builders/compilepkg.go
emmaxy: Since the purpose for `goSrcsNogo` and `cgoSrcsNogo` is to retain the original source files un-modified by coverage, I'm only making deep copies when coverage is enabled, as deep copy is more expensive in terms of performance and memory usage. But I also agree the code will look cleaner if we always make a deep copy, wdyt?

go/tools/builders/compilepkg.go
emmaxy: This else clause is necessary, because `goSrcsNogo` was referencing to `goSrcs` before cgo compilation, and here it's updated to `goSrcs` that are cgo compiled, this will ensure cgo enabled targets can run static analysis successfully.

.bazelci/presubmit.yml
emmaxy: Thanks for the suggestion, that makes sense.

fwiw, I was referring to the coverage tests, and I favored this approach as it doesn't limit to CI, but also ensures the incompatible tests won't be triggered on Windows platform if running locally.  https://github.com/bazelbuild/rules_go/blob/d5bc6adb81d3aba03ff625e0c7e57504caf49222/tests/core/coverage/BUILD.bazel#L16-L18

But since the yaml list is the preferred approach, I would love to switch to that.

go/tools/builders/compilepkg.go
```diff
@@ -343,6 +342,15 @@ func compileArchive(
 		if err != nil {
 			return err
 		}
+		if coverMode != "" {
+			// Compile original source files, not coverage instrumented, for nogo
```
fmeum, line 346: Do we need this `cgo2` call if nogo isn't used? 

It looks like we are now compiling cgo sources twice - is this true and if so, is it really necessary?

fmeum: @emmaxy Friendly ping, I would call this the last blocker to merging this.

emmaxy: Good call, thanks! This additional cgo compile is not needed if nogo doesn't run, I updated the condition accordingly.

fmeum (approved):

PR #3771, opened 2023-11-29
cgo2: refactor for readability

sluongng:
This is a no-op refactoring of cgo2 func:
- Rename from cgo2 to cgo
- Reorganize function's arguments into a cgoOption struct
- Replace usages of deprecated ioutil with `io` and `os` equivalents

No functionality change is expected.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -23,22 +23,41 @@ package main
 import (
 	"bytes"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


sluongng:
We could also rename the file from cgo2.go -> cgo.go, but I was afraid of the noisy diff during code review so I decided against it.

sluongng:
This came out of my reading of https://github.com/bazelbuild/rules_go/pull/3770 to help aid my understanding of the current code.

Zooming out a bit, we already have some of the basic structs such as `archiveSrcs` in `filter.go` that we should make use of.

PR #3775, opened 2023-12-02
Use `tools.go` trick to make `go.mod` work with `go mod tidy`

fmeum:
This allows us to use `go mod tidy` and `go get` to maintain rules_go's `go.mod` file, which is used to specify its Go dependencies with Bzlmod.

`go.mod` files are added to helpers that don't ship with rules_go to remove unnecessary dependencies.


tyler-french (approved):
Looks good!

PR #3779, opened 2023-12-07
fix(gpd): don't panic on invalid root

JamyDev:
**What type of PR is this?**

Bug fix

In rare cases the root won't actually have a package in the registry, causing the lookup to panic, and subsequently cause gopls (or other tools) to fail their resolution, and likely try multiple times. 

Fixes #

**Other notes for review**

Will add a test case shortly, hence the draft

### DIFF
--- a/go/tools/gopackagesdriver/packageregistry.go
+++ b/go/tools/gopackagesdriver/packageregistry.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


tyler-french:
I think this a very simple and safe change that is safe merging as is (cc @fmeum). Adding a test would be good, but I don't think this bug fix should necessarily need one, as it seems pretty trivial to avoid panicking.

fmeum (approved):
@JamyDev If you do write a test for this, we can always merge it as a separate PR.

PR #3782, opened 2023-12-11
Add Bzlmod support for `nogo`

### LINKED ISSUE #3529: Support nogo with Bzlmod
It would be nice for there to be docs on how to use `nogo` with the `go_sdk` bzlmod extension. I'm still a newbie with the new Bazel module system and it's not obvious to me how to translate `go_register_toolchains(nogo = ...)` into it.

I'd also be happy to hear there's another preferred way of applying `vet` and such

fmeum:
**What type of PR is this?**

Feature

See `bzlmod.md` for the details. This change is backwards compatible for WORKSPACE setups, but allows finer control over the set of targets nogo is applied to when using Bzlmod.

Fixes #3529 

**Other notes for review**

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -19,6 +19,7 @@ go_sdk.download(
 use_repo(
     go_sdk,
     "go_toolchains",
+    # This name is ugly on purpose to avoid a conflict with a user-named SDK.
     "io_bazel_rules_nogo",
 )
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -11,18 +11,16 @@ bazel_dep(name = "platforms", version = "0.0.4")
 bazel_dep(name = "rules_proto", version = "4.0.0")
 bazel_dep(name = "protobuf", version = "3.19.2", repo_name = "com_google_protobuf")
 
-non_module_dependencies = use_extension("//go/private:extensions.bzl", "non_module_dependencies")
-use_repo(
-    non_module_dependencies,
-    "io_bazel_rules_nogo",
-)
-
 go_sdk = use_extension("//go:extensions.bzl", "go_sdk")
 go_sdk.download(
     name = "go_default_sdk",
     version = "1.21.1",
 )
-use_repo(go_sdk, "go_toolchains")
+use_repo(
+    go_sdk,
+    "go_toolchains",
+    "io_bazel_rules_nogo",
+)
 
 register_toolchains("@go_toolchains//:all")
 
@@ -40,4 +38,5 @@ use_repo(
     "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
     "org_golang_google_protobuf",
     "org_golang_x_net",
+    "org_golang_x_tools",
 )
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -14,7 +14,7 @@
 
 load("@bazel_features//:features.bzl", "bazel_features")
 load("//go/private:sdk.bzl", "detect_host_platform", "go_download_sdk_rule", "go_host_sdk_rule", "go_multiple_toolchains")
-load("//go/private:repositories.bzl", "go_rules_dependencies")
+load("//go/private:nogo.bzl", "DEFAULT_NOGO", "go_register_nogo")
 
 def host_compatible_toolchain_impl(ctx):
     ctx.file("BUILD.bazel")
@@ -67,6 +67,32 @@ _host_tag = tag_class(
     },
 )
 
+_NOGO_DEFAULT_INCLUDES = ["@@//:__subpackages__"]
+_NOGO_DEFAULT_EXCLUDES = []
+
+_nogo_tag = tag_class(
+    attrs = {
+        "nogo": attr.label(
+            doc = "The nogo target to use when this module is the root module.",
+        ),
+        "includes": attr.label_list(
+            default = _NOGO_DEFAULT_INCLUDES,
+            doc = """
+A Go target is checked with nogo if its package matches at least one of the entries in 'includes'
+and none of the entries in 'excludes'. By default, nogo is applied to all targets in the main
+repository.
+
+Uses the same format as 'visibility', i.e., every entry must be a label that ends with ':__pkg__' or
+':__subpackages__'.
+""",
+        ),
+        "excludes": attr.label_list(
+            default = _NOGO_DEFAULT_EXCLUDES,
+            doc = "See 'includes'.",
+        ),
+    },
+)
+
 # A list of (goos, goarch) pairs that are commonly used for remote executors in cross-platform
 # builds (where host != exec platform). By default, we register toolchains for all of these
 # platforms in addition to the host platform.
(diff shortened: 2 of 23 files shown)


fmeum:
@sluongng I would be interested in your opinion on this.

sluongng (commented):
This PR does a few things:

1. Introduce calling of `go_register_nogo` inside a new tag of `go_sdk`. This help user start registering their nogo binary 👍

2. A `_get_nogo` mechanism to exclude and include which Bazel packages can be subjected to `nogo` analysis. This could be easier reviewed if it’s introduced in a separate PR, but now that you have included it here, no need to spend more time on splitting it.

Overall I think (2) provides a better, more sensible default than what we have today. Though it’s a bit overlap with the `nogo_config.json` mechanism we current have in rules `nogo()`. I think we should add a deprecation notice to the include/exclude in `nogo_config.json` so we could phase it out in favor of this one.

Finally the `bcr` test demonstrated how `nogo` could be setup in a repo and how some certain packages could be excluded. I think we are missing a “test for nogo failure” case here, but I am not quite sure if the current test setup would allow that. Is it possible to setup a test like that under `//tests/core/nogo`?

Overall this LGTM. Thanks a lot for working on it 🙏

fmeum:
> I think we are missing a “test for nogo failure” case here, but I am not quite sure if the current test setup would allow that. Is it possible to setup a test like that under //tests/core/nogo?

It doesn't right now, but I will add a regular `go_bazel_test` covering the inclusion/exclusion logic as well as some Starlark tests.

fmeum:
Making `go_bazel_test` usable for Bzlmod tests will take some effort, depending on the level of test runtime efficiency we aim for. I will see whether I can produce something that works quickly.

tyler-french (approved):
Tested at Uber. LGTM!

go/private/extensions.bzl
```diff
@@ -67,6 +67,32 @@ _host_tag = tag_class(
...
+        "includes": attr.label_list(
+            default = _NOGO_DEFAULT_INCLUDES,
+            doc = """
+A Go target is checked with nogo if its package matches at least one of the entries in 'includes'
+and none of the entries in 'excludes'. By default, nogo is applied to all targets in the main
+repository.
+
+Uses the same format as 'visibility', i.e., every entry must be a label that ends with ':__pkg__' or
+':__subpackages__'.
+""",
```
sluongng, line 87: Do we want to document the magic value `"all"` here? or it's not intended for usage in bzlmod?

go/tools/bazel_testing/bazel_testing.go
```diff
@@ -384,8 +389,9 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
 
 	// If there's no WORKSPACE file, create one.
 	workspacePath := filepath.Join(mainDir, "WORKSPACE")
-	if _, err := os.Stat(workspacePath); os.IsNotExist(err) {
-		w, err := os.Create(workspacePath)
+	if _, err = os.Stat(workspacePath); os.IsNotExist(err) {
+		var w *os.File
+		w, err = os.Create(workspacePath)
```
sluongng, line 394: not quite sure this hunk is needed 🤔

sluongng: are you using some static linter that enforces a certain style?

sluongng (approved):
Few style nits/questions, but LGTM overall

MODULE.bazel
```diff
@@ -5,31 +5,56 @@ module(
...
 go_sdk.download(
     name = "go_default_sdk",
     version = "1.21.1",
 )
-use_repo(go_sdk, "go_toolchains")
+
+use_repo(
+    go_sdk,
+    "go_toolchains",
+    "io_bazel_rules_nogo",
```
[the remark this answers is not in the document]
fmeum: It's actually better to choose a long name here: It is internal to rules_go and we don't want it to conflict with user-named SDK repos.

go/private/extensions.bzl
fmeum: I haven't documented it anywhere since I think it's ultimately problematic with Bzlmod: If you use it, the introduction of a new transitive Go dep would be a breaking change. We need to have users explicitly list the deps that they want nogo to be applied to.

go/tools/bazel_testing/bazel_testing.go
fmeum: IntelliJ showed me that the `err :=` here shadowed the named return, which breaks the `defer` func below. I fixed this here since I also do this right now in the hunk below, which follows the same pattern.

go/tools/bazel_testing/bazel_testing.go
```diff
@@ -416,6 +422,46 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
...
+		}
+		var w *os.File
+		w, err = os.Create(moduleBazelPath)
+		if err != nil {
+			return "", cleanup, err
+		}
+		defer func() {
+			if cerr := w.Close(); err == nil && cerr != nil {
+				err = cerr
+			}
```
tyler-french, line 439: nit: 
```suggestion
			err = errors.Join(err, w.Close())
```

go/tools/bazel_testing/bazel_testing.go
```diff
@@ -416,6 +422,46 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
...
+			return "", cleanup, fmt.Errorf("could not find relative path from %q to %q for io_bazel_rules_go", mainDir, rulesGoAbsPath)
+		}
+		rulesGoPath = filepath.ToSlash(rulesGoPath)
+		info := moduleFileTemplateInfo{
+			RulesGoPath: rulesGoPath,
+			Suffix:      args.ModuleFileSuffix,
+		}
+		if err := defaultModuleBazelTpl.Execute(w, info); err != nil {
+			return "", cleanup, err
+		}
```
tyler-french, line 453: optional: you can inline this
```suggestion
		if err := defaultModuleBazelTpl.Execute(w, moduleFileTemplateInfo{
			RulesGoPath: rulesGoPath,
			Suffix:      args.ModuleFileSuffix,
		}); err != nil {
			return "", cleanup, err
		}
```

tests/core/nogo/bzlmod/includes_excludes_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestNotIncluded(t *testing.T) {
+	if err := bazel_testing.RunBazel("build", "//:lib"); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestIncluded(t *testing.T) {
+	if err := bazel_testing.RunBazel("build", "//go:lib"); err == nil {
+		t.Fatal("Expected build to fail")
+	} else if !strings.Contains(err.Error(), "lib.go:6:3: declaration of \"foo\" shadows declaration at line 4 (shadow)") {
```
tyler-french, line 117: (aside) This whole function could be simplified with assert 😄, we may want to consider using it:

e.g. https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains
```
assert.ErrorContains(t, bazel_testing.RunBazel(..), "message")
```

tyler-french (approved):
Looks great! All of my comments are just Go style nits.
I tested the latest and it works great!

go/tools/bazel_testing/bazel_testing.go
fmeum: Thanks for the playground link, I didn't know that named returns worked like that.

fmeum: On second thought, I think that this is still problematic: The problem isn't that the `defer` above uses the wrong `err`, it's that the `defer` *below* (line 398-402) sets the wrong `err`, namely the one in the inner scope. Is my thinking correct here?

go/tools/bazel_testing/bazel_testing.go
```diff
@@ -416,6 +422,46 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
...
+		var w *os.File
+		w, err = os.Create(moduleBazelPath)
+		if err != nil {
+			return "", cleanup, err
+		}
+		defer func() {
+			if cerr := w.Close(); err == nil && cerr != nil {
+				err = cerr
+			}
+		}()
```
motiejus, line 440: This is what I used for such case (you can probably still find this pattern used quite a bit under `src/code.uber.internal/infra/...`, but it was never made official). First, prefix all return variables with `_`:

```go
func setupWorkspace(args Args, files []string) (_dir string, _cleanup func() error, _err error) {
```

That way, it's always clear which value we are referring to: the returned one or the one in the current scope? With that, you are allowed to use the return values *only* in `defer`. Like this:

```suggestion
		defer func() {
			if err := w.Close(); err == nil {
				_err = errors.Join(_err, err)
			}
		}()
```

Then nesting, shadowing works as expected. I didn't have time to spread this pattern, nor create a linter. :)

fmeum: Thanks, I like that pattern. I'll leave this as is for now since we'll likely refactor all of this when we make it ready for running in Bzlmod (not just running a Bzlmod project).

PR #3783, opened 2023-12-11
Expose protoc-gen-go-grpc as a go_proto_compiler option

ryanpbrewster:
**What type of PR is this?**

Feature

In https://github.com/bazelbuild/rules_go/issues/3022 it was noted that the default `go_proto_compiler` exposed by `rules_go` is no longer actively being maintained. Most recent development effort is going into google.golang.org/protobuf/cmd/protoc-gen-go

You can see this in the gRPC tutorial: https://grpc.io/docs/languages/go/quickstart/ which recommends
```
$ go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
$ go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2
```

Specifically there are a few changes to how this other gRPC compiler plugin generates code that are desireable:
- it simplifies forwards-compatibility by introducing an `UnimplementedXyzServer` struct that implementers must embed
- it modifies the `RegisterXyzServer` function so that it accepts a `grpc.ServiceRegistrar` interface rather than a concrete `*grpc.Server` pointer

Unfortunately, these exact benefits make it a bit rocky to transition from the old plugin to the new one. Rather than try to swap out the default proto compiler, I've opted in this PR to simply _expose an additional option_. This should be backwards compatible, but will allow `rules_go` users to begin incrementally transitioning their gRPC code over to the new plugin.

In a future release, you might consider fully deprecating the old plugin, but that will be a pretty significant breaking change and should be undertaken carefully.

Partially addresses https://github.com/bazelbuild/rules_go/issues/3022

### DIFF
--- a/proto/BUILD.bazel
+++ b/proto/BUILD.bazel
@@ -39,6 +39,19 @@ go_proto_compiler(
     ],
 )
 
+go_proto_compiler(
+    name = "go_grpc_v2",
+    plugin = "@org_golang_google_grpc_cmd_protoc_gen_go_grpc//:protoc-gen-go-grpc",
+    suffix = "_grpc.pb.go",
+    visibility = ["//visibility:public"],
+    deps = WELL_KNOWN_TYPES_APIV2 + [
+        "@org_golang_google_grpc//:go_default_library",
+        "@org_golang_google_grpc//codes:go_default_library",
+        "@org_golang_google_grpc//status:go_default_library",
+        "@org_golang_x_net//context:go_default_library",
+    ],
+)
+
 GOGO_VARIANTS = [
     "combo",
     "gogo",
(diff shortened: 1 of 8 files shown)


proto/BUILD.bazel
```diff
@@ -39,6 +39,18 @@ go_proto_compiler(
     ],
 )
 
+go_proto_compiler(
+    name = "go_grpc_v2",
```
ryanpbrewster, line 43: I don't love this name. I want to gently nudge callers towards it without forcing them to deal with the internals of which plugin is which.

Any suggestions would be welcome

fmeum:
Thanks for sending the PR! There is already another one that deals with this update: https://github.com/bazelbuild/rules_go/pull/3761 
Since you have looked into this in detail, could you take a look at the other PR? Does it fully address the problem?

mering:
I agree that the integrations test is valuable and it would be great if you could propose it again once my PR has been merged.

PR #3785, opened 2023-12-13
Pass env to gentestmain so it will correctly filter

### LINKED ISSUE #2763: go_test: filter source files for target platform when generating testmain.go

0.20.0

0.22.2

3.7.1

I haven't tested, but source code indicates so.

Linux, macOS, etc.

### What did you do?

I made a package that contains some `*_test.go` source files with build tags at the top:

```
// +build darwin linux
```

That those source files are ignored on other platforms and that `bazel test` still works, only running tests for the other source files.

### What did you see instead?

A build error, due to testmain.go referencing nonexistent source files.

The change made in bazelbuild/bazel-gazelle#152 seems fine for targets like go_binary() and go_library(), but for go_test() it ends up generating something that rules_go is unable to deal with. If I patch up Gazelle as follows:

```
diff --git a/language/go/generate.go b/language/go/generate.go
index 5f73b1a..4fef14e 100644
--- a/language/go/generate.go
+++ b/language/go/generate.go
@@ -651,7 +651,7 @@ func (g *generator) maybeGenerateExtraLib(lib *rule.Rule, pkg *goPackage) *rule.
 
 func (g *generator) setCommonAttrs(r *rule.Rule, pkgRel string, visibility []string, target goTarget, embed string) {
 	if !target.sources.isEmpty() {
-		r.SetAttr("srcs", target.sources.buildFlat())
+		r.SetAttr("srcs", target.sources.build())
 	}
 	if target.cgo {
 		r.SetAttr("cgo", true)
```

Then it generates more convoluted go_test() declarations, but those do work properly in case of cross compilation for different platforms.

patrickmscott:
This is a fairly simple change and the test is also simple. Without the addition test.bzl, the new cgo_required test fails to compile. It's more of a regression test than anything else.

**What type of PR is this?**

Bug fix

Correctly filters test functions

Fixes #2763 

**Other notes for review**

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -120,6 +120,7 @@ def _go_test_impl(ctx):
         executable = go.toolchain._builder,
         arguments = [arguments],
         toolchain = GO_TOOLCHAIN_LABEL,
+        env = go.env,
     )
 
     test_gc_linkopts = gc_linkopts(ctx)
--- a/tests/core/cgo/BUILD.bazel
+++ b/tests/core/cgo/BUILD.bazel
@@ -472,3 +472,9 @@ cc_binary(
     srcs = ["use_c_symbol_through_go.c"],
     deps = [":use_transitive_symbol"],
 )
+
+go_test(
+    name = "cgo_required",
+    srcs = ["cgo_required_test.go"],
+    pure = "on",
+)
--- a/tests/core/cgo/cgo_required_test.go
+++ b/tests/core/cgo/cgo_required_test.go
@@ -1,0 +1,10 @@
+//go:build cgo
+// +build cgo
+
+package cgo_required
+
+import "testing"
+
+// Without correct filtering, gentestmain will try to link against this test
+// that does not exist with pure = True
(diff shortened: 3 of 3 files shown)


go/private/rules/test.bzl
```diff
@@ -120,6 +120,7 @@ def _go_test_impl(ctx):
         executable = go.toolchain._builder,
         arguments = [arguments],
         toolchain = GO_TOOLCHAIN_LABEL,
+        env = go.env,
```
fmeum, line 123: This looks good, but I would like to understand better why it fixes the issue. Which variable do we end up passing in here that matters?

patrickmscott: The issue I faced was `CGO_ENABLED`. `build.Context` does not properly filter out cgo files when generating testmain when cgo is disabled.

fmeum (approved):

fmeum:

tyler-french (approved):

tyler-french:
> @tyler-french Could you test this?

Tested. LGTM!

PR #3788, opened 2023-12-15
prepare release 0.44.0

tyler-french:
This PR prepares the new release of `rules_go` version `0.44.0`.

At this time we can't upgrade `x/tools` because releases after `0.15.0` remove [the persistent package](https://pkg.go.dev/golang.org/x/tools/internal/persistent)

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -50,7 +50,7 @@ def go_rules_dependencies(force = False):
     wrapper(
         http_archive,
         name = "bazel_skylib",
-        # 1.5.0, latest as of 2023-11-16
+        # 1.5.0, latest as of 2023-12-15
         urls = [
             "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.5.0/bazel-skylib-1.5.0.tar.gz",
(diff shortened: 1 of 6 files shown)


go/private/repositories.bzl
```diff
@@ -244,18 +244,18 @@ def go_rules_dependencies(force = False):
     _maybe(
         http_archive,
         name = "com_github_golang_mock",
-        # v1.7.0-rc.1, latest as of 2023-11-16
+        # v1.6.0, latest as of 2023-12-15
```
fmeum, line 247: This is a downgrade, not sure whether we want that.

fmeum (commented):
I'm wondering what our policy for `go.mod` updates should be: On the one hand we don't want to unnecessarily bump version requirements for end user projects, on the other hand we could aim for aligning the default versions between WORKSPACE and Bzlmod builds.

What do you think?

tyler-french:
> I'm wondering what our policy for `go.mod` updates should be: On the one hand we don't want to unnecessarily bump version requirements for end user projects, on the other hand we could aim for aligning the default versions between WORKSPACE and Bzlmod builds.
> 
> What do you think?

In this case, I'm not sure there's any we need to bump. The only one is `x/tools`, but we need to keep this at `0.15.0` due to the removal of a package.

fmeum (approved):

PR #3789, opened 2023-12-16
Emit nogo facts into a separate archive

fmeum:
**What type of PR is this?**

Refactoring

Stores nogo facts in a separate archive in preparation for running nogo in a separate action. This gets rid of a lot of code related to packaging nogo facts into the regular export archive.

Note: This commit breaks compatibility with Go 1.16 and earlier.

**Other notes for review**


go/tools/builders/compilepkg.go
```diff
@@ -449,18 +454,25 @@ func compileArchive(
...
-		nogoChan = make(chan error)
-		go func() {
-			nogoChan <- runNogo(ctx, workDir, nogoPath, nogoSrcs, deps, packagePath, importcfgPath, outFactsPath)
-		}()
-		defer func() {
-			if nogoChan != nil {
-				cancel()
-				<-nogoChan
+	if nogoPath != "" {
+		if len(nogoSrcs) == 0 {
```
linzhp, line 458: Can we move this check into `runNogo`? It's the function that calls the nogo tool and it should be responsible for making sure that the list of srcs passed to the tool is not empty.

linzhp (approved):

go/tools/builders/compilepkg.go
fmeum: That looks much nicer, thanks for the suggestion. In fact, I think that we might be able to replace `runNogo` with a pure Starlark implementation in a follow-up PR.

@tyler-french Could you test this in case @linzhp hasn't yet?

linzhp (approved):

go/tools/builders/compilepkg.go
```diff
@@ -517,44 +522,25 @@ func compileArchive(
...
-	if err = extractFileFromArchive(outPath, workDir, pkgDef); err != nil {
-		return err
-	}
-	pkgDefPath := filepath.Join(workDir, pkgDef)
-	if nogoStatus == nogoSucceeded {
-		return appendFiles(goenv, outXPath, []string{pkgDefPath, outFactsPath})
 	}
-	return appendFiles(goenv, outXPath, []string{pkgDefPath})
+
+	return nil
```
sluongng, line 540: could you clarify whether if the old TODO is still relevant?
Afaik we are not using linkobj flag anywhere. Was it replaced by something else?

sluongng: oh nvm I found the flag a few files below 😮‍💨

go/tools/builders/compilepkg.go
```diff
@@ -76,8 +77,9 @@ func compilePkg(args []string) error {
 	fs.StringVar(&nogoPath, "nogo", "", "The nogo binary. If unset, nogo will not be run.")
 	fs.StringVar(&packageListPath, "package_list", "", "The file containing the list of standard library packages")
 	fs.StringVar(&coverMode, "cover_mode", "", "The coverage mode to use. Empty if coverage instrumentation should not be added.")
-	fs.StringVar(&outPath, "o", "", "The output archive file to write compiled code")
-	fs.StringVar(&outFactsPath, "x", "", "The output archive file to write export data and nogo facts")
+	fs.StringVar(&outPath, "o", "", "The full output archive file required by the linker")
+	fs.StringVar(&outXPath, "x", "", "The export-only output archive required to compile dependent packages")
```
sluongng, line 81: In my head, the flags here supposed to be reverse:

Exported files, or ‘x’, are needed by the linker as it let the linker know which object is being exported from which package to “link” the final binary.

Perhaps, to avoid future confusion, we should use “-o” and “-lo” for “object” and “linkobject” instead of “-o” and “-x”. The precedent for this convention is in Go source:

https://cs.opensource.google/go/go/+/master:test/linkobj.go;l=88;drc=b2fd76ab8dd72b9f2135c1e9b15130e6028fc838

Wdyt?

go/tools/builders/compilepkg.go
```diff
@@ -612,6 +603,15 @@ func runNogo(ctx context.Context, workDir string, nogoPath string, srcs []string
 	return nil
 }
 
+func appendToArchive(goenv *env, outPath string, objFiles []string) error {
+	// Use abs to work around long path issues on Windows.
+	// TODO(jayconrod): copy cmd/internal/archive and use that instead of
+	// shelling out to cmd/pack.
```
sluongng, line 609: Nit: I don’t think we should consider this TODO as it will increase the maintenance effort for rules_go. Shelling out help us ensure compatibility against upstream Go SDK.

go/tools/builders/utils.go
```diff
@@ -0,0 +1,53 @@
...
+	return os.Symlink(inPath, outPath)
+}
+
+func copyOrLinkFile(inPath, outPath string) error {
+	if runtime.GOOS == "windows" {
+		return copyFile(inPath, outPath)
+	} else {
+		return linkFile(inPath, outPath)
+	}
+}
```
sluongng, line 53: these are mostly used in cgo2.go?

Do you want to put them into that file instead of using a dedicated utils file?

sluongng (commented):
This is a super cool and important PR. It addressed one of the most intriguing TODO in rules_go that I know of.

The result is much cleaner than original and pave way for future refactoring of nogo. Great job!

go/tools/builders/compilepkg.go
fmeum: Sounds good, I also renamed some variables.

go/private/actions/archive.bzl
```diff
@@ -58,9 +58,13 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         pre_ext += _recompile_suffix
     out_lib = go.declare_file(go, name = source.library.name, ext = pre_ext + ".a")
 
-    # store __.PKGDEF and nogo facts in .x
+    # store export information for compiling dependent packages separately
     out_export = go.declare_file(go, name = source.library.name, ext = pre_ext + ".x")
```
sluongng, line 62: Nit: it would be nice to have these 2 renamed as well. 
Im thinking `s/out_lib/out_archive/` and `s/out_export/out_linkobj/`.

Could be done in a separate PR though to avoid noise.

go/private/actions/compilepkg.bzl
```diff
@@ -108,10 +124,13 @@ def emit_compilepkg(
         args.add("-p", importmap)
     args.add("-package_list", go.package_list)
 
-    args.add("-o", out_lib)
-    args.add("-x", out_export)
-    nogo = go.get_nogo(go)
+    args.add("-lo", out_lib)
+    args.add("-o", out_export)
```
sluongng, line 128: Just to make sure we are on the same page:

- `go tool compile -o` should write to `*.a` or `*.o` files. If `-pack` is set, which we do set it unconditionally, then it's always `*.a` file.

- `go tool compile -linkobj` should write to `*.lo` files.

This seems to be mapping it in reverse(?):

- `-lo` -> out_lib -> `.a`
- `-o` -> out_export -> `.x`

There is not a strong precedent for these, and the extension suppose to change based on usage of `-pack` and `-linkobj`.

https://cs.opensource.google/search?q=%22-linkobj%22&ss=go

I suggest we stick to the example provided in https://cs.opensource.google/go/go/+/master:test/linkobj.go;l=75-90;drc=b2fd76ab8dd72b9f2135c1e9b15130e6028fc838

```bash
go tool compile ... -pack=1 ... -o p1.a -linkobj p1.lo
```

WDYT?

sluongng: Re-reading https://go-review.googlesource.com/c/go/+/22500

> The first part is for reading by future compiles.
The second part is for reading by the final link step.
Splitting the file this way allows distributed build systems
to ship the compile-input part only to compile steps and
the linker-input part only to linker steps.

The implication here is that for subsequent compile actions, we should only need `.a`. And for downstream link actions, we should only need `.lo`.

Right now, we are including both output types in both compile and link actions?
Potentially, this could be split into different providers for downstream action to consume separately, making the input root trees for those actions relatively smaller.
No need to change it in this PR though.

sluongng: Hmm now im not sure if the linker requires the link object file to be ".a" for it to work 🤔

tyler-french (approved):
Tested this at Uber. We were able to get it working by adjusting certain rules to not depend on things being in the archives, and using the export files instead. We may want to add some notes to the release with this.

go/private/actions/compilepkg.bzl
fmeum: > Right now, we are including both output types in both compile and link actions?

The compilepkg action only depends on the export files of its dependencies, so we should already be doing this right.

sluongng: Perhaps dropbox’s rules could be a good point of reference 

https://github.com/dropbox/dbx_build_tools/blob/ee7d614c85217adb061162ed0e5fd852d893b939/build_tools/go/go.bzl#L426-L431

fmeum: > Hmm now im not sure if the linker requires the link object file to be ".a" for it to work 🤔

That's what I'm thinking as well: the full link object may be handed over to external linkers not under our control, so I would say it's safer to keep the standard `.a` extension.

We can always rename the files in a separate change to improve consistency with vanilla Go.

sluongng: > The compilepkg action only depends on the export files of its dependencies, so we should already be doing this right.

I will test this PR out and double check this. I think you might be right though.

avagin:
This pull request has not been tested and broke users. The motivation of the change is unclear. In gVisor, we have a few addition nogo analyzers that require to have access to object files. Here is one of them: https://cs.opensource.google/gvisor/gvisor/+/master:tools/checkescape/checkescape.go.

PR #3661, opened 2023-08-15
cgo packages with assembly: Support CGO_ENABLED=0

evanj:
**What type of PR is this?**

> Uncomment one line below and remove others.
>
Bug fix

**What does this PR do? Why is it needed?**

Go supports building cgo packages both with and without Cgo. It uses build constraints to exclude the appropriate files. When building a Cgo package with assembly files, we must exclude them. Previous to this change, rules_go would run the Go assembler on them. This meant that packages which had "conditional" imports of cgo libraries with assembly would fail when compiled with cgo disabled.

Add tests for these two cases:
* asm_conditional_cgo: A Go package that compiles both with and without Cgo.
* asm_dep_conditional_cgo: A Go package that conditionally imports a package with Cgo.

**Which issues(s) does this PR fix?**

Fixes #3411

**Other notes for review**

I tested this with the DataDog/zstd package. This now works both with and without cgo/cross compilation: https://github.com/evanj/ddzstdbazel

PR #3798, opened 2023-12-24
Exclude assembly files in cgo packages in pure mode

hugelgupf:
This matches `go build` behavior: assembly files in packages with cgo are processed through cgo first. This means that noncgo compilation excludes assembly files in packages with cgo.

Note https://github.com/golang/go/blob/2184a394777ccc9ce9625932b2ad773e6e626be0/src/cmd/go/internal/work/exec.go#L758 in Go build source

Example package: https://github.com/cloudflare/circl/tree/main/dh/x25519 curve_amd64.(h|s)

Does not compile correctly before this patch when pure = "on".

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -119,9 +119,16 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         )
     else:
         cgo_deps = depset()
+        sources = split.go + split.c + split.cxx + split.objc + split.headers
+        # If the package has cgo, cgo processes the Assembly files first.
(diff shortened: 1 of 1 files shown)


fmeum:
Could you explain how this relates to https://github.com/bazelbuild/rules_go/pull/3661, which I thought had fixed this issue? Sorry, I'm not very familiar with how the vanilla Go toolchain handles cgo.

albertocavalcante:
Hi @hugelgupf, did you have a chance to check @fmeum's comment? 

If this by any chance fixes depending on `cloudflare/circl` with `pure = on` this will be greatly appreciated.

Thanks!

PR #3806, opened 2023-12-29
gopackagesdriver: skip root packages that can't be built

### LINKED ISSUE #3805: gopackagesdriver: returns roots for incompatible targets

`v0.44.0`, `master`

n/a

`8.0.0-pre.20231030.2`

yes

macOS, arm64

n/a

### What did you do?

In VSCode configured to use `gopackagesdriver`, open any Go file.

My repository contains some platform-specific targets in our `third_party` (vendor) directory. We don't want `bazel test //...` to match those, so they're marked with `target_compatible_with`. For example:

```

package(default_applicable_licenses = ["//third_party/go/golang.org/x/sys:license"])

go_library(
    name = "registry",
    srcs = [
        "key.go",
        "syscall.go",
        "value.go",
        "zsyscall_windows.go",
    ],
    importpath = "golang.org/x/sys/windows/registry",
    target_compatible_with = [
        "@platforms//os:windows",
    ],
    deps = select({
        "@io_bazel_rules_go//go/platform:windows": [
            "//third_party/go/golang.org/x/sys/windows",
        ],
        "//conditions:default": [],
    }),
)

go_test(
    name = "registry_test",
    srcs = [
        "export_test.go",
        "registry_test.go",
    ],
    embed = [":registry"],
    target_compatible_with = [
        "@platforms//os:windows",
    ],
)
```

No errors. VSCode should load metadata about Go packages and should support editor features like go-to-definition.

### What did you see instead?

In the bottom status bar, I see the error:

```
packages.Load error: root package @//third_party/go/golang.org/x/sys/windows/registry:registry is missing
```

### Analysis

This happens because the response from `gopackagesdriver` includes `@//third_party/go/golang.org/x/sys/windows/registry:registry` in the list of roots (which loading metadata for all packages), but it's not in the package list. The error message is printed by `packages.Load`. I think we should not include the package in the roots because it's not buildable on the host platform. `bazel query` does match it, but `bazel build` can't analyze it. There's no `.pkg.json` file produced for it.

jayconrod:
**What type of PR is this?**

> Bug fix

Omits package IDs from gopackagesdriver's roots list if we can't build package metadata for them.

Fixes #3805

**Other notes for review**


jayconrod:
Are there tests for `gopackagesdriver`? I didn't see a place to add a test case. If I missed it, please let me know, happy to add.

fmeum:
@JamyDev added a basic test that you could extend (if possible): tests/integration/gopackagesdriver/gopackagesdriver_test.go

jayconrod:
Thanks! Not sure how I missed that. Added a test case there.

fmeum (approved):

tests/integration/gopackagesdriver/gopackagesdriver_test.go
```diff
@@ -186,6 +197,35 @@ func TestExternalTests(t *testing.T) {
...
+	out, _, err := bazel_testing.BazelOutputWithInput(reader, "run", "@io_bazel_rules_go//go/tools/gopackagesdriver", "--", "./...")
+	if err != nil {
+		t.Fatalf("error running bazel: %v", err)
+	}
+	var resp response
+	if err := json.Unmarshal(out, &resp); err != nil {
+		t.Fatalf("unmarshaling response: %v", err)
+	}
+
+	helloLabel := "@@//:hello"
```
fmeum, line 213: Whether this label has leading ats or not depends on whether tests are run with Bzlmod. Could you try to make the test independent of that?

jayconrod: Done, should work with and without `@@` now. It passed locally with Bazel 6.4.0 and 7.0.0.

fmeum: Sorry, I missed that we are also testing with Bazel 5.4.0, which doesn't yet preserve the `@` for main repo labels. You probably need yet another case.

It's also interesting to see the Windows test fail with Bazel 7.0.0. I am not sure why that is the case, it may be an actual bug.

jayconrod: The Windows failure does seem like an actual bug. I added some logging, and it looks like it's traversing the `bazel-main` junction in the test's workspace directory. Doesn't seem to happen when I run `bazel query` there directly.

I'll investigate that further as soon as I can, probably in a day or two. Will most likely need a separate PR to go before this one.

JamyDev (approved):

fmeum:
@jayconrod There is still a failure on Bazel 5.4.0. I think it's fine if this fix doesn't support that version, so if you rather want to disable the test in the 5.4.0 CI run, feel free to do so.

jayconrod:
Sorry I haven't gotten back to this yet! I'll take some time this weekend and try to finish it.

I ran into some trouble with the testing framework and was planning to submit another PR ahead of this one that would fix it. Windows, as always, is being troublesome though.

jayconrod:
In #3856, I've moved `gopackagesdriver_test` into `//go/tools/gopackagesdriver`. Since it embeds the library instead of calling via `bazel run`, the test is considerably easier to write, debug, and faster to iterate on.

I've rebased this one, but GitHub won't let me set the base to a branch in my own fork. So... the diff will look ugly until #3856 is merged and this is rebased again. Sorry for the extra confusion.

JamyDev (approved):

fmeum (approved):

jayconrod:
I'm getting a CI error on the job `BCR test module (--incompatible_enable_proto_toolchain_resolution) on :mac: macOS arm64`. It seems unrelated to this PR but doesn't sound like a flake. Actually it seems like a problem with the upstream bazelci.py script.

@fmeum WDYT?

```
$ curl -sS https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/bazelci.py?1739548865 -o bazelci.py
--
  | python3 bazelci.py runner --task=bcr_tests_proto_config_01
  | /Users/buildkite/builds/bk-macos-arm64-ngkq/bazel/rules-go-golang/bazelci.py:833: SyntaxWarning: invalid escape sequence '\$'
  | return re.match("^\${{\s*(\w+)\s*}}$", s)
  | Traceback (most recent call last):
  | File "/Users/buildkite/builds/bk-macos-arm64-ngkq/bazel/rules-go-golang/bazelci.py", line 34, in <module>
  | import requests
  | ModuleNotFoundError: No module named 'requests'
  | 🚨 Error: The command exited with status 1
  | ^^^ +++
  | user command error: exit status 1
```

jayconrod:
It looks like there are a few other reports of CI flakes on macOS arm64 from other projects. I pushed an empty commit and it succeeded. 🤷

PR #3808, opened 2024-01-02
Fix a race detected only if a test times out

patrickmscott:
I believe this can happen because the goroutine is started before the command is started. Start the process first, then on success, start the signal handler. Add a test for regression.

I verified the test fails without my change to wrap.go and the race is from the signal handling goroutine.

**What type of PR is this?**

Bug fix

It prevents a race detection when a test is timed out. We saw this a bunch and devs were really confused thinking there was a legit race detected.

Fixes #

**Other notes for review**

### DIFF
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -124,6 +124,10 @@ func Wrap(pkg string) error {
 	if !filepath.IsAbs(exePath) && strings.ContainsRune(exePath, filepath.Separator) && chdir.TestExecDir != "" {
 		exePath = filepath.Join(chdir.TestExecDir, exePath)
 	}
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGTERM)
+
 	cmd := exec.Command(exePath, args...)
 	cmd.Env = append(os.Environ(), "GO_TEST_WRAP=0")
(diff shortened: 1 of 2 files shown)


fmeum:
@patrickmscott Thanks for catching this. I'm still not sure I understand why there is a race in the first place. Could you post the race detector trace you see?

go/tools/bzltestutil/wrap.go
```diff
@@ -128,19 +128,22 @@ func Wrap(pkg string) error {
...
-	err := cmd.Run()
+	err := cmd.Start()
+	if err == nil {
+		go func() {
+			// When the Bazel test timeout is reached, Bazel sends a SIGTERM that
+			// we need to forward to the inner process.
+			// TODO: This never triggers on Windows, even though Go should simulate
+			//  a SIGTERM when Windows asks the process to close. It is not clear
+			//  whether Bazel uses the required graceful shutdown mechanism.
+			c := make(chan os.Signal, 1)
```
fmeum, line 140: Could you try moving this line and the next out of the Go routine and before the call to `Start`? It's probably just a theoretical concern, but I realize that we otherwise may lose `SIGTERM`s sent to the process before the goroutine executes.

patrickmscott: Done

fmeum (approved):

PR #3820, opened 2024-01-09
Fix coverage collection with modified `COVERAGE_DIR` env variable

### LINKED ISSUE #3819: Coverage is not working if `os` module is used in test
While running coverage on a pkg that use the `os` module, coverage is not outputed. 

0.44.2

0.35.0

7.0.0

Yes

Darwin / arm64

### What did you do?
Please see this repository for reproduction:
https://github.com/supercairos/bazelgoperfectcoverage

running 
`bazel coverage //...` don't produce any coverage. 

Some coverage produced

### What did you see instead?
No coverage produced

fmeum:
**What type of PR is this?**

Bug fix

Fixes #3819 

**Other notes for review**

### DIFF
--- a/go/tools/bzltestutil/lcov.go
+++ b/go/tools/bzltestutil/lcov.go
@@ -28,6 +28,9 @@ import (
 	"testing/internal/testdeps"
 )
 
+// Lock in the COVERAGE_DIR during test setup in case the test uses e.g. os.Clearenv.
+var coverageDir = os.Getenv("COVERAGE_DIR")
+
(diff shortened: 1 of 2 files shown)


fmeum:
I tried to add a test for this, but it failed with race detection, presumably because `os.Clearenv` is prone to races. I verified manually that this fixes the issue.

linzhp (approved):
Maybe you can turn the race detection off for this test?

fmeum:
Found a way to do this by using `Setenv` for the same purpose.

linzhp (approved):
Nice

PR #3827, opened 2024-01-12
go_test: Allow not registering a SIGTERM handler

illicitonion:
**What type of PR is this?**

Bug fix
Feature

The rules_go SIGTERM handler causes tests which themselves test
responding to SIGTERM to panic and fail.


go/private/rules/test.bzl
```diff
@@ -112,6 +112,9 @@ def _go_test_impl(ctx):
         "l_test=" + external_source.library.importpath,
     )
     arguments.add("-pkgname", internal_source.library.importpath)
+
+    # Boolean Go flags require =, so we can't just use the attr as an arg value, we need to convert it ourselves.
+    arguments.add("-register_timeout_handler={}".format("true" if ctx.attr.register_timeout_handler else "false"))
```
fmeum, line 117: This allocates a new string for each target. Instead, you could just have an `if ... else`. Alternatively, since the flag has a default, only adding the flag when it's not the default would be an option as well.

fmeum (commented):
Cc @linzhp Would this help with your tests?

linzhp:
This helps, but we simply add `signal.Reset(syscall.SIGTERM)` to the test cases or `TestMain` of them that register `SIGTERM` handler. @illicitonion did you try that approach?

go/private/rules/test.bzl
illicitonion: Done

illicitonion:
> This helps, but we simply add `signal.Reset(syscall.SIGTERM)` to the test cases or `TestMain` of them that register `SIGTERM` handler. @illicitonion did you try that approach?

Ooh that's a nice idea! Yeah, this approach works, and it seems reasonable (though a little noisy) for tests which are explicitly handling signals themselves, but it feels much noisier to have to add to all tests if they use `goleak`...

fmeum:
goleak will come with a default ignorelist entry in the future as far as I understood @linzhp. 

I do prefer the Go-only solution in that case if we document it.

linzhp:
Good point on goleak. Uber maintains a internal patch of all known leaks, but there is no plan to put the list in the open source goleak project. It's unfair to ask every user of goleak to maintain a patch. So in combination of the SIGTERM reset and goleak opt-out, this PR is quite useful, unless we have better implementation of the timeout handler.

linzhp:
I am not a maintainer of goleak, based on [this reply](https://github.com/uber-go/goleak/issues/119#issuecomment-1892702927), they don't want a default ignore list, given that goleak supports [IgnoreAnyFunction](https://pkg.go.dev/go.uber.org/goleak#IgnoreAnyFunction). 

So tests that uses goleak can call `VerifyNone(t, goleak.IgnoreAnyFunction("github.com/bazelbuild/rules_go/go/tools/bzltestutil.RegisterTimeoutHandler.func1")))` or 

```go
func TestMain(m *testing.M) {
  goleak.VerifyTestMain(m, goleak.IgnoreAnyFunction("github.com/bazelbuild/rules_go/go/tools/bzltestutil.RegisterTimeoutHandler.func1"))
}
```

fmeum:
I followed up on the goleak thread, maybe we can solve this by introducing an ignore mechanism that rules_go can use and goleak doesn't have to maintain.

PR #3832, opened 2024-01-16
Fix regression in empty `.go` file filtering for nogo

### LINKED ISSUE #3828: nogo fails using WORKSPACE.bazel after upgrade to 0.45.0

0.45.0

0.35.0

```
bazel version
Build label: 7.0.0- (@non-git)
Build time: Tue Jan 1 00:00:00 1980 (315532800)
Build timestamp: 315532800
Build timestamp as int: 315532800
```

ya

Linux amd64

### What did you do?

Upgrade rules_go to 0.45.0 from 0.44.2.

Upon upgrading, nogo fails on things that are wrong and returns garbage output like:

```
bazel-out/k8-fastbuild/bin/path/to/file/1437484352.go:1:1: at least one file in a package should have a package comment (ST1000)
```

Note that only the file path was replaced, not the file name. Bazel literally returns a random number here.  Downgrading to 0.44.2 makes the issue go away.

no nogo failures

### What did you see instead?

gibberish  nogo failures

fmeum:
**What type of PR is this?**

Bug fix

6236dd80670cdfe1477a484e98e3f82dbcc28fe9 regressed the nogo exclusion logic for the empty `.go` file generated for targets that otherwise don't contain any `.go` files. The test had been silently disabled a while ago when the name pattern for the empty file was changed.

This is fixed by reintroducing the exclusion logic and making the test more strict.

Fixes #3828

**Other notes for review**

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -272,7 +272,7 @@ func compileArchive(
 	compilingWithCgo := haveCgo && cgoEnabled
 
 	filterForNogo := func(slice []string) []string {
-		var filtered []string
+		filtered := make([]string, 0, len(slice))
 		for _, s := range slice {
 			// Do not subject the generated empty .go file to nogo checks.
 			if s != emptyGoFilePath {
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -203,6 +203,7 @@ func compileArchive(
 	}
 	defer cleanup()
 
+	emptyGoFilePath := ""
 	if len(srcs.goSrcs) == 0 {
 		// We need to run the compiler to create a valid archive, even if there's nothing in it.
 		// Otherwise, GoPack will complain if we try to add assembly or cgo objects.
@@ -227,7 +228,7 @@ func compileArchive(
 			matched:  true,
 			pkg:      "empty",
 		})
-
+		emptyGoFilePath = emptyGoFile.Name()
 	}
 	packageName := srcs.goSrcs[0].pkg
 	var goSrcs, cgoSrcs []string
@@ -270,18 +271,24 @@ func compileArchive(
 	// constraints.
 	compilingWithCgo := haveCgo && cgoEnabled
 
+	filterForNogo := func(slice []string) []string {
+		var filtered []string
+		for _, s := range slice {
+			// Do not subject the generated empty .go file to nogo checks.
+			if s != emptyGoFilePath {
+				filtered = append(filtered, s)
+			}
+		}
+		return filtered
+	}
 	// When coverage is set, source files will be modified during instrumentation. We should only run static analysis
 	// over original source files and not the modified ones.
 	// goSrcsNogo and cgoSrcsNogo are copies of the original source files for nogo to run static analysis.
-	goSrcsNogo := goSrcs
-	cgoSrcsNogo := cgoSrcs
+	goSrcsNogo := filterForNogo(goSrcs)
+	cgoSrcsNogo := append([]string{}, cgoSrcs...)
 
 	// Instrument source files for coverage.
 	if coverMode != "" {
-		// deep copy original source files for nogo static analysis, avoid being modified by coverage.
-		goSrcsNogo = append([]string{}, goSrcs...)
-		cgoSrcsNogo = append([]string{}, cgoSrcs...)
-
 		relCoverPath := make(map[string]string)
 		for _, s := range coverSrcs {
 			relCoverPath[abs(s)] = s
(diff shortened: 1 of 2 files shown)


fmeum:
cc @sluongng

go/tools/builders/compilepkg.go
```diff
@@ -270,16 +271,17 @@ func compileArchive(
 	// constraints.
 	compilingWithCgo := haveCgo && cgoEnabled
 
+	keepForNogo := func(s string) bool { return !ignoreForNogo[s] }
```
sluongng, line 274: Nit: i think you could inline the whole filterSlice func here instead of declaring a utility func + a filter func.

sluongng (approved):
Great fix. Thanks!

go/tools/builders/compilepkg.go
fmeum: Done

go/tools/builders/compilepkg.go
```diff
@@ -203,6 +203,7 @@ func compileArchive(
 	}
 	defer cleanup()
 
+	ignoreForNogo := make(map[string]bool)
```
sluongng, line 206: Sorry for not spotting this earlier, but is this a length == 1 map?
If that’s the case, could we use a simple `string` var instead?

go/tools/builders/compilepkg.go
```diff
@@ -270,16 +271,25 @@ func compileArchive(
...
 	// goSrcsNogo and cgoSrcsNogo are copies of the original source files for nogo to run static analysis.
-	goSrcsNogo := goSrcs
+	goSrcsNogo := filterForNogo(goSrcs)
 	cgoSrcsNogo := cgoSrcs
 
 	// Instrument source files for coverage.
 	if coverMode != "" {
 		// deep copy original source files for nogo static analysis, avoid being modified by coverage.
-		goSrcsNogo = append([]string{}, goSrcs...)
+		goSrcsNogo = filterForNogo(append([]string{}, goSrcs...))
```
sluongng, line 292: We already filter once outside the `if` clause, let’s just re-use the result?

```suggestion
		goSrcsNogo = filterForNogo(append([]string{}, goSrcsNogo...))
```

sluongng: and since the filterForNogo is creating a copied slice anyway, I think we could simplify this further.

go/tools/builders/compilepkg.go
linzhp: nit for naming: we should probably call this `emptyFiles`, just in case there are other usage of them in the future. If we need to use map as a set, we prefer `map[string]struct{}`, because map[string]bool has 3 states for every entry: true,false,absent, while map[string]struct{} only has two: present,absent, which has less ambiguity.

go/tools/builders/compilepkg.go
tyler-french: +1, `filterForNogo` is effectively a deep copy

tests/core/nogo/generate/empty_test.go
```diff
@@ -77,12 +75,11 @@ var Analyzer = &analysis.Analyzer{
 
 func run(pass *analysis.Pass) (interface{}, error) {
 	for _, f := range pass.Files {
-		pos := pass.Fset.PositionFor(f.Pos(), false)
-
-		if strings.HasSuffix(pos.Filename, filepath.Join(".", "_empty.go")) {
+		// Fail on any package that isn't the test's package 'simple' or 'main'.
```
tyler-french, line 78: clarification: what is `simple`?

go/tools/builders/compilepkg.go
tyler-french: To @sluongng's point, could we avoid needing a set when there is a maximum of one empty file? I think a `string` or `*string` would indeed

go/tools/builders/compilepkg.go
fmeum: Changed this by making both deep copies unconditional, the alternative just seemed too confusing for little gain.

go/tools/builders/compilepkg.go
```diff
@@ -270,18 +271,24 @@ func compileArchive(
 	// constraints.
 	compilingWithCgo := haveCgo && cgoEnabled
 
+	filterForNogo := func(slice []string) []string {
+		var filtered []string
```
tyler-french, line 275: optional, pre-allocate
```suggestion
		filtered := make([]string, 0, len(slice))
```

tyler-french (approved):
Tested at Uber and LGTM! Added a small optional suggestion

PR #3833, opened 2024-01-16
Print stack traces without panicking on test timeout

fmeum:
**What type of PR is this?**

Bug fix
Feature

It replaces a call to `panic` in the timeout handler to just printing all stack traces, which removes the need to deregister the `SIGTERM` handler when the test relies on particular `SIGTERM` behavior at the cost of having Bazel kill the process after the grace period instead of the `panic` doing it right away.

Also removes the `SIGTERM` forwarding logic from the wrapper as it isn't needed: Bazel sends `SIGTERM` to all child processes.

Related to #3827 

**Other notes for review**

### DIFF
--- a/go/tools/bzltestutil/timeout.go
+++ b/go/tools/bzltestutil/timeout.go
@@ -15,22 +15,25 @@
 package bzltestutil
 
 import (
+	"fmt"
 	"os"
 	"os/signal"
-	"runtime/debug"
+	"runtime"
 	"syscall"
 )
 
 func RegisterTimeoutHandler() {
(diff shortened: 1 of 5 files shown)


fmeum:
@illicitonion Does this resolve the most pressing part of the issue for you?

illicitonion:
Yeah, this resolves the "tests may handle" part - the goleak part is still an issue but I'm on board with it potentially having a different solution.

Thanks!

fmeum:
I simplified this further and bumped the limit for stack sizes.

PR #3839, opened 2024-01-20
nogo: Allow not running analyzers for packages.

### LINKED ISSUE #3838: nogo: Allow not running analyzers for certain packages
Currently, nogo is run unconditionally for all go code. The exclude/include escape hatches only ignore diagnostic errors from running analyzers, however, the analyzers are still run.

This is problematic because some external code like AWS's EC2 go client are huge and take a lot of memory to compile. This causes our build to OOM. It would be nice to have a mechanism to not run an analyzer at all.

This would allow builds to run must faster since you aren't linting external code that you don't care to lint.

I propose adding two new fields to the nogo config:
```
"exclude_pkg": {
  "github.com/DolceTriade/generated": "this is generated code"
},
"only_pkg": {
  "github.com/DolceTriade": "only lint my code"
}
```

The semantics of exclude_pkg and only_pkg would be the same as only_files and exclude_files, except they operate at the package level and instead of ignoring errors from an analyzer, they don't run that analyzer all together.

DolceTriade:
Currently, nogo is run unconditionally for all go code. The exclude/include escape hatches only ignore diagnostic errors from running analyzers, however, the analyzers are still run.

This is problematic because some external code like AWS's EC2 go client are huge and take a lot of memory to compile. This causes our build to OOM. It would be nice to have a mechanism to not run an analyzer at all.

This would allow builds to run must faster since you aren't linting external code that you don't care to lint.

This PR adds two new fields to the nogo config: "exclude_pkg" and "only_pkg". These semantics are the same as "only_file" and "exclude_file" except they operate at the package level.

An example config might look like:
```
 {
        "_base": {
            "exclude_files": {
                "external/": "third_party",
                "proto/": "generated protobuf",
                "testmain\\.go": "generated test main for go tests"
            },
            "only_pkgs": {
                "github.com/DolceTriade": "Only run lints for my code"
            }
        }
}
```

Fixes #3838

**What type of PR is this?**

> Feature

### DIFF
--- a/go/tools/builders/generate_nogo_main.go
+++ b/go/tools/builders/generate_nogo_main.go
@@ -81,6 +81,26 @@ var configs = map[string]config{
 			{{- end}}
 		},
 		{{- end}}
+		{{- if $config.OnlyPkgs}}
+		onlyPkgs: []*regexp.Regexp{
+			{{- range $path, $comment := $config.OnlyPkgs}}
+			{{- if $comment}}
+			// {{$comment}}
+			{{end -}}
+			{{printf "regexp.MustCompile(%q)" $path}},
+			{{- end}}
+		},
+		{{- end -}}
+		{{- if $config.ExcludePkgs}}
+		excludePkgs: []*regexp.Regexp{
+			{{- range $path, $comment := $config.ExcludePkgs}}
+			{{- if $comment}}
+			// {{$comment}}
+			{{end -}}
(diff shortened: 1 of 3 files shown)


DolceTriade:
Based on discussion in the issue, we want to go in another direction since nogo already supports this funcationlity, but not exposed to WORKSPACE users.

PR #3842, opened 2024-01-22
nogo: Create a go_register_nogo wrapper for WORKSPACE users.

### LINKED ISSUE #3838: nogo: Allow not running analyzers for certain packages

DolceTriade:
**What type of PR is this?**
> Feature

Currently, workspace users register nogo targets by abusing the go_register_toolchain() function, which calls go_register_nogo(). Instead, we should expose go_register_nogo to workspace users directly. This will allow workspace users and bzlmod users to use the same underlying code because go_register_nogo allows users to specify includes and excludes which you currently can't do with WORKSPACE.

Add a test to verify this functionality works in WORKSPACE too.

Fixes #3838

### DIFF
--- a/go/nogo.rst
+++ b/go/nogo.rst
@@ -71,18 +71,24 @@ want to run.
         visibility = ["//visibility:public"],
     )
 
-Pass a label for your `nogo`_ target to ``go_register_toolchains`` in your
+Pass a label for your `nogo`_ target to ``go_register_nogo`` in your
 ``WORKSPACE`` file. When using ``MODULE.bazel``, see the Bzlmod_ documentation
 instead.
 
 .. code:: bzl
 
-    load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
+    load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_nogo")
     go_rules_dependencies()
-    go_register_toolchains(nogo = "@//:my_nogo") # my_nogo is in the top-level BUILD file of this workspace
+    go_register_toolchains(version = "1.20.7")
+    go_register_nogo(
+      nogo = "@//:my_nogo"  # my_nogo is in the top-level BUILD file of this workspace
+      includes = ["@//:__subpackages__"],  # Labels to lint. By default only lints code in workspace.
+      excludes = ["@//generated:__subpackages__"],  # Labels to exclude.
+    )
 
 **NOTE**: You must include ``"@//"`` prefix when referring to targets in the local
-workspace.
+workspace. Also note that you cannot use this to refer to bzlmod repos, as the labels
+don't go though repo mapping.
 
 The `nogo`_ rule will generate a program that executes all the supplied
 analyzers at build-time. The generated ``nogo`` program will run alongside the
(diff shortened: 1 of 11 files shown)


fmeum (approved):
Looks good to me. Could you look into the failing tests?

go/nogo.rst
```diff
@@ -71,15 +71,19 @@ want to run.
...
+    load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_nogo")
     go_rules_dependencies()
-    go_register_toolchains(nogo = "@//:my_nogo") # my_nogo is in the top-level BUILD file of this workspace
+    go_register_nogo(
+      nogo = "@//:my_nogo"  # my_nogo is in the top-level BUILD file of this workspace
+      includes = ["@//:__subpackages__"],  # Labels to lint. By default only lints code in workspace.
+      excludes = ["@//generated:__subpackages__"],  # Labels to exclude.
+    )
 
 **NOTE**: You must include ``"@//"`` prefix when referring to targets in the local
```
fmeum, line 88: Could you add a note saying that this can't refer to Bzlmod repos? The labels don't go through repo mapping.

go/nogo.rst
```diff
@@ -71,15 +71,19 @@ want to run.
...
 ``WORKSPACE`` file. When using ``MODULE.bazel``, see the Bzlmod_ documentation
 instead.
 
 .. code:: bzl
 
-    load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
+    load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_nogo")
     go_rules_dependencies()
-    go_register_toolchains(nogo = "@//:my_nogo") # my_nogo is in the top-level BUILD file of this workspace
+    go_register_nogo(
```
fmeum, line 82: How does this interact with `go_register_toolchain` calls that users make to e.g. select a different Go version? Could you document an example?

DolceTriade: Added an example. Previously you would have two calls to `go_register_toolchains()` and now you have one call to `go_register_toolchains` and one to `go_register_nogo`

go/nogo.rst
DolceTriade: done

DolceTriade:
Tests should be fixed...at least they don't fail locally.

DolceTriade:
@fmeum I'm not sure the includes/excludes support works on Bazel 5.4.0. I suspect this was never tested because it was only used by bzlmod until now.

Is it important for the includes/excludes support to work for 5.4.0? Is there a way we can ignore this test on older versions of bazel?

fmeum:
Sure, just exclude it here: https://github.com/bazelbuild/rules_go/blob/30099a6add3c43706b4eec82b773b78310874935/.bazelci/presubmit.yml#L29

DolceTriade:
Ok! Now all tests pass!

PR #3848, opened 2024-02-06
go_test: ensure external source compilation has data

sluongng:
go_test is split into 2 compilation actions, one for internal test,
where package name of the test is the same with library package.

```bash
package foo

package foo
```

and external test where the package name is different.
```bash
package foo

package foo_test
```

Ensure that the compilation for external test has access to go_test's
data attribute.

### DIFF
--- a/tests/core/go_test/x_defs/BUILD.bazel
+++ b/tests/core/go_test/x_defs/BUILD.bazel
@@ -49,22 +49,36 @@ go_library(
     x_defs = {"Qux": "Qux"},
 )
 
+genrule(
+    name = "data_dep",
+    outs = ["data_dep.txt"],
+    cmd = "touch $@",
+)
+
 go_library(
     name = "x_defs_lib",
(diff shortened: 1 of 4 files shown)


tests/core/go_test/x_defs/BUILD.bazel
```diff
@@ -49,22 +49,36 @@ go_library(
...
 go_test(
     name = "x_defs_test",
     srcs = ["x_defs_test.go"],
-    data = ["x_defs_test.go"],
+    data = [
+        "data_dep.txt",
+        "x_defs_test.go",
+    ],
     x_defs = {
         "BinGo": "$(rlocationpath x_defs_test.go)",
```
fmeum, line 80: Interesting, I didn't know `srcs` was always read by `ctx.expand_locations`. That made the test pretty weak.

fmeum (approved):

PR #3852, opened 2024-02-08
runfiles: provide Rlocations to look up multiple paths

sluongng:
Typically, we will be using `rlocationpath` and it's plural form,
`rlocationpaths` to generate the runfile location path from the target
label.  In the case of `rlocationpaths` (plural), the result string
will include multiple location paths, white-space separated.

Provide `runfiles.Rlocations()` to handle the string generated from
`rlocationpaths`.

### DIFF
--- a/go/runfiles/global.go
+++ b/go/runfiles/global.go
@@ -28,12 +28,49 @@ func Rlocation(path string) (string, error) {
 	return RlocationFrom(path, CallerRepository())
 }
 
+// Rlocations returns the absolute paths name of a set of runfiles.
+// The input "paths" is expected to be a list of runfile names that match the
+// criterias documented in runfiles.Rlocation(), separated by a single white-space
+// character.
+// Usually this input value could be generated from Bazel's
+//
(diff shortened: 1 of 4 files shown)


sluongng:
@fmeum ping because you know the runfiles better than anyone

fmeum:
The plural forms of the location functions are unfortunately quite problematic: They can't really work with runfiles paths containing spaces, which folks are working on getting supported, and whether they result in space-separated paths or multiple arguments depends on whether (and how) the rule chooses to shell tokenize the `args` after expansion.

I'm actually thinking of introducing a new way to perform location expansion that instead returns an `Args` object, but I of course would first need to get a design doc out.

All of that is to explain why I wouldn't want to commit to an "official" runfiles function that hard-codes "spaces as separators" in its behavior. Conceptually speaking, since this is essentially a `map` over a slice obtained via `strings.Split`, I have a preference for just having users roll their own utility function for this if needed.

sluongng:
sounds reasonable to me

PR #3856, opened 2024-02-11
gopackagesdriver: move and simplify test

jayconrod:
**What type of PR is this?**

> Bug fix

Moved //tests/integration/gopackagesdriver:gopackagesdriver_test to //go/tools/gopackagesdriver:gopackagesdriver_test.

Previously the test invoked gopackagesdriver with 'bazel run' in a temporary workspace, which forced Bazel to build it, slowing down the test significantly, especially on Windows. Now, the test embeds the gopackagesdriver library and invokes it directly. The driver still runs in a temporary workspace and still needs to invoke bazel, so the test is still slow, but it's faster than before.

Also, a number of other improvements:

- gopackagesdriver now sets --consistent_labels if the bazel version is 6.4.0 or later instead of guessing based on the link-stamped rules_go repo name.
- bazel_testing.BazelCmd now writes --output_user_root to a .bazelrc file instead of passing it to commands. gopackagesdriver invokes bazel directly, so this lets it use the same output root.

For #3805
Prerequisite for test in #3806

**Other notes for review**


fmeum:
Thanks, slow tests are a pain.

The failing test suggests that there could be some issue related to flags changing between `build` and `config` or another command. `bazel config` just dumps Skyframe state, so any kind of invalidation happening prior to it (e.g. even touching a file) can lead to a failure.

jayconrod:
Hmm, the only thing in this PR that might have affected this test was moving `--output_user_root` to `.bazelrc`. I would think that would have no effect on `bazel cquery` and `bazel config`. I have noticed that Bazel shuts down after every command in `go_bazel_test`, maybe because the working directory is next to the exec root. I'd expect `bazel config` not to work at all, but that should have been an issue earlier.

I actually can't reproduce this: instead, I get

```
--- FAIL: TestGoBinaryNonGoAttrsAreReset (10.87s)
    hermeticity_test.go:202: bazel build //:main //:helper: Starting local Bazel server and connecting to it...
        Computing main repo mapping:
        WARNING: --enable_bzlmod is set, but no MODULE.bazel file was found at the workspace root. Bazel will create an empty MODULE.bazel file. Please consider migrating your external dependencies from WORKSPACE to MODULE.bazel. For more details, please refer to https://github.com/bazelbuild/bazel/issues/18958.       
        Computing main repo mapping:
        Computing main repo mapping:
        Loading:
        Loading: 0 packages loaded
        Analyzing: 2 targets (1 packages loaded, 0 targets configured)
        Analyzing: 2 targets (70 packages loaded, 676 targets configured)
        Analyzing: 2 targets (72 packages loaded, 9952 targets configured)
        Analyzing: 2 targets (72 packages loaded, 9958 targets configured)
        ERROR: C:/_b/zrbrqgaf/external/io_bazel_rules_go/BUILD.bazel:42:7: in stdlib rule @@io_bazel_rules_go//:stdlib:                                         
                File "C:/_b/zrbrqgaf/external/io_bazel_rules_go/go/private/rules/stdlib.bzl", line 33, column 20, in _stdlib_impl
                        go = go_context(ctx)
                File "C:/_b/zrbrqgaf/external/io_bazel_rules_go/go/private/context.bzl", line 458, column 20, in go_context
                        mode = get_mode(ctx, toolchain, cgo_context_info, go_config_info)
                File "C:/_b/zrbrqgaf/external/io_bazel_rules_go/go/private/mode.bzl", line 112, column 13, in get_mode
                        fail(("linkmode '{}' can't be used when cgo is disabled. Check that pure is not set to \"off\" and that a C/C++ toolchain is configured for " +
        Error in fail: linkmode 'c-archive' can't be used when cgo is disabled. Check that pure is not set to "off" and that a C/C++ toolchain is configured for your current platform. If you defined a custom platform, make sure that it has the @io_bazel_rules_go//go/toolchain:cgo_on constraint value.
```

That fails on `master`. So something is more deeply broken. I'll try, but I'm probably not going to have more time to fix it this weekend or in the next two weeks. Okay if we leave these PRs open for a bit?

JamyDev:
Did a quick scan over the code, looks good. I'll do a more thorough review tomorrow.

jayconrod:
All right, tests are green.

The reason why I was seeing `hermeticity_test` fail on master was that I was missing some environment variables that are set in CI.

After finding those I could reproduce the error, caused by using the wrong output root directory.

go/tools/gopackagesdriver/gopackagesdriver_test.go
```diff
@@ -186,7 +167,71 @@ func TestExternalTests(t *testing.T) {
...
+		"SYSTEMROOT":  true,
+		"TEMP":        true,
+		"TMP":         true,
+		"TZ":          true,
+		"USER":        true,
+	}
+	var oldEnv []string
+	for _, env := range os.Environ() {
+		key, value, ok := strings.Cut(env, "=")
+		if ok && !allowEnv[key] {
```
achew22, line 199:
```suggestion
	allowEnv := map[string]struct{}{
		"HOME":        {},
		"PATH":        {},
		"PWD":         {},
		"SYSTEMDRIVE": {},
		"SYSTEMROOT":  {},
		"TEMP":        {},
		"TMP":         {},
		"TZ":          {},
		"USER":        {},
	}
	var oldEnv []string
	for _, env := range os.Environ() {
		key, value, ok := strings.Cut(env, "=")
		if allowed := allowEnv[key]; ok && allowed {
```

go/tools/gopackagesdriver/json_packages_driver.go
```diff
@@ -23,7 +23,7 @@ type JSONPackagesDriver struct {
 	registry *PackageRegistry
 }
 
-func NewJSONPackagesDriver(jsonFiles []string, prf PathResolverFunc, bazelVersion string) (*JSONPackagesDriver, error) {
+func NewJSONPackagesDriver(jsonFiles []string, prf PathResolverFunc, bazelVersion bazelVersion) (*JSONPackagesDriver, error) {
```
achew22, line 26: There is no function that returns a `bazelVersion` exposed in this package. This is implicitly making this constructor package private. If that's the intent, can you make the whole thing private `newJSONPackagesDriver`?

achew22 (approved):

go/tools/gopackagesdriver/json_packages_driver.go
jayconrod: Mind if I address that in a follow-up?

This is a `main` package so nothing can import it, but it has a lot of exported symbols. They all ought to be renamed.

go/tools/gopackagesdriver/gopackagesdriver_test.go
jayconrod: Done, though I needed to adjust it a bit and didn't commit it directly.

JamyDev (approved):

jayconrod:
I'd be grateful if someone can merge this. Google revoked all my powers here when I left, so I'm just a regular citizen now.

PR #3866, opened 2024-02-15
[go_library] add .x file to declared output files in DefaultInfo

tyler-french:
After https://github.com/bazelbuild/rules_go/pull/3789, there's no easy way to analyze the package information of a library file from a test using [`bazel.ListRunfiles`](https://pkg.go.dev/github.com/bazelbuild/rules_go/go/tools/bazel#ListRunfiles) without manually copying over the `.x` file somewhere.

This file is arguably more important to users than the machine code, and is generally small, so we should consider exporting it.

### DIFF
--- a/go/private/rules/library.bzl
+++ b/go/private/rules/library.bzl
@@ -44,7 +44,7 @@ def _go_library_impl(ctx):
         source,
         archive,
         DefaultInfo(
-            files = depset([archive.data.file]),
+            files = depset([archive.data.export_file]),
         ),
         coverage_common.instrumented_files_info(
             ctx,


go/private/rules/library.bzl
```diff
@@ -44,7 +44,10 @@ def _go_library_impl(ctx):
         source,
         archive,
         DefaultInfo(
-            files = depset([archive.data.file]),
+            files = depset([
+                archive.data.file,
```
fmeum, line 48: How about only advertising the `export_file` here? Having multiple file outputs always makes a rule more tedious to use.

tyler-french: Sounds good, as long as we're not depending on this file output.

PR #3872, opened 2024-02-24
Remove Latin-1 workaround on Bazel 6.4.0+

mattyclarkson:
The fix for Latin-1 encoded files was landed in [6.4.0] after being landed on Bazel [master].

We can conditionally use the old, unhermetic, `tar` workaround on modern Bazel versions.

[master]: https://github.com/bazelbuild/bazel/pull/18448
[6.4.0]: https://github.com/bazelbuild/bazel/pull/19765

**What type of PR is this?**

Bug fix

We currently use a unhermetic `tar` workaround when a Go SDK URL ends with `.tar.gz` but modern Bazel versions work fine unpacking the `tar` archive. We can conditionally use the workaround.

N/A

**Other notes for review**

None

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -16,7 +16,6 @@ load("//go/private:common.bzl", "executable_path")
 load("//go/private:nogo.bzl", "go_register_nogo")
 load("//go/private/skylib/lib:versions.bzl", "versions")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "patch", "read_user_netrc", "use_netrc")
-load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
 
 MIN_SUPPORTED_VERSION = (1, 14, 0)
 
@@ -442,7 +441,7 @@ def _remote_sdk(ctx, urls, strip_prefix, sha256):
     # in Bazel 6.0.0+ (bazelbuild/bazel#16052). The only situation where
     # .zip files are needed seems to be a macOS host using a Windows toolchain
     # for remote execution.
-    if bazel_features.external_deps.extract_supports_unicode_filenames:
+    if versions.is_at_least("6.4.0", versions.get() or "6.4.0"):
         ctx.download_and_extract(
             url = urls,
             stripPrefix = strip_prefix,
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -16,6 +16,7 @@ load("//go/private:common.bzl", "executable_path")
 load("//go/private:nogo.bzl", "go_register_nogo")
 load("//go/private/skylib/lib:versions.bzl", "versions")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "patch", "read_user_netrc", "use_netrc")
+load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
 
 MIN_SUPPORTED_VERSION = (1, 14, 0)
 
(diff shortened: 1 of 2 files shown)


go/private/sdk.bzl
```diff
@@ -441,7 +441,10 @@ def _remote_sdk(ctx, urls, strip_prefix, sha256):
     # in Bazel 6.0.0+ (bazelbuild/bazel#16052). The only situation where
     # .zip files are needed seems to be a macOS host using a Windows toolchain
     # for remote execution.
-    if urls[0].endswith(".tar.gz"):
+    if (urls[0].endswith(".tar.gz") and
+          # Development versions of Bazel have an empty version string. We assume that they are
+          # more recent than the version that introduced the fix.
+          not versions.is_at_least("6.4.0", versions.get() or "6.4.0")):
```
fmeum, line 447: Thanks for sending this, I completely forgot to clean this up after the fix.

We aim to consolidate all feature detection in https://github.com/bazel-contrib/bazel_features. Could you submit a PR to that repo adding a new feature (e.g. `extract_supports_unicode_filenames`) here:
https://github.com/bazel-contrib/bazel_features/blob/697d54eaf05ed83972204f8deee9817dcd6edfde/features.bzl#L21

We can then bump the `bazel_features` dependency and use the named feature.

mattyclarkson:
@fmeum this follows up on your `tar` workaround from #2771 to make it conditional on Bazel 6.4.0+ as that has bazelbuild/bazel#19765 so we can be hermetic for the Go SDK download.

go/private/sdk.bzl
fmeum: In fact, if we do this, we can also get rid of the `zip` workaround in this case.

mattyclarkson: bazel-contrib/bazel_features#41

mattyclarkson: > In fact, if we do this, we can also get rid of the zip workaround in this case.
Sorry, I don't understand. I was going to change the logic to:
```
if bazel_features.external_deps.extract_supports_unicode_filenames:
  # Do `download_and_extract`
elif urls[0].endswith(".tar.gz"):
  # Do `tar` workaround
elif urls[0].endswith(".zip") and host_goos != "windows":
  # Do `zip` workaround
```
What should I change it too. I'm happy for you to pick this up if it would be more expedient.

fmeum: Yeah, that logic is exactly what I had in mind!

The feature is available in bazel_features v1.6.0 now.

mattyclarkson:
```
ERROR: Failed to load Starlark extension '@io_bazel_rules_go_bazel_features//:features.bzl'.

Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @io_bazel_rules_go_bazel_features
This could either mean you have to add the '@io_bazel_rules_go_bazel_features' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
```

Is this anything to do with the double-step setup polyfilling we're doing in the `WORKSPACE`.

I'm a bit in the dark what's going on here. Any tips?

go/private/sdk.bzl
```diff
@@ -470,12 +479,7 @@ def _remote_sdk(ctx, urls, strip_prefix, sha256):
             auth = auth,
         )
     else:
-        ctx.download_and_extract(
-            url = urls,
-            stripPrefix = strip_prefix,
-            sha256 = sha256,
-            auth = auth,
-        )
+        fail("No supported workaround for extracting Go SDK non-ASCII filenames. More recent versions of Bazel have correct support for unpacking the Go SDK.")
```
fmeum, line 482: Let's make this more actionable by mentioning Bazel 6.4.0 specifically.

go/private/sdk.bzl
fmeum: Ah, yes, I think it's because the call that registers our deps ends up loading the same file and that leads to the cycle. I will take a look. If that doesn't work, we may have to revert back to the hardcoded version check. Sorry for the trouble, WORKSPACE is painful.

mattyclarkson: I've push a commit on top that refactors back the `versions` check. Feel free to remove that commit, revert or whatever else if you can get `bazel_features` to work!

mattyclarkson: Had to add a final statement to logic statement to pick up Windows.

mattyclarkson: There is a chance with a change in logic that a setup that used to fall through to `download_and_extract` now will fail: that setup just might not be in BuildKite.

fmeum: The Bazel 5 pipeline is failing. If we can also make that work, this is definitely good to go. If not, we can still consider dropping support for Bazel 5.

mattyclarkson: Seems like I caught all the edge cases.

fmeum (approved):

PR #3874, opened 2024-02-26
fix sdk download for common execution platforms

scaiper:
**What type of PR is this?**

Bug fix

This PR fixes a bug where host platform is passed in `go_download_sdk_rule` instead of execution platform.

It is very small bug fix, not sure if issue is needed.

**Other notes for review**


scaiper:
Not sure why `//tests/core/cgo:opts_test` fails on 11 Bullseye (OpenJDK 17, gcc 10.2.1).

fmeum (approved):
Well, that's embarrassing. Thanks!

PR #3885, opened 2024-03-11
Cleanup temp dirs in GoToolchainBinaryBuild

### LINKED ISSUE #3884: GoToolchainBinaryBuild leaves temp gopath dir (linux), or uploads to cache (windows)

0.46.0

0.35.0

6.4

Linux & Windows

### What did you do?

Build anything with rules_go.

### What did you see instead?

Each `GoToolchainBinaryBuild` execution leaves 2 entries like `/tmp/tmp.CGmiYEaaeP/`  under linux. Those correspond to throwaway cache created by `mktemp -d` that is never cleaned up:
  https://github.com/bazelbuild/rules_go/blob/d45e218910fbecb00867e3a4dde3fe14d4a1e00a/go/private/rules/binary.bzl#L462

Under Windows the same rule uses `declare_directory` to create space for cache and returns the declared directory from the rule, what causes uploading temporary gocache entries to remote cache and subsequent downloads.
https://github.com/bazelbuild/rules_go/blob/d45e218910fbecb00867e3a4dde3fe14d4a1e00a/go/private/rules/binary.bzl#L440

Cache contains ~500 files / 50MB, which are not really problematic numbers as those action usually do not execute often (maybe other than in CI), but still those temporary files should be cleaned up.

hauserx:
**What type of PR is this?**

Bug fix

Fixes #3884

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -436,11 +436,20 @@ def _go_tool_binary_impl(ctx):
 
     out = ctx.actions.declare_file(name)
     if sdk.goos == "windows":
-        gopath = ctx.actions.declare_directory("gopath")
-        gocache = ctx.actions.declare_directory("gocache")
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Thanks, especially for also handling Windows!

PR #3902, opened 2024-03-26
Update go vet with latest analyzers

sluongng:
Upgrade golang.org/x/tools to v0.18.0.
Note that the latest v0.19.0 has a bug that we should avoid using.

Update the analyzers (passes) in TOOLS_NOGO and the ones automatically
included when `vet = True` is configured.
The vet analyzers are taken from `go tool vet help` documentation.


fmeum:
The nogo failure looks like it's easier to fix than ignore.

sluongng:
@fmeum is it though?

The `fieldalignment` analyzer is quite noisy and is triggered on proto-generated code? 🤔 

Im in favor of disabling it with a TODO.

go/private/rules/nogo.bzl
```diff
@@ -140,11 +140,41 @@ def nogo(name, visibility = None, **kwargs):
...
     if kwargs.get("vet"):
         kwargs["deps"] = kwargs.get("deps", []) + [
-            Label("@org_golang_x_tools//go/analysis/passes/atomic:go_default_library"),
-            Label("@org_golang_x_tools//go/analysis/passes/bools:go_default_library"),
-            Label("@org_golang_x_tools//go/analysis/passes/buildtag:go_default_library"),
-            Label("@org_golang_x_tools//go/analysis/passes/nilfunc:go_default_library"),
-            Label("@org_golang_x_tools//go/analysis/passes/printf:go_default_library"),
+            Label("@org_golang_x_tools//go/analysis/passes/{}".format(p))
+            for p in [
+                "appends",
```
fmeum, line 145:
```suggestion
                # keep sorted
                "appends",
```

go/private/rules/nogo.bzl
```diff
@@ -140,11 +140,41 @@ def nogo(name, visibility = None, **kwargs):
...
+                "stdmethods",
+                "stringintconv",
+                "structtag",
+                "testinggoroutine",
+                "tests",
+                "timeformat",
+                "unmarshal",
+                "unreachable",
+                "unsafeptr",
+                "unusedresult",
```
fmeum, line 176: This list differs from the other one. Is this intentional? Could we share a common Starlark constant?

fmeum:
> @fmeum is it though?
> 
> The `fieldalignment` analyzer is quite noisy and is triggered on proto-generated code? 🤔 
> 
> Im in favor of disabling it with a TODO.

True, I missed that this was the generated code, not the test. In that case I agree that disabling it by default is the right move.

go/private/rules/nogo.bzl
sluongng: Correct.

TOOLS_NOGO document said:

> # TOOLS_NOGO is a list of all analysis passes in
> # golang.org/x/tools/go/analysis/passes.

so I pretty much got it from `bazel query` on that package. It contains a bit more analyzers than the ones used in `go vet` (which from the help document I noted in the PR summary).

fmeum: Okay, could you add a comment to this list explaining the difference?

fmeum (approved):

sluongng:
Hmm, these dependencies upgrades turn out to be a huge PITA.

Especially, for the `popular_repo` test, I think we should consider dropping `x/tools` testing entirely. The repo is where `gopls` is being actively developed and the recent code quality is not that high. v0.19.0 contains a nil pointer bug and a bunch of build/test targets in that repo are not gazelle/rules_go compatible.

I also had to patch Gazelle temporarily to avoid analyzer duplication. It should be a temp workaround until (a) we remove that in Gazelle, then (b) release a new Gazelle version, and then (c) upgrade the Gazelle version in rules_go.

sluongng:
closing this, gona try for a bigger change

PR #3911, opened 2024-04-03
Mark `go_sdk` extension as `reproducible`

fmeum:
Setting the new `reproducible` parameter of [`module_ctx.extension_metadata`](https://bazel.build/rules/lib/builtins/module_ctx#extension_metadata) to `True` results in the `go_sdk` extension being omitted from the lockfile. This is especially useful since it depends on OS/arch and thus makes lockfile maintenance more difficult if present.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,7 +7,7 @@ module(
 
 # The custom repo_name is used to prevent our bazel_features polyfill for WORKSPACE builds from
 # conflicting with the real bazel_features repo.
-bazel_dep(name = "bazel_features", version = "1.1.1", repo_name = "io_bazel_rules_go_bazel_features")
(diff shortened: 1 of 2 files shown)


malt3:
I was trying to understand this change. It uses this feature, if enabled: https://github.com/bazel-contrib/bazel_features/blob/888ffa12f47b604d679e4f2d66e28c89abac991a/features.bzl#L47-L50

See also https://github.com/bazelbuild/bazel/commit/c796aba6ee36970956ea32b46a2f121bb4d1818a and https://github.com/bazelbuild/bazel/commit/e730201e6bf8d6c1c80433b5b42305c3167a8660

fmeum:
@malt3 I added a PR description, hope that helps.

tyler-french (approved):

PR #3916, opened 2024-04-08
GoLink action. Use -extar <ar path> if cc_toolchain provides ar path

OparinE:
**What type of PR is this?**

 Bug fix

This PR solves  GoLink actions issue, when we can't create static library if go toolchain depends on portable cc toolchain.

Prerequisites:
- was found on MacOS, but behavior is general
- go toolchain uses cc **portable** toolchain

GoLink action uses `link` tool under the hood. `link` tool calls `ar` tool for archiving jobs and `ar` tool is called by name.
GoLink action envs contains PATH variable containing path to cc_toolchain.
If `ar` tool is found in PATH absolute path, everything is fine. In case of relative path, `link` raises an error: "cannot execute tool by relative path".

This PR forces `link` tool to explicitly use `ar` tool from cc_toolchain if cc_toolchain provides it.

Issue is not created

**Other notes for review**

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -89,6 +89,10 @@ def emit_link(
     builder_args = go.builder_args(go, "link")
     tool_args = go.tool_args(go)
 
+    # use ar tool from cc toolchain if cc toolchain provides it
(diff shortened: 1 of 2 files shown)


OparinE:
@fmeum , CI is passed

PR #3918, opened 2024-04-11
Don't ignore the return value of runfiles.merge_all()

EdSchouten:
**What type of PR is this?**

Bug fix

I'm running into an issue where data dependencies of C/C++ libraries don't end up being placed in the input roots of Go based tests that depend on these libraries via cgo. It turns out that this is caused by emit_archive() not properly merging runfiles from all GoArchive dependencies.

I tried to make a small reproduction case of this under tests/core/cgo, but for the life of me I can't seem to reproduce this. Maybe this can only be observed when cross compiling or using custom C++ toolchains?

Couldn't find any.

**Other notes for review**


fmeum (approved):
Nice catch!

PR #3919, opened 2024-04-12
Add support for `--incompatible_enable_proto_toolchain_resolution`

### LINKED ISSUE #3895: Support --incompatible_enable_proto_toolchain_resolution
Current

7.0.0 or later

### What did you do?

Try to build a `go_library` target with the flag enabled and a different toolchain registered for `@rules_proto//proto:toolchain_type`

When `--incompatible_enable_proto_toolchain_resolution` is enabled, rules should consult the relevant toolchain_type to locate protoc. 
Example PR from rules_python:
https://github.com/bazelbuild/rules_python/pull/1577

That means I ought to be able to add a Go example to https://github.com/alexeagle/toolchains_protoc/tree/main/examples
which is a project that ensures we never have to build the `cc_binary` target for protoc from sources.

### What did you see instead?
https://github.com/bazelbuild/rules_go/blob/cac338e5c869116cbe86758001dced886a38b39c/proto/BUILD.bazel#L129 depends on `@com_google_protobuf//:protoc` directly.

In this case I registered a pre-built protoc toolchain (using https://github.com/alexeagle/toolchains_protoc) but get an error:

```
$ bazel query --output=build //cli/core/pkg/aspect/config:config
# /home/alex/Projects/silo/cli/core/pkg/aspect/config/BUILD.bazel:3:11
go_library(
  name = "config",
  generator_name = "config",
  generator_function = "go_library_macro",
  generator_location = "cli/core/pkg/aspect/config/BUILD.bazel:3:11",
  srcs = ["//cli/core/pkg/aspect/config:config.go"],
  deps = ["//cli/core/pkg/bazel:bazel", "//cli/core/pkg/ioutils:ioutils", "@com_github_spf13_cobra//:cobra"],
  importpath = "github.com/aspect-build/silo/cli/core/pkg/aspect/config",
)
# Rule config instantiated at (most recent call last):
#   /home/alex/Projects/silo/cli/core/pkg/aspect/config/BUILD.bazel:3:11                                                                in <toplevel>
#   /home/alex/.cache/bazel/_bazel_alex/cf34354d28c19c286eb12fd35d42d565/external/io_bazel_rules_go/go/private/rules/wrappers.bzl:45:15 in go_library_macro
# Rule go_library defined at (most recent call last):
#   /home/alex/.cache/bazel/_bazel_alex/cf34354d28c19c286eb12fd35d42d565/external/io_bazel_rules_go/go/private/rules/library.bzl:61:18 in <toplevel>

alex@a:~/Projects/silo$ code $(bazel info output_base)/external
alex@a:~/Projects/silo$ bazel build //cli/core/pkg/aspect/config:config
ERROR: /home/alex/.cache/bazel/_bazel_alex/cf34354d28c19c286eb12fd35d42d565/external/com_google_protobuf/BUILD: no such target '@@com_google_protobuf//:protoc': target 'protoc' not declared in package '' defined by /home/alex/.cache/bazel/_bazel_alex/cf34354d28c19c286eb12fd35d42d565/external/com_google_protobuf/BUILD (Tip: use `query "@com_google_protobuf//:*"` to see all the targets in that package)
ERROR: /home/alex/.cache/bazel/_bazel_alex/cf34354d28c19c286eb12fd35d42d565/external/io_bazel_rules_go/proto/BUILD.bazel:127:20: no such target '@@com_google_protobuf//:protoc': target 'protoc' not declared in package '' defined by /home/alex/.cache/bazel/_bazel_alex/cf34354d28c19c286eb12fd35d42d565/external/com_google_protobuf/BUILD (Tip: use `query "@com_google_protobuf//:*"` to see all the targets in that package) and referenced by '@@io_bazel_rules_go//proto:protoc'
```

fmeum:
**What type of PR is this?**

Feature

With `--incompatible_enable_proto_toolchain_resolution`, users of `protoc` are expected to obtain it through `@rules_proto//proto:toolchain_type` instead of the Bazel command line flag `--proto_compiler`. This change adds support for this in a backwards compatible way.

Note that the above is actually a lie: The proper way to use `protoc` would be to go through a Go-specific `proto_lang_toolchain` and use the methods on `proto_common` to interact with `protoc`. Since rules_go has a very bespoke setup with customizable compilers and the need to apply reset transitions in case `protoc` *is* built from source, this would require major changes. We should revisit this after Proto toolchainization has been enabled by default.

Fixes #3895

**Other notes for review**

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3,9 +3,60 @@ workspace(name = "io_bazel_rules_go")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
 
+# Required by toolchains_protoc.
+http_archive(
+    name = "platforms",
+    sha256 = "5eda539c841265031c2f82d8ae7a3a6490bd62176e0c038fc469eabf91f6149b",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
+        "https://github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
+    ],
+)
+
+# The non-polyfill version of this is needed by rules_proto below.
+http_archive(
+    name = "bazel_features",
+    sha256 = "d7787da289a7fb497352211ad200ec9f698822a9e0757a4976fd9f713ff372b3",
+    strip_prefix = "bazel_features-1.9.1",
+    url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.9.1/bazel_features-v1.9.1.tar.gz",
+)
+
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
+bazel_features_deps()
+
 go_rules_dependencies()
 
 go_register_toolchains(version = "1.21.8")
+
+http_archive(
+    name = "rules_proto",
+    sha256 = "71fdbed00a0709521ad212058c60d13997b922a5d01dbfd997f0d57d689e7b67",
+    strip_prefix = "rules_proto-6.0.0-rc2",
+    url = "https://github.com/bazelbuild/rules_proto/releases/download/6.0.0-rc2/rules_proto-6.0.0-rc2.tar.gz",
+)
+
+load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
+
+rules_proto_dependencies()
+
+load("@rules_proto//proto:toolchains.bzl", "rules_proto_toolchains")
+
+rules_proto_toolchains()
+
+http_archive(
+    name = "toolchains_protoc",
+    sha256 = "b312e6de6485e01f753cb87fa09b4193f1762593141790dd0a90abf5e520b1d7",
+    strip_prefix = "toolchains_protoc-0.2.1",
+    url = "https://github.com/alexeagle/toolchains_protoc/releases/download/v0.2.1/toolchains_protoc-v0.2.1.tar.gz",
+)
+
+load("@toolchains_protoc//protoc:toolchain.bzl", "protoc_toolchains")
+
+protoc_toolchains(
+    name = "protoc_toolchains",
+    version = "v25.3",
+)
 
 http_archive(
     name = "com_google_protobuf",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -6,10 +6,10 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 # Required by toolchains_protoc.
 http_archive(
     name = "platforms",
-    sha256 = "5eda539c841265031c2f82d8ae7a3a6490bd62176e0c038fc469eabf91f6149b",
+    sha256 = "218efe8ee736d26a3572663b374a253c012b716d8af0c07e842e82f238a0a7ee",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
-        "https://github.com/bazelbuild/platforms/releases/download/0.0.9/platforms-0.0.9.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/platforms/releases/download/0.0.10/platforms-0.0.10.tar.gz",
+        "https://github.com/bazelbuild/platforms/releases/download/0.0.10/platforms-0.0.10.tar.gz",
     ],
 )
 
@@ -31,9 +31,9 @@ go_register_toolchains(version = "1.21.8")
 
 http_archive(
     name = "rules_proto",
-    sha256 = "71fdbed00a0709521ad212058c60d13997b922a5d01dbfd997f0d57d689e7b67",
-    strip_prefix = "rules_proto-6.0.0-rc2",
-    url = "https://github.com/bazelbuild/rules_proto/releases/download/6.0.0-rc2/rules_proto-6.0.0-rc2.tar.gz",
+    sha256 = "303e86e722a520f6f326a50b41cfc16b98fe6d1955ce46642a5b7a67c11c0f5d",
+    strip_prefix = "rules_proto-6.0.0",
+    url = "https://github.com/bazelbuild/rules_proto/releases/download/6.0.0/rules_proto-6.0.0.tar.gz",
 )
 
 load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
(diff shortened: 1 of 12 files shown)


fmeum:
@alexeagle None of us have worked with proto toolchains so far, so it would be great if you could also take a look.

fmeum:
FYI @sluongng You might find this interesting :-)

proto/compiler.bzl
```diff
@@ -31,6 +35,15 @@ load(
...
 )
 
+# This is actually a misuse of Proto toolchains: The proper way to use `protoc` would be to go
+# through a Go-specific `proto_lang_toolchain` and use the methods on `proto_common` to interact
+# with `protoc`. Since rules_go has a very bespoke setup with customizable compilers and the need
+# to apply reset transitions in case `protoc` *is* built from source, this would require major
+# changes.
+# TODO: Revisit this after --incompatible_enable_proto_toolchain_resolution has been enabled by
+#  default.
+_PROTO_TOOLCHAIN_TYPE = "@rules_proto//proto:toolchain_type"
```
sluongng, line 45: Neither https://bazel.build/rules/lib/toplevel/proto_common
or https://github.com/search?q=repo%3Abazelbuild%2Frules_proto%20%2Fproto_common%2F&type=code provide me with any inspiration.

Took me a while to realize that `proto_common` here is not a starlark struct like `cc_common` or `java_common`, but is a file located in `@rules_proto//proto:proto_common.bzl`. This inconsistency is quite rough... I wonder if we could do `s/proto_toolchains/proto_common/` here (or in rules_proto).

WORKSPACE
```diff
@@ -3,10 +3,61 @@ workspace(name = "io_bazel_rules_go")
...
+    strip_prefix = "toolchains_protoc-0.2.1",
+    url = "https://github.com/alexeagle/toolchains_protoc/releases/download/v0.2.1/toolchains_protoc-v0.2.1.tar.gz",
+)
+
+load("@toolchains_protoc//protoc:toolchain.bzl", "protoc_toolchains")
+
+protoc_toolchains(
+    name = "protoc_toolchains",
+    version = "v25.3",
+)
```
sluongng, line 59: Im slightly confused by this.

Are we using `toolchains_protoc` in rules_go now? Or this is only needed for tests?

What if the user want to use the non-prebuilt toolchain?

tests/bcr/MODULE.bazel
```diff
@@ -21,6 +21,10 @@ local_path_override(
 bazel_dep(name = "gazelle", version = "0.33.0")
 bazel_dep(name = "protobuf", version = "3.19.6")
 
+# Required with --incompatible_enable_proto_toolchain_resolution.
+# Avoids building protoc from source, which speeds up CI runs.
+bazel_dep(name = "toolchains_protoc", version = "0.2.1")
```
sluongng, line 26: Should we add 2 separate tests: 1 with the pre-built toolchain and 1 with the normal toolchain?

proto/compiler.bzl
```diff
@@ -193,7 +211,7 @@ def _go_proto_compiler_impl(ctx):
 
 _go_proto_compiler = rule(
     implementation = _go_proto_compiler_impl,
-    attrs = {
+    attrs = dict({
```
sluongng, line 214: I wonder if we could use the `{} | {}` to merge the 2 dicts, instead of `dict({}, **{})`?

WORKSPACE
```diff
@@ -3,10 +3,61 @@ workspace(name = "io_bazel_rules_go")
...
 go_rules_dependencies()
 
 go_register_toolchains(version = "1.21.8")
 
+http_archive(
+    name = "rules_proto",
+    sha256 = "71fdbed00a0709521ad212058c60d13997b922a5d01dbfd997f0d57d689e7b67",
+    strip_prefix = "rules_proto-6.0.0-rc2",
+    url = "https://github.com/bazelbuild/rules_proto/releases/download/6.0.0-rc2/rules_proto-6.0.0-rc2.tar.gz",
+)
```
sluongng, line 37: Do we need to update https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl for these new repos?

sluongng (commented):
A few questions, mostly because I am not that familiar with the new flag implication.

proto/compiler.bzl
fmeum: `proto_toolchains` is a helper struct that just happens to be colocated with `proto_common`. `proto_common` is the equivalent of `cc_common`, with the difference being that it is already loaded from its canonical post-Starlarkification location.

I know there is a lot going on here, so please let me know if I should add any comments.

proto/compiler.bzl
fmeum: It's not available on Bazel 5.4.0, which is our current minimum.

tests/bcr/MODULE.bazel
fmeum: I did, see https://github.com/bazelbuild/rules_go/pull/3919/files#diff-b03048100ab56455b5b377a5c20ffd3d17f80c719befca45253af622eb9395efR76.

WORKSPACE
fmeum: This is needed when running with `--incompatible_enable_proto_toolchain_resolution` in our "incompatible flags" CI run. With that flag, users are free to supply any proto toolchain of their choice and could also produce one built from source.

WORKSPACE
fmeum: That macro contains an explicit comment about not wanting rules_proto in there. The relationship with the repo doesn't change with this PR: If you want `go_proto_library` to work, you need to supply it yourself. However, users might need to update to a newer rules_proto version, so we should wait until the release of rules_proto 6.0.0.

WORKSPACE
alexeagle: Note, I'd like to upstream that toolchains_protoc into rules_proto after the 6.0.0 release.

fmeum:
I'll take another stab at fixing the integration tests next week. Everything else is stable for review.

proto/compiler.bzl
sluongng: Oh haha, I forgot about Bazel LTS still has 5.4 as supported 😓 

[One ~~day~~year more](https://www.youtube.com/watch?v=Sv-BxH3SVS8)

fmeum:
I updated rules_proto to 6.0.0 and fixed the remaining tests, PTAL @linzhp @tyler-french

linzhp (approved):

sluongng:
Hmm i cannot resolve the conversations. Perhaps @fmeum can. 

The PR LGTM. I hope the legacy toolchain still provide an alternative option to use protoc from source.

Cheers!

avdv:
Hi,

when trying to upgrade to 0.48.0 / 0.48.1 I see this error:
```

fmeum:
Yes. Since proto dependencies can't be handled for the users without surprising version conflicts, we let you supply them yourself when you are using protos (in this case indirectly through buildtools). You could also consider switching to `MODULE.bazel`, which automates this once again with proper version resolution.

PR #3920, opened 2024-04-18
Setting test.timeout according to TEST_TIMEOUT

### LINKED ISSUE #3875: Test binaries swallow SIGTERM; printed stack traces misleading

v0.46.0

v0.35.0

6.2.1

Yes

macOS/arm64

None

### What did you do?

Run a test with `--test_timeout=5`

The test will time out after 5 seconds

### What did you see instead?

The test times out after 20 seconds. I tried a few different values and saw that the test will time out 15 seconds after I request it to, for example, if I set the timeout to 10 seconds, it will time out in fact after 25 seconds, etc. It seems like whatever the timeout "should" be, `rules_go` is adding 15 seconds to it. This behavior wasn't there in `rules_go` v0.43.

linzhp:
**What type of PR is this?**
Bug fix

This is another attempt to make test timeout experience close to "go test". It meets the following requirements:
1. gorountine stack at the time of test termination is printed out
2. "bazel test" doesn't add 15s to the configured timeout (#3875)
3. test.xml reports the test cases executed at the time of timeout if `test.v` is passed
4. Bazel reports the timed out test as TIMEOUT as opposed to FAIL

This implements the approach described in https://github.com/bazelbuild/rules_go/issues/3875#issuecomment-1974892818

Fixes #3875

**Other notes for review**
cc @rickystewart


go/tools/builders/generate_test_main.go
```diff
@@ -235,7 +237,16 @@ func main() {
...
 	{{end}}
-	bzltestutil.RegisterTimeoutHandler()
+
+	testTimeout := os.Getenv("TEST_TIMEOUT")
+	if testTimeout != "" {
+		flag.Lookup("test.timeout").Value.Set(testTimeout+"s")
+		// If Bazel sends a SIGTERM because the test timed out, it sends it to all child processes. Because
+		// we already set -test.timeout according to the TEST_TIMEOUT, we need to ignore the signal so the
+		// test has time to clean up. It will be killed by Bazel after the grace period instead.
+		signal.Ignore(syscall.SIGTERM)
```
fmeum, line 247: Could you also update this comment?
https://github.com/bazelbuild/rules_go/blob/e6cb73792b77e3b595ba456e92211ae7e4a82346/go/tools/bzltestutil/wrap.go#L128-L132

fmeum (approved):
That's a clean solution to our problems, thanks!

sluongng:
This change broke one of our tests which runs `sh -c kill -TERM $$` to test the termination of the shell.

We have to register our own `signal.Notify` as a workaround for that test.
https://github.com/buildbuddy-io/buildbuddy/pull/6436/

I think this was the original intention of `go/tools/bzltestutil/timeout.go`?

go/tools/builders/generate_test_main.go
```diff
@@ -235,7 +237,19 @@ func main() {
...
+	testTimeout := os.Getenv("TEST_TIMEOUT")
+	if testTimeout != "" {
+		flag.Lookup("test.timeout").Value.Set(testTimeout+"s")
+		// If Bazel sends a SIGTERM because the test timed out, it sends it to all child processes. Because
+		// we set -test.timeout according to the TEST_TIMEOUT, we need to ignore the signal so the test has
+		// time to properly produce the output (e.g. stack trace). It will be killed by Bazel after the grace
+		// period (15s) expires.
+		// If TEST_TIMEOUT is not set (e.g., when the test binary is run by Delve for debugging), we don't
+		// ignore SIGTERM so it can be properly terminated.
+		signal.Ignore(syscall.SIGTERM)
```
linzhp, line 250: @sluongng @fmeum Maybe we should do this instead here?

```
	c := make(chan os.Signal, 1)
	signal.Notify(c, syscall.SIGTERM)
	go func() {
		<-c
	}()
```

fmeum: With the intent of this being resettable by `signal.Reset`? I like that.

PR #3923, opened 2024-04-19
prepare rules_go v 0.47

tyler-french:
This PR prepares the `rules_go` release `v0.47.0`. A few tests are no longer applicable in the `x/tools` repo.


go/private/repositories.bzl
```diff
@@ -244,13 +244,13 @@ def go_rules_dependencies(force = False):
     wrapper(
         http_archive,
         name = "org_golang_google_genproto",
-        # main, as of 2023-12-15
+        # main, as of 2024-04-19
         urls = [
-            "https://mirror.bazel.build/github.com/googleapis/go-genproto/archive/995d672761c0c5b9ac6127b488b48825f9a2e5fb.zip",
-            "https://github.com/googleapis/go-genproto/archive/995d672761c0c5b9ac6127b488b48825f9a2e5fb.zip",
+            "https://mirror.bazel.build/github.com/googleapis/go-genproto/archive/8c6c420018be7d99c6336d84554d856523d514bf.zip",
```
tyler-french, line 249: @linzhp should we be worried about any of these upgrades causing downstream issues?

tests/integration/popular_repos/BUILD.bazel
```diff
@@ -173,7 +172,6 @@ test_suite(
         "@org_golang_x_tools//go/callgraph:callgraph_test",
         "@org_golang_x_tools//go/callgraph/static:static_test",
         "@org_golang_x_tools//go/callgraph/vta/internal/trie:trie_test",
-        "@org_golang_x_tools//go/cfg:cfg_test",
```
tyler-french, line 176: @fmeum This test was the only one that was removed because it was failing (the others were removed).

This one failed with:
```
==================== Test output for @org_golang_x_tools//go/cfg:cfg_test:
--- FAIL: TestSmoke (0.01s)
    cfg_test.go:180: go tool not available: 'go env GOROOT' does not match runtime.GOROOT:
                go env: /home/user/.cache/bazel/_bazel_tfrench/eb80ee7746c8e42f65d76b9b55d4ad41/external/go_sdk
                GOROOT: GOROOT
FAIL
Target @org_golang_x_tools//go/cfg:cfg_test up-to-date:
```
Is this expected behavior

go/private/repositories.bzl
linzhp: It should be fine as long as `go_rules_dependencies()` is called in WORKSPACE after loading external repos needed by the repo.

However, we should not upgrade any deps in bzlmod unless necessary, because that will force all downstream consumers to upgrade.

With that in mind, @fmeum, do you think we should upgrade WORKSPACE deps as part of every release?

fmeum: On average I would say it's slightly better to keep updating WORKSPACE deps - it makes it less likely that rules_go pulls in an outdated version of something that overrides some other ruleset's deps. But this can still happen when unstable deps contain breaking changes, so it's a mixed bag and we could very well choose to stay out of that.

tests/integration/popular_repos/BUILD.bazel
fmeum: Yes: We build the binaries with a fake GOROOT as they could otherwise contain absolute paths. If we could pass in the relative path to the SDK instead, that may be better

fmeum (approved):

go/private/repositories.bzl
tyler-french: Sounds good. For now let's upgrade them as normal

PR #3927, opened 2024-04-24
Automatically run `bazel mod tidy` in `@rules_go//go`

### LINKED ISSUE #3900: `@rules_go//go get` should trigger `bazel mod tidy`
Bazel 7.1.0 adds a new `bazel mod tidy` that collects and applies `buildozer` fixups and updates the lock file. As [suggested](https://docs.google.com/document/d/13LbK_1WhA4la0eH7yISjnMvXs2cKFXD-adKPu0i0RK0/edit?disco=AAAA2ThnB8U) in the proposal doc, it will be a nice quality of like improvement if `rules_go` trigger `bazel mod tidy` on dependency updates:
> rulesets can provide an improved experience by wrapping the language-native mechanism for adding a dependency and automatically running bazel mod tidy. In the case of rules_go, this simplifies the workflow further to add the dependency via bazel run @rules_go//go get example.com, a wrapper around go get, which automatically runs bazel mod tidy.

fmeum:
**What type of PR is this?**

Feature

`bazel mod tidy` is run if supported and if relevant files change due to the command.

Fixes #3900
Fixes https://github.com/bazelbuild/bazel-gazelle/issues/1749

**Other notes for review**

### DIFF
--- a/go/tools/go_bin_runner/main.go
+++ b/go/tools/go_bin_runner/main.go
@@ -1,21 +1,29 @@
 package main
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
+	"fmt"
+	"io"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
+	"strings"
 
 	"github.com/bazelbuild/rules_go/go/runfiles"
 )
 
 var GoBinRlocationPath = "not set"
 var ConfigRlocationPath = "not set"
+var HasBazelModTidy = "not set"
 
 // Produced by gazelle's go_deps extension.
 type Config struct {
-	GoEnv map[string]string `json:"go_env"`
+	GoEnv     map[string]string `json:"go_env"`
+	DepsFiles []string          `json:"dep_files"`
 }
 
 func main() {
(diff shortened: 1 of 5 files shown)


stefanpenner:
Just checking, will this also work for users of go.work workspaces?

fmeum:
> Just checking, will this also work for users of go.work workspaces?

This requires adding the relevant files to https://github.com/bazelbuild/bazel-gazelle/blob/7c84c339075a992ea1a6f720899d91c0e047efe2/internal/bzlmod/go_deps.bzl#L657. Maybe that already happens with the fake `go.mod` files we create for a workspace?

go/tools/go_bin_runner/main.go
```diff
@@ -34,8 +42,35 @@ func main() {
...
+	hashesAfter, err := hashWorkspaceRelativeFiles(cfg.DepsFiles)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	diff := diffMaps(hashesBefore, hashesAfter)
+	if len(diff) > 0 {
+		if HasBazelModTidy == "True" {
+			_, _ = fmt.Fprintf(os.Stderr, "\nrules_go: Running 'bazel mod tidy' since %s changed...\n", strings.Join(diff, ", "))
+			err = runProcess([]string{"bazel", "mod", "tidy"}, os.Environ(), cwd)
```
sluongng, line 66: FYI: not everybody use the `bazel` binary from $PATH.

Some folks prefer using `bazelisk`, `b`, `bb`, `blaze`.

Could we make this configurable somehow? Or better, could we expose a starlark API for this?

linzhp (approved):

go/tools/go_bin_runner/main.go
fmeum: Good point, I'm now honoring the `BAZEL` environment variable.

PR #3929, opened 2024-04-29
bzltestutil: restore timeout signal handler

sluongng:
In https://github.com/bazelbuild/rules_go/pull/3920, a new mechanism to
handle test timeout was introduced. However this broke existing tests
that use SIGTERM inside.

Restore the original behavior.


sluongng:
@linzhp FYI

go/tools/builders/generate_test_main.go
```diff
@@ -247,7 +247,14 @@ func main() {
 		// period (15s) expires.
 		// If TEST_TIMEOUT is not set (e.g., when the test binary is run by Delve for debugging), we don't
 		// ignore SIGTERM so it can be properly terminated.
-		signal.Ignore(syscall.SIGTERM)
+		// We do not panic (like native go test does) because users may legitimately want to use SIGTERM
+		// in tests.
+		// See https://github.com/golang/go/blob/e816eb50140841c524fd07ecb4eaa078954eb47c/src/testing/testing.go#L2351
+		c := make(chan os.Signal, 1)
```
fmeum, line 253: Could you also explain the difference compared to `signal.Ignore` in the comment?

sluongng: added to PR summary.

fmeum: I think this makes sense in a comment since the first thing I think when seeing this code is "why not `signal.Ignore`?". It looks like a "reimplementation", but so far we don't even know why it makes a difference. But maybe @linzhp does?

sluongng: I would attribute it to unfamiliarity with how `signal` package works underneath. It also took me a bit to read through the implementation and re-parsing this single sentence `Ignore undoes the effect of any prior calls to Notify for the provided signals.`.

However, readers who are familiar with `signal` package should be able to tell from a glance. And because it's a bit distracting to improve `signal` package usability from rules_go code 😅, I opted to keep that comment out.

fmeum: https://github.com/buildbuddy-io/buildbuddy/pull/6436#discussion_r1583764882 has a good explanation for the troubles this courses.

sluongng: Added a line explaining `signal.Ignore` with reference back to this PR to help future readers 😅

go/tools/builders/generate_test_main.go
```diff
@@ -245,9 +245,23 @@ func main() {
...
+		//
 		// If TEST_TIMEOUT is not set (e.g., when the test binary is run by Delve for debugging), we don't
-		// ignore SIGTERM so it can be properly terminated.
-		signal.Ignore(syscall.SIGTERM)
+		// ignore SIGTERM so it can be properly terminated. (1)
+		// We do not panic (like native go test does) because users may legitimately want to use SIGTERM
+		// in tests.
+		//
+		// signal.Notify is used to ensure that there is a no-op signal handler registered.
+		// Avoid using signal.Ignore here: despite the name, it's only used to unregistered handler(s) that
```
fmeum, line 255:
```suggestion
		// Avoid using signal.Ignore here: despite the name, it's only used to unregister handlers that
```

fmeum (approved):

PR #3939, opened 2024-05-08
generate_test_main: Move timeout handling back to bzltestutil

### LINKED ISSUE #3938: goleak broken by recent timeout changes

0.47.1

latest

bazel 6

yes

Linux amd64

not relevant

### What did you do?

Ran tests using goleak

successful tests

### What did you see instead?

tests failing with a lingering goroutine

This is because in the past we would register the go channel handler in a deterministic location (ie, `bzltestutil`), but now since its injected into `main()`, the name depends on the name of the target, which is not deterministic to ignore.

We should bring back the `RegisterTimeoutHandler` so we have a deterministic location to ignore again.

DolceTriade:
**What type of PR is this?**

Bug fix

This gives us a determinstic location to ignore for goleak, which we previously had, but was removed as part of #3920.

Fixes #3938

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -98,13 +98,11 @@ import (
 	"log"
 	"os"
 	"os/exec"
-	"os/signal"
 {{if .TestMain}}
 	"reflect"
 {{end}}
 	"strconv"
 	"strings"
-	"syscall"
 	"testing"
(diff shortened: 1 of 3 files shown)


fmeum (approved):
Cc @linzhp @sluongng

linzhp (approved):
LGTM

malt3 (approved):
LGTM

PR #3956, opened 2024-06-07
Allow unused parameters when building the stdlib

cramertj:
**What type of PR is this?**

Bug fix

The stdlib build does not support building with -Wextra or -Wunused-parameter since it has unused parameters. This patch removes these flags from the stdlib build.

### DIFF
--- a/go/platform/apple.bzl
+++ b/go/platform/apple.bzl
@@ -44,5 +44,6 @@ def apple_ensure_options(ctx, env, _tags, compiler_option_lists, linker_option_l
     min_version = _apple_version_min(ctx, platform, platform_type)
     for compiler_options in compiler_option_lists:
         compiler_options.append(min_version)
+        compiler_options.append('-Wno-unused-parameter')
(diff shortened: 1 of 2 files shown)


cramertj:
CQ failure looks like maybe an infra issue?

go/platform/apple.bzl
```diff
@@ -44,5 +44,6 @@ def apple_ensure_options(ctx, env, _tags, compiler_option_lists, linker_option_l
     min_version = _apple_version_min(ctx, platform, platform_type)
     for compiler_options in compiler_option_lists:
         compiler_options.append(min_version)
+        compiler_options.append('-Wno-unused-parameter')
```
fmeum, line 47:
```suggestion
        compiler_options.append("-Wno-unused-parameter")
```
should fix the remaining CI failure.

Is it clear that this only ends up in the stdlib build and not in other places were it may not be the desired behavior? Could you add a comment?

PR #3965, opened 2024-06-19
make `go_reset_target` public

tyler-french:
`go_reset_target` is helpful when certain plugins or tool attributes to rules can't build under certain configurations. For example, many thrift plugins fail to build when a dependency of a `go_binary` with `c-shared` instrumentation.

To alleviate this issue, you can pass the tool to `go_reset_target`.

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -318,21 +318,26 @@ go_reset_target = rule(
         "dep": attr.label(
             mandatory = True,
             cfg = go_tool_transition,
+            doc = """The target to forward providers from and apply go_tool_transition to.""",
         ),
         "_allowlist_function_transition": attr.label(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
     },
-    doc = """Forwards providers from a target and applies go_tool_transition.
+    doc = """Forwards providers from a target and default Go binary settings.
 
-go_reset_target depends on a single target, built using go_tool_transition. It
+go_reset_target depends on a single target and builds it to be a Go tool binary. It
 forwards Go providers and DefaultInfo.
 
-This is used to work around a problem with building tools: Go tools should be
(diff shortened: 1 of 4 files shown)


go/private/rules/transition.bzl
```diff
@@ -319,8 +320,8 @@ forwards Go providers and DefaultInfo.
 
 This is used to work around a problem with building tools: Go tools should be
 built with 'cfg = "exec"' so they work on the execution platform, but we also
-need to apply go_tool_transition so that e.g. a tool isn't built as a shared
-library with race instrumentation. This acts as an intermediate rule that allows
+need to apply go_tool_transition so that, for example, a tool isn't built as a shared
+library with race instrumentation. This acts as an intermediate rule that allows users
 to apply both both transitions.
```
fmeum, line 325:
```suggestion
to apply both transitions.
```

go/private/rules/transition.bzl
```diff
@@ -319,8 +320,8 @@ forwards Go providers and DefaultInfo.
 
 This is used to work around a problem with building tools: Go tools should be
 built with 'cfg = "exec"' so they work on the execution platform, but we also
-need to apply go_tool_transition so that e.g. a tool isn't built as a shared
-library with race instrumentation. This acts as an intermediate rule that allows
+need to apply go_tool_transition so that, for example, a tool isn't built as a shared
```
fmeum, line 323: This mentions a symbol that itself isn't public.

fmeum (approved):
I generally support this, we should thus make sure that the docs comment makes sense for external users.

go/private/rules/transition.bzl
tyler-french: Good point, tried to update it to be a bit more clear.

fmeum (approved):

linzhp (approved):

PR #3966, opened 2024-06-20
Upgrade Stardoc to 0.6.2 .

### LINKED ISSUE #3947: rules_go + protobufs + experimental_sibling_repository_layout fails to build

0.48. The issue also reproduces at commit 354a98f4acf2333b7603ede50dd5fbc20ae315b1.

N/A

7.1.2

Yes

Debian-based Linux variant, x86_64 architecture.

N/A

### What did you do?

See repro at https://github.com/ted-xie/rules_go_proto_bug. TL;DR building a go_proto_library rule that's part of an external dependency with `--experimental_sibling_repository_layout` fails. Failure message:

```
Could not find file in descriptor database: ../maybe_rules_go_proto_bug~/proto/foo.proto: No such file or directory
```

I expected the build to succeed.

### What did you see instead?

The build fails due to the go proto compiler not being able to find the proto definition.

/cc @ahumesky

lberki:
This is in turn to make rules_go work with
`--experimental_sibling_repository_layout` or at least makes `bazel build --experimental_sibling_repository_layout //docs:docs_go_extras_extras-docgen` pass.

It required some re-shuffling of the stanzas in the WORKSPACE file for rules_proto, Stardoc and rules_go itself. I did not investigate deeper given that the plan of record is to migrate to bzlmod.
**What type of PR is this?**

> Bug fix

Makes `rules_go` work with `--experimental_sibling_repository_layout`. Or at least fixes *one* problem.

Fixes #3947 

**Other notes for review**

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,7 +1,54 @@
 workspace(name = "io_bazel_rules_go")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
+
+http_archive(
+    name = "rules_proto",
+    sha256 = "303e86e722a520f6f326a50b41cfc16b98fe6d1955ce46642a5b7a67c11c0f5d",
+    strip_prefix = "rules_proto-6.0.0",
+    url = "https://github.com/bazelbuild/rules_proto/releases/download/6.0.0/rules_proto-6.0.0.tar.gz",
+)
+
+load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies")
+
+rules_proto_dependencies()
+
+load("@rules_proto//proto:toolchains.bzl", "rules_proto_toolchains")
+
+rules_proto_toolchains()
+
+# For API doc generation
+# This is a dev dependency, users should not need to install it
+# so we declare it in the WORKSPACE
+http_archive(
+    name = "io_bazel_stardoc",
+    sha256 = "62bd2e60216b7a6fec3ac79341aa201e0956477e7c8f6ccc286f279ad1d96432",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/stardoc/releases/download/0.6.2/stardoc-0.6.2.tar.gz",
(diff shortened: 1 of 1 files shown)


lberki:
Argh. The CI failure seems to be reproducible by

`bazel build --nobuild --incompatible_enable_proto_toolchain_resolution @@io_bazel_stardoc//stardoc/proto:stardoc_output_java_proto`

My guess: the upgrade to Stardoc also requires an upgrade of `rules_java` on their side so that it works with `--incompatible_enable_proto_toolchain_resolution`.

fmeum:
Could you also add the flag here:
https://github.com/bazelbuild/rules_go/blob/d3ecd76f872a96745ca2b68c1598628458dcc54c/.bazelrc#L40
and here:
https://github.com/bazelbuild/rules_go/blob/d3ecd76f872a96745ca2b68c1598628458dcc54c/.bazelrc#L36

Without this, we won't have good CI coverage as most tests are integration tests.

The error about invalid escapes is related to rules_go using custom templates. I could look into getting rid of them if that simplifies the update.

lberki:
> Could you also add the flag here:
It's there if you scroll right, on both of these lines :)

I haven't looked at the error about invalid escapes, since, well, fixing the other one seems to be complicated enough so I'll go with the generic uninformed opinion that it'd be great help; otherwise I don't know what exactly in the Stardoc update broke this; I wanted to look at it *after* I figure out the proto toolchain issue.

lberki:
The Stardoc issue is fixed in https://github.com/bazelbuild/stardoc/pull/237 .

fmeum:
Okay, I can look into it after the stardoc fix has been released, together with the general migration required at that point.

lberki:
Well, I was hoping it'd be a simple thing but turns out, not so; this is my first encounter with the magnificent Velocity Template Language and all my naive attempts to circumvent the backspace limitation failed.

fmeum:
I would be in favor of getting rid of the custom templates in favor of whatever stardoc ships by default. If that gets rid of the escaping issue, that's a plus.

lberki:
That is beyond my mandate here, I'm afraid; I was trying to do some adversarial programming to find a way to get a backslash into that string literal, to no avail, so the best I could do is to add some hackish method to `MarkdownUtil` to inject backslashes somehow.

Ain't nice, but it would work.

dzbarsky:
We are now on 0.8

PR #3968, opened 2024-06-21
docs: Discourage use `go_sdk.host()` in bzlmod.md (see #713)

### LINKED ISSUE #3967: go_sdk.host() leads to unstable unreproducible builds

I was using 0.46.0 when this problem occurred. I have meanwhile [switched](https://github.com/enola-dev/enola/pull/714) to 0.48.1, but I have no reason to believe anything specifically related to this changed in the meantime.

I was using 0.36.0 when this problem occurred. I have meanwhile [switched](https://github.com/enola-dev/enola/pull/714) to 0.37.0, but I have no reason to believe anything specifically related to this changed in the meantime.

7.1.1

I suspect so, but I'm not 100% sure, because I have no reason to believe anything specifically related to this problem changed in the meantime, but I gave up and have worked around this problem with another approach before trying to reliably reproduce it with the very very latest versions of rules_go & gazelle & Bazel itself. Based on my understanding of the problem, I doubt it just got fixed.

Debian Linux (derivation) with `apt` package manager on amd64 = x86_64.

Not relevant.

### What did you do?

I have used `go_sdk.host()`, as documented on https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/bzlmod.md, and my build would regularly break, see https://github.com/enola-dev/enola/issues/626, both on GitHub Action CI and locally on Debian & Fedora Linux distros, due to:

`ERROR: /home/runner/work/enola/enola/core/lib/BUILD:131:21: While resolving toolchains for target //core/lib:core_proto_doc (c4aba0f): invalid registered toolchain '@go_toolchains//:all': while parsing '@go_toolchains//:all': no such package '@@rules_go~~go_sdk~main___host_0//': Inconsistent filesystem operations. /opt/hostedtoolcache/go/1.21.8/x64/pkg/tool is no longer an existing directory. Did you delete it during the build? The results of the build are not guaranteed to be correct. You should probably run 'bazel clean' and investigate the filesystem inconsistency (likely due to filesystem updates concurrent with the build)`

The "during the build" & "likely due to filesystem updates concurrent with the build" parts of the error message are somewhat misleading; the root cause of this problem is simply that `go_sdk.host()` records `/opt/hostedtoolcache/go/1.21.8/x64/pkg/tool` somewhere, but that is not "stable", because an OS package manager (à la apt or dnf) regularly updates that "host" go version - and the paths change... e.g. from `/opt/hostedtoolcache/go/1.21.8/x64/pkg/tool` or `/usr/lib/go-1.21/pkg/tool` or whatever to a newer release.

_[{ Fast, Correct } — Choose two](https://bazel.build/install)_ 😈  LOL

### What did you see instead?

My builds regularly break "just by themselves" when I'm using `go_sdk.host()`.

vorburger:
see https://github.com/enola-dev/enola/issues/713

**What type of PR is this?**

Documentation

Fixes #3967

### DIFF
--- a/docs/go/core/bzlmod.md
+++ b/docs/go/core/bzlmod.md
@@ -44,6 +44,10 @@ go_sdk.download(
 go_sdk.host()
 ```
 
+Nota bene: The use of `go_sdk.host()` [may break builds](https://github.com/enola-dev/enola/issues/713) whenever the host Go version is upgraded
+(because many OS package managers, such as Debian/Ubuntu's `apt`, distribute Go into a directory which contains the version, such as `/usr/lib/go-1.22/`).
+As package upgrades happen outside of Bazel's control, this will lead to non-reproducible builds. Due to this, use of `go_sdk.host()` is discouraged.
+
 You can register multiple Go SDKs and select which one to use on a per-target basis using [`go_cross_binary`](rules.md#go_cross_binary).
 As long as you specify the `version` of an SDK, it will be downloaded lazily, that is, only when it is actually needed during a particular build.
 The usual rules of [toolchain resolution](https://bazel.build/extending/toolchains#toolchain-resolution) apply, with SDKs registered in the root module taking precedence over those registered in dependencies.


fmeum (approved):

vorburger:
I've just noticed that I copy/pasted the wrong issue ID in the commit message, converted to Draft to fix that...

PR #3969, opened 2024-06-21
Add a full `fs.FS` implementation to `runfiles`

### LINKED ISSUE #3375: New runfiles library is missing `listRunfiles`

v0.36.0

5.3.2

Yes

Linux

### What did you do?

Started using the new recommended runfiles library and tried to call ListRunfiles().

Same feature available as in the `github.com/bazelbuild/rules_go/go/tools/bazel` library.

### What did you see instead?

A `listRunfiles` function in the new library.

### My typical usage

I often mess up runfile paths. To make the error more readable, I suggest a similarly named file in the error message.

```go
package runfileutil

import (
	"errors"
	"fmt"
	"os"
	"path"
	"strings"

	runfiles1p "github.com/bazelbuild/rules_go/go/runfiles"
	"github.com/bazelbuild/rules_go/go/tools/bazel"
)

// Runfile returns an absolute path to the file named by "path", which should be
// a relative path from the workspace root to the file within the bazel
// workspace.
//
// Runfile may be called from tests invoked with 'bazel test' and binaries
// invoked with 'bazel run'. On Windows, only tests invoked with 'bazel test'
// are supported.
//
// If the runfile isn't found, an error is returned that suggests
// similarly-named runfiles.
func Runfile(filePath string) (string, error) {
	got, err := runfiles1p.Rlocation(filePath)
	if err == nil {
		return got, nil
	}
	if err != nil && strings.Contains(err.Error(), "could not locate file") {
		err = fmt.Errorf("%s: %w", err, os.ErrNotExist)
	}

	if !errors.Is(err, os.ErrNotExist) {
		return "", fmt.Errorf("unknown error locating runfile %q: %w", filePath, err)
	}
	base := path.Base(filePath)
	entries, err := bazel.ListRunfiles()
	if err != nil {
		return "", fmt.Errorf("could not locate %q: failed to list runfile entries: %w", filePath, err)
	}
	sameBaseEntries := filter(entries, func(entry bazel.RunfileEntry) bool {
		return path.Base(entry.ShortPath) == base
	})

	if len(sameBaseEntries) == 0 {
		return "", fmt.Errorf("could not locate %q among %d runfiles:\n  %s", filePath, len(entries),
			strings.Join(mapSlice(entries, func(e bazel.RunfileEntry) string { return e.ShortPath }), "\n  "))
	}
	return "", fmt.Errorf("could not locate %q among %d runfiles; suggested match(es):\n  %s", filePath, len(entries),
		strings.Join(mapSlice(sameBaseEntries, func(e bazel.RunfileEntry) string {
			return fmt.Sprintf("%q", e.ShortPath)
		}), "\n  "))

	// var sameBaseShortPaths []*bazel.RunfileEntry
	// for _, e := range entries {
	// 	if path.Base(e.ShortPath)
	// }
}

func mapSlice[T, R any](s []T, fn func(T) R) []R {
	var out []R
	for _, t := range s {
		out = append(out, fn(t))
	}
	return out
}

func filter[T any](s []T, fn func(T) bool) []T {
	var out []T
	for _, t := range s {
		if fn(t) {
			out = append(out, t)
		}
	}
	return out
}

```

### LINKED ISSUE #3830: Getting a runfiles directory with runfiles.Rlocation instead of bazel.Runfile

v0.45.0

v0.35.0

6.1.2

rules_go and gazelle are at the latest, Bazel isn't but I don't think it should affect this

Linux, x86-64

### What did you do?

I noticed that [`bazel.Runfile` was deprecated](https://pkg.go.dev/github.com/bazelbuild/rules_go/go/tools/bazel#Runfile) and recommended [the `runfiles` package as a replacement](https://pkg.go.dev/github.com/bazelbuild/rules_go/go/runfiles).

The problem I have is that `runfiles.Rlocation`, which I believe is the replacement, doesn't work for directories as far as I can tell.

I expected `bazel.Runfile("some/directory")` and `runfiles.Rlocation("some/directory")` to return the same result.

### What did you see instead?

 Instead, I get the error `runfile some/directory: file does not exist`. 

I'm able to work around this with:

```go
p, err := runfiles.Rlocation("__main__/some/directory/a_file_I_know_exists")
if err != nil { ... }
p = filepath.Dir(p)
// p is now what bazel.Runfile("some/directory") used to return
```

I'm not sure if I'm just using the wrong helper/missing something, but this workaround isn't great, as `a_file_I_know_exists` is a pretty hacky and specific construct, and I also am not sure where `__main__` comes from (was a shot in the dark based on printing a `runfiles.Runfiles`)

fmeum:
**What type of PR is this?**

Feature

Runfiles instances now implement `fs.FS` in a platform-agnostic way, regardless of whether they are manifest- or directory-based. This is meant to finally provide a full replacement for runfiles functionality offered by the legacy `bazel` package.

Fixes #3375
Fixes #3830 

**Other notes for review**


fmeum:
cc @sluongng

go/runfiles/manifest.go
```diff
@@ -97,10 +107,121 @@ func (m manifest) path(s string) (string, error) {
...
+	})
+	return &manifestReadDirFile{dirFile(path.Base(name)), entries}, nil
+}
+
+func (m *manifest) initTrie() {
+	if m.trie != nil {
+		return
+	}
+	m.trie = make(map[string]trie)
+	for k := range m.m {
```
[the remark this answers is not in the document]
fmeum: This will only trigger if the call to `m.path` at the beginning of the method fails, which means that either the user requested a file that doesn't exist (should be rare in passing tests) or explicitly wants to access a directory rather than a file. In this sense it is already lazy.

tests/runfiles/fs_test.go
```diff
@@ -12,59 +12,185 @@
...
 
 	"github.com/bazelbuild/rules_go/go/runfiles"
+	"github.com/bazelbuild/rules_go/go/tests/runfiles/testfs"
 )
 
-func TestFS(t *testing.T) {
-	fsys, err := runfiles.New()
+var isBzlmodEnabled bool
+
+func init() {
```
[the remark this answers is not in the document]
sluongng: I think even just a simple `fs.WalkDir` test that verifies all dirs and files inside a small runfiles tree are visited would be sufficient. I can then point folks to the test as an example of how to use the API. It would also help document the expected tree traversal order and contents, which we could compare against future rules_go/bazel changes.

go/runfiles/manifest.go
```diff
@@ -97,10 +107,119 @@ func (m manifest) path(s string) (string, error) {
...
+func (m *manifest) open(name string) (fs.File, error) {
+	if name != "." {
+		r, err := m.path(name)
+		if err == ErrEmpty {
+			return emptyFile(name), nil
+		} else if err == nil {
+			return os.Open(r)
+		}
+	}
+
```
sluongng, line 127: So once we are done with this `if` block, we can conclude that `name == "." || err == os.ErrNotExist` right?

To make it more clear:

```suggestion
	if name != "." {
		r, err := m.path(name)
		if err == ErrEmpty {
			return emptyFile(name), nil
		}
		if err == nil {
			// found
			return os.Open(r)
		}
		// err == os.ErrNotExist, do a full-tree scan
	}

```

go/runfiles/manifest.go
```diff
@@ -97,10 +107,119 @@ func (m manifest) path(s string) (string, error) {
...
+		}
+		if dir == nil {
+			return nil, os.ErrNotExist
+		}
+	}
+
+	var entries []manifestDirEntry
+	for e := range dir {
+		entries = append(entries, manifestDirEntry{e, m.index[path.Join(name, e)]})
+	}
```
sluongng, line 162: Prealloc:

```suggestion
	entries := make([]manifestDirEntry, 0, len(dir))
	for e := range dir {
		entries = append(entries, manifestDirEntry{e, m.index[path.Join(name, e)]})
	}
```

go/runfiles/manifest.go
```diff
@@ -97,10 +107,119 @@ func (m manifest) path(s string) (string, error) {
...
+	sort.Slice(entries, func(i, j int) bool {
+		return entries[i].name < entries[j].name
+	})
+	return &manifestReadDirFile{dirFile(path.Base(name)), entries}, nil
+}
+
+func (m *manifest) initTrie() {
+}
```
sluongng, line 170: Redundant

```suggestion
```

go/runfiles/manifest.go
```diff
@@ -97,10 +107,119 @@ func (m manifest) path(s string) (string, error) {
...
+		return nil, io.EOF
+	}
+	if n <= 0 || n > len(m.entries) {
+		n = len(m.entries)
+	}
+	entries := m.entries[:n]
+	m.entries = m.entries[n:]
+
+	var dirEntries []fs.DirEntry
+	for _, e := range entries {
```
sluongng, line 193: Prealloc

```suggestion
	dirEntries := make([]fs.DirEntry, 0, len(entries))
	for _, e := range entries {
```

sluongng (commented):
a few more nits

fmeum:
Thanks, I addressed your comments and am debugging the remaining test failure on Windows.

tests/runfiles/fs_test.go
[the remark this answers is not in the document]
fmeum: I added such a test and it unearthed that walking source directories in runfiles behaved differently between manifest and directory. This was the case because `fs.WalkDir` doesn't descend into directory symlinks, which are used in this case by the directory-based but not the manifest-based runfiles. We could fix this either way, I went with resolving symlinks here. Let me know what you think.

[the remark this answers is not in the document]
sluongng: I would imagine supporting symlink traversal would be more useful here. Though Im not sure if there could be any recursive setup that screws with Go's `fs.WalkDir` assumption. Perhaps if that ever happened, then the caller could implement tracking in their WalkFunc and return SkipDirErr when a seen dir is revisited.

So the current approach sgtm.

linzhp (approved):

fmeum:
@meteorcloudy The new macOS machines frequently run into timeouts: https://buildkite.com/bazel/rules-go-golang/builds/6622#01908e0d-57ff-4850-8b3c-cb9fe2fff52f

Are we perhaps lacking any settings?

meteorcloudy:
@fmeum Yes, please check https://github.com/bazelbuild/continuous-integration/issues/1981#issuecomment-2217609837

.bazelci/presubmit.yml
```diff
@@ -108,6 +108,7 @@ tasks:
       - "--apple_crosstool_top=@local_config_apple_cc//:toolchain"
       - "--crosstool_top=@local_config_apple_cc//:toolchain"
       - "--host_crosstool_top=@local_config_apple_cc//:toolchain"
+      - "--local_test_jobs=2"
```
meteorcloudy, line 111: Sorry I just check the job already runs with `--local_test_jobs=2`, which is added automatically by the CI script
https://buildkite.com/bazel/rules-go-golang/builds/6622#019096cb-8b58-4881-840b-42f5466ad948

I think you might have to actually bump the test sizes of those timeout tests.

meteorcloudy:
@fmeum To unblock this, you can also temporarily switch back to the old Mac fleet by using `macos_legacy` or `macos_arm64_legacy`.

The timeout looks weird, does it look like some network issue? One major difference is that the new Mac fleet is ipv6 only, not sure if it's relevant.

fmeum:
@meteorcloudy Thanks, I'll go with this workaround for now until I can properly investigate this (limited to GitHub web UI right now).

The failing test downloads a few old Go releases from the official download page, I don't know why this would cause issues. The relevant tests are tagged `local`.

meteorcloudy:
> One major difference is that the new Mac fleet is ipv6 only, not sure if it's relevant.

Indeed it's caused by the ipv6-only network. 

- We are setting `--host_jvm_args=-Djava.net.preferIPv6Addresses=true` in `/private/etc/bazel.bazelrc`
- But it only affected the outer Bazel, inner Bazel running in the test has [`--nosystem_rc`](https://github.com/bazelbuild/rules_go/blob/634fc283f8d84ea61253a62b99a61edb5509f5de/go/tools/bazel_testing/bazel_testing.go#L182) specified.

I'll let you decide how to inject this for macos ci, here is an example for Bazel's integration test https://github.com/bazelbuild/bazel/commit/2c51a0c8590e73c88d6a678f1b0583c6aee313c3

PR #3980, opened 2024-07-03
Add stamping support to go_binary

### LINKED ISSUE #2154: Add stamp attribute in go_binary rule

0.19.0

commit: 7a19100a6d86a392a2f88e337fbcc60a3e4b154a

Near master

0.24.1

Yes

Linux amd64

Hope the stamp behavior to keep consistent with other language toolchains in https://docs.bazel.build/versions/master/be/overview.html

Many [lang]_binary rule have stamp attribute.

ian-h-chamberlain:
**What type of PR is this?**

Feature

Adds a `stamp` attribute to `go_binary` which can be used to override `--[no]stamp` behavior for a specific binary target.

Semantics match [those of `cc_binary`](https://bazel.build/reference/be/c-cpp#cc_binary.stamp), and are limited to just go_binary for now.

Closes #2154

**Other notes for review**

Open questions / tasks:
- [ ] I am not 100% sure where documentation needs to be updated, if someone can point me in the right direction I'm happy to add that.
- [ ] Should `go_test` have this attribute as well? I didn't think it made sense for libraries have it, since conflicts between dependencies would need to be resolved, but I could probably add it for `_test` without needing to address those.


fmeum (commented):
The ternary value for `stamp` is generally pretty confusing, even if consistent with `cc_binary`. Could you explain what you would like to use this new attribute for? I wonder whether we really need to support all three values.

ian-h-chamberlain:
Sure, the use case we have in my organization is for iterating while manually testing one of our apps. We basically have two main use cases:

- Building for production via CI (we use `--stamp` in the build)
- Developer build for local testing (requires remembering `--stamp` every time to get a stamped build)

We don't want to stamp everything (non-Go, unit tests, etc.) all of the time, to maximize cache hits, so we don't have `--stamp` in our `.bazelrc`. However, we would ideally like to use stamping for developer builds to embed e.g. git data, build timestamp just for the final `go_binary`. 

As it stands without this change (or equivalent), users have to remember to pass `--stamp` every time they build this particular target to get the embedded metadata, but with this change we could set `stamp = 1` and forget it. For our purposes we would also be fine with `stamp = True` to force-stamp a particular target (where the default behavior for `False` would be to just follow `--[no]stamp`).

I mainly kept the ternary in case there would ever be users that want to forcefully _disable_ stamping for certain targets, but I'm not sure if is particularly common. A quick search on GitHub shows _some_ usage like this, but it seems fairly limited (mostly forks of `llvm-bazel`, I guess): https://github.com/search?q=%2F%5Cbstamp+%3D+0%2F+path%3A%2FBUILD%28%5C.bazel%29%3F%24%2F&type=code

gpanders:
I was surprised to learn today that `go_binary` does not already have a `stamp` attribute, and found this PR after trying to figure out why.

Our use case is similar to the one described above. Having this attribute would be helpful to us as well.

PR #3984, opened 2024-07-08
Break reliance on GOROOT_FINAL

### LINKED ISSUE #969: Full stdlib file paths in binary
Seems like paths to standard library packages are put into the binary. That means builds are not reproducible, right?
```
==================
WARNING: DATA RACE
Write at 0x00c4205d00cc by main goroutine:
  internal/race.Write()
      /private/var/tmp/_bazel_ash2k/ea451d98a8bdaad3872b7aa995bea02c/bazel-sandbox/1936151865359060414/execroot/__main__/bazel-out/host/bin/external/go_stdlib_darwin_amd64_cgo_race/src/internal/race/race.go:41 +0x38
  sync.(*WaitGroup).Wait()
      /private/var/tmp/_bazel_ash2k/ea451d98a8bdaad3872b7aa995bea02c/bazel-sandbox/1936151865359060414/execroot/__main__/bazel-out/host/bin/external/go_stdlib_darwin_amd64_cgo_race/src/sync/waitgroup.go:129 +0xf4
  github.com/ash2k/stager/wait.(*Group).Wait()
      vendor/github.com/ash2k/stager/wait/group.go:14 +0x3a
  github.com/atlassian/smith/pkg/controller.(*BundleController).Run()
      pkg/controller/controller.go:104 +0x51f
  github.com/atlassian/smith/cmd/smith/app.(*App).Run()
      cmd/smith/app/app.go:148 +0x169f
  main.runWithContext()
      cmd/smith/main.go:81 +0x417
  main.run()
      cmd/smith/main.go:36 +0xd9
  main.main()
      cmd/smith/main.go:26 +0x33

Previous read at 0x00c4205d00cc by goroutine 115:
  internal/race.Read()
      /private/var/tmp/_bazel_ash2k/ea451d98a8bdaad3872b7aa995bea02c/bazel-sandbox/1936151865359060414/execroot/__main__/bazel-out/host/bin/external/go_stdlib_darwin_amd64_cgo_race/src/internal/race/race.go:37 +0x38
  sync.(*WaitGroup).Add()
      /private/var/tmp/_bazel_ash2k/ea451d98a8bdaad3872b7aa995bea02c/bazel-sandbox/1936151865359060414/execroot/__main__/bazel-out/host/bin/external/go_stdlib_darwin_amd64_cgo_race/src/sync/waitgroup.go:71 +0x16f
  github.com/ash2k/stager/wait.(*Group).Start()
      vendor/github.com/ash2k/stager/wait/group.go:35 +0x43
  github.com/ash2k/stager/wait.(*Group).StartWithChannel()
      vendor/github.com/ash2k/stager/wait/group.go:20 +0xd0
  github.com/atlassian/smith/pkg/controller.(*crdEventHandler).ensureWatch()
      pkg/controller/controller_crd_event_handler.go:111 +0xaaf
  github.com/atlassian/smith/pkg/controller.(*crdEventHandler).OnAdd()
      pkg/controller/controller_crd_event_handler.go:36 +0x57
  k8s.io/client-go/tools/cache.(*processorListener).run()
      vendor/k8s.io/client-go/tools/cache/shared_informer.go:545 +0x3b2
  k8s.io/client-go/tools/cache.(*processorListener).(k8s.io/client-go/tools/cache.run)-fm()
      vendor/k8s.io/client-go/tools/cache/shared_informer.go:381 +0x41
  k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1()
      vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:71 +0x5c

Goroutine 115 (running) created at:
  k8s.io/apimachinery/pkg/util/wait.(*Group).Start()
      vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:69 +0x6f
  k8s.io/client-go/tools/cache.(*sharedProcessor).addAndStartListener()
      vendor/k8s.io/client-go/tools/cache/shared_informer.go:381 +0x2dd
  k8s.io/client-go/tools/cache.(*sharedIndexInformer).AddEventHandlerWithResyncPeriod()
      vendor/k8s.io/client-go/tools/cache/shared_informer.go:331 +0x281
  k8s.io/client-go/tools/cache.(*sharedIndexInformer).AddEventHandler()
      vendor/k8s.io/client-go/tools/cache/shared_informer.go:267 +0x67
  github.com/atlassian/smith/pkg/controller.(*BundleController).Run()
      pkg/controller/controller.go:89 +0x2e3
  github.com/atlassian/smith/cmd/smith/app.(*App).Run()
      cmd/smith/app/app.go:148 +0x169f
  main.runWithContext()
      cmd/smith/main.go:81 +0x417
  main.run()
      cmd/smith/main.go:36 +0xd9
  main.main()
      cmd/smith/main.go:26 +0x33
==================
Found 1 data race(s)
```

### LINKED ISSUE #3983: Full GOROOT paths in Go 1.23 binaries

v0.48.0

v0.37.0

7.2.0

Yes

linux amd64

N/A

### What did you do?

Ran this rules_go test on Go 1.23rc1:
```shell
$ bazel test //tests/core/strip:strip_test
```

Target passes.

### What did you see instead?

All cases failed with messages like the following. For brevity I'll only show one case.
```
    --- FAIL: Test/run_sometimes_opt (0.95s)
        strip_test.go:228: running: bazel run --strip sometimes --compilation_mode opt //:strip -- -wantstrip=false
        strip_test.go:240: unexpected error: exit status 2
            stderr:
            WARNING: The following rc files are no longer being read, please transfer their contents or import their path into one of the standard rc files:
            /etc/bazel.bazelrc
            Computing main repo mapping:
            Loading:
            Loading: 0 packages loaded
            WARNING: Build option --compilation_mode has changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
            Analyzing: target //:strip (0 packages loaded, 0 targets configured)
            Analyzing: target //:strip (0 packages loaded, 0 targets configured)
            [0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
            INFO: Analyzed target //:strip (0 packages loaded, 10813 targets configured).
            INFO: Found 1 target...
            Target //:strip up-to-date:
              bazel-bin/strip_/strip
            INFO: Elapsed time: 0.908s, Critical Path: 0.26s
            INFO: 2 processes: 1 internal, 1 processwrapper-sandbox.
            INFO: Build completed successfully, 2 total actions
            INFO: Running command line: bazel-bin/strip_/strip '-wantstrip=false'
            panic: got unexpected stack trace line "\tbazel-out/k8-opt/bin/external/io_bazel_rules_go/stdlib_/src/runtime/debug/stack.go:26 +0x5e" at index 2

            goroutine 1 [running]:
            main.main()
                strip.go:33 +0x2b9
```

This is likely due to https://github.com/golang/go/issues/62047, which drops support for `GOROOT_FINAL`, which rules_go uses to replace actual GOROOTs to keep builds reproducible (see #971).

JacobOaks:
**What type of PR is this?**

Bug fix

Due to https://github.com/golang/go/issues/62047, Go 1.23 won't support `GOROOT_FINAL`. This means that #971 will no longer fix #969,
meaning linked binaries will contain full GOROOT paths, making them non-reproducible.

Instead of using `GOROOT_FINAL`, this change directly sets `GOROOT` when invoking the linker, which will cause the linker to continue to write `GOROOT` into the binary, keeping builds consistent.

[An existing test](https://github.com/bazelbuild/rules_go/blob/master/tests/core/strip/strip_test.go#L247-L260) that asserts on paths containing GOROOT passes with this PR. This passes on Go 1.23rc1 as well (it fails without this PR).

I'm not a rules_go expert and I'm not married to this particular solution. If there's a clever way we can set `-trimpath` when invoking the compiler, that might be better - but we already use that flag to trim off the bazel sandbox I believe.

Fixes #3983


sluongng:
I suspect this might require a minimum Go SDK version bump to 1.21, in rules_go setup as well as in the test.

Is it possible to make the behavior conditional based on the current SDK version?

JacobOaks:
Hey @sluongng - can you elaborate on why you suspect that? As far as I can tell, rules_go seems to be set up for 1.21 already.

sluongng:
It's from the issue you linked.

> As of Go 1.21, in order to provide reproducible builds of the Go toolchain (https://github.com/golang/go/issues/57120), cmd/dist started building releases with the -trimpath flag set, which incidentally caused GOROOT_FINAL to have no effect (https://github.com/golang/go/issues/61921). (Development versions of toolchains are still built without -trimpath by default, so GOROOT_FINAL may affect those builds somewhat.)

So we could safely ignore this variable starting from 1.21.

However, some folks could be using rules_go with older Go SDK versions who wish to get the latest rules_go features still. It's best if we could avoid forcing folks to upgrade their Go SDK when upgrading rules_go by making rules_go code work with older versions as well as newer ones.

Effectively, I think you could undo your change in go/private/context.bzl and leave the variable as is.

In link.go, you could call a `func manageGoRoot() (func())` func (return is the defer func) with 2 different implementations: one using 
```
//go:build go1.21
// +build go1.21
```
and one using
```
//go:build !go1.21
// +build !go1.21
```

sluongng:
On second thought, if this is only a breaking behavior in 1.23, I think we should make it a 1.23 forward feature instead of 1.21. 🤔

JacobOaks:
Hey @sluongng - so I did try this solution on a couple pre-Go1.23 versions, and it worked on the most recent older versions too - but I agree it's probably a good idea to hide this behind 1.23+ just to be safe.

I tried to add build constraints, but the problem I ran into is that `go_tool_binary` targets (which the builder binary happens to  be) [do not support build constraints](https://github.com/bazelbuild/rules_go/blob/master/go/private/rules/binary.bzl#L509-L512) due to the way [they invoke go build](https://github.com/bazelbuild/rules_go/blob/master/go/private/rules/binary.bzl#L472). So it seems like we have a couple options:
* Change `go_tool_binary` to materialize a temporary go module
* Switch off of something like `runtime.Version()`, which is probably quite brittle and undesirable
* Verify that this change works on all Go versions supported by rules_go and drop the version-switching idea

Let me know if you have thoughts/other ideas.

sluongng:
ah it's part of nogo which is built after a transition? 🤔 

Hmm...

JacobOaks:
Yeah, I think nogo is a separate type of rule that does support build constraints.

I updated the PR to switch based off `runtime.Version` for now, which like I mentioned isn't quite as satisfying as build constraints would be, but does seem to technically work.

fmeum (approved):

tyler-french (approved):
We've tested this at Uber

go/tools/builders/link.go
```diff
@@ -161,3 +175,20 @@ func link(args []string) error {
 
 	return nil
 }
+
+var versionExp = regexp.MustCompile(`.*go1\.(\d+).*$`)
```
sywhang, line 179: Consider avoiding regex magic with stdlib: https://pkg.go.dev/go/version

JacobOaks: I decided to avoid this to keep the logic compilable on older Go versions than 1.22.

sywhang: makes sense -- disregard this comment. Thanks for the clarification 👍

PR #3986, opened 2024-07-09
rm crosstool

sluongng:
**What type of PR is this?**

Documentation

Crosstool is the old way in Bazel to define a C/C++ toolchain.
It has been deprecated and replaced with the new toolchain + platforms auto resolve system in recent years.

The current docs in rules_go are out-of-date with a huge disclaimer up top.
Several links still point to the old Wiki page which has already been moved to ./docs/ dir.

This PR does 2 things:

1. Rename `crosstool` attribute in Go Context in starlark to `cc_toolchain_files`.
This new name better communicates the actual usage of the attribute, which is to propagate all
CC toolchain files to Go actions.

2. Remove `go/crosstool.rst` doc entirely and update out-of-date crosstool references to point to
the new platforms and toolchains doc.

**Other notes for review**


WORKSPACE
```diff
@@ -83,7 +83,7 @@ http_archive(
 )
 
 # For manual testing against an LLVM toolchain.
-# Use --crosstool_top=@llvm_toolchain//:toolchain
+# Use --extra_toolchains=@llvm_toolchain//:toolchain
```
fmeum, line 86: Is this still the right target (`toolchain`, not `cc_toolchain`)?

go/toolchains.rst
```diff
@@ -601,7 +601,7 @@ Fields
 | The suffix to use for shared libraries in this build mode. Mostly used when                      |
 | generating output filenames of binary rules.                                                     |
 +--------------------------------+-----------------------------------------------------------------+
-| :param:`crosstool`             | :type:`list of File`                                            |
+| :param:`cc_toolchain_files`    | :type:`list of File`                                            |
```
fmeum, line 604: As much as I dislike this fact, this is public API. It's probably still fine to make this change, but we should remember to mention it in the release notes.

WORKSPACE
sluongng: Good call. It was previously pointing to a cc_toolchain_suite. Fixed it to point to `toolchain` targets.

go/toolchains.rst
sluongng: Yeah. I am highly skeptical that anyone is using this API.

I briefly considered marking it as deprecated, and adding a new attribute called `cc_toolchain_files`. But I think that would still result in the same "potential" breakages.

I think it's better just to rip off the bandage here so that the rules are cleaner and easier to maintain. If there are downstream complains about this, we could take note for the next time we do these cleanups.

fmeum (approved):

fmeum:
I'll try to fix ci over on the FS PR first.

sluongng:
I think this PR should be safe to merge now?

PR #3990, opened 2024-07-18
Migrate to macos_arm64

meteorcloudy:
Bazel CI is going to remove `macos_legacy` soon.

Removing `--nosystem_rc` in test setup to pick up `--host_jvm_args=-Djava.net.preferIPv6Addresses=true` set in `/private/etc/bazel.bazelrc` on Bazel CI's Mac VMs (ipv6-only).

Context: https://github.com/bazelbuild/continuous-integration/issues/1981#issuecomment-2230296824

### DIFF
--- a/go/tools/bazel_testing/bazel_testing.go
+++ b/go/tools/bazel_testing/bazel_testing.go
@@ -179,9 +179,9 @@ func TestMain(m *testing.M, args Args) {
 // hide that this code is executing inside a bazel test.
 func BazelCmd(args ...string) *exec.Cmd {
 	cmd := exec.Command("bazel")
-    // --nosystem_rc isn't used here because Bazel may need essential flags set in
-    // system rc to be able to work correctly
-    // See https://github.com/bazelbuild/rules_go/pull/3969#issuecomment-2220405416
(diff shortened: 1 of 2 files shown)


go/tools/bazel_testing/bazel_testing.go
```diff
@@ -179,7 +179,7 @@ func TestMain(m *testing.M, args Args) {
 // hide that this code is executing inside a bazel test.
 func BazelCmd(args ...string) *exec.Cmd {
 	cmd := exec.Command("bazel")
-	cmd.Args = append(cmd.Args, "--nosystem_rc", "--nohome_rc")
+	cmd.Args = append(cmd.Args, "--nohome_rc")
```
fmeum, line 182: Could you add a comment explaining why `--nosystem_rc` isn't used here? This is a good example for why we don't want it, even though it appears to be a must for integration tests otherwise.

fmeum (approved):

go/tools/bazel_testing/bazel_testing.go
meteorcloudy: Done!

fmeum (approved):

PR #3995, opened 2024-07-26
Run nogo in a separate validation action

### LINKED ISSUE #3619: Nogo: provide a way to ignore external cgo source

v0.40.1

Irrelevant

6.2.1

Yes

Darwin / ARM64

N/A

### What did you do?

We have a dependency over `github.com/shirou/gopsutil/v3/process` package.

When enable `copylocks` analyzer by including `"@org_golang_x_tools//go/analysis/passes/copylock:go_default_library"` in `nogo` config, the code fail to compile.

We have already set `nogo_config.json` with

```
    "copylocks": {
        "exclude_files": {
            "external[\\\\,\\/]": "third_party"
        }
    },
```

However, the `cgo` source path does not include `external` and therefore, escaped the current regex.

We also want to avoid specific regex like this

```
    "copylocks": {
        "exclude_files": {
            "cgo[\\\\,\\/]github.com[\\\\,\\/]shirou[\\\\,\\/]gopsutil[\\\\,\\/]": "third_party cgo",
            "external[\\\\,\\/]": "third_party"
        }
    },
```

Even though it works, it might be harder to maintain.

If possible, external `cgo` sources should be included into `nogo` with `external` inside the source path.
This would allow us to ignore those files using existing regex.

### What did you see instead?

```
ERROR: /private/var/tmp/_bazel_sluongng/06e573a93bc2d6a9cad4ad41f00b4310/external/com_github_shirou_gopsutil_v3/process/BUILD.bazel:3:11: GoCompilePkg external/com_github_shirou_gopsutil_v3/process/process.a failed: (Exit 1): builder failed: error executing command (from target @com_github_shirou_gopsutil_v3//process:process) bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/go_sdk_darwin_arm64/builder_reset/builder compilepkg -sdk external/go_sdk_darwin_arm64 -installsuffix darwin_arm64 -src ... (remaining 95 arguments skipped)

/var/folders/ly/gjsqx9wd0cv1ck7xdslrt6t00000gn/T/rules_go_work-1366086039/cgo/github.com/shirou/gopsutil/v3/process/process.go:144:9: String passes lock by value: github.com/shirou/gopsutil/v3/process.Process contains sync.RWMutex (copylocks)
/var/folders/ly/gjsqx9wd0cv1ck7xdslrt6t00000gn/T/rules_go_work-1366086039/cgo/github.com/shirou/gopsutil/v3/process/process.go:145:23: call of json.Marshal copies lock value: github.com/shirou/gopsutil/v3/process.Process contains sync.RWMutex (copylocks)
```

Note how `rules_go_work-1366086039/cgo/github.com/shirou/gopsutil/v3/process/process.go` does not contain external inside.

### LINKED ISSUE #3695: Add validation actions for nogo outputs

### LINKED ISSUE #3846: compilepkg for cgo targets fails with nogo and --collect_code_coverage enabled
`0.45.1`

`0.35.0`

`6.4.0`

yes

```
➜  cgo_test git:(main) go env GOHOSTARCH GOHOSTOS
amd64
linux
```

issue is reproducible, with or without hermetic C toolchain

### What did you do?

minimal repro:  https://github.com/xytan0056/cgo_test 
force enable nogo [here](https://github.com/xytan0056/cgo_test/blob/main/WORKSPACE#L155) 
```
bazel build ":project" --collect_code_coverage
```

build pass 

### What did you see instead?
```
compilepkg: open bazel-out/k8-fastbuild/bin/project_/_cgo_install.h: file exists
```
with context
```
ERROR: /home/user/cgo_test/BUILD.bazel:40:10: GoCompilePkg project.a failed: (Exit 1): builder failed: error executing command bazel-out/path/to/bin/external/go_sdk/builder_reset/builder compilepkg -sdk external/go_sdk -installsuffix linux_amd64 -src main.go -src add.c -src add.h -embedroot '' -embedroot ... (remaining 49 arguments skipped)

compilepkg: open bazel-out/k8-fastbuild/bin/project_/_cgo_install.h: file exists
Target //:project failed to build
INFO: Elapsed time: 7.345s, Critical Path: 1.49s
INFO: 41 processes: 6 internal, 35 processwrapper-sandbox.
```
> [!NOTE]  
> However after disabling `nogo`, build can pass

fmeum:
**What type of PR is this?**

Feature

`nogo` is run in a separate action that can be toggled with `--run_validations`. This avoids rerunning compilation if the `nogo` tool changes and also allows to collect `nogo` findings of targets that depend on other targets that themselves have `nogo` findings (with `--keep_going`).

Fixes #3619
Fixes #3695 
Fixes #3846 
Closes #3707

**Other notes for review**


fmeum:
@joeljeske Sorry for the long delay, but this should finally be ready (and is certainly heavily inspired by your PR!).

tests/core/nogo/custom/custom_test.go
```diff
@@ -455,7 +455,7 @@ func Test(t *testing.T) {
 			wantSuccess: false,
 			includes: []string{
 				// note the cross platform regex :)
-				`.*[\\/]cgo[\\/]examplepkg[\\/]pure_src_with_err_calling_native.go:.*function must not be named Foo \(foofuncname\)`,
+				`examplepkg[\\/]pure_src_with_err_calling_native.go:.*function must not be named Foo \(foofuncname\)`,
```
fmeum, line 458: I'm not entirely sure why I needed to make this change. Any ideas?

fmeum:
CC @tingilee as you were raised the original issue

go/tools/builders/nogo_validation.go
```diff
@@ -0,0 +1,27 @@
...
+	"errors"
+	"os"
+)
+
+func nogoValidation(args []string) error {
+	validationOutput := args[0]
+	logFile := args[1]
+	// Always create the output file, but fail if the log file is non-empty, to
+	// avoid a "action failed to create outputs" error.
+	if err := os.WriteFile(validationOutput, []byte{}, 0644); err != nil {
```
linzhp, line 13: nit: we can use `os.Create` for short.

Optionally, we can write logContent to `validationOutput` to preserve the log

go/private/actions/compilepkg.bzl
```diff
@@ -194,3 +202,89 @@ def emit_compilepkg(
...
+    )
+
+    validation_args = go.actions.args()
+    validation_args.add("nogovalidation")
+    validation_args.add(out_validation)
+    validation_args.add(out_log)
+    go.actions.run(
+        inputs = [out_log],
+        outputs = [out_validation],
+        mnemonic = "GoNogoValidation",
```
linzhp, line 285: nit: we can call this `ValidateNogo` to be consistent with `CompilePkg`.

go/tools/builders/nogo.go
```diff
@@ -0,0 +1,166 @@
...
+	cmd.Stdout, cmd.Stderr = out, out
+	// Always create this file as a Bazel-declared output, but keep it empty
+	// if nogo finds no issues.
+	outLog, err := os.Create(outLogPath)
+	if err != nil {
+		return fmt.Errorf("error creating nogo log file: %v", err)
+	}
+	defer outLog.Close()
+	if err := cmd.Run(); err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
```
linzhp, line 141:
```suggestion
	err := cmd.Run()
	if err == nil {
		return nil
	}
	if exitErr, ok := err.(*exec.ExitError); ok {
```

reduce the nesting.

go/tools/builders/nogo.go
```diff
@@ -0,0 +1,166 @@
...
+			if !exitErr.Exited() {
+				cmdLine := strings.Join(args, " ")
+				return fmt.Errorf("nogo command '%s' exited unexpectedly: %s", cmdLine, exitErr.String())
+			}
+			prettyOut := relativizePaths(out.Bytes())
+			if exitErr.ExitCode() == 2 {
+				// Do not fail the action if nogo has findings so that facts are
+				// still available for downstream targets.
+				_, err := outLog.Write(prettyOut)
```
linzhp, line 150:
```suggestion
			if exitErr.ExitCode() != 2 {
				return errors.New(string(prettyOut))
			}
		// Do not fail the action if nogo has findings so that facts are
		// still available for downstream targets.
		_, err := outLog.Write(prettyOut)
```
reduce nesting

go/tools/builders/compilepkg.go
```diff
@@ -17,29 +17,17 @@
...
 	"sort"
 	"strings"
 )
 
-type nogoResult int
-
-const (
-	nogoNotRun nogoResult = iota
-	nogoError
-	nogoFailed
```
linzhp, line 39: Let's keep these constants (maybe put them in a new file constant.go), so we can do `if exitErr.ExitCode() == nogoFailed`, which is more readable.

We probably want to make `nogoSucceeded` to be 0.

linzhp (approved):

go/private/actions/compilepkg.bzl
fmeum: What would you call the action that actually runs `Nogo`? Just `RunNogo`?

linzhp (approved):

go/private/actions/compilepkg.bzl
linzhp: yeah, `RunNogo` sounds good

tests/core/nogo/custom/custom_test.go
sluongng: The previous path with `cgo/` prefix was created from https://github.com/bazelbuild/rules_go/blob/86a7f025d2fc052ab6154603b481cb7e4419598f/go/tools/builders/cgo2.go#L68-L71

Could it be because of the new `cgoGoSrcsPath` is being used instead?

In some analyzers setup, folks would have a regex to specifically exclude cgo-generated source code from being scanned. I think if we cannot ensure that cgo-generated source are not being excluded by default, we should provide some consistent naming patterns for folks to exclude them downstream.

go/private/actions/compilepkg.bzl
```diff
@@ -194,3 +202,89 @@ def emit_compilepkg(
...
+    validation_args.add(out_log)
+    go.actions.run(
+        inputs = [out_log],
+        outputs = [out_validation],
+        mnemonic = "ValidateNogo",
+        executable = go.toolchain._builder,
+        arguments = [validation_args],
+        execution_requirements = SUPPORTS_PATH_MAPPING_REQUIREMENT,
+        progress_message = "Validating nogo output for %{label}",
+    )
```
sluongng, line 290: Could you please add a comment here to clarify why we use 2 instead of 1 action? And if there is a "failing" analysis, then which of the 2 actions will fail?

It's slightly confusing to me because if we were to run this with `--sandbox_debug`, then the sandbox would not contain the failing scanned ".go" source file, but only the output log from `RunNogo`. This could make troubleshooting nogo failure a bit more difficult.

go/tools/builders/compilepkg.go
```diff
@@ -277,20 +257,10 @@ func compileArchive(
...
-				filtered = append(filtered, s)
-			}
-		}
-		return filtered
-	}
 	// When coverage is set, source files will be modified during instrumentation. We should only run static analysis
 	// over original source files and not the modified ones.
 	// goSrcsNogo and cgoSrcsNogo are copies of the original source files for nogo to run static analysis.
-	goSrcsNogo := filterForNogo(goSrcs)
+	goSrcsNogo := append([]string{}, goSrcs...)
```
sluongng, line 263: Nit: since 1.21, we have `slices.Clone()`.

Could you please add a comment up top to remind us about it when we raise the minimum supported Go version?

go/tools/builders/compilepkg.go
```diff
@@ -348,23 +318,26 @@ func compileArchive(
...
+			}
+			_, _, _, err = cgo2(goenv, goSrcsNogo, cgoSrcsNogo, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSrcs, hSrcs, packagePath, packageName, cc, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags, cgoExportHPath, cgoGoSrcsForNogoPath)
 			if err != nil {
 				return err
 			}
 		} else {
-			goSrcsNogo = goSrcs
+			// If the package uses Cgo, compile .s and .S files with cgo2, not the Go assembler.
+			// Otherwise: the .s/.S files will be compiled with the Go assembler later
+			srcDir, goSrcs, objFiles, err = cgo2(goenv, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSrcs, hSrcs, packagePath, packageName, cc, cppFlags, cFlags, cxxFlags, objcFlags, objcxxFlags, ldFlags, cgoExportHPath, cgoGoSrcsForNogoPath)
```
sluongng, line 336: Nit: the comment is supposed to apply to both the if and else clauses, not exclusively to the else clause.

Perhaps we could move this to outside of the else clause like this

```
cgoSrcsForNogo := cgoGoSrcsForNogoPath
if coverMode != "" && cgoGoSrcsForNogoPath != "" {
  cgoSrcsForNogo = ""
  ...
}
srcDir, goSrcs, objFiles, err = cgo2(..., cgoSrcsForNogo)
...
```

wdyt?

go/tools/builders/constants.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+const (
+	nogoSuccess int = iota
+	nogoError
+	nogoViolation
+)
```
sluongng, line 21: Nit: It's a bit weird to see us share a source file like this between 2 binaries.

Should we make a new `tools/nogo` package for the nogo binary and import this as a shared library?

sluongng: Please also add a comment here to note that these are not just internal states, but action exit code (0, 1, 2).

Having the word `exit code` somewhere close would help newer folks grep through the source a bit easier.

sluongng (commented):
Few nits, but the current approach lgtm

tests/core/nogo/custom/custom_test.go
fmeum: I don't see how cgo-generated sources could be excluded by us: They may be needed by an analyzer to understand where symbols originating from cgo come from. If you see a way to send them to the nogo binary but still somehow exclude them from actual analysis, that would be a great way to resolve this.

Files generated by cgo either have a filename that starts with an underscore or contain the comment `// Code generated by cmd/cgo; DO NOT EDIT.`. As that is also true with regular Go, I would prefer folks to just depend on that to exclude these files if needed.

go/private/actions/compilepkg.bzl
fmeum: I added comments to both actions that should address your concerns: `--sandbox_debug` debugging is still possible.

go/tools/builders/constants.go
fmeum: I added a comment to explain that these are exit codes and another one to explain why (and that) we share the source file. There are ways to break the cyclic dependency, but source file sharing seems simple enough and other targets in this package already to it.

go/tools/builders/compilepkg.go
fmeum: I thought about doing that, but failed to find a layout that communicates that we are running `cgo` twice and ignoring its outputs in such a clear way. I added a TODO about getting rid of the unnecessarily duplicated compilation and also duplicated the comment on when to use which assembler.

fmeum: Nvm about the TODO, cgo needs the compilation output to generate code.

sluongng: Here is an example cgo regex we have used to exclude out nogo failures on generated code 

https://github.com/buildbuddy-io/buildbuddy/blob/d8461cff320849f1f9dc8aef322a15a525f76f10/BUILD#L39-L41

I don't mind using a different regex, but ideally, there should be a regex that we could reliably exclude generated code.

go/private/actions/compilepkg.bzl
sluongng: So there are 2 failure cases:

- Rules/analyzer author: the analyzer failed to run and die after the first action. This can be sandbox debugged.

- Typical developer: they changed some Go source code, did build, and failed the "linter". This check happens in the second action and currently, the sandbox for this would only include the log file. If user try to debug with sandbox debug, they will only be able to see the log file and nothing else.

What do you think if we reuse all the inputs of the first action in the second action?
I think that would enable users to use sandbox debug to inspect the `.go` source files that are failing the style check.

fmeum: Would you consider it a common case to `--sandbox_debug` a linter failure? Wouldn't the output of `--verbose_failures` or `-s` be sufficient to debug this? `--spawn_strategy=local` is also always an option when you aren't interested in the sandboxing aspect.

I'm hesitant to add unnecessary inputs to an action that can easily lie on the critical path, but it's true that it is very fast and reusing inputs shouldn't be too costly.

sluongng: In the past, we have had nogo running on all sorts of unexpected source files: external deps, coverage generated, cgo generated, etc... So it was common for me to use `--sandbox_debug` to check what was inside the GoCompilePkg sandbox: inputs and leftover outputs.

As nogo improved over time, this has happened less frequently. Still, I would prefer to keep the ability to debug the action and to let users access what caused the failures vs only being able to access the action that processed the failure log.

sluongng (approved):
Lgtm. Let's ship this while the pr is still small 😂

PR #4000, opened 2024-07-31
Make `go_bazel_test` pass with Bzlmod enabled

fmeum:
This is a quick fix for the test runner only, the tests themselves still rely on WORKSPACE (via local_repository). The rules_go repo is also hardcoded, which means that other external repos likely won't work.

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**

### DIFF
--- a/go/tools/bazel_testing/bazel_testing.go
+++ b/go/tools/bazel_testing/bazel_testing.go
@@ -101,6 +101,8 @@ type Args struct {
 // instead of running tests.
 const debug = false
 
+var bzlmodEnabled = os.Getenv("BZLMOD_ENABLED") == "True"
+
 // outputUserRoot is set to the directory where Bazel should put its internal files.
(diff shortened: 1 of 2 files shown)


fmeum:
@tyler-french @linzhp This is the minimal change needed to get `go_bazel_test` to pass in the Gazelle repo. 

The automatic collection of repos from runfiles is problematic with Bzlmod since we have no easy way of knowing the apparent name. If we decide to stick with it, I can probably make it work by generating and parsing `repo_mapping` files, but even that may not work due to the differences between `local_repository` and a proper Bazel module.

A better solution would be to rewrite the tests to use Bzlmod exclusively and, with the exception of the current repo, load all deps from the BCR instead of from runfiles. That's quite a non-trivial task though.

PR #4005, opened 2024-08-02
Rewrite `//go:cgo_ldflag` comments to relative paths in stdlib action

### LINKED ISSUE #3994: GoStdlib includes absolute paths in the output

0.48.0

N/A

`7.2.1`

Yes

Linux x86_64, but can repro on any platform

Custom crosstool, using LLVM toolchain and custom sysroot. The crosstool setup find the sysroot and sets -I and -L to the correct _relative_ paths

### What did you do?
I am building a simple go binary, but the generated stdlib archives contain absolute paths. Looks like paths passed to the linker are expanded to absolute paths and stored in the object files. [repro.zip](https://github.com/user-attachments/files/16383084/repro.zip) contains a minimal repro:
```
$ unzip repro.zip
$ bazel build :example
$ strings $(find bazel-out/ -name cgo.a) | grep external/llvm
```

Fully hermetic and reproducible builds

### What did you see instead?
The full path in the output causes the output digest to change, reducing cache hits

fmeum:
**What type of PR is this?**

Bug fix

The linker flags passed to the stdlib action need to contain absolute paths to work in Go's own build environment, but should not pollute the resulting archives with absolute paths. Since cgo persists the linker flags in `//go:cgo_ldflag` comments, those are now rewritten by the (renamed) filterbuildid command.

Fixes #3994

**Other notes for review**


fmeum:
#4009 is a more general solution

PR #4009, opened 2024-08-05
Lazy expand absolute path to avoid including them in the output

### LINKED ISSUE #3994: GoStdlib includes absolute paths in the output

AlessandroPatti:
**What type of PR is this?**

Bug fix

It fixes reproducibility issues reported in https://github.com/bazelbuild/rules_go/issues/3994 by lazy-expanding paths to absolute paths

Fixes #3994

### DIFF
--- a/go/tools/builders/cc.go
+++ b/go/tools/builders/cc.go
@@ -26,6 +26,10 @@ func cc(args []string) error {
 			trimmed := strings.TrimPrefix(s, cgoAbsPlaceholder)
 			abspath := filepath.Join(ccroot, trimmed)
 			if _, err := os.Stat(abspath); err == nil {
+                                // Only return the abspath if it exists, otherwise it
+                                // means that either it won't have any effect or the original
+                                // value was not a relpath (e.g. a path with a XCODE placehold from
+                                // macos cc_wrapper)
 				return abspath
 			}
 			return trimmed
--- a/go/tools/builders/cc.go
+++ b/go/tools/builders/cc.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+)
+
+func cc(args []string) error {
+	cc := os.Getenv("GO_CC")
+	if cc == "" {
+		errors.New("GO_CC environment variable not set")
+	}
+	ccroot := os.Getenv("GO_CC_ROOT")
+	if ccroot == "" {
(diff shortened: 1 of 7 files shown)


fmeum:
@AlessandroPatti Tests are failing on macOS, presumably because the wrapper isn't fully transparent.

go/tools/builders/cc.go
```diff
@@ -0,0 +1,43 @@
...
+	if ccroot == "" {
+		errors.New("GO_CC_ROOT environment variable not set")
+	}
+	normalized := []string{cc}
+	normalized = append(normalized, args...)
+	transformArgs(normalized, cgoAbsEnvFlags, func(s string) string {
+		if strings.HasPrefix(s, cgoAbsPlaceholder) {
+			trimmed := strings.TrimPrefix(s, cgoAbsPlaceholder)
+			abspath := filepath.Join(ccroot, trimmed)
+			if _, err := os.Stat(abspath); err == nil {
```
fmeum, line 28: Could you add a comment explaining why this is needed?

tests/integration/reproducibility/reproducibility_test.go
```diff
@@ -200,49 +202,92 @@ func Test(t *testing.T) {
...
+			root, err := os.Readlink(filepath.Join(dir, "bazel-out/"))
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+				if err != nil {
+					return err
+				}
+				if !strings.Contains(path, "io_bazel_rules_go/stdlib_/pkg") {
```
fmeum, line 270: This will break with Bzlmod. Could we make this check less strict, e.g. only go by an `stdlib_/pkg` prefix?

fmeum (approved):

fmeum:
Thanks, this is a really clean and well-tested solution!

PR #4010, opened 2024-08-06
Various starlark cleanups

DavidZbarsky-at:
**What type of PR is this?**
**> Other**

Cleans up various Stalark inefficiencies. This saves around .5s analysis time in our repo, which does not have too much Go code yet. It is likely more noticeable in repos that have more go code

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -82,9 +82,10 @@ def emit_compilepkg(
     if bool(nogo) != bool(out_facts):
         fail("nogo must be specified if and only if out_facts is specified")
 
-    inputs = (sources + embedsrcs + [go.sdk.package_list] +
-              [archive.data.export_file for archive in archives] +
-              go.sdk.tools + go.sdk.headers + go.stdlib.libs)
+    inputs_direct = (sources + embedsrcs + [go.sdk.package_list] +
+                     [archive.data.export_file for archive in archives] +
+                     go.sdk.headers)
+    inputs_transitive = [go.sdk.tools, go.stdlib.libs]
     outputs = [out_lib, out_export]
 
     args = go.builder_args(go, "compilepkg", use_path_mapping = True)
@@ -105,7 +106,7 @@ def emit_compilepkg(
         expand_directories = False,
     )
     if cover and go.coverdata:
-        inputs.append(go.coverdata.data.export_file)
+        inputs_direct.append(go.coverdata.data.export_file)
         args.add("-arc", _archive(go.coverdata))
         if go.mode.race:
             args.add("-cover_mode", "atomic")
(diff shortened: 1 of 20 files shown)


go/private/actions/compilepkg.bzl
```diff
@@ -82,7 +82,7 @@ def emit_compilepkg(
     if bool(nogo) != bool(out_facts):
         fail("nogo must be specified if and only if out_facts is specified")
 
-    inputs = (sources + embedsrcs + [go.package_list] +
+    inputs = (sources + embedsrcs + [go.sdk.package_list] +
```
fmeum, line 85: Only vaguely related: I wonder what we could save if we used depsets here.

go/private/context.bzl
```diff
@@ -578,14 +570,9 @@ def go_context(ctx, attr = None):
         root = goroot,
         go = binary,
         stdlib = stdlib,
```
fmeum, line 572: Yay for making this smaller. It's technically public API, but I don't want to view it as such.

go/private/mode.bzl
```diff
@@ -58,34 +58,18 @@ def mode_string(mode):
...
-        if v == "off":
-            return False
-        if v == "auto":
-            continue
-        fail("Invalid value {}".format(v))
-    fail("_ternary failed to produce a final result from {}".format(values))
-
 def get_mode(ctx, go_toolchain, cgo_context_info, go_config_info):
-    static = _ternary(go_config_info.static if go_config_info else "off")
+    static = go_config_info.static if go_config_info else False
```
fmeum, line 62: Why is this a safe change?

proto/def.bzl
```diff
@@ -90,6 +94,8 @@ def _proto_library_to_source(_go, attr, source, merge):
             merge(source, compiler[GoSource])
 
 def _go_proto_library_impl(ctx):
+    check_importpaths(ctx)
```
fmeum, line 97: This seems new?

go/private/actions/compilepkg.bzl
DavidZbarsky-at: we can try in a follow-up :)

go/private/mode.bzl
DavidZbarsky-at: I believe these values come from `//go:config:static` so they are already bools, though it's certainly possible that I am not fully following how this works. Tests seem to pass, not sure how our coverage is

proto/def.bzl
DavidZbarsky-at: Yeah so one of the things bothering me is that there are some differences between the library (base, proto, tool) and binary/test attributes/codepaths, but they both funnel into `go_context`. Then we end up having a bunch of `getattr`/`hasattr` and fallbacks to handle both of these. I wonder if it would be cleaner to have a dedicated lib_context and bin_context that have a shared shape?

Absent that, this change hoists `check_importpath` calls to the library entrypoints, because these attrs don't exist on binaries, and this allows simplifying the check function to avoid `hasattr` which speeds it up.

Without:
image

With:
image

DavidZbarsky-at:
Added another commit from testing in @jbedard's private repo;
Before: 
image

After:
image

PR #4016, opened 2024-08-08
Simplify duplicate importmap handling

dzbarsky:
Starlark cleanup and perf improvement - saves around 250ms on buildbuddy

- It's more efficient to ignore a duplicate archive inside the action vs in analysis phase
- I belive `_transitive_archives_without_test_archives` is no longer needed due to the `recompile_external_deps` machinery for tests
- This allows to remove some data from the providers as well

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -106,13 +106,11 @@ def emit_link(
     if go.mode.link == LINKMODE_PLUGIN:
         tool_args.add("-pluginpath", archive.data.importpath)
 
+    if go.coverage_enabled and go.coverdata:
+        test_archives = list(test_archives) + [go.coverdata.data]
+
     arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])
     builder_args.add_all(arcs, before_each = "-arc", map_each = _format_archive)
-
-    if (go.coverage_enabled and go.coverdata):
-        # The linker knows to exclude the coverdata archive if it's already included in `arcs`.
-        builder_args.add("-arc", _format_archive(go.coverdata.data))
-
     builder_args.add("-package_list", go.sdk.package_list)
 
     # Build a list of rpaths for dynamic libraries we need to find.
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -106,10 +106,24 @@ def emit_link(
     if go.mode.link == LINKMODE_PLUGIN:
         tool_args.add("-pluginpath", archive.data.importpath)
 
-    if go.coverage_enabled and go.coverdata:
-        test_archives = list(test_archives) + [go.coverdata.data]
+    # TODO(zbarsky): Bazel versions older than 7.2.0 do not properly deduplicate this dep
+    # Can replace with the following once we support Bazel 7.2.0+ only:
+    #    if go.coverage_enabled and go.coverdata:
+    #        test_archives = list(test_archives) + [go.coverdata.data]
+    #    arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])
 
-    arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])
+    if go.coverage_enabled and go.coverdata:
+        potentially_duplicated_arcs = depset(test_archives + [go.coverdata.data], transitive = [d.transitive for d in archive.direct]).to_list()
+        importmaps = {}
+        arcs = []
+        for arc in potentially_duplicated_arcs:
+            if arc.importmap in importmaps:
+                continue
+            importmaps[arc.importmap] = True
+            arcs.append(arc)
+    else:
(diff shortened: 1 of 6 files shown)


dzbarsky:
This one needs more work

go/private/actions/link.bzl
```diff
@@ -241,41 +212,3 @@ def _extract_extldflags(gc_linkopts, extldflags):
         else:
             filtered_gc_linkopts.append(opt)
     return filtered_gc_linkopts, extldflags
-
-def _check_conflicts(arcs):
-    importmap_to_label = {}
-    for arc in arcs:
-        if arc.importmap in importmap_to_label:
```
dzbarsky, line 248: `link.go` checks for this while computing `dedupedArchives`

go/private/actions/link.bzl
```diff
@@ -241,41 +212,3 @@ def _extract_extldflags(gc_linkopts, extldflags):
...
-	{}
-Set "importmap" to different paths or use 'bazel cquery' to ensure only one
-package with this path is linked.""".format(
-                arc.importmap,
-                importmap_to_label[arc.importmap],
-                arc.label,
-            )
-        importmap_to_label[arc.importmap] = arc.label
-    for arc in arcs:
-        for dep_importmap, dep_label in zip(arc._dep_importmaps, arc._dep_labels):
```
dzbarsky, line 260: I think we'll just get linker errors for these? seems ok?

go/private/actions/link.bzl
```diff
@@ -241,41 +212,3 @@ def _extract_extldflags(gc_linkopts, extldflags):
...
-	  was compiled with: {}
-	but was linked with: {}""".format(
-                    arc.importmap,
-                    dep_importmap,
-                    dep_label,
-                    importmap_to_label[dep_importmap],
-                )
-                if importmap_to_label[dep_importmap].name.endswith("_test"):
-                    err += """
-This sometimes happens when an external test (package ending with _test)
```
dzbarsky, line 277: I don't think this message is correct anymore

go/private/actions/link.bzl
```diff
@@ -130,12 +106,13 @@ def emit_link(
...
 
-    arcs = _transitive_archives_without_test_archives(archive, test_archives)
-    arcs.extend(test_archives)
-    if (go.coverage_enabled and go.coverdata and
-        not any([arc.importmap == go.coverdata.data.importmap for arc in arcs])):
-        arcs.append(go.coverdata.data)
+    arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])
     builder_args.add_all(arcs, before_each = "-arc", map_each = _format_archive)
+
+    if (go.coverage_enabled and go.coverdata):
```
fmeum, line 112:
```suggestion
    if go.coverage_enabled and go.coverdata:
```

go/private/actions/link.bzl
fmeum: Can we reproduce this case in an example or even a test?

go/tools/builders/link.go
```diff
@@ -51,16 +50,47 @@ func link(args []string) error {
...
+		}
+
+		if arc.packagePath == "github.com/bazelbuild/rules_go/go/tools/coverdata" {
+			// We can end up with this archive twice if the test transitively depends on coverdata.
+			// See https://github.com/bazelbuild/rules_go/issues/3017 and
+			// https://github.com/bazelbuild/rules_go/pull/3032
+			// Note that `arc.file` != `seen.file`:
+			//   _main/bazel-out/darwin_arm64-fastbuild-ST-92a08095f520/bin/external/io_bazel_rules_go/go/tools/coverdata/coverdata.a
+			//   _main/bazel-out/darwin_arm64-fastbuild/bin/external/io_bazel_rules_go/go/tools/coverdata/coverdata.a
+			// Perhaps path mapping can help here?
```
fmeum, line 77: It can help avoid the overhead, but not the potential correctness problem in case any of these is affected by different Go settings. Could you diff the configs of the two files? Maybe we should add them in a different place that isn't subject to this duplication.

dzbarsky: ok I pushed an extra commit fixing it by transitioning `coverdata` to non_go_tool cfg, wdyt?

go/tools/builders/importcfg.go
```diff
@@ -179,10 +179,21 @@ func buildImportcfgFileForLink(archives []archive, stdPackageListPath, installSu
...
+    %s
+    %s
+Set "importmap" to different paths or use 'bazel cquery' to ensure only one
+package with this path is linked.`,
+				arc.packagePath,
+				arc.importPath,
+				prevLabel)
 		}
-		depsSeen[arc.packagePath] = arc.label
+		// TODO(zbarsky): The labels are empty, and `importPath` contains the label.
```
fmeum, line 193: We can take care of this in a follow-up change.

fmeum (approved):

PR #4020, opened 2024-08-09
Optimize CcInfo computation

dzbarsky:
This shows up around 4% of buildbuddy analysis phase
image

After this change, it goes away.

It seems like Bazel is able to make this lazy, which is good since it's only used for building LINKMODE_C_ARCHIVE/LINKMODE_C_SHARED which is fairly rare.

**What type of PR is this?**
Performance improvement

Avoid computing CcInfo for every intermediate library when it's not needed

Fixes #

**Other notes for review**


go/private/rules/binary.bzl
```diff
@@ -94,6 +95,24 @@ def new_cc_import(
         ),
     )
 
+def _go_cc_aspect_impl(target, _ctx):
```
fmeum, line 98: Why `_ctx`? It's not unused.

fmeum (commented):
I'm surprised to see that this reduces CPU usage as I would have expected the aspect to be evaluated eagerly. Could you check the memory implications of having a new aspect?

go/private/rules/binary.bzl
```diff
@@ -94,6 +95,24 @@ def new_cc_import(
...
+    if GoSource in target:
```
fmeum, line 99: Let's make this an early return instead.

go/private/rules/binary.bzl
dzbarsky: oh it was leftover from when i was using the GoSource, i'll fix

dzbarsky:
> I'm surprised to see that this reduces CPU usage as I would have expected the aspect to be evaluated eagerly. Could you check the memory implications of having a new aspect?

I did as well, but I guess it must be lazy somehow? (And if it's not, I think we have a path to make it so; the `go_binary` macro already splits to 2 rules for executable vs non-executable, and we only need the aspect on the non-executable one)

I'll poke at memory usage as well

dzbarsky:
> > I'm surprised to see that this reduces CPU usage as I would have expected the aspect to be evaluated eagerly. Could you check the memory implications of having a new aspect?
> 
> I did as well, but I guess it must be lazy somehow? (And if it's not, I think we have a path to make it so; the `go_binary` macro already splits to 2 rules for executable vs non-executable, and we only need the aspect on the non-executable one)
> 
> I'll poke at memory usage as well

@fmeum so I tried `bazel shutdown && bazel build --nobuild //... && bazel info used-heap-size-after-gc` on buildbuddy repo. I saw 215-216MB as baseline, and 220MB with this PR. I made the aspect only apply to non-executable binary and now it is down to 213-214MB, so it's actually an improvement :)

fmeum (approved):
Great job!

PR #4022, opened 2024-08-09
Tweak how we pass GOEXPERIMENT to actions

dzbarsky:
**What type of PR is this?**
Small cleanup. We can just encode the string in the repo rule instead of creating it for every go action.

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -62,9 +62,9 @@ go_sdk(
 go_tool_binary(
     name = "builder",
     srcs = ["@io_bazel_rules_go//go/tools/builders:builder_srcs"],
+    exec_compatible_with = {exec_compatible_with},
     ldflags = "-X main.rulesGoStdlibPrefix={}".format(RULES_GO_STDLIB_PREFIX),
     sdk = ":go_sdk",
-    exec_compatible_with = {exec_compatible_with},
 )
 
 non_go_reset_target(
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -350,7 +350,7 @@ def _collect_cc_infos(deps, cdeps):
 def _check_binary_dep(go, dep, edge):
     """Checks that this rule doesn't depend on a go_binary or go_test.
 
-    go_binary and go_test may return provides with useful information for other
+    go_binary and go_test may return providers with useful information for other
     rules (like go_path), but go_binary and go_test may not depend on other
     go_binary and go_binary targets. Their dependencies may be built in
     different modes, resulting in conflicts and opaque errors.
@@ -483,7 +483,7 @@ def go_context(ctx, attr = None):
     env = {
         "GOARCH": mode.goarch,
         "GOOS": mode.goos,
-        "GOEXPERIMENT": ",".join(toolchain.sdk.experiments),
+        "GOEXPERIMENT": toolchain.sdk.experiments,
         "GOROOT": goroot,
         "GOROOT_FINAL": "GOROOT",
         "CGO_ENABLED": "0" if mode.pure else "1",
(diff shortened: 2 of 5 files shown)


go/private/BUILD.sdk.bazel
```diff
@@ -47,7 +47,7 @@ filegroup(
 go_sdk(
     name = "go_sdk",
     srcs = [":srcs"],
-    experiments = {experiments},
+    experiments = "{experiments}",
```
fmeum, line 50: Let's keep using `repr` on the joined string so that we don't need to manually add quotes.

dzbarsky: This matches `goarch`, `goos`, and `version`. I'm fine with using `repr` but we should we swap them all for consistency? (And then it's a bit unclear what to do with `go = "bin/go{exe}",`

fmeum: `repr` is preferable for `experiments` as its value is directly set by users and may thus mess with manual escaping. We can leave the other attributes as is for now or also move them to `repr`. `{exe}` could be replaced with logic in the repo rules.

dzbarsky: gotcha, I noticed we were already using `repr` on `exec_compatible_with` so I only changed `experiments` for now; `goos` and `gooarch` are also used to build paths for a glob

fmeum (approved):

jtacoma:
Is this the right place to ask for suggestions about how to deal with this change? Asking because it is a backwards-incompatible change. I understand this is allowed between releases that share the same major version number if that number is zero, as it was in the previous v0.50.1 and following v0.51.0 releases. For leaf monorepos this might not be a real problem because they can switch from list-of-strings to comma-delimited-string in the same commit that upgrades from v0.50.1 to v0.51.0 (or later). Unfortunately, that's not an option for repos that (like this one) are used in many _other_ repos.

For example, I'm trying out Bazel + Nix + Go in a hobby project where I ran into this issue when using [rules_nixpkgs](https://github.com/tweag/rules_nixpkgs) because its toolchains/go/go.bzl uses experiments. I considered sending a patch to update to the new API before realizing that this would force all users of rules_nixpkgs to upgrade rules_go to v0.51.0 or later. Any idea how to bridge this gap with minimal churn?

fmeum:
@jtacoma Thanks for raising this. What's the part of this PR that is backwards incompatible? My understanding is that it only modifies how attributes are passed around internally.

jtacoma:
I was a little confused too since I noticed `go_sdk` is defined and used only within the "private" subdirectory of rules_go. However, it is also [public in go/def.bzl](https://github.com/bazel-contrib/rules_go/blob/00e2be6396f2609aad37cb02cc12e61d68dd5155/go/def.bzl#L133). I ran into this because [rules_nixpkgs loads go_sdk from `@rules_go//go:def.bzl`](https://github.com/tweag/rules_nixpkgs/blob/4b193df3b1a03415db74628944beb05c3320aae3/toolchains/go/go.bzl#L72).

dzbarsky:
> I was a little uncertain since I noticed `go_sdk` is defined and used only within the "private" subdirectory of rules_go. However, it is also [public in go/def.bzl](https://github.com/bazel-contrib/rules_go/blob/00e2be6396f2609aad37cb02cc12e61d68dd5155/go/def.bzl#L133). I ran into this because [rules_nixpkgs loads go_sdk from `@rules_go//go:def.bzl`](https://github.com/tweag/rules_nixpkgs/blob/4b193df3b1a03415db74628944beb05c3320aae3/toolchains/go/go.bzl#L72).

Ah, we didn't realize that was exported. We should probably turn that export into a macro that rewrites the list into a comma-separated string to keep it backwards compatible.

fmeum:
@dzbarsky We could also handle the conversion in the go_sdk rule and don't change its attributes.

dzbarsky:
Yep, makes sense. I can send a PR Thursday to restore the interface

PR #4023, opened 2024-08-10
Tweak nogo archive handling

dzbarsky:
This is one of the sources of feeding structs into `deps`.

**What type of PR is this?**
Starlark cleanup

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/rules/nogo.bzl
+++ b/go/private/rules/nogo.bzl
@@ -46,8 +46,9 @@ def _nogo_impl(ctx):
     if ctx.attr.debug:
         nogo_args.add("-debug")
     nogo_inputs = []
-    for dep in ctx.attr.deps:
-        nogo_args.add("-analyzer_importpath", dep[GoArchive].data.importpath)
+    analyzer_archives = [dep[GoArchive] for dep in ctx.attr.deps]
+    analyzer_importpaths = [archive.data.importpath for archive in analyzer_archives]
+    nogo_args.add_all(analyzer_importpaths, before_each = "-analyzer_importpath")
     if ctx.file.config:
         nogo_args.add("-config", ctx.file.config)
         nogo_inputs.append(ctx.file.config)
@@ -74,7 +75,7 @@ def _nogo_impl(ctx):
     nogo_source = go.library_to_source(go, struct(
         srcs = [struct(files = [nogo_main])],
         embed = [ctx.attr._nogo_srcs],
-        deps = ctx.attr.deps,
+        deps = analyzer_archives,
     ), nogo_library, False)
     _, executable, runfiles = go.binary(
         go,
--- a/go/private/rules/nogo.bzl
+++ b/go/private/rules/nogo.bzl
@@ -25,7 +25,6 @@ load(
     "EXPORT_PATH",
     "GoArchive",
     "GoLibrary",
-    "get_archive",
 )
 load(
     "//go/private/rules:transition.bzl",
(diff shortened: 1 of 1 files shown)


go/private/rules/nogo.bzl
```diff
@@ -76,7 +74,7 @@ def _nogo_impl(ctx):
     nogo_source = go.library_to_source(go, struct(
         srcs = [struct(files = [nogo_main])],
         embed = [ctx.attr._nogo_srcs],
-        deps = analyzer_archives,
+        deps = ctx.attr.deps,
```
fmeum, line 77: Doesn't this have a different type (list of `Target`s instead of list of providers)? How does this work?

go/private/rules/nogo.bzl
```diff
@@ -47,9 +46,8 @@ def _nogo_impl(ctx):
     if ctx.attr.debug:
         nogo_args.add("-debug")
     nogo_inputs = []
-    analyzer_archives = [get_archive(dep) for dep in ctx.attr.deps]
-    analyzer_importpaths = [archive.data.importpath for archive in analyzer_archives]
-    nogo_args.add_all(analyzer_importpaths, before_each = "-analyzer_importpath")
+    for dep in ctx.attr.deps:
+        nogo_args.add("-analyzer_importpath", dep[GoArchive].data.importpath)
```
fmeum, line 50: Is this really an improvement memory-wise?

**Old version**
* reference to a vector arg
* reference to `"-analyzer_importpath"`
* number of import paths
* for each import path
  * reference to import path

`12 + 4*N` bytes

**New version**
* for each import path
  * reference to `"-analyzer_importpath"`
  * reference to import path

`8*N` bytes

dzbarsky: Doesn't the old version also have an extra `8 + 4*N` bytes for the `analyzer_importpaths` list as well? Or does that not matter because it's not retained?

go/private/rules/nogo.bzl
dzbarsky: The code can handle both; I think it's nicer to pass the `Target`s and then grab the providers as needed; code like `srcs = [struct(files = [nogo_main])],` seems a bit roundabout. Is there a performance reason not to do that?

go/private/rules/nogo.bzl
fmeum: It's not retained. `add_all` drains all lists (but not depsets) into its internal array. This makes it more efficient than related `add` calls (unless you use none of the keyword args).

go/private/rules/nogo.bzl
fmeum: I don't know whether there is a memory cost to retaining targets here. Probably not as Bazel retains them elsewhere. But it's definitely unusual and a bit of a code smell as it allows downstream consumers to read arbitrary (possibly even non-Go providers). It also won't play with Starlark types. 

If it doesn't matter much, let's keep the single provider in this field.

go/private/rules/nogo.bzl
dzbarsky: ah ok, I tried one more variant with the `map_each` wdyt?

go/private/rules/nogo.bzl
dzbarsky: that's fair; the part that I don't like is that the code ends up handling both `Target` and provider list. It seems it will need to change either way to support typing, though maybe it should stop supporting Target and push the extraction into all the callers. I'll revert this part for now; we'll need to change it anyway if we refactor the provider setup.

go/private/rules/nogo.bzl
fmeum: The additional memory usage is constant, but I don't see this part as an improvement to code structure or readability. `map_each` should be used only when it avoids retaining certain objects, but the import paths are already retained. How about just inlining the `get_archive` call?

dzbarsky: ah, I thought it would push work out of the analysis phase. I'll just simplify it in that case

fmeum (approved):

PR #4027, opened 2024-08-11
Mark deps explicitly added with `//go get` as direct

fmeum:
**What type of PR is this?**

Feature

When running `go get example.com/foo@v1.2.3`, the module `example.com/foo` will be added to `go.mod` with an `// indirect` comment, just like its transitive dependencies. This results in `go_deps` not being able to distinguish this explicitly requested new dep from other indirect deps that shouldn't be imported by the root module. Running `go mod tidy` or `go get` without arguments is required to have the comment removed after adding a reference to the new module in code.

This change makes it so that `bazel run @rules_go//go get example.com/foo@v1.2.3` marks the requested module as a direct dependency. This realizes https://github.com/golang/go/issues/68593 in our custom wrapper, thus making this command the only one needed to add a new module dependency and have it work with a subsequent Bazel build (assuming that also updates the BUILD files with Gazelle). This [matches the behavior of `gopls` when adding a new dependency](https://github.com/golang/tools/blob/ec1a81bfec7cc6563474b7aa160db30df9bfce6b/gopls/internal/server/command.go#L805-L809).

### DIFF
--- a/go/tools/go_bin_runner/main.go
+++ b/go/tools/go_bin_runner/main.go
@@ -20,7 +20,7 @@ var GoBinRlocationPath = "not set"
 var ConfigRlocationPath = "not set"
 var HasBazelModTidy = "not set"
 
-var bazelWorkingDir = os.Getenv("BUILD_WORKSPACE_DIRECTORY")
+var bazelWorkingDir = os.Getenv("BUILD_WORKING_DIRECTORY")
 var bazelWorkspaceDir = os.Getenv("BUILD_WORKSPACE_DIRECTORY")
 
 // Produced by gazelle's go_deps extension.
--- a/go/tools/go_bin_runner/main.go
+++ b/go/tools/go_bin_runner/main.go
@@ -20,6 +20,9 @@ var GoBinRlocationPath = "not set"
 var ConfigRlocationPath = "not set"
 var HasBazelModTidy = "not set"
 
+var bazelWorkingDir = os.Getenv("BUILD_WORKSPACE_DIRECTORY")
+var bazelWorkspaceDir = os.Getenv("BUILD_WORKSPACE_DIRECTORY")
+
 // Produced by gazelle's go_deps extension.
 type Config struct {
 	GoEnv     map[string]string `json:"go_env"`
@@ -48,10 +51,14 @@ func main() {
 	}
 
 	args := append([]string{goBin}, os.Args[1:]...)
-	cwd := os.Getenv("BUILD_WORKING_DIRECTORY")
-	err = runProcess(args, env, cwd)
-	if err != nil {
+	if err = runProcess(args, env); err != nil {
 		log.Fatal(err)
+	}
+
+	if len(os.Args) > 1 && os.Args[1] == "get" {
+		if err = markRequiresAsDirect(goBin, os.Args[2:]); err != nil {
+			log.Fatal(err)
+		}
 	}
 
 	hashesAfter, err := hashWorkspaceRelativeFiles(cfg.DepsFiles)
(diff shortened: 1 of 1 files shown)


go/tools/go_bin_runner/main.go
```diff
@@ -20,6 +20,9 @@ var GoBinRlocationPath = "not set"
 var ConfigRlocationPath = "not set"
 var HasBazelModTidy = "not set"
 
+var bazelWorkingDir = os.Getenv("BUILD_WORKSPACE_DIRECTORY")
```
sluongng, line 23:
```suggestion
var bazelWorkingDir = os.Getenv("BUILD_WORKING_DIRECTORY")
```

sluongng (approved):
LGTM.

linzhp:
I am not sure if we should diverge from the official `go get` behavior. The workflow we recommends in Uber is to add a (potentially dummy) import to the code before running `go mod tidy`. This makes sure that people don't add new deps to the repo without using it

go/tools/go_bin_runner/main.go
```diff
@@ -48,12 +51,16 @@ func main() {
 	}
```
tyler-french, line 51: Random stylistic thing: 

I would be good to separate out the logic into a `run()` function that returns an error, so that we only need to call `log.Fatal` once. Then we can test this stuff in the future:

```
func main() {
  if err := run(os.Args, os.Stderr); err != nil {
    log.Fatal(err)
  }
}

func run(args []string, stderr io.Writer) error {
  .. everything else
}
```

go/tools/go_bin_runner/main.go
```diff
@@ -20,6 +20,9 @@ var GoBinRlocationPath = "not set"
 var ConfigRlocationPath = "not set"
 var HasBazelModTidy = "not set"
 
+var bazelWorkingDir = os.Getenv("BUILD_WORKING_DIRECTORY")
+var bazelWorkspaceDir = os.Getenv("BUILD_WORKSPACE_DIRECTORY")
```
tyler-french, line 24: There's an `os.Getenv("BAZEL")` call below. Should we maybe extract this to a function loadEnv() that returns a structure and is called once?

go/tools/go_bin_runner/main.go
```diff
@@ -115,12 +121,76 @@ func getGoEnv(goBin string, cfg Config) ([]string, error) {
...
+	}
+
+	if len(editArgs) > 0 {
+		_, _ = fmt.Fprintln(os.Stderr, "rules_go: Marking requested modules as direct dependencies...")
+		cmd = exec.Command(goBin, append([]string{"mod", "edit"}, editArgs...)...)
+		cmd.Dir = bazelWorkingDir
+		err = cmd.Run()
+		if err != nil {
+			return err
+		}
```
tyler-french, line 184:
```suggestion
		if err := cmd.Run(); err != nil {
			return err
		}
```

tyler-french (approved):

fmeum:
> I am not sure if we should diverge from the official go get behavior. The workflow we recommends in Uber is to add a (potentially dummy) import to the code before running go mod tidy. This makes sure that people don't add new deps to the repo without using it

I also don't like this deviation (especially now that the Go issue has been closed), but note that we are already deviating in an aspect that's arguably worse: If you `go get` a dep, start to use it in code and then `go build ./...`, this will pass. If you `bazel run @rules_go//go get` a dep, start to use it in code and then `bazel build //...` (including running Gazelle), you will get an error due to the dep not having been added as explicit.

linzhp (approved):
I see. Thanks for the explanation

PR #4028, opened 2024-08-11
Remove mode from GoArchive

dzbarsky:
**What type of PR is this?**
Cleanup

Looks like we can use the mode from the `source` instead and reduce the chance of inconsistencies

Fixes #

**Other notes for review**


go/private/actions/archive.bzl
```diff
@@ -212,5 +212,4 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         cgo_deps = depset(transitive = [cgo_deps] + [a.cgo_deps for a in direct]),
         cgo_exports = cgo_exports,
         runfiles = runfiles,
-        mode = go.mode,
```
fmeum, line 215: Please also update the providers.rst file

dzbarsky: thanks, done

fmeum (approved):

PR #4032, opened 2024-08-12
Cleanup GoArchiveData usage of as_tuple

dzbarsky:
**What type of PR is this?**
Starlark cleanup

`as_tuple` was showing up in profiles, we can just call `tuple` directly. This exposed that we were flattening a depset which isn't necessary.

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -13,10 +13,6 @@
 # limitations under the License.
 
 load(
-    "//go/private:common.bzl",
-    "as_tuple",
-)
-load(
     "//go/private:context.bzl",
     "get_nogo",
 )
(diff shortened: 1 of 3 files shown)


go/private/actions/archive.bzl
```diff
@@ -171,19 +167,19 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         pathtype = source.library.pathtype,
 
         # GoSource fields
-        srcs = as_tuple(source.srcs),
-        orig_srcs = as_tuple(source.orig_srcs),
-        _cover = as_tuple(source.cover),
-        _embedsrcs = as_tuple(source.embedsrcs),
+        srcs = tuple(source.srcs),
```
fmeum, line 170: Do we even need these `tuple` calls? Provider contents are frozen when the target has been evaluated, so we wouldn't need to worry about immutability.

fmeum (approved):

go/private/actions/archive.bzl
dzbarsky: hmm are you sure? Here's what I see without `tuple`:
```
	File "/Users/dzbarsky/rules_go/go/private/rules/library.bzl", line 40, column 25, in _go_library_impl
	File "/Users/dzbarsky/rules_go/go/private/actions/archive.bzl", line 212, column 28, in emit_archive
		transitive = depset([data], transitive = [a.transitive for a in direct]),
Error in depset: depset elements must not be mutable values
```

fmeum: Right, didn't think of `depset`. The Java rules use complex logic where they join all args into an escaped string and then store that in a depset only to tokenize it later but I would prefer not to do that here.

fmeum:
There are conflicts now.

dzbarsky:
> Some other merge resulted in conflicts.

Rebased!

fmeum (approved):

PR #4034, opened 2024-08-12
Remove some fields from Go context

dzbarsky:
**What type of PR is this?**
Starlark cleanup

All of these are either unused or can easily be retrieved from another context field.
We were already retrieving them from both so this change makes things more consistent.

Fixes #

**Other notes for review**


sluongng:
Im a bit iffy about this.

On one hand, this is a nice cleanup, probably overdue.

On the other hand, we have been advertising Go context wrapper as the canonical way to extend rules_go. I can see some downstream rules depending on `go_ctx.go` to access the binary directly for example.

Perhaps leaving the removed fields in the source as comments, each pointing to their replacement would help folks migrate more easily? WDYT?

dzbarsky:
> Im a bit iffy about this.
> 
> On one hand, this is a nice cleanup, probably overdue.
> 
> On the other hand, we have been advertising Go context wrapper as the canonical way to extend rules_go. I can see some downstream rules depending on `go_ctx.go` to access the binary directly for example.
> 
> Perhaps leaving the removed fields in the source as comments, each pointing to their replacement would help folks migrate more easily? WDYT?

That's a good idea. If we want to avoid breakage, we can also add a deprecated_props kwarg or otherwise shim things so that internal rules get a slimmed down context and external world has a chance to migrate

fmeum:
I'm also in favor of cleaning up rules_go's own usages while still providing the same props, but moved to a separate "deprecated" section in the docs. We can call out their imminent removal in the release notes.

dzbarsky:
> I'm also in favor of cleaning up rules_go's own usages while still providing the same props, but moved to a separate "deprecated" section in the docs. We can call out their imminent removal in the release notes.

I've updated to allow opting out of deprecated properties and added to the docs. WDYT of this approach?

fmeum (approved):
This is great, thanks for simplifying the migration!

PR #4037, opened 2024-08-13
Move PATH env-var computation to GoContextInfo

dzbarsky:
**What type of PR is this?**
Starlark cleanup

We can compute the PATH once instead of in every go action. We check for `PATH` in the existing env but since the existing env came from hardcoded dict and cgo_context_info.env, this should be equivalent. 

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -516,41 +516,13 @@ def go_context(ctx, attr = None):
     if mode.arm:
         env["GOARM"] = mode.arm
 
-    if not cgo_context_info:
-        cc_toolchain_files = depset()
-        cgo_tools = None
-    else:
+    if cgo_context_info:
         env.update(cgo_context_info.env)
         cc_toolchain_files = cgo_context_info.cc_toolchain_files
-
-        # Add C toolchain directories to PATH.
-        # On ARM, go tool link uses some features of gcc to complete its work,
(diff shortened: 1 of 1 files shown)


go/private/context.bzl
```diff
@@ -827,6 +799,34 @@ def _cgo_context_data_impl(ctx):
         cc_toolchain.target_gnu_system_name,
     )
 
+    # Add C toolchain directories to PATH.
```
fmeum, line 802: I'm planning to get rid of `cgo_context` in favor of resolving the toolchain information for each individual target. Should I take this change to mean that this could have a relevant impact on analysis time?

dzbarsky: This change is measurable but not super impactful. Perhaps 50-100ms on buildbuddy repo. But if we do more toolchain resolution per target, it might have more impact. Why do you want to go that route?

fmeum (approved):

go/private/context.bzl
fmeum: We can still merge this though as it improves the situation and doesn't make this future change more difficult.

dzbarsky: Ah ok, the desire makes sense. Let's profile with your change once it's ready and see how things look

PR #4038, opened 2024-08-13
Remove orig_srcs

dzbarsky:
**What type of PR is this?**
Starlark cleanup

The documentation says it can differ from `srcs` "if processing tools such as cgo or cover are applied." but in practice I did not see it mutating in the code; I suspect this may have been true in an earlier iteration of the rules. I also put in an assert in the provider creation code if they are different at the end and it didn't fire.

Fixes #

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -50,23 +50,20 @@ def file_path(f):
     return paths.join(prefix, f.path)
 
 def _go_archive_to_pkg(archive):
+    go_files = [
+        file_path(src)
+        for src in archive.data.srcs
+        if src.path.endswith(".go")
+    ]
     return struct(
         ID = str(archive.data.label),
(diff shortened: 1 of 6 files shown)


go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -50,23 +50,20 @@ def file_path(f):
...
-            file_path(src)
-            for src in archive.data.orig_srcs
-            if src.path.endswith(".go")
-        ],
-        CompiledGoFiles = [
-            file_path(src)
-            for src in archive.data.srcs
-            if src.path.endswith(".go")
-        ],
+        GoFiles = go_files,
```
dzbarsky, line 62: I wasn't sure if these are both needed, but seems safer to keep both

fmeum:
This has conflicts

dzbarsky:
> This has conflicts

Rebased!

fmeum (approved):

PR #4044, opened 2024-08-15
Remove GoSource/GoLibrary from test/binary rules

dzbarsky:
**What type of PR is this?**
Starlark cleanup

We don't want go_binary or go_test targets to be included in `deps` or `embed`. We have a check for it today, but we can push this into the starlark type system by ensuring they don't have these providers.

Fixes #

**Other notes for review**


fmeum:
I am concerned that this will make it more difficult for other rules to interop with Go rules. Could we instead make it so that `go_binary` doesn't emit any Go providers?

tests/core/starlark/provider_tests.bzl
```diff
@@ -0,0 +1,36 @@
...
+    target = analysistest.target_under_test(env)
+    asserts.false(env, GoSource in target, "Should not have GoSource")
+    asserts.false(env, GoLibrary in target, "Should not have GoLibrary")
+
+    return analysistest.end(env)
+
+providers_test = analysistest.make(_providers_test_impl)
+
+def provider_test_suite():
+    go_binary(
```
dzbarsky, line 18: I tried to test this more directly like so:
```
def _providers_test_impl(ctx):
    env = analysistest.begin(ctx)
    asserts.expect_failure(env, "not allowed")
    return analysistest.end(env)

providers_test = analysistest.make(_providers_test_impl)

def provider_test_suite():
    go_binary(
        name = "go_binary",
        tags = ["manual"],
    )

    go_library(
        name = "binary_lib",
        deps = [":go_binary"],
        tags = ["manual"],
    )

    providers_test(
        name = "go_binary_providers_test",
        target_under_test = ":binary_lib",
    )
```

but that results in an error in analysis itself, not in test phase:
```
ERROR: /Users/dzbarsky/rules_go/tests/core/starlark/BUILD.bazel:10:20: in deps attribute of go_library rule //tests/core/starlark:binary_lib: '//tests/core/starlark:go_binary' does not have mandatory providers: 'GoLibrary'. Since this rule was created by the macro 'provider_test_suite', the error might have been caused by the macro implementation
```

any ideas how to do this better?

dzbarsky:
@fmeum ok I like your suggestion better, thanks. PTAL

fmeum (approved):

PR #4048, opened 2024-08-16
Cleanup _merge_embed

dzbarsky:
**What type of PR is this?**
Starlark cleanup

Since only one the lib or one of the embeds can specify cgo, we can simplify this a bit

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -222,14 +222,16 @@ def _merge_embed(source, embed):
     source["x_defs"].update(s.x_defs)
     source["gc_goopts"] = source["gc_goopts"] + s.gc_goopts
     source["runfiles"] = source["runfiles"].merge(s.runfiles)
-    if s.cgo and source["cgo"]:
-        fail("multiple libraries with cgo enabled")
-    source["cgo"] = source["cgo"] or s.cgo
-    source["cdeps"] = source["cdeps"] or s.cdeps
-    source["cppopts"] = source["cppopts"] or s.cppopts
-    source["copts"] = source["copts"] or s.copts
-    source["cxxopts"] = source["cxxopts"] or s.cxxopts
-    source["clinkopts"] = source["clinkopts"] or s.clinkopts
+
+    if s.cgo:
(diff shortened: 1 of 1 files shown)


go/private/context.bzl
```diff
@@ -259,9 +261,7 @@ def _library_to_source(go, attr, library, coverage_instrumented):
     srcs = attr_srcs + generated_srcs
     embedsrcs = [f for t in getattr(attr, "embedsrcs", []) for f in as_iterable(t.files)]
     attr_deps = getattr(attr, "deps", [])
-    generated_deps = getattr(library, "deps", [])
-    deps = attr_deps + generated_deps
-    deps = [get_archive(dep) for dep in deps]
+    deps = [get_archive(dep) for dep in attr_deps]
```
fmeum, line 264: Why is this change safe?

dzbarsky: currently the only supported way to make a `GoLibrary` is with `new_library`, which does not have a `deps` attribute. While it does take a `kwargs` and assign anything from it into the provider, `deps` is not documented in either `GoLibrary` (providers.rst) or `new_library` (toolchains.rst), so I consider this to be relying on undocumented behavior.

In terms of impact, in order to be affected by this change, code would need to:
1. Be passing `ctx.attr` to `attr` in `_library_from_source` but having additonal deps that are not in `ctx.attr.deps`
2. Be passing a custom struct to `attr` in `_library_from_source` and wanting to have deps.

The 2nd case is trivially fixable by passing deps, and the first case is also fixable by switching to constructing a struct. It seems like a fairly contrived scenario though.

We could make the workaround even simpler by taking an explicit `deps` in `_library_to_source` and falling back to `getattr` if it's not passed?

dzbarsky: I'm not sure why this ended up in this PR, I can split to a separate one if you prefer

fmeum:
This has conflicts

dzbarsky:
> This has conflicts

rebased!

fmeum (approved):

PR #4050, opened 2024-08-16
builder: work around change in filepath.EvalSymlinks behavior

### LINKED ISSUE #4049: Go 1.23.0 Windows build fails with "stdlib: The system cannot find the path specified."

0.49.0

Also reproduced at `master`, which is currently `520bee5039bcaf831ed3e79c513f35f629c04624`.

n/a

7.3.0

yes

Windows / amd64

n/a

### What did you do?

1. Create a minimal workspace, as below (use [txtar](https://pkg.go.dev/golang.org/x/exp/cmd/txtar) to extract).
2. `bazel build //:hello`

```
-- WORKSPACE --
-- MODULE.bazel --
module(
    name = "scratch",
    version = "0.0.0",
)

bazel_dep(
    name = "rules_go",
    version = "0.49.0",
)

go_sdk = use_extension("@rules_go//go:extensions.bzl", "go_sdk")
go_sdk.download(
    name = "go_sdk",
    version = "1.23.0",
)

use_repo(go_sdk, "go_sdk")
-- BUILD --
load("@rules_go//go:def.bzl", "go_library")

go_library(
    name = "hello",
    srcs = ["hello.go"],
    importpath = "example.com/hello",
)
-- hello.go --
package hello
```

Successful build

### What did you see instead?

```
ERROR: C:/_b/fhie23dh/external/rules_go~/BUILD.bazel:42:7: GoStdlib external/rules_go~/stdlib_/pkg failed: (Exit 1): builder.exe failed: error executing GoStdlib command (from target @@rules_go~//:stdlib) bazel-out\x64_windows-opt-exec-ST-d57f47055a04\bin\external\rules_go~~go_sdk~go_sdk\builder_reset\builder.exe stdlib -sdk external/rules_go~~go_sdk~go_sdk -goroot external/rules_go~~go_sdk~go_sdk ... (remaining 8 arguments skipped)
stdlib: The system cannot find the path specified.
Target //:hello failed to build                                                                                                                         
```

jayconrod:
**What type of PR is this?**

> Bug fix

filepath.EvalSymlinks works differently on Windows in Go 1.23.0: mount points and reparse points are no longer treated as symbolic links by os.Stat. Bazel uses junctions (implemented as reparse points) for inputs, so this means EvalSymlinks can fail with some kinds of inputs where it didn't before.

replicateTree calls EvalSymlinks before constructing an input tree for `GoStdlib`. This seems to be necessary in some cases, though I'm not altogether sure why. EvalSymlinks fails due to the new behavior though.

This change sets `GODEBUG=winsymlink=0` when invoking the builder binary to revert to the old behavior. This should not affect the behavior of user code compiled using the builder. This may not be a permanent solution, but it should work at least through Go 1.25.

Fixes #4049

**Other notes for review**

No test, but verified locally. Simply running `GoStdlib` with Go 1.23.0 will cover this. rules_go currently builds with Go 1.21.8 (no longer a version supported by Google!).

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -486,6 +486,14 @@ def go_context(
         #
         # See https://go.dev/doc/toolchain for more info.
         "GOTOOLCHAIN": "local",
+
+        # NOTE(#4049): Since Go 1.23.0, os.Readlink (and consequently
+        # filepath.EvalSymlinks) stopped treating Windows mount points and
+        # reparse points (junctions) as symbolic links. Bazel uses junctions
+        # when constructing exec roots, and we use filepath.EvalSymlinks in
+        # GoStdlib, so this broke us. Setting GODEBUG=winsymlink=0 restores
+        # the old behavior.
+        "GODEBUG": "winsymlink=0",
     }
 
     # Path mapping can't map the values of environment variables, so we pass GOROOT to the action
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -486,14 +486,6 @@ def go_context(
         #
         # See https://go.dev/doc/toolchain for more info.
         "GOTOOLCHAIN": "local",
-
-        # NOTE(#4049): Since Go 1.23.0, os.Readlink (and consequently
-        # filepath.EvalSymlinks) stopped treating Windows mount points and
-        # reparse points (junctions) as symbolic links. Bazel uses junctions
-        # when constructing exec roots, and we use filepath.EvalSymlinks in
-        # GoStdlib, so this broke us. Setting GODEBUG=winsymlink=0 restores
-        # the old behavior.
-        "GODEBUG": "winsymlink=0",
     }
 
     # Path mapping can't map the values of environment variables, so we pass GOROOT to the action
--- a/go/tools/builders/replicate.go
+++ b/go/tools/builders/replicate.go
@@ -30,13 +30,15 @@ const (
 	softlinkMode
 )
 
-type replicateOption func(*replicateConfig)
-type replicateConfig struct {
-	removeFirst bool
-	fileMode    replicateMode
-	dirMode     replicateMode
-	paths       []string
-}
+type (
+	replicateOption func(*replicateConfig)
+	replicateConfig struct {
+		removeFirst bool
+		fileMode    replicateMode
+		dirMode     replicateMode
+		paths       []string
+	}
+)
 
 func replicatePaths(paths ...string) replicateOption {
 	return func(config *replicateConfig) {
@@ -128,11 +130,6 @@ func replicateDir(src, dst string, config *replicateConfig) error {
 func replicateTree(src, dst string, config *replicateConfig) error {
 	if err := os.RemoveAll(dst); err != nil {
 		return fmt.Errorf("Failed to remove file at destination %s: %v", dst, err)
-	}
-	if l, err := filepath.EvalSymlinks(src); err != nil {
-		return err
-	} else {
-		src = l
 	}
 	if s, err := os.Stat(src); err != nil {
 		return err
--- a/go/tools/builders/replicate.go
+++ b/go/tools/builders/replicate.go
@@ -30,15 +30,13 @@ const (
 	softlinkMode
 )
 
-type (
-	replicateOption func(*replicateConfig)
-	replicateConfig struct {
-		removeFirst bool
-		fileMode    replicateMode
-		dirMode     replicateMode
-		paths       []string
-	}
-)
+type replicateOption func(*replicateConfig)
+type replicateConfig struct {
+	removeFirst bool
+	fileMode    replicateMode
+	dirMode     replicateMode
+	paths       []string
+}
 
 func replicatePaths(paths ...string) replicateOption {
 	return func(config *replicateConfig) {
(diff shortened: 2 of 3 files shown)


go/tools/builders/replicate.go
```diff
@@ -30,13 +30,15 @@ const (
...
 )
 
-type replicateOption func(*replicateConfig)
-type replicateConfig struct {
-	removeFirst bool
-	fileMode    replicateMode
-	dirMode     replicateMode
-	paths       []string
-}
+type (
```
jayconrod, line 33: It looks like this repo uses gofumpt, set in .vscode/settings.json. I was surprised it made this change, but maybe it's intentional? 🤷

go/tools/builders/replicate.go
```diff
@@ -129,11 +131,6 @@ func replicateTree(src, dst string, config *replicateConfig) error {
 	if err := os.RemoveAll(dst); err != nil {
 		return fmt.Errorf("Failed to remove file at destination %s: %v", dst, err)
 	}
-	if l, err := filepath.EvalSymlinks(src); err != nil {
```
fmeum, line 132: CI is failing pretty badly and I don't understand all of the failures. If this is too weird, we could fix this for now by setting `GODEBUG=winsymlink=0` during compilation.

jayconrod: I'll take a look at the failures today. `GODEBUG=winsymlink=0` might be a good backup option, as you say.

Incidentally, is there a reference for the CI images? I've sometimes found rules_go CI failures very hard to reproduce locally, especially on Windows.

fmeum: You could try to run https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/setup-windows.ps1 in a Windows VM.

jayconrod: > You could try to run https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/setup-windows.ps1 in a Windows VM.

Thanks, that's probably what I need.

jayconrod:
I've rolled back the changes to `replicate.go` and switched to `GODEBUG=winsymlink=0`. That worked at least for `no_srcs_test` when using Go 1.23.0 in `WORKSPACE` (not part of this PR).

The error before sounded like we had opened a directory then called `io.Copy` on it. I started debugging but ran out of time today (lot of plates spinning right now). Probably better to go with a smaller workaround than to attempt a deeper change in `replicate.go`.

go/private/context.bzl
```diff
@@ -439,6 +439,14 @@ def go_context(
...
         # See https://go.dev/doc/toolchain for more info.
         "GOTOOLCHAIN": "local",
+
+        # NOTE(#4049): Since Go 1.23.0, os.Readlink (and consequently
+        # filepath.EvalSymlinks) stopped treating Windows mount points and
+        # reparse points (junctions) as symbolic links. Bazel uses junctions
+        # when constructing exec roots, and we use filepath.EvalSymlinks in
+        # GoStdlib, so this broke us. Setting GODEBUG=winsymlink=0 restores
+        # the old behavior.
+        "GODEBUG": "winsymlink=0",
```
fmeum, line 449: Wouldn't this also apply to compilation of user code? If we only need it for the builder binary, maybe we could move it closer to that.

jayconrod: Do you have an idea on how this should work?

I'm actually not familiar with the `GODEBUG` mechanism. I thought it mainly applied at run-time, so setting it here would only affect rules_go actions. But [Go, Backwards Compatibility, and GODEBUG](https://go.dev/doc/godebug) says this:

> When a GODEBUG setting is not listed in the environment variable, its value is derived from three sources: the defaults for the Go toolchain used to build the program, amended to match the Go version listed in go.mod, and then overridden by explicit //go:debug lines in the program.

This is a bit vague: I'm not sure if "the Go toolchain" here means the compiler, linker, std, or `go` tool. `go.mod` doesn't apply here. I'm also not sure if the `//go:debug` lines are interpreted by the `go` tool or compiler.

I'm guessing we should set `GODEBUG=winsymlink=0` in the scripts run by `go_tool_binary`, since that builds the builder binary and the `std` packages it depends on, especially `runtime`, which seems to be responsible for parsing `GODEBUG`. If I understand correctly, that would have no side effect on the `GoStdlib` action (other than working around the bug) or on compiled user code. Does that sound workable?

fmeum: Yes, that's my understanding as well: we should only set the variable when building the builder (in `go_tool_binary`).

jayconrod: I think this works: I've added a `//go:debug winsymlink=0` directive in builder.go. These directives can be added to `main` packages and control the default for that binary.

Setting `GODEBUG=winsymlink=0` in the `go_tool_binary` scripts didn't seem to work: that only seems to apply at run-time, not compile-time.

fmeum (approved):

go/private/context.bzl
jayconrod: Nope, does not work. Versions of Go before 1.23.0 error on `//go:debug winsymlink=0` because of the unknown build constraint.

jayconrod: I suspect the *only* way to make this work is to set `GODEBUG` in the environment when running the builder, at least for the `GoStdLib` action. As I understand, setting `GODEBUG` in the environment affects only run-time behavior, not behavior of compiled code. WDYT?

fmeum: You are right, I misunderstood how `GODEBUG` works. This sounds right, sorry!

jayconrod: All right, switched back to setting `GODEBUG` for all actions. I think `GoStdlib` and `GoStdlibList` are the only ones that need it, but if we kept it so narrow, someone would get a very confusing error message if they tried to call `replicate` anywhere else.

fmeum (approved):

alexeagle:
Thanks @jayconrod !! I was running into this also

PR #4052, opened 2024-08-17
Cleanup importpath/importmap handling in go_context

dzbarsky:
**What type of PR is this?**
Starlark cleanup

This looks like a 50-100ms speedup from avoiding the array allocation in infer_importpaths and reduced `getattr` calls. Is `getattr` slower than direct property access? (If not, we can avoid passing in the values from go_library)

Fixes #

**Other notes for review**


fmeum (approved):
This is also a win in explicitness.

Property access and in particular operators are faster than method calls as they don't involve the fully general construction of argument arrays.

PR #4054, opened 2024-08-18
Remove more hasattr/getattr calls in go_context

dzbarsky:
**What type of PR is this?**
starlark cleanup

The current logic to grab all the providers is confusing; it runs the fallback logics even in the case when `go_context_data` is found. Easier to follow when we pass the data in explicitly, and a bit faster as well.

I also considered making a `go_context_v2` api that doesn't have all the fallbacks and using that to power `go_context`, but didn't do it for now.

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/rules/library.bzl
+++ b/go/private/rules/library.bzl
@@ -39,8 +39,11 @@ def _go_library_impl(ctx):
         include_deprecated_properties = False,
         importpath = ctx.attr.importpath,
         importmap = ctx.attr.importmap,
+        importpath_aliases = ctx.attr.importpath_aliases,
         embed = ctx.attr.embed,
+        go_context_data = ctx.attr._go_context_data,
     )
+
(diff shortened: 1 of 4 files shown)


go/private/rules/library.bzl
```diff
@@ -204,8 +207,37 @@ go_library = rule(
...
+    validation_output = archive.data._validation_output
+
+    return [
+        library,
+        source,
+        archive,
+        DefaultInfo(
+            files = depset([archive.data.export_file]),
+        ),
+        coverage_common.instrumented_files_info(
```
dzbarsky, line 226: I split this since it has a different set of attirbutes. this rule is only really used for `go/tools/bzltestutil/chdir`, right? Do we actually need the coverage provider or the OutputGroupInfo?

fmeum (approved):

PR #4056, opened 2024-08-19
Speedup get_mode some more

dzbarsky:
**What type of PR is this?**
starlark cleanup

Looks like we were adding the `gotags` from ctx.var multiple times when cgo was enabled. This led me to realize we can compute all tags at `GoConfigInfo` creation time.

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/mode.bzl
+++ b/go/private/mode.bzl
@@ -105,16 +105,6 @@ def get_mode(ctx, go_toolchain, cgo_context_info, go_config_info):
             fail(("linkmode '{}' can't be used when cgo is disabled. Check that pure is not set to \"off\" and that a C/C++ toolchain is configured for " +
                   "your current platform. If you defined a custom platform, make sure that it has the @io_bazel_rules_go//go/toolchain:cgo_on constraint value.").format(linkmode))
 
-    tags = list(go_config_info.tags)
-    if "gotags" in ctx.var:
-        tags += ctx.var["gotags"].split(",")
-    if cgo_context_info:
-        tags += cgo_context_info.tags
-    if race:
-        tags.append("race")
-    if msan:
-        tags.append("msan")
-
     return struct(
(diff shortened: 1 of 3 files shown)


go/private/mode.bzl
```diff
@@ -105,16 +105,6 @@ def get_mode(ctx, go_toolchain, cgo_context_info, go_config_info):
             fail(("linkmode '{}' can't be used when cgo is disabled. Check that pure is not set to \"off\" and that a C/C++ toolchain is configured for " +
                   "your current platform. If you defined a custom platform, make sure that it has the @io_bazel_rules_go//go/toolchain:cgo_on constraint value.").format(linkmode))
 
-    tags = list(go_config_info.tags)
-    if "gotags" in ctx.var:
-        tags += ctx.var["gotags"].split(",")
```
fmeum, line 110: I think that this changes behavior: `ctx.var` is caller-sensitive and e.g. implicitly picks up variables from the `toolchains` attribute. I doubt that's the intended behavior though, so I'm fine with this change if we can be reasonably certain that it wasn't introduced for this purpose.

dzbarsky: Good catch! We have tests that configure tags on a `go_binary`:
`tests/core/go_binary/BUILD.bazel:    gotags = ["good"],`

They pass with this change; I think what happens is that the use the `go_transition` on the `go_context_data` which picks up the `gotags` from ctx.attr and then they end up in the `go_config_info`.

Does that resolve your concern?

fmeum: That's tangential: `ctx.var` picks up tags from `--define=gotags` (horrible, but also global) and the `toolchains` attribute (no idea why anybody would use that for Go build tags). I'm fine with breaking the latter.

fmeum (approved):

PR #4058, opened 2024-08-21
More go_context speedups

dzbarsky:
**What type of PR is this?**
Starlark cleanup

After the previous cleanups, `mode` and `GoConfigInfo` are pretty similar, the only differences were `link` vs `linkmode` and goos/goarch were missing on `GoConfigInfo`. So we add the default ones from the toolchain, and this lets us use the `GoConfigInfo` in the majority of cases without having to copy all the properties to `mode`. Another option is to move the `goos`/`goarch` from `mode` to top-level properties on `go_context`, though this is potentially a breaking change. Having them on GoConfigInfo seems like a reasonable direction to move in anyway if we plan to use Bazel's platforms to control what we target. 

This required aligning the `linkmode` name, which is good for consistency anyway.

We pass in `goos`/`goarch` to `go_context` since they are only set in terminal rules and we don't need to look them up for intermediate libs.

I also noticed `getattr` from the proto libs showing up in profiles, so we clean that up as well.

This speeds up go_context another 2x, saving 100ms in buildbuddy repo

Fixes #

**Other notes for review**


dzbarsky:
I suppose this is breaking either way since `mode.link` became `mode.linkmode`. Feels pretty minor, maybe we just mention that in release notes?

go/private/context.bzl
```diff
@@ -952,6 +998,7 @@ go_config = rule(
     doc = """Collects information about build settings in the current
     configuration. Rules may depend on this instead of depending on all
     the build settings directly.""",
+    toolchains = [GO_TOOLCHAIN],
```
fmeum, line 1001: Adding a toolchain dependency here likely means that in case of a missing Go toolchain, Bazel will fail on this target (as a transitive dep of all Go targets) instead of an actual user-defined Go target. That's only marginally worse (or maybe even better?) than failing on some random actual `go_library` that's a leaf in the dep tree.

Could you check how the "missing toolchain" error changes with this, e.g. by targeting an exotic platform?

fmeum (approved):

go/private/context.bzl
dzbarsky: It's the same as on master, it looks like the first failure at the top-level target is reported

```
platform(
    name = "exotic_platform",
    constraint_values = [
        "@platforms//os:haiku",
        "@platforms//cpu:mips64",
    ],
)
```

```
➜  rules_go git:(zbarsky/more-mode2) ✗ bazel build //tests/core/build_stdlib:build_stdlib_test  --platforms=//:exotic_platform                                                                                                                                                     ~/rules_go
ERROR: /Users/dzbarsky/rules_go/tests/core/build_stdlib/BUILD.bazel:3:14: While resolving toolchains for target //tests/core/build_stdlib:build_stdlib_test (dd044de): No matching toolchains found for types //go:toolchain.
To debug, rerun with --toolchain_resolution_debug='//go:toolchain'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
```

dzbarsky: `bazel build //tests/core/...  --platforms=//:exotic_platform` fails on a random test in that package when run a few times

fmeum: Nice, guess I still haven't fully understood how toolchain resolution works :-) Thanks for confirming this!

dzbarsky:
@tyler-french You can band-aid this by adding the following attributes to `_go_thriftrw_library`:
```
        "importmap": attr.string(),
        "importpath_aliases": attr.string_list(),  # experimental, undocumented
```

I didn't think this part would be breaking, because this aspect is only applied to `deps` in `go_proto_library` and propagates along `deps` and `embed`. It's surprising to me that you have a thrift library that ends up as a dep of a proto library, I would expect it to be a tree of `go_proto_libraries`. Any idea what's going on here? (My guess was that maybe you have a proto plugin that depends on thrift, but I don't see how those deps would get the aspect applied...)

I supposed we could revert part of this PR to continue to support these as optional, though the direct property access is slightly faster than `getattr`. I'd like to understand what you're hitting though before we change that, since it's possible your graph is larger than you intended

PR #4060, opened 2024-08-21
Remove usage of _flatten_possibly_transitioned_attr in common case

dzbarsky:
**What type of PR is this?**
Starlark cleanup

Making the caller deal with a transitioned attribute is cleaner and will be necessary for proper starlark typing anyway

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -410,9 +410,8 @@ def go_context(
         if hasattr(attr, "_go_config"):
             go_config_info = attr._go_config[GoConfigInfo]
         if hasattr(attr, "_stdlib"):
-            stdlib = _flatten_possibly_transitioned_attr(attr._stdlib)[GoStdLib]
+            stdlib = attr._stdlib[GoStdLib]
     else:
-        go_context_data = _flatten_possibly_transitioned_attr(go_context_data)
         if CgoContextInfo in go_context_data:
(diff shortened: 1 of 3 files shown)


go/private/context.bzl
```diff
@@ -412,7 +412,6 @@ def go_context(
         if hasattr(attr, "_stdlib"):
             stdlib = _flatten_possibly_transitioned_attr(attr._stdlib)[GoStdLib]
```
fmeum, line 413: Can we also remove this?

fmeum (approved):

go/private/context.bzl
dzbarsky: there's only 1 transition of `_stdlib` in this repo and it isn't passed to `go_context`, so yes. But it could technically be a breaking change for consumers. I suspect everyone uses `_go_context_data` instead though. I'm happy to remove it if you're ok with that, though I think we should just deprecate this lookup mechanism instead (in future releases)

fmeum: I think that we can clean this up now. If we are already making a few minor changes to the API, we might as well do it in one release so that folks don't need to look into their rules in detail multiple times. This is low-effort enough to fix if it even comes up as a breakage.

dzbarsky: ok, updated

fmeum (approved):

PR #4063, opened 2024-08-26
Handle targets provided in deps as a result of GoSource.resolver

dzbarsky:
**What type of PR is this?**
Starlark fixup

**What does this PR do? Why is it needed?**
Make the migration process a little smoother

**Which issues(s) does this PR fix?**
https://github.com/bazelbuild/rules_go/pull/4030#issuecomment-2310491357

**Other notes for review**

PR #4064, opened 2024-08-26
prepare rules_go release 0.50.0

tyler-french:
This will be merged, but we will end up cherry picking certain commits onto an `release-0.50` branch, which will not include all of the commits on master.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "rules_go",
-    version = "0.49.0",
+    version = "0.50.0",
     compatibility_level = 0,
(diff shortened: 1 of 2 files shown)


dzbarsky:
I think we should take https://github.com/bazelbuild/rules_go/pull/4063 first to make this a bit less breaking

tyler-french:
> I think we should take #4063 first to make this a bit less breaking

Oh good call, let me test it

tyler-french:
There's still some issues we're having with https://github.com/bazelbuild/rules_go/pull/4058 and Nogo validations

fmeum (approved):

PR #4066, opened 2024-08-27
gomock: allow passing extra flags to mockgen_tool

### LINKED ISSUE #4065: gomock: no way to pass extra flags to mockgen
### What version of rules_go/gazelle/Bazel are you using?

latest (v0.49.0 / v0.38.0 / 7.3.1)

`go.uber.org/mock v0.4.1-0.20240710190148-70a5e8c8f892`

### What did you do?

I'm using `@rules_go//extras:gomock.bzl` to create mocks, and then putting them back in the git tree via [write_source_file](https://github.com/aspect-build/bazel-lib/blob/main/docs/write_source_files.md#write_source_file) because some code editors don't enjoy generated files without being able to see them.

`mockgen` includes a comment on top of the file to aid with regeneration, but that includes the `bazel-out/` path which isn't stable. Obviously this means I get unnecessarily large diffs, which isn't ideal.

To fix this I'd have to pass extra flags to gomock, but `sh_binary` has multiple outputs when trying to wrap the mockgen binary which conflicts with `allow_single_file = True` in gomock's rule().

So, I think rules_go's gomock() should allow for passing either these specific flags (`--write_command_comment`, `--write_source_comment`), or arbitrary flags.

TvdW:
**What type of PR is this?**

> Feature

Adds the ability to pass extra arguments to gomock

Fixes #4065
Closes #4067

**Other notes for review**

#4067 is another way to do it, but I think accepting an argument list is cleaner


fmeum (approved):
I like this approach more, adding attributes for single args doesn't scale.

Could you also revert https://github.com/bazelbuild/rules_go/commit/4feb7b8fe2a047e23c6327659896a2637cd0ca38 in this PR?

TvdW:
Sure, done!

fmeum (approved):

PR #4030, opened 2024-08-11
Merge GoLibrary and GoSource providers

dzbarsky:
**What type of PR is this?**
Code simplification

**What does this PR do? Why is it needed?**
These providers are 1:1, and we always create a library followed by `source_from_library`. Merging them allows to remove the `library` field from `GoSource` as well as the the `resolve` field, since we don't need to smuggle the closure between the providers anymore.

If we do this, we will also need to fix up Gazelle. We can either have a patch in this repo so CI passes and then followup with the Gazelle fix, or somehow shim the API until we release new versions in lockstep. Open to ideas how to best do this.

**Which issues(s) does this PR fix?**
One step toward https://github.com/bazelbuild/rules_go/issues/2578

Fixes #

**Other notes for review**

PR #4074, opened 2024-08-28
Speedup compilepkg args handling

dzbarsky:
**What type of PR is this?**
Starlark cleanup

A lot of the args between compilepkg and nogo are shared, so we can reuse the `Args` object. Cleaned up a few other things that showed up in profiles while I was here. 

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -170,7 +170,6 @@ def _dirname(file):
 def _builder_args(go, command = None, use_path_mapping = False):
     args = go.actions.args()
     args.use_param_file("-param=%s")
-    args.set_param_file_format("shell")
     if command:
         args.add(command)
     sdk_root_file = go.sdk.root_file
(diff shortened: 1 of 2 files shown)


fmeum:
This is a nice improvement, including to retained memory. I considered doing this right in the nogo change, but wanted to keep it small. 

We should wait for the nogo release before we merge this.

dzbarsky:
> This is a nice improvement, including to retained memory. I considered doing this right in the nogo change, but wanted to keep it small.
> 
> We should wait for the nogo release before we merge this.

Yep, makes sense.

dzbarsky:
@fmeum shall we revisit this after bazel8 ships?

fmeum:
Yes, I'll also cut the long due rules_go release!

dzbarsky:
> Yes, I'll also cut the long due rules_go release!

OK, I've rebased it. Maybe we can try to land https://github.com/bazelbuild/rules_go/pull/4030 after the release is cut?

go/private/context.bzl
```diff
@@ -187,14 +186,15 @@ def _builder_args(go, command = None, use_path_mapping = False):
         # Use a file rather than goroot as the latter is just a string and thus
         # not subject to path mapping.
         args.add_all("-goroot", [goroot_file], map_each = _dirname, expand_directories = False)
-    args.add("-installsuffix", installsuffix(go.mode))
-    args.add_joined("-tags", go.mode.tags, join_with = ",")
+    mode = go.mode
+    args.add("-installsuffix", installsuffix(mode))
+    if mode.tags:
+        args.add_joined("-tags", mode.tags, join_with = ",")
```
fmeum, line 192:
```suggestion
    args.add_joined("-tags", mode.tags, join_with = ",")
```

Bazel already has an early return for this case in `args.add_joined`.

fmeum (approved):

go/private/context.bzl
dzbarsky: I saw it show up but it was only 10ms and inconsistent. I'll revert this bit

fmeum: I hit commit :⁠-⁠)

dzbarsky: Ok thanks!

PR #4093, opened 2024-09-09
Remove unneeded packages from stdlib

dzbarsky:
**What type of PR is this?**
perf improvement

This cuts down the time to build the stdlib from 8.1 to 7.4s in my testing and it also reduces the size of the `sdk.libs` depset and the resulting linkcfg files.

I wasn't sure if it's ok to exclude all of `cmd/`; I suspect it should be ok since the toolchain has prebuilt tools anyway, in which case we can simplify the glob pattern.

Fixes #

**Other notes for review**


fmeum (approved):
Something something colocation of tests is great 😅

dzbarsky:
> Something something colocation of tests is great 😅

Sure is for the authors :)

Issue #4078, opened 2024-09-02
[Bazel CI] FATAL: Bazel crashed due to an internal error while evaluating node



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4093#0191abc3-d4cb-40d4-91ac-5d7e907b960f

 Platform: Multiple

 Logs: 
```
bazel --nosystem_rc --nohome_rc info
WARNING: Option 'experimental_enable_bzlmod' is deprecated: Use --enable_bzlmod instead
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.IllegalStateException: java.lang.RuntimeException: Unrecoverable error while evaluating node 'PlatformMappingValue.Key{path=platform_mappings, wasExplicitlySetByUser=false}' (requested by nodes 'BuildConfigurationKeyValue.Key{buildOptions=1a1e367614cf4c436a40b59b0da811d24bd9acb481ab2d9d559457b7aabecffd}')
	at com.google.devtools.build.lib.skyframe.SkyframeExecutor.evaluateSkyKeys(SkyframeExecutor.java:2030)
	at com.google.devtools.build.lib.skyframe.SkyframeExecutor.evaluateSkyKeys(SkyframeExecutor.java:2006)
	at com.google.devtools.build.lib.skyframe.SkyframeExecutor.createBuildConfigurationKey(SkyframeExecutor.java:1962)
	at com.google.devtools.build.lib.skyframe.SkyframeExecutor.getConfiguration(SkyframeExecutor.java:1886)
	at com.google.devtools.build.lib.runtime.commands.InfoCommand.lambda$exec$0(InfoCommand.java:158)
	at com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:181)
	at com.google.devtools.build.lib.runtime.commands.InfoCommand.exec(InfoCommand.java:219)
	at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.execExclusively(BlazeCommandDispatcher.java:679)
	at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.exec(BlazeCommandDispatcher.java:252)
	at com.google.devtools.build.lib.server.GrpcServerImpl.executeCommand(GrpcServerImpl.java:607)
	at com.google.devtools.build.lib.server.GrpcServerImpl.lambda$run$1(GrpcServerImpl.java:676)
[...]

PR #4095, opened 2024-09-10
Add --enable_workpace in .bazelrc

meteorcloudy:
**What type of PR is this?**

Dealing with incompatible change

Due to https://github.com/bazelbuild/bazel/issues/23023

Fixes [#](https://github.com/bazelbuild/rules_go/issues/4078)

**Other notes for review**


dzbarsky:
We don't need this anymore

PR #4096, opened 2024-09-10
try and add an xdef to mimic what go toolset does for testing.Testing…

### LINKED ISSUE #3686: support Go1.21 testing.Testing() function
Well, I don't know much about bazel, hope here is the right place for this issue.

```
http_archive(
    name = "io_bazel_rules_go",
    sha256 = "278b7ff5a826f3dc10f04feaf0b70d48b68748ccd512d7f98bf442077f043fe3",
    urls = [
        "http://bazel-cache.pingcap.net:8080/bazelbuild/rules_go/releases/download/v0.41.0/rules_go-v0.41.0.zip",
        "http://ats.apps.svc/bazelbuild/rules_go/releases/download/v0.41.0/rules_go-v0.41.0.zip",
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.41.0/rules_go-v0.41.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.41.0/rules_go-v0.41.0.zip",
    ],
)
```

```
http_archive(
    name = "bazel_gazelle",
    sha256 = "b8b6d75de6e4bf7c41b7737b183523085f56283f6db929b86c5e7e1f09cf59c9",
    urls = [
        "http://bazel-cache.pingcap.net:8080/bazelbuild/bazel-gazelle/releases/download/v0.31.1/bazel-gazelle-v0.31.1.tar.gz",
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.31.1/bazel-gazelle-v0.31.1.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.31.1/bazel-gazelle-v0.31.1.tar.gz",
        "http://ats.apps.svc/bazelbuild/bazel-gazelle/releases/download/v0.31.1/bazel-gazelle-v0.31.1.tar.gz",
    ],
)
```

```
bazel version
Bazelisk version: v1.14.0
Build label: 6.3.2
Build time: Tue Aug 8 15:48:33 2023 (1691509713)
Build timestamp: 1691509713
Build timestamp as int: 1691509713
```

I think so.

linux, x86_64

### What did you do?

I'm trying to use the `testing.Testing()` function in https://github.com/pingcap/tidb/pull/46194
And I found some weird phenomenon.
Some tests are stable when I run it on my local environment, using the standard Go.
But they failed to run in our CI environment. Our CI is using bazel...

After debugging a while, I find that it seems `testing.Testing()` is not working as expected when using bazel.
As you can see from here, https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/testing/testing.go;l=656-661

```
// Testing reports whether the current code is being run in a test.
// This will report true in programs created by "go test",
// false in programs created by "go build".
func Testing() bool {
	return testBinary == "1"
}
```

This function works by change the link flag when calling "go test"
Grep the go repo source code you can find this:

```
find . -name '*.go'|xargs grep 'testBinary'
grep: ./go/parser/testdata/issue42951/not_a_file.go: Is a directory
./debug/gosym/pclntab_test.go:  pclinetestBinary string
./debug/gosym/pclntab_test.go:  pclinetestBinary = filepath.Join(pclineTempDir, "pclinetest")
./debug/gosym/pclntab_test.go:  cmd := exec.Command(testenv.GoToolPath(t), "build", "-o", pclinetestBinary)
./debug/gosym/pclntab_test.go:          pclinetestBinary = ""
./debug/gosym/pclntab_test.go:  f, tab := crack(pclinetestBinary, t)
./testing/testing.go:// testBinary is set by cmd/go to "1" if this is a binary built by "go test".
./testing/testing.go:// because this is possible, the compiler will not optimize testBinary
./testing/testing.go:var testBinary = "0"
./testing/testing.go:   return testBinary == "1"
./cmd/go/internal/test/test.go:                 testBinary := testBinaryName(p)
./cmd/go/internal/test/test.go:                 pkgsForBinary[testBinary] = append(pkgsForBinary[testBinary], p)
./cmd/go/internal/test/test.go:         for testBinary, pkgs := range pkgsForBinary {
./cmd/go/internal/test/test.go:                         base.Errorf("cannot write test binary %s for multiple packages:\n%s", testBinary, buf.String())
./cmd/go/internal/test/test.go: testBinary := testBinaryName(p)
./cmd/go/internal/test/test.go: a.Target = testDir + testBinary + cfg.ExeSuffix
./cmd/go/internal/test/test.go:                 if strings.Contains(testBinary, bad) {
./cmd/go/internal/test/test.go:         target := filepath.Join(base.Cwd(), testBinary+cfg.ExeSuffix)
./cmd/go/internal/test/test.go:                                 target = filepath.Join(target, testBinary+cfg.ExeSuffix)
./cmd/go/internal/test/test.go:                                 target = filepath.Join(base.Cwd(), target, testBinary+cfg.ExeSuffix)
./cmd/go/internal/test/test.go:// testBinaryName can be used to create name for test binary executable.
./cmd/go/internal/test/test.go:func testBinaryName(p *load.Package) string {
./cmd/go/internal/load/test.go: ldflags := append(p.Internal.Ldflags, "-X", "testing.testBinary=1")
```

bazel test do not set the "testing.testBinary" appropriately.
A workaround for this is modify the BUILD.bazel and add     `x_defs = {"testing.testBinary": "1"},` to the rule go_test,
but maybe a better fix would be in the tools.

Go1.21 testing.Testing() function return true when using bazel test.

### What did you see instead?

It get false, the wrong value.

paulgmiller:
See  https://pkg.go.dev/testing\#Testing
Go test sets an -X so that code knows if its in a UT

**What type of PR is this?**
 Feature? 

Adds an xdef to mimic go test more precisely. 

Should we only do for go > 1.23?

Fixes #3686

**Other notes for review**

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -150,7 +150,8 @@ def _go_test_impl(ctx):
     # before user code. See comment above the init function
     # in bzltestutil/init.go.
     test_gc_linkopts.extend(["-X", "+initfirst/github.com/bazelbuild/rules_go/go/tools/bzltestutil/chdir.RunDir=" + run_dir])
-
+    #post golang 1.23 testing.Testing is expected to be set for go tests
+    test_gc_linkopts.extend(["-X", "testing.Testing=1"])
     # Now compile the test binary itself
     test_library = GoLibrary(
         name = go.label.name + "~testmain",
@@ -311,7 +312,6 @@ _go_test_kwargs = {
             doc = """Map of defines to add to the go link command.
             See [Defines and stamping] for examples of how to use these.
             """,
-            "testing.testBinary" = "1",
         ),
         "linkmode": attr.string(
             default = "auto",
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -311,6 +311,7 @@ _go_test_kwargs = {
             doc = """Map of defines to add to the go link command.
             See [Defines and stamping] for examples of how to use these.
             """,
+            "testing.testBinary" = "1",
         ),
         "linkmode": attr.string(
             default = "auto",


go/private/rules/test.bzl
```diff
@@ -150,7 +150,8 @@ def _go_test_impl(ctx):
     # before user code. See comment above the init function
     # in bzltestutil/init.go.
     test_gc_linkopts.extend(["-X", "+initfirst/github.com/bazelbuild/rules_go/go/tools/bzltestutil/chdir.RunDir=" + run_dir])
-
+    #post golang 1.23 testing.Testing is expected to be set for go tests
```
fmeum, line 153: Please run buildifier and add a reference to the `testing.Testing()` function here.

apsaltis-ddog:
👋 What's the status of this PR?

I have a patch locally that should be enough to add in this functionality, and it wouldn't take much polish to make it ready for upstreaming.  I'd be happy to create a new PR for this if y'all would like.

PR #4102, opened 2024-09-12
feat: Saving the nogo fixes

peng3141:
**What type of PR is this?**

> Feature

nogo analyzers may produce fixes as analysis.[Diagnostic](https://pkg.go.dev/golang.org/x/tools/go/analysis#Diagnostic). This PR allows rules_go to save the fixes as patches, one per target. The patch and the command to apply the patch is printed out to the terminal for users to manually apply. The patch is also available in the "nogo_fix" output group. This allows people to get patches for all targets without failing the build by passing `--norun_validations --output_groups nogo_fix`.

Example output:

```

nogo: errors found by nogo during build-time code analysis:
src/example.com/send_request.go:133:4: Add and Done should not both exist inside the same goroutine block (example_nogo_analyzer)

-------------------Suggested Fix-------------------
--- src/example.com/send_request.go
+++ src/example.com/send_request.go
@@ -123,6 +123,7 @@
 
 	for !isComplete {
 		<-ticker.C
+		concurrentJobs.Add(1)
 		go func() {
 			text, hasMore := <-line
 			if !hasMore {
@@ -130,7 +131,6 @@
 				return
 			}
 
-			concurrentJobs.Add(1)
 			defer concurrentJobs.Done()
 
 			atomic.AddInt32(&totalSend, 1)

-----------------------------------------------------

To apply the suggested fix, run the following command:
$ patch -p1 < bazel-out/k8-fastbuild/bin/src/example.com/send_request.nogo.patch

Target //src/example.com:send_request failed to build
```

**Other notes for review**
An analyzer may suggest multiple alternative fixes to one issue. Only the first one is selected by default, unless it conflicts with other fixes, in which case it moves on to try the next alternative. If all alternatives are tried but still have conflicts, they will be skipped. In such case, the user will have to apply the patch first, and run nogo again to get the fix to the issue.


go/tools/builders/nogo.go
```diff
@@ -82,10 +85,11 @@ func nogo(args []string) error {
 		return err
 	}
 
-	return runNogo(workDir, nogoPath, goSrcs, ignoreSrcs, facts, importPath, importcfgPath, outFactsPath, outLogPath)
+	return runNogo(workDir, nogoPath, goSrcs, ignoreSrcs, facts, importPath, importcfgPath, outFactsPath, outLogPath, nogoFixPath)
 }
 
-func runNogo(workDir string, nogoPath string, srcs, ignores []string, facts []archive, packagePath, importcfgPath, outFactsPath string, outLogPath string) error {
+func runNogo(workDir string, nogoPath string, srcs, ignores []string, facts []archive, packagePath, importcfgPath, outFactsPath string, outLogPath string, nogoFixPath string) error {
```
linzhp, line 91:
```suggestion
func runNogo(workDir string, nogoPath string, srcs, ignores []string, facts []archive, packagePath, importcfgPath, outFactsPath string, outLogPath string, outFixPath string) error {
```

go/private/rules/test.bzl
```diff
@@ -68,13 +68,19 @@ def _go_test_impl(ctx):
...
 
     # Compile the library to test with internal white box tests
     internal_library = go.new_library(go, testfilter = "exclude")
     internal_source = go.library_to_source(go, ctx.attr, internal_library, ctx.coverage_instrumented())
     internal_archive = go.archive(go, internal_source)
     if internal_archive.data._validation_output:
         validation_outputs.append(internal_archive.data._validation_output)
+    if internal_archive.data._out_nogo_fix:
+        # Note we do not include external_archive.data._out_nogo_fix since
+        # they correspond to the packages external to the current workspace.
```
fmeum, line 81: The distinction between `internal` and `external` archive refers to same package vs. separate package tests here. Could you fix the comment or explain why it's accurate?

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// DiagnosticEntry represents a diagnostic entry with the corresponding analyzer.
+type DiagnosticEntry struct {
+	analysis.Diagnostic
+	*analysis.Analyzer
+}
+
+// An Edit describes the replacement of a portion of a text file.
+type Edit struct {
```
fmeum, line 20: How would users apply these fixes? I wonder whether we shouldn't just export a regular patch file instead, possibly one per analyzer in a tree artifact in case different analyzers produce patches that conflict with each other.

This was my previous attempt at exporting fixes as patches: https://github.com/bazelbuild/rules_go/commit/998b738105e4fcefcf432f5d53723826073691d2

go/private/actions/compilepkg.bzl
```diff
@@ -87,6 +88,8 @@ def emit_compilepkg(
         fail("nogo must be specified if and only if out_nogo_log is specified")
     if bool(nogo) != bool(out_nogo_validation):
         fail("nogo must be specified if and only if out_nogo_validation is specified")
+    if bool(nogo) != bool(out_nogo_fix):
+        fail("nogo must be specified if and only if out_nogo_fix is specified")
```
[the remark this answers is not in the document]
peng3141: no, they do not need to provide a fix.
Even for the linters that have fixing rules, we do not know in advance whether it will produce fixes or not for the build target, depending on whether the target has error to fix or not. 

due to this uncertainty, conservatively, we create out_nogo_fix for each target as long as nogo is present. 

The declaration happens in archive.bzl:
    nogo = get_nogo(go)
    if nogo:
        out_facts = go.declare_file(go, name = source.library.name, ext = pre_ext + ".facts")
        out_nogo_log = go.declare_file(go, name = source.library.name, ext = pre_ext + ".nogo.log")
        out_nogo_validation = go.declare_file(go, name = source.library.name, ext = pre_ext + ".nogo")
        out_nogo_fix = go.declare_file(go, name = source.library.name, ext = pre_ext + ".nogo.fix")
    else:
        out_facts = None
        out_nogo_log = None
        out_nogo_validation = None
        out_nogo_fix = None

go/private/actions/compilepkg.bzl
```diff
@@ -271,6 +276,7 @@ def _run_nogo(
     args.add_all(archives, before_each = "-facts", map_each = _facts)
     args.add("-out_facts", out_facts)
     args.add("-out_log", out_log)
+    args.add("-fixpath", out_nogo_fix)
```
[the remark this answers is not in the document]
peng3141: Done

go/tools/builders/nogo.go
```diff
@@ -25,6 +25,7 @@ func nogo(args []string) error {
 	var importPath, packagePath, nogoPath, packageListPath string
 	var testFilter string
 	var outFactsPath, outLogPath string
+	var nogoFixPath string
```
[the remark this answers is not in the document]
peng3141: Done

go/tools/builders/nogo.go
```diff
@@ -39,6 +40,8 @@ func nogo(args []string) error {
 	fs.StringVar(&nogoPath, "nogo", "", "The nogo binary")
 	fs.StringVar(&outFactsPath, "out_facts", "", "The file to emit serialized nogo facts to")
 	fs.StringVar(&outLogPath, "out_log", "", "The file to emit nogo logs into")
+	fs.StringVar(&nogoFixPath, "fixpath", "", "The path of the file that stores the nogo fixes")
```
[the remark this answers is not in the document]
peng3141: Done

go/tools/builders/nogo.go
```diff
@@ -82,10 +85,11 @@ func nogo(args []string) error {
 		return err
 	}
 
-	return runNogo(workDir, nogoPath, goSrcs, ignoreSrcs, facts, importPath, importcfgPath, outFactsPath, outLogPath)
+	return runNogo(workDir, nogoPath, goSrcs, ignoreSrcs, facts, importPath, importcfgPath, outFactsPath, outLogPath, nogoFixPath)
```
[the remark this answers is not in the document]
peng3141: Done

go/tools/builders/nogo.go
peng3141: Done

go/private/rules/test.bzl
peng3141: thanks, 

currently, we retrieve fix via --output_groups=nogo_fix, which is only for the target being built, not including those dependent targets.

thinking about alternatives where we dump fixes for the dependent targets too. maybe by including the fix into the validation action.

go/tools/builders/nogo_change.go
peng3141: thanks for sharing, I agree your proposal is a more end-to-end solution. Will include this. 

We have a separate CLI command for patching, which shares some common part with your version:
internally, it reads the edits, sorts them, applies them in memory to get the 'after' file, then produces the patch file by comparing before/after, then applies the patch file.

It holds the edits for each analyzer separately, in the patch file, edits of all analyzers are merged. If they overlap, it would fail.

peng3141:
thanks for the comments, i am in the process of addressing them.

go/private/rules/test.bzl
peng3141: addressed in the new version.

now, we wire the fix as the output of validation action. This way, --run_validations ensure every target (not only the input target, but also the dependent target) has their fixes if any produced.

The previous implementation is incorrect, since it wires the fix as the outputgroup of the input target only, as a result, we cannot get the fix for the dependent target even if they are linted.

go/tools/builders/nogo_change.go
peng3141: addressed in the new commit. 

one caveat is that, currently it copies/pastes https://github.com/pmezard/go-difflib/blob/master/LICENSE under its permission of copyright. 

after one day of attempt, I could not get test passing due to the dependency missing of https://github.com/pmezard/go-difflib. Can we address this at last or in a separate PR?

peng3141:
this shows the log of the latest version:
https://gist.github.com/peng3141/bdaefac333434cf2ecbef4edfd8d0200

go/private/rules/test.bzl
peng3141: back to your original question, still think we should not include external package, given external package is validated independently during which nogo fixes are already produced.

peng3141:
@fmeum @linzhp the PR is ready for review.

could you focus on the high-level design, once it is agreed upon, we can have one pass to address the readability nits.

go/tools/builders/nogo_validation.go
```diff
@@ -8,21 +8,53 @@ import (
...
+	nogoFixContent, err := os.ReadFile(nogoFixFileTmp)
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(nogoFixFile, nogoFixContent, 0755)
+	if err != nil {
+		return err
+	}
+
 	if len(logContent) > 0 {
```
linzhp, line 34: reduce nesting: https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

go/tools/builders/nogo_main.go
```diff
@@ -543,20 +541,39 @@ func checkAnalysisResults(actions []*action, pkg *goPackage) string {
...
 	if numSkipped > 0 {
 		errs = append(errs, fmt.Errorf("%d analyzers skipped due to type-checking error: %v", numSkipped, pkg.typeCheckError))
 	}
+
+	if nogoFixPath != "" {
+		// If the nogo fixes are requested, we need to save the fixes to the file even if they are empty.
+		// Otherwise, bazel will complain "not all outputs were created or valid"
+		change, err := NewChangeFromDiagnostics(diagnostics, pkg.fset)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("errors in dumping nogo fix, specifically in converting diagnostics to change %v", err))
```
linzhp, line 557: If there are errors here, does it make sense to call `ToPatches` and `SavePatchesToFile` below?

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,325 @@
...
+		out = append(out, edit.New...)
+		lastEnd = edit.End
+	}
+	out = append(out, src[lastEnd:]...)
+
+	if len(out) != size {
+		panic("wrong size")
+	}
+
+	return string(out), nil
```
linzhp, line 111: can we return bytes instead? The `out` is converted to string here and immediately converted back to []byte by its only caller

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,325 @@
...
+		lastEnd = edit.End
+	}
+
+	return edits, size, nil
+}
+
+// The following is about the `Change`, a high-level abstraction of edits.
+// Change represents a set of edits to be applied to a set of files.
+type Change struct {
+	AnalyzerToFileToEdits map[string]map[string][]Edit `json:"analyzer_file_to_edits"`
```
linzhp, line 150: Do we need 3 levels of nesting, only to be flattened later?

linzhp (commented):
Only half way through, posting some comments so far. I will continue later this week

go/tools/builders/nogo_validation.go
peng3141: It already followed the "reduce-nesting" and handled the error case first, right? 
+	nogoFixContent, err := os.ReadFile(nogoFixFileTmp)
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(nogoFixFile, nogoFixContent, 0755)
+	if err != nil {
+		return err
+	}

go/tools/builders/nogo_main.go
peng3141: yes, that was by design, although we may need to discuss about design choices here. 

Btw, we need to write to empty string to nogoFixPath when there are errors. See nogo.go (line 100) for similar logic. Otherwise, bazel complains that the declared file is not defined.

Coming back to the design choices, here is another design in comparison:
`
if nogoFixPath != "" {
	// If nogo fixes are requested, save the fixes to the file even if they are empty.
	// This prevents Bazel from complaining about missing or invalid outputs.
	change, err := NewChangeFromDiagnostics(diagnostics, pkg.fset)
	if err != nil {
		// Ensure an empty patch file is saved when there's an error in generating the change.
		errs = append(errs, fmt.Errorf("error converting diagnostics to change: %v", err))
		if saveErr := SavePatchesToFile(nogoFixPath, nil); saveErr != nil {
			errs = append(errs, fmt.Errorf("error saving empty patches file: %v", saveErr))
		}
	} else {
		fileToPatch, err := ToPatches(Flatten(*change))
		if err != nil {
			errs = append(errs, fmt.Errorf("error generating patches: %v", err))
			if saveErr := SavePatchesToFile(nogoFixPath, nil); saveErr != nil {
				errs = append(errs, fmt.Errorf("error saving empty patches file: %v", saveErr))
			}
		} else {
			if err := SavePatchesToFile(nogoFixPath, fileToPatch); err != nil {
				errs = append(errs, fmt.Errorf("error saving patches to file: %v", err))
			}
		}
	}
}

`

In my current design, when NewChangeFromDiagnostics returns error, the change has partial result of fixes which can be still applied. Let us consider this case: 
There are high-quality analyzers that produce great fixes, and there are some poorly written analyzers that produce wrong fixes, e.g., they produced corrupted offsets. Current design still allows the fixes from the well-written analyzers to be applied. 

There is one caveat case:  the change may already include some edits from the bad analyzer, which have valid offsets but are of poor quality. 

In my opinion, the nogo framework should faithfully apply the fixes that are applicable (with valid offsets). It should not ban fixes from all analyzers if one analyzer is bad. Also it is the responsibility of the monorepo owners to remove/fix bad analyzers.

peng3141: also if we adopt my current design, NewChangeFromDiagnostics can be further improved to include as many fixes as possible, instead of early return.
https://gist.github.com/peng3141/7e4eee8cf23452fa9bdfa107215fbc08

go/tools/builders/nogo_change.go
peng3141: yes, we need this.

the two levels file:edits is required, since we track and apply patch per file.

besides, see the Flatten() function, it considers the cases that different analyzers produce conflicting edits, i.e., edits that overlap with each other. In this case, it is impossible to apply both edits.

we will ignore the latter analyzer (sorted already for determinism) but still allow the former analyzer to proceed. 
This is why we add the extra indirection of indexing by analyzer.

go/tools/builders/nogo_main.go
linzhp: I like the ability to let user choose analyzers to trust

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,326 @@
...
+		// Initialize the map for the analyzer if it doesn't exist
+		c.AnalyzerToFileToEdits[analyzer] = make(map[string][]Edit)
+	}
+
+	// Append the edit to the list of edits for the specific file under the analyzer
+	c.AnalyzerToFileToEdits[analyzer][file] = append(c.AnalyzerToFileToEdits[analyzer][file], edit)
+}
+
+// Flatten takes a Change and returns a map of FileToEdits, merging edits from all analyzers.
+func Flatten(change Change) map[string][]Edit {
```
linzhp, line 217: Can we produce one patch file per analyzer? That way, we don't need to merge edits to the same file from different analyzers. Like you said, some analyzers may produce bad edits. When users apply patches one by one, they can ignore those patched produced by bad analyzers.

go/tools/builders/nogo_change.go
linzhp: Let's avoid overusing maps and create more informative data structure: https://abhinav.github.io/future-proof-packages-2023/#/%EF%B8%8F-map-overuse

lpxz:
has been a while, back to this full time.

go/tools/builders/nogo_change.go
lpxz: done

go/tools/builders/nogo_change.go
lpxz: there will be a single file per target, as created at the .bzl files (the go files only fill in contents).
but still, in that single file, we can index patches with analyzer. 

asking users to pick which analyzer to patch may be too much for developers. Ideally we want them simply say "yes/no". Also the conflicts may not happen often, we may ask only in the case of conflicts.

will think further.

go/tools/builders/nogo_main.go
lpxz: done, this support of letting user choose has to happen on the patching side, where we show preview of diff and users answer "yes" or "no". 

this is now supported.

go/tools/builders/nogo_change.go
lpxz: good point, especially given this is across package boundary.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,326 @@
...
+	// Inner function: returns true if the given string contains any non-whitespace characters.
+	hasNonWhitespaceCharacter := func(s string) bool {
+		return strings.ContainsFunc(s, func(r rune) bool {
+			return !unicode.IsSpace(r)
+		})
+	}
+
+	// Trim left
+	for i := 0; i < len(lines); i++ {
+		if hasNonWhitespaceCharacter(lines[i]) {
```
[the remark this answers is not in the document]
lpxz: no longer relevant, moved out of rules_go

go/tools/builders/nogo_change_serialization.go
```diff
@@ -0,0 +1,65 @@
...
+	err = os.WriteFile(filename, jsonData, 0644)
+	if err != nil {
+		return fmt.Errorf("error writing to file: %v", err)
+	}
+
+	return nil
+}
+
+// LoadPatchesFromFile loads the map[string]string (file paths to patch content) from a JSON file.
+// Note LoadPatchesFromFile is used for testing only.
```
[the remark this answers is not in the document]
lpxz: no longer relevant, moved out of rules_go

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,326 @@
...
+			return nil, err
+		}
+
+		out, err := ApplyEditsBytes(contents, edits)
+		if err != nil {
+			return nil, err
+		}
+
+		diff := UnifiedDiff{
+			// difflib.SplitLines does not handle well the whitespace at the beginning or the end.
```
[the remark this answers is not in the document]
lpxz: no longer relevant, moved out of rules_go

go/tools/builders/nogo_change.go
lpxz: done, the patching is moved out of rules_go, 

fyi, it has the support of letting users preview the diff and pick. At the end, it applies the selected edits. In case of overlapping, we skip the first overlapping one for max progress, the next build can fix that one too.

lpxz:
addressed comments from @linzhp.

made the following design-level change:
- move most logic out of rules_go into a third_party patcher tool (@linzhp let us discuss where we can host this, initially we can put it in my personal github repo, but in the long run, we may want to host it more properly).
- rules_go does the minimal work now to export the []edits for each file and analyzer.

This way, we make sure we do not impact rules_go's performance. 

Accordingly, we shall change the expectation of this PR as follows:
this PR does not provide e2e patching support.  It only exports the fixes. 

ready for review, @linzhp and @fmeum , thanks!

go/tools/builders/nogo_change.go
linzhp: > asking users to pick which analyzer to patch may be too much for developers.

I don't think so. The output from rules_go can be simply:

```
nogo: errors found by nogo during build-time code analysis:
src/example.com/foo/compute.go:28:2: self-assignment of x to x (assign)

To apply the fixes run:

$ patch -p1 < bazel-out/k8-fastbuild/bin/src/example.com/foo/go_default_library.nogo.assign.patch
```

When there are multiple analyzers failing, we can list multiple patch command for users to copy:

```
To apply the fixes run:

$ patch -p1 < bazel-out/k8-fastbuild/bin/src/example.com/foo/go_default_library.nogo.assign.patch
$ patch -p1 < bazel-out/k8-fastbuild/bin/src/example.com/foo/go_default_library.nogo.composite.patch
```

I agree with what @fmeum said here: https://github.com/bazel-contrib/rules_go/pull/4102#discussion_r1779996815

I don't want rules_go to depend on a custom patch tool, while we can use a standard Unix utility.

> Ideally we want them simply say "yes/no".

I don't think users' input matters there. When two fixes conflict, there is nothing the tool can do even the user reply "yes", right? The only thing we can do is to reject the patch and ask user to manually apply the fix, which is exactly what the standard `patch` command does.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Change represents a collection of analyzer edits.
+type Change struct {
+	AnalyzerToEdits map[string]AnalyzerEdits `json:"analyzer_to_edits"` // Analyzer as the key, files-to-edits mapping as the value
+}
+
+// DiagnosticEntry represents a diagnostic entry with the corresponding analyzer.
+type DiagnosticEntry struct {
+	analysis.Diagnostic
```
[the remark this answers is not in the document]
linzhp: Oh, I see, you just moved an existing struct here. That's fine then, but you should still make this struct private, otherwise people may start using it and we will run into issues mentioned in the link above

lpxz:
@linzhp 
let me address your high-level feedback first. 

your proposal does not really work:
$ patch -p1 < bazel-out/k8-fastbuild/bin/src/example.com/foo/go_default_library.nogo.assign.patch
$ patch -p1 < bazel-out/k8-fastbuild/bin/src/example.com/foo/go_default_library.nogo.composite.patch

patch is highly unreliable tool that is based on the surrounding context. After you apply the first patch, the 2nd patching may get the context messed up and is not able to proceed. 

I do not like such unreliable patching. The only reliable way is to perform the precise offset-based editing. This imposes a strict requirement though: the edits users want to apply should not overlap. This means we have to have some tool that is better than patch. 

Also I think this PR is only for exporting edits, right? this is already an improvement over existing status that no fixes can be exported.

go/tools/builders/nogo_main.go
peng3141: discussed offline, users already select which linters to run in the bazel.

go/tools/builders/nogo_change.go
peng3141: done

go/tools/builders/nogo_change.go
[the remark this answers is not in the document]
peng3141: done, simplified.

go/tools/builders/nogo_change_serialization.go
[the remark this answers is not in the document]
peng3141: removed.

go/tools/builders/nogo_change.go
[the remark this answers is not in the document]
peng3141: SplitLines is inevitable, see another patching lib also does it: https://github.com/sergi/go-diff/blob/master/diffmatchpatch/patch.go#L483.

the  problem here is the whitespace prefix or suffix, which is handled by the Trim

go/tools/builders/nogo_change.go
peng3141: discussed offline. 

agreed that we should use standard tool like 'patch -p1'.

we cannot have multiple patch files, each for one analyzer, since the bazel creates a single file for each target. Also it is not easy to tell the analyzers at bazel resolving time. 

what we currently do is:
- for each file, we apply the edits (offset-based) and drop the analyzers that introduce overlapping edits with added edits. Analyzers are sorted for order. After this, we can compute the new content after fixing. Also, we can compute the patch from orig and new content. 
- we merge patches for all files into one, so that "patch -p1 < file" works.

users' inputs do not matter, for the analyzers they pick, they may still have overlapping edits that we need to resolve.

go/private/actions/archive.bzl
```diff
@@ -64,10 +64,21 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         out_facts = go.declare_file(go, name = source.name, ext = pre_ext + ".facts")
         out_nogo_log = go.declare_file(go, name = source.name, ext = pre_ext + ".nogo.log")
         out_nogo_validation = go.declare_file(go, name = source.name, ext = pre_ext + ".nogo")
+
+        # out_nogo_fix_tmp holds the fixes produced by the RunNogo action, out_nogo_fix holds the fixes produced by the ValidateNogo action.
+        # They have the same content, but ValidateNogo propagates the fixes and eventually externalizes the fixes via `_validation` in the OutputGroupInfo section.
+        # --run_validations (default=True) ensures nogo validation is applied to not only the input targets but also their dependent targets,
+        # thereby producing available fixes for all targets.
+        # Otherwise, if we externalize out_nogo_fix_tmp (not going through the ValidateNogo action) by putting it into a field (e.g., `nogo_fix`) in the OutputGroupInfo section of the input targets,
+        # we can see the fix for the input targets, but will miss the fixes for the dependent targets.
```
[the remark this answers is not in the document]
peng3141: question about output_group vs nogo validation action:
what is the benefit of this output_group option over validation?

also, I feel validation is more flexible, e.g., we can dump fixes immediately after error message there.

go/tools/builders/.nogo_change.go.swp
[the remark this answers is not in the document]
peng3141: vim swap file... removed.

go/tools/builders/nogo_change.go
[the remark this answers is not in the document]
peng3141: done

peng3141:
addressed the comments, tested with locally with
- multiple files have errors of different analyzers.

one outstanding question is about output_group vs nogo_validation, I am not sure about advantage of output_group.
also i think nogo_validation offers more flexibility, e.g., it can print patch message immediately after error message.

peng3141:
discussed offline and made these improvements:

- also will add the printing of patch in validation action immediately after the errors. 
- Also add the nogo_fix to output_group support the case of not running validations.

      - tested via $ bazel build --output_groups=nogo_fix --norun_validations 

- there is no need to have two fix files, one for run_nogo and one for validation action. We just need one file. Actually validation does not need to write to it at all.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+				if err != nil {
+					fileRelativePath = file.Name() // fallback logic
+				}
+				c.AddEdit(fileRelativePath, analyzer, edit)
+			}
+		}
+	}
+
+	if len(allErrors) > 0 {
+		return c, fmt.Errorf("errors: %v", allErrors)
```
linzhp, line 203: The errors will be appended to other errors and printed out one by one, one per line. We should either return a list of errors here so they will be appended properly, or format them in the same way as other errors and return one error

go/private/actions/archive.bzl
```diff
@@ -64,10 +64,13 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         out_facts = go.declare_file(go, name = source.name, ext = pre_ext + ".facts")
         out_nogo_log = go.declare_file(go, name = source.name, ext = pre_ext + ".nogo.log")
         out_nogo_validation = go.declare_file(go, name = source.name, ext = pre_ext + ".nogo")
+
+        out_nogo_fix = go.declare_file(go, name = source.name, ext = pre_ext + ".nogo.fix")
```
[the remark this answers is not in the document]
peng3141: good point, done.

go/tools/builders/difflib.go
```diff
@@ -0,0 +1,792 @@
...
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+// This file is copied from https://github.com/pmezard/go-difflib, under the permission of the above copyright.
```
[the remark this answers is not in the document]
peng3141: yes. now fixed. will upload soon.

long story:
it is beyond my capability to add the lib, so i copied it over. 
luckily, @fmeum shared the [diff](https://github.com/bazel-contrib/rules_go/commit/998b738105e4fcefcf432f5d53723826073691d2#diff-79cc840404662d3000bbd76e53f1ba1130554a46223b3b42ca07e59bc4dcd7bbR109) made before, now I copied it over and it works (with minor change to adapt to the recent new_go_info, [GoArchive] changes).

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+		sort.Strings(analyzers)
+
+		mergedEdits := make([]Edit, 0)
+
+		for _, analyzer := range analyzers {
+			edits := fileEdits.AnalyzerToEdits[analyzer]
+
+			// Deduplicate and sort edits for the current analyzer
+			edits, _ = UniqueEdits(edits)
```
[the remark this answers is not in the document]
peng3141: removed. 

we were considering the worst case that a crappy analyzer did so. 
it should not happen if the administrator does a decent job in ruling out bad analyzers.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	return a[i].End < a[j].End
+}
+func (a editsSort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+
+// UniqueEdits returns a list of edits that is sorted and
+// contains no duplicate edits. Returns the index of some
+// overlapping adjacent edits if there is one and <0 if the
+// edits are valid.
+// Deduplication helps in the cases where two analyzers produce duplicate edits.
```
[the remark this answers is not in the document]
peng3141: not really, duplicate and overlapping edit do not have the same effect.

we know how to handle the duplicate edit safely, so we handle it and UniqueEdits still return "valid.
we do not know how to handle overlapping edit, UniqueEdits has to return "invalid", meaning something wrong just happened and we do not know how to handle it.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+			if invalidIndex >= 0 {
+				// Skip edits from this analyzer if merging them would cause overlaps.
+				// Apply the non-overlapping edits first. After that, a rerun of bazel build will
+				// allows these skipped edits to be applied separately.
+				// Note the resolution happens to each file independently.
+				continue
+			}
+
+			// Update the merged edits
+			mergedEdits = candidateEdits
```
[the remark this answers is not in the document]
peng3141: that has no much difference, given that, in either case, the users need to run another build anyway to resolve the overlapping ones from some analyzer. 

I prefer the current design for two reasons:
- simplicity of coding/readability.
- not to confuse users by fixing partially the errors from an analyzer, we either fix all or none for that analyzer.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+func UniqueEdits(edits []Edit) ([]Edit, int) {
+	if len(edits) == 0 {
+		return nil, -1
+	}
+	equivalent := func(x, y Edit) bool {
+		return x.Start == y.Start && x.End == y.End && x.New == y.New
+	}
+	SortEdits(edits)
+	unique := []Edit{edits[0]}
+	invalid := -1
```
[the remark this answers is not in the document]
peng3141: good point, done.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+	for len(lines) > 0 && strings.TrimSpace(lines[len(lines)-1]) == "" {
+		lines = lines[:len(lines)-1]
+	}
+
+	return lines
+}
+
+
+
+func SaveToFile(filename string, combinedPatch string) error {
```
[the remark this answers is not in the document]
peng3141: removed.

SaveToFile used to be a big func, after our discussion and simplification, it no longer needs to exist as a func.

go/tools/builders/nogo_change.go
peng3141: prefer to return a single error, since all these are in one logical group.

formatted same way as in nogo_main.go

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+		// Store the final merged edits for the file
+		fileToEdits[file] = mergedEdits
+	}
+
+	return fileToEdits
+}
+
+
+// ToCombinedPatch converts all edits to a single consolidated patch.
+func ToCombinedPatch(fileToEdits map[string][]Edit) (string, error) {
```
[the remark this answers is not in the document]
peng3141: done. 

I tried to make structs types private too, but that make the type name like variables such as 'edit', 'fileEdits', so I still keep the uppercase for each, to avoid naming collision or accidental use, I prefix them with Nogo.

go/private/rules/library.bzl
```diff
@@ -49,6 +49,13 @@ def _go_library_impl(ctx):
     go_info = new_go_info(go, ctx.attr)
     archive = go.archive(go, go_info)
     validation_output = archive.data._validation_output
+    nogo_fix_output = archive.data._nogo_fix_output
+
+    nogo_validation_outputs = []
+    if validation_output:
+        nogo_validation_outputs.append(validation_output)
+    if nogo_fix_output:
+        nogo_validation_outputs.append(nogo_fix_output)
```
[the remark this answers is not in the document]
peng3141: removed here and in binary.bzl/test.bzl.
tested that bazel build still works in generating the fixes. 

why I added it:
remember we had two fix files, one from run_nogo, one from validation. 
I thought we need to add explicitly the one from validation into the _validation output group:
https://bazel.build/extending/rules#using_the_validations_output_group

now we have a single file, updated in run_nogo, and also validation action runs anyway, there is no need.

go/tools/builders/difflib.go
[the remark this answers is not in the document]
peng3141: actually, now tests start to fail due to missing dep of "@com_github_pmezard_go_difflib//difflib:go_default_library", e.g., 
$ bazel test //tests/core/nogo/config:config_test

do you happen to know how to add dep there? @linzhp

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+		if lastEnd < edit.Start {
+			out = append(out, src[lastEnd:edit.Start]...)
+		}
+		out = append(out, edit.New...)
+		lastEnd = edit.End
+	}
+	out = append(out, src[lastEnd:]...)
+
+	if len(out) != size {
+		return nil, fmt.Errorf("applyEditsBytes: unexpected output size, got %d, want %d", len(out), size)
```
linzhp, line 121: How can this happen?

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+	// Remove trailing newline
+	result := combinedPatch.String()
+	if len(result) > 0 && result[len(result)-1] == '\n' {
+		result = result[:len(result)-1]
+	}
+
+	return result, nil
+}
+
+func trimWhitespaceHeadAndTail(lines []string) []string {
```
linzhp, line 331: Why do we need this? The line numbers in the unified diff will be wrong if we remove the empty lines

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+		if err != nil {
+			return "", fmt.Errorf("failed to generate patch for file %s: %v", filePath, err)
+		}
+
+		// Append the patch for this file to the giant patch
+		combinedPatch.WriteString(patch)
+		combinedPatch.WriteString("\n") // Ensure separation between file patches
+	}
+
+	// Remove trailing newline
```
linzhp, line 322: Why do we need this? The new lines at the end of the patch don't seem harmful. If we don't need to delete the new line, we can pass a `io.Writer` into this function and call `difflib.WriteUnifiedDiff` instead, so we don't need to hold the patch in memory.

go/tools/builders/nogo_change_test.go
```diff
@@ -0,0 +1,786 @@
...
+}
+
+// Mock analyzers for the test
+var (
+	analyzer1 = &analysis.Analyzer{Name: "analyzer1"}
+	analyzer2 = &analysis.Analyzer{Name: "analyzer2"}
+)
+
+// ApplyEdits() and validate() here provide the reference implementation for testing
+// applyEditsBytes() from the refactored nogo_change code, now using NogoEdit.
```
linzhp, line 39: Please don't do this. If this test fail, people have to debug BOTH your test code AND your production code to see which one has a bug.

Instead, please manually prepare an expected output to compare with the production code

go/tools/builders/nogo_change_test.go
```diff
@@ -0,0 +1,786 @@
...
+			sortEdits(tt.edits)
+			if !reflect.DeepEqual(tt.edits, tt.sorted) {
+				t.Fatalf("expected %v, got %v", tt.sorted, tt.edits)
+			}
+		})
+	}
+}
+
+// TestCases uses NogoEdit now instead of Edit
+var TestCases = []struct {
```
linzhp, line 272: Why is this a global variable?

go/private/rules/binary.bzl
```diff
@@ -152,13 +152,19 @@ def _go_binary_impl(ctx):
...
+    nogo_validation_outputs = []
+    if validation_output:
+        nogo_validation_outputs.append(validation_output)
 
     providers = [
         archive,
         OutputGroupInfo(
             cgo_exports = archive.cgo_exports,
             compilation_outputs = [archive.data.file],
-            _validation = [validation_output] if validation_output else [],
```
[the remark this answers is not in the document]
peng3141: done

go/private/rules/library.bzl
```diff
@@ -65,7 +70,8 @@ def _go_library_impl(ctx):
         OutputGroupInfo(
             cgo_exports = archive.cgo_exports,
             compilation_outputs = [archive.data.file],
-            _validation = [validation_output] if validation_output else [],
```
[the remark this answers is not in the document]
peng3141: done

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+type NogoEdit struct {
+	New   string // the replacement
+	Start int    // starting byte offset of the region to replace
+	End   int    // (exclusive) ending byte offset of the region to replace
+}
+
+// NogoFileEdits represents the mapping of analyzers to their edits for a specific file.
+type NogoFileEdits struct {
+	AnalyzerToEdits map[string][]NogoEdit // Analyzer as the key, edits as the value
+}
```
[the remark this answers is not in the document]
peng3141: done, 

NogoEdit -> nogoEdit
NogoFileEdits -> analyzerToEdits
NogoChange -> nogoChange

later, flatten() and toCombinedPatch shared a type alias
type fileToEdits map[string][]nogoEdit // File path as the key, list of nogoEdit as the value

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+	return fmt.Sprintf("{Start:%d,End:%d,New:%q}", e.Start, e.End, e.New)
+}
+
+// sortEdits orders a slice of NogoEdits by (start, end) offset.
+// This ordering puts insertions (end = start) before deletions
+// (end > start) at the same point, but uses a stable sort to preserve
+// the order of multiple insertions at the same point.
+// (applyEditsBytes detects multiple deletions at the same point as an error.)
+func sortEdits(edits []NogoEdit) {
+	sort.Stable(editsSort(edits))
```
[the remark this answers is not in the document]
peng3141: it is a little helper func that hides the implementation details like the type conversion for complying with Sort interface. 

also the comments above can be helpful for readers.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+type editsSort []NogoEdit
```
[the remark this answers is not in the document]
peng3141: sounds good,

changed to byStartEnd, since it is first by start then end (if starts are same).

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+
+func (a editsSort) Len() int { return len(a) }
+func (a editsSort) Less(i, j int) bool {
+	if cmp := a[i].Start - a[j].Start; cmp != 0 {
+		return cmp < 0
+	}
```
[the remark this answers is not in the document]
peng3141: done.

peng3141:
half way, will handle the rest tmr.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+
+	if len(allErrors) > 0 {
+		var errMsg bytes.Buffer
+		sep := ""
+		for _, err := range allErrors {
+			errMsg.WriteString(sep)
+			sep = "\n"
+			errMsg.WriteString(err.Error())
+		}
+		return c, fmt.Errorf("errors:\n%s", errMsg.String())
```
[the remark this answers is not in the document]
peng3141: done

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+		for _, sf := range entry.Diagnostic.SuggestedFixes {
+			for _, edit := range sf.TextEdits {
+				start, end := edit.Pos, edit.End
+				if !end.IsValid() {
+					end = start
+				}
+
+				file := fileSet.File(edit.Pos)
+				if file == nil {
+					allErrors = append(allErrors, fmt.Errorf("invalid fix: missing file info for pos %v", edit.Pos))
```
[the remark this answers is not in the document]
peng3141: good point. done.

go/tools/builders/nogo_change.go
peng3141: it cannot happen. 
code greatly simplified now.

go/tools/builders/nogo_change_test.go
peng3141: done.

go/tools/builders/nogo_change_test.go
peng3141: done.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+	return a[i].End < a[j].End
+}
+func (a editsSort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+
+
+// validateBytes checks that edits are consistent with the src byte slice,
+// and returns the size of the patched output. It may return a different slice if edits are sorted.
+func validateBytes(src []byte, edits []NogoEdit) ([]NogoEdit, int, error) {
+	if !sort.IsSorted(editsSort(edits)) {
+		edits = append([]NogoEdit(nil), edits...)
```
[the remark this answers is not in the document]
peng3141: removed. given we already sort and validate before ApplyEdits.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	return a[i].End < a[j].End
+}
+func (a editsSort) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+
+
+// validateBytes checks that edits are consistent with the src byte slice,
+// and returns the size of the patched output. It may return a different slice if edits are sorted.
+func validateBytes(src []byte, edits []NogoEdit) ([]NogoEdit, int, error) {
+	if !sort.IsSorted(editsSort(edits)) {
```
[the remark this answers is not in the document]
peng3141: code simplified, take a look there.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+	edits, size, err := validateBytes(src, edits)
+	if err != nil {
+		return nil, err
+	}
+
+	// Apply the edits.
+	out := make([]byte, 0, size)
+	lastEnd := 0
+	for _, edit := range edits {
+		if lastEnd < edit.Start {
```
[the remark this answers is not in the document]
peng3141: code simplified. put a summary here:
- NewChangeFromDiagnosticEdits: check each edit is valid, eg.., within the [filestart, fileend).
- uniqueSortedEdits inside flatten ensures sorted, non-overlapping. 

prefer to keep uniqueSortedEdits, since:
- it accepts the edits of an analyzer atomically (either all or none).
- also, it happens early to prune invalid cases.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,350 @@
...
+
+		for _, analyzer := range analyzers {
+			edits := fileEdits.AnalyzerToEdits[analyzer]
+
+			// Deduplicate and sort edits for the current analyzer
+			edits, _ = UniqueEdits(edits)
+
+			// Merge edits into the current list, checking for overlaps
+			candidateEdits := append(mergedEdits, edits...)
+			candidateEdits, invalidIndex := UniqueEdits(candidateEdits)
```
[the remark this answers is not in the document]
peng3141: no, this checks the edits from multiple analyzers for a single file.

go/tools/builders/nogo_change.go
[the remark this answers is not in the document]
peng3141: done, return an error additionally if it applies.

yes, both are partial, so they do not have much difference fundamentally.

The benefit of "accepting all or none for each file" is that, we can have a clear error message to deliver to users, e.g., fixes from some analyzer are dropped for some file.

go/tools/builders/nogo_main.go
```diff
@@ -543,20 +541,39 @@ func checkAnalysisResults(actions []*action, pkg *goPackage) string {
...
 	if numSkipped > 0 {
 		errs = append(errs, fmt.Errorf("%d analyzers skipped due to type-checking error: %v", numSkipped, pkg.typeCheckError))
 	}
+
+	if nogoFixPath != "" {
+		// If the nogo fixes are requested, we need to save the fixes to the file even if they are empty.
+		// Otherwise, bazel will complain "not all outputs were created or valid"
+		change, err := newChangeFromDiagnostics(diagnostics, pkg.fset)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("errors in dumping nogo fix, specifically in converting diagnostics to change: %v", err))
```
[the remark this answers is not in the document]
peng3141: done, similarly handled the following errors that already have enough context from the function.

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+
+		// Store the final merged edits for the file
+		fileToEdits[file] = mergedEdits
+	}
+
+	return fileToEdits
+}
+
+// toCombinedPatch converts all edits to a single consolidated patch.
+func toCombinedPatch(fileToEdits map[string][]NogoEdit) (string, error) {
```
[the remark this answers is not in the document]
peng3141: the main message is toPatch, 
the full name would be combineFixesAsPatch, will leave this to you :)

go/tools/builders/nogo_change.go
```diff
@@ -0,0 +1,344 @@
...
+		return edits, hasOverlap
+	}
+	equivalent := func(x, y NogoEdit) bool {
+		return x.Start == y.Start && x.End == y.End && x.New == y.New
+	}
+	sortEdits(edits)
+	unique := []NogoEdit{edits[0]}
+	for i := 1; i < len(edits); i++ {
+		prev, cur := edits[i-1], edits[i]
+		if !equivalent(prev, cur) { // equivalent ones are safely skipped
```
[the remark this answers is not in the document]
peng3141: done

go/tools/builders/nogo_change.go
peng3141: it is because difflib.SplitLines blindly adds an extra "\n" at the end: https://github.com/pmezard/go-difflib/blob/master/difflib/difflib.go#L766C1-L772C2, I am not sure why it does so (does not look like a bug). 

some examples: 
"line1\nline2\nline3" ->  ["line1\n", "line2\n", "line3\n"].
"line1\nline2\nline3\n" -> ["line1\n", "line2\n", "line3\n\n"].

This unintended additional newline (\n\n) can lead to incorrect diff outputs or unexpected behavior in downstream processes that consume these lines.

there may be a better workaround this.

go/tools/builders/nogo_change.go
peng3141: good point.

go/private/rules/nogo.bzl
```diff
@@ -65,7 +65,7 @@ def _nogo_impl(ctx):
         go,
         struct(
             embed = [ctx.attr._nogo_srcs],
-            deps = analyzer_archives,
+            deps = analyzer_archives + [ctx.attr._go_difflib[GoArchive]],
```
fmeum, line 68: Adding this together with `ctx.attr.deps` in line 49 would be less surprising.

go/private/rules/test.bzl
```diff
@@ -99,6 +102,10 @@ def _go_test_impl(ctx):
     external_archive = go.archive(go, external_go_info, is_external_pkg = True)
     if external_archive.data._validation_output:
         validation_outputs.append(external_archive.data._validation_output)
+    if external_archive.data._nogo_fix_output:
+        # internal vs external archive refers to the same package vs separate package.
+        # we include the nogo fixes for transitive dependency too.
```
fmeum, line 107: What does “transitive dependency" mean here? Aren't we just reporting no-go fixes for both sources of internal and of external tests?

go/tools/builders/nogo_fix.go
```diff
@@ -0,0 +1,241 @@
...
+	return a[i].End < a[j].End
+}
+func (a byStartEnd) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+
+
+// applyEditsBytes applies a sequence of nogoEdits to the src byte slice and returns the result.
+// Edits are applied in order of start offset; edits with the same start offset are applied in the order they were provided.
+// applyEditsBytes returns an error if any edit is out of bounds, or if any pair of edits is overlapping.
+func applyEditsBytes(src []byte, edits []nogoEdit) ([]byte, error) {
+	// assumption: at this point, edits should be unique, sorted and non-overlapping.
```
fmeum, line 64: This somewhat contradicts the method comment. Could it be merged into it?

go/tools/builders/nogo_fix.go
```diff
@@ -0,0 +1,241 @@
...
+	for fileName, edits := range finalChanges {
+		finalFileChanges = append(finalFileChanges, fileChange{fileName: fileName, changes: edits})
+	}
+
+	if len(allErrors) == 0 {
+		return finalFileChanges, nil
+	}
+
+	var errMsg bytes.Buffer
+	errMsg.WriteString("some suggested fixes are invalid or have conflicts with other fixes:")
```
fmeum, line 160: Please add a suggested follow-up such as applying other fixes and rerunning the build.

go/tools/builders/nogo_main.go
```diff
@@ -87,27 +88,65 @@ func run(args []string) (error, int) {
...
+	}
+	return nil, exitCode
+}
+
+func saveSuggestedFixes(nogoFixPath string, diagnostics []diagnosticEntry, pkg *goPackage) []error {
+	if nogoFixPath == "" {
+		return nil
+	}
+	var errs []error
+	// the patch file has be created even if there is no fix.
```
fmeum, line 135:
```suggestion
	// the patch file has to be created even if there is no fix.
```

go/tools/builders/nogo_validation.go
```diff
@@ -8,21 +8,42 @@ import (
...
 	}
+
+	fixContent, err := os.ReadFile(fixFile)
+	if err != nil {
+		return err
+	}
+
 	if len(logContent) > 0 {
+		var fixMessage string
+		// See nogo_change_serialization.go, if the patches are empty, then fixContent is empty by design, rather than an empty json like {}.
```
fmeum, line 31: This file doesn't seem to exist

fmeum (approved):

go/tools/builders/nogo_validation.go
linzhp: Removed. The code is self-explanatory. No comment needed

go/private/rules/nogo.bzl
linzhp: We can't do that. `analyzer_archives` from Line 49 is used to populate `analyzer_importpaths` in Line 50. We don't want `_go_difflib` to be in `analyzer_importpaths`, because it cannot be imported as a nogo analyzer

go/private/rules/test.bzl
linzhp: Removed the comment. The code is self-explanatory enough

fmeum (approved):

go/tools/builders/nogo_change.go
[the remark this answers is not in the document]
peng3141: marking conservations as resolved here and after.

go/tools/builders/nogo_fix.go
```diff
@@ -0,0 +1,238 @@
...
+		if !foundApplicableFix {
+			allErrors = append(allErrors, fmt.Errorf(
+				"ignoring suggested fixes from analyzer %q at %s",
+				entry.analyzerName, fileSet.Position(entry.Pos),
+			))
+		}
+	}
+
+	var finalFileChanges []fileChange
+	for fileName, edits := range finalChanges {
```
peng3141, line 150: iteration over a map is in random order, as a result, finalFileChanges may be in different orders, leading to different patch files. 

shall we sort them to avoid non-determistic results?

go/tools/builders/nogo_fix.go
```diff
@@ -0,0 +1,238 @@
...
+					New: string(edit.NewText),
+					analyzerName: entry.analyzerName,
+				}
+				candidateChanges[file.Name()] = append(candidateChanges[file.Name()], fix)
+			}
+			// validating the edits from current SuggestedFix. All edits from a SuggestedFix must be
+			// either accepted or discarded atomically, because a SuggestedFix may move a statement from one place
+			// to the other. If we only accept part of the edits, the statement may either appear twice or disappear.
+			for fileName, edits := range candidateChanges {
+				edits = append(edits, finalChanges[fileName]...)
```
peng3141, line 125: in terms of performance, this shall be faster:
the rationale is that, finalChanges[fileName] is a growing list that is sorted already (see validate func), appending few edits to a sorted list can make the new sorting in validate fast.  

combinedEdits := make([]Edit, len(finalChanges[fileName])+len(edits))
copy(combinedEdits, finalChanges[fileName])               // Copy finalChanges[fileName] first
copy(combinedEdits[len(finalChanges[fileName]):], edits)  // Append edits
edits = combinedEdits

go/tools/builders/nogo_main.go
```diff
@@ -87,27 +88,65 @@ func run(args []string) (error, int) {
...
+	return nil, exitCode
+}
+
+func saveSuggestedFixes(nogoFixPath string, diagnostics []diagnosticEntry, pkg *goPackage) []error {
+	if nogoFixPath == "" {
+		return nil
+	}
+	var errs []error
+	// the patch file has to be created even if there is no fix.
+	patchFile, err := os.Create(nogoFixPath)
```
peng3141, line 136: `patchFile, err := os.OpenFile(nogoFixPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)`

the previously working code wrote to the file, assuming it is created. 

https://github.com/bazel-contrib/rules_go/pull/4212/commits/8f837f7530c36f040cf4d281c3f951744b1cf70e#diff-feece2c0b8a961ba0b9b6054398956c1841470bbe2a72b3eecc0f78c3cedeb3eL567

peng3141: followed the [way](https://github.com/bazel-contrib/rules_go/pull/4102/files#diff-feece2c0b8a961ba0b9b6054398956c1841470bbe2a72b3eecc0f78c3cedeb3eR98) the facts file was written to.

go/tools/builders/nogo_fix.go
peng3141: nvm, saw there is sorting in writePatch.

peng3141:
after fixing the above IO permission problem above, we will see new issues for MacOS Arm64: https://buildkite.com/bazel/rules-go-golang/builds/7415#01942a79-a909-4c01-acca-4fb018ca0b8b

https://github.com/bazel-contrib/rules_go/pull/4203 seems to be relevant, since after reverting the change like [this](https://github.com/bazel-contrib/rules_go/pull/4212/commits/f581f9497756e639ab20e314870792bd1c9b9500), the PR passed all: https://github.com/bazel-contrib/rules_go/pull/4212/

will look into https://github.com/bazel-contrib/rules_go/pull/4203. My guess is that the selection of toolchain may not work properly for macOS arm64.

go/tools/builders/nogo_fix.go
peng3141: this is an optional nit.

go/private/repositories.bzl
```diff
@@ -102,6 +102,25 @@ def go_rules_dependencies(force = False):
...
     )
 
+    # Needed for nogo to generate unified diff
+    # releaser:upgrade-dep pmezard go-difflib
+    wrapper(
+        http_archive,
+        name = "com_github_pmezard_go_difflib",
+        # v1.0.0, latest as of 2024-12-19
+        urls = [
+            "https://mirror.bazel.build/github.com/pmezard/go-difflib/archive/refs/tags/v1.0.0.tar.gz",
```
peng3141, line 112: this link leads to nothing. 

based on the URL, this seems to be normal, since it is like a cache and we have not populated the entry yet.

go/tools/builders/nogo_main.go
peng3141: it is an issue of CI infra:

https://github.com/bazelbuild/continuous-integration/issues/2153
https://github.com/bazel-contrib/rules_go/pull/4216

go/private/repositories.bzl
peng3141: @linzhp looks like they suggest this: GitHub doesn't guarantee a stable checksum of source archives in the form of https://github.com/<org>/<repo>/archive/...

see more: https://blog.bazel.build/2023/02/15/github-archive-checksum.html

i think it is fine, since https://github.com/pmezard/go-difflib/archive/refs/tags/v1.0.0.tar.gz is not dynamically generated, it has not been changed since 2016.

peng3141: created the mirror request here: https://github.com/bazelbuild/bazel/issues/24824

fmeum: All these archives are dynamically generated by GitHub, hence the warning. It's definitely better to use a mirror URL if we can.

peng3141:
landing is blocked by the central CI infra issue: https://github.com/bazelbuild/continuous-integration/issues/2153.

go/tools/builders/nogo_fix.go
linzhp: >  this shall be faster

Not necessarily:
1. if `edits` has enough capacity to hold `finalChanges[fileName]`, my implementation is faster, because it doesn't need to copy items in `edits`
2. if `edits` doesn't have enough capacity, reallocations will happen. If one reallocation happens, the speed for both implementation will be the same, because they both involves creating another array and copying all items in `edits` and `finalChanges[fileName]` over. Your implementation is only faster only when multiple reallocations happen.

Since we don't know which case is more frequent, I prefer a simpler implementation

Issue #4113, opened 2024-09-21
gopackagesdriver strips all compiled files in the CompiledGoFiles list.


lbcjbb:
The list returned the `stdliblist` action contains well several compiled files:

``` bash
$ bazel build \
     --aspects=@rules_go//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect \
     --output_groups=go_pkg_driver_stdlib_json_file \
     @rules_go//:stdlib
$ jq -r 'select(.Name == "net") | .CompiledGoFiles[]' <'bazel-bin/external/rules_go+/stdlib_/stdlib.pkg.json'
...
__BAZEL_OUTPUT_BASE__/external/rules_go++go_sdk+go_sdk/src/net/unixsock_readmsg_cmsg_cloexec.go
__BAZEL_OUTPUT_BASE__/external/rules_go++go_sdk+go_sdk/src/net/writev_unix.go
__BAZEL_EXECROOT__/bazel-out/k8-fastbuild/bin/external/rules_go+/stdlib_/gocache/70/707ee66f89be0179bf372c6598510d4c2595e7f6ae85448b325dac26399a3540-d
__BAZEL_EXECROOT__/bazel-out/k8-fastbuild/bin/external/rules_go+/stdlib_/gocache/21/21a53368c5f9c496f433c668c3801fabbcec9dd92c5f35a21f3529ee6c8377a7-d
__BAZEL_EXECROOT__/bazel-out/k8-fastbuild/bin/external/rules_go+/stdlib_/gocache/95/95f68a9a66a06f04518b91afd6be0f383df32608e04dbf489e1ccc385824c963-d
__BAZEL_EXECROOT__/bazel-out/k8-fastbuild/bin/external/rules_go+/stdlib_/gocache/27/2765873aea41cf96ec1b48d36012b1fbd6ec5e0028d4a84799f9cc00facf6b96-d
__BAZEL_EXECROOT__/bazel-out/k8-fastbuild/bin/external/rules_go+/stdlib_/gocache/8d/8db549e22eaae16aa05c6cbffa30967189d31835d293b054696ee35dfe8eb0eb-d
__BAZEL_EXECROOT__/bazel-out/k8-fastbuild/bin/external/rules_go+/stdlib_/gocache/17/17018c2c3907184b0b56e4f58e17d4d1d7ce403f6c54b8a72e25ebcf5a91b258-d
[...]

PR #4114, opened 2024-09-21
fix(gpd): keep compiled files from stdlib packages

lbcjbb:
**What type of PR is this?**

Bug fix

When we create a new dedicated package for all external tests, we should not override the list of CompiledGoFiles by the list of GoFiles.

Fixes: #4113

### DIFF
--- a/go/tools/gopackagesdriver/gopackagesdriver_test.go
+++ b/go/tools/gopackagesdriver/gopackagesdriver_test.go
@@ -4,18 +4,15 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 	"testing"
+
 	"github.com/bazelbuild/rules_go/go/tools/bazel_testing"
 )
-
-type response struct {
-	Roots    []string `json:",omitempty"`
-	Packages []*FlatPackage
-}
 
 func TestMain(m *testing.M) {
 	bazel_testing.TestMain(m, bazel_testing.Args{
@@ -81,6 +78,29 @@ func main() {
 	fmt.Fprintln(os.Stderr, "Subdirectory Hello World!")
 }
 		`,
+		// Explicitly disable bzlmod which is enabled by default with bazel >=7.0.
+		// Remove this setup function when the test will be launched with bzlmod and
+		// the RULES_GO_REPO_NAME_FOR_TEST will be the canonical name of this rules_go module.
+		// This is required to match the stdlib label returned by the 'bazel query --consistent_labels',
+		// otherwise the gopackagesdriver will not return any stdlib packages.
+		SetUp: func() error {
+			f, err := os.OpenFile(".bazelrc", os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
+			if err != nil {
+				return fmt.Errorf("can't open bazelrc file: %v", err)
+			}
+
+			_, err = fmt.Fprintf(f, "common --noexperimental_enable_bzlmod\n")
+			if err != nil {
+				_ = f.Close()
+				return fmt.Errorf("can't update bazelrc file: %v", err)
+			}
+
+			if err := f.Close(); err != nil {
+				return fmt.Errorf("can't close bazelrc file: %v", err)
+			}
+
+			return nil
+		},
 	})
 }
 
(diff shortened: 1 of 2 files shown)


go/tools/gopackagesdriver/gopackagesdriver_test.go
```diff
@@ -81,6 +78,29 @@ func main() {
 	fmt.Fprintln(os.Stderr, "Subdirectory Hello World!")
 }
 		`,
+		// Explicitly disable bzlmod which is enabled by default with bazel >=7.0.
+		// Remove this setup function when the test will be launched with bzlmod and
+		// the RULES_GO_REPO_NAME_FOR_TEST will be the canonical name of this rules_go module.
+		// This is required to match the stdlib label returned by the 'bazel query --consistent_labels',
+		// otherwise the gopackagesdriver will not return any stdlib packages.
+		SetUp: func() error {
+			f, err := os.OpenFile(".bazelrc", os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
```
fmeum, line 87: Does this write into the runfiles tree? Could we pass an explicit arg to all Bazel invocations instead?

go/tools/gopackagesdriver/gopackagesdriver_test.go
```diff
@@ -89,6 +109,33 @@ const (
...
+
+	for _, pkg := range resp.Packages {
+		// net, plugin and user stdlib packages seem to have compiled files only for Linux.
+		if pkg.Name != "cgo" {
+			continue
+		}
+
+		var hasCompiledFiles bool
+		for _, x := range pkg.CompiledGoFiles {
+			if filepath.Ext(x) == "" {
```
fmeum, line 127: Why is an empty extension the correct condition here?

go/tools/gopackagesdriver/gopackagesdriver_test.go
lbcjbb: Yes. I append a new config param in the existing `.bazelrc` file created by [go/tools/bazel_testing/bazel_testing.go#L326-L338](https://github.com/bazelbuild/rules_go/blob/master/go/tools/bazel_testing/bazel_testing.go#L326-L338).
I couldn't use ` GO_BAZEL_TEST_BAZELFLAGS` because it's only for the 'bazel build' command.

But now Bazel is able to ignore incompatible parameters for all commands, I guess we could modify `bazel_testing`  as below:

``` diff
diff --git a/go/tools/bazel_testing/bazel_testing.go b/go/tools/bazel_testing/bazel_testing.go
index a414b40c..093231a4 100644
--- a/go/tools/bazel_testing/bazel_testing.go
+++ b/go/tools/bazel_testing/bazel_testing.go
@@ -331,7 +331,7 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
                fmt.Fprintf(bazelrcBuf, "startup --output_user_root=%s\n", outputUserRoot)
        }
        if flags := os.Getenv("GO_BAZEL_TEST_BAZELFLAGS"); flags != "" {
-               fmt.Fprintf(bazelrcBuf, "build %s\n", flags)
+               fmt.Fprintf(bazelrcBuf, "common %s\n", flags)
        }
        if err := os.WriteFile(bazelrcPath, bazelrcBuf.Bytes(), 0666); err != nil {
                return "", cleanup, err

````
In this way, we could use `GO_BAZEL_TEST_BAZELFLAGS` for this specific test.

go/tools/gopackagesdriver/gopackagesdriver_test.go
lbcjbb: Maybe I didn't understand what you meant, but no, it's precisely the opposite. The test fails if the driver response returns an empty list of packages for the stdlib.

lbcjbb: > Why is an empty extension

Ok, I've just understood your question, sorry. This condition (empty file extension) is already used in `gopackagesdriver` to identify this kind of compiled file:

- [tools/gopackagesdriver/build_context.go#L26-L31](https://github.com/bazelbuild/rules_go/blob/master/go/tools/gopackagesdriver/build_context.go#L26-L31)

go/tools/gopackagesdriver/gopackagesdriver_test.go
fmeum: Could you send a PR to rules_go with this change? We can leave this as is here and simplify it after the next update.

go/tools/gopackagesdriver/gopackagesdriver_test.go
fmeum: Thanks for the explanation

fmeum (approved):

Issue #4127, opened 2024-10-02
Go rules run cpp toolchain from cgo_context_data on an incompatible execution platform


jesses-canva:
### What version of rules_go are you using?

v0.47

### What version of gazelle are you using?

v0.36.0

### What version of Bazel are you using?

7.3.0

### Does this issue reproduce with the latest releases of all the above?

Untested

### What operating system and processor architecture are you using?

Ubuntu 22.04 x86_64

### Any other potentially useful information about your toolchain?

Remote execution

### What did you do?

We have a remote execution set up with two execution platforms (registered with `--host_platform` and `--extra_execution_platforms`). There is a Go toolchain registered and compatible with both execution platforms, however only the second execution platform has a compatible cpp toolchain registered.

```
Execution platform    | Available toolchains
----------------------+----------------------------------------------------------------------------
1. A                  | @io_bazel_rules_go//go:toolchain
2. B                  | @io_bazel_rules_go//go:toolchain, @bazel_tools//tools/cpp:toolchain_type
```

What we observed is that `rules_go` will try to use the cpp toolchain, that is only compatible with the second execution platform, in an action running on the first one, failing because the cpp toolchain isn't installed on that platform.

Note the execution platform ordering is important, the error we get is because Bazel prefers the first one if it thinks it is compatible with the action.

### What did you expect to see?

Successful build

### What did you see instead?

```
ERROR: /var/lib/blah/bazel/a8584ebfb3d6ff0dfe61abfbfa5bb4d3/external/io_bazel_rules_go/BUILD.bazel:42:7: GoStdlib external/io_bazel_rules_go/stdlib_/pkg failed: (Exit 1): builder failed: error executing GoStdlib command (from target @@io_bazel_rules_go//:stdlib)
...
cgo: C compiler "/usr/bin/clang-13" not found: exec: "/usr/bin/clang-13": stat /usr/bin/clang-13: no such file or directory
```

## Discussion

[...]

PR #4128, opened 2024-10-02
Add dependency on cpp toolchain to rules that also depend on cgo_context_data

jesses-canva:
**What type of PR is this?**

Bug fix

This diff fixes the below issue by adding a cpp toolchain dependency to all the rules that depend on `cgo_context_data`, so they also have their execution platform constrained to the platforms compatible with the selected cpp toolchain.

Fixes https://github.com/bazelbuild/rules_go/issues/4127

**Other notes for review**

A more ideal fix for this problem would be to turn `cgo_context_data` into a toolchain, however that would be a bigger diff.

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -132,6 +132,12 @@ _UNSUPPORTED_FEATURES = [
     "rules_go_unsupported_feature",
 ]
 
+# In pure mode, a C++ toolchain isn't needed when transitioning.
+# But if we declare a mandatory toolchain dependency here, a cross-compiling C++ toolchain is required at toolchain resolution time.
+# So we make this toolchain dependency optional, so that it's only attempted to be looked up if it's actually needed.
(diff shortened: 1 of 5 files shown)


jesses-canva:
Oh, this fails CI because it is testing it with Bazel 5.4.0 which doesn't support optional toolchains. 😕

fmeum:
This is a change that I have wanted to make for a long time, thanks for sending it!

I would be in favor of obtaining `cgo_context_info` from the toolchain directly, even if that results in a larger diff. I think that it will make the user-visible parts of this change less confusing by not mixing toolchains in different configurations (that of the context dep and that of the current target).

jesses-canva:
@fmeum Yep that would also work, I don't think I will be able to spend any more time on this for a while though.

dzbarsky:
I think Fabian fixed this but please yell at me if I got it wrong :)

PR #4134, opened 2024-10-07
BUGFIX: Adds the static libraries from cgo_deps to the link command line

filmil:
Not sure why this wasn't added here.

**What type of PR is this?**

Bug fix repro.
At least I think.

It adds the `.a` libraries when linking a go program with C or C++ FFI archive.
I am not seeing why this is not being done when linking the binary.

An alternative is to add `linkmode = "c-shared"`, or `linkmode = "c-archive"`, but this is not correct if you want an actual executable, and not a library.

Fixes #

**Other notes for review**


fmeum:
Could you add a test case that exercises this logic and fails without it? cgo is complicated enough that I wouldn't want to make a change without a test.

filmil:
Sorry, this is not a polished fix. This just illustrates what seems to be a glaring omission in how cgo binaries are linked.
I am not 100% sure what *the* correct fix should be.
Do y'all have someone who's well versed in the innards of link.bzl who could advise?

PR #4140, opened 2024-10-09
Make it easier to parse rules_go's .bzl files with Go Starlark

EdSchouten:
**What type of PR is this?**

Feature

I'm doing some experiments, where I'm trying to parse rules_go's source code with Go Starlark so that I can analyze it. In the process I've noticed that there are some small constructs that Go Starlark doesn't seem to like. In both those cases It looks like Go Starlark works as intended.

n/a

**Other notes for review**


EdSchouten:
Dear lord. Bazel itself doesn't support \u, even though it was added to Starlark three years ago:

https://github.com/bazelbuild/starlark/commit/7f53743af10f87f0915a2e941baf2e795ea8f3e3

Let me remove the commit that was converting \x to \u for now.

fmeum (approved):

PR #4144, opened 2024-10-15
fix: properly expand absolute path placeholder for link calls

### LINKED ISSUE #3886: cgo: `go link` detects all linker flags as unsupported when using custom cc toolchain

master sync'd to latest

```
mike.danese@ip-10-110-30-195:~/rules_go$ git show
commit 36e04e9f1e9b56c865332edfb49e1620beaedf2c
```

```
mike.danese@ip-10-110-30-195:~/rules_go$ bazel version
Bazelisk version: development
Build label: 7.1.0
Build time: Mon Mar 11 17:55:31 2024 (1710179731)
Build timestamp: 1710179731
Build timestamp as int: 1710179731
mike.danese@ip-10-110-30-195:~/rules_go$ git show
commit 36e04e9f1e9b56c865332edfb49e1620beaedf2c
```

Linux

### What did you see instead?

I have a custom cc toolchain based on LLVM. This toolchain sets a sysroot link option, e.g. `--sysroot=external/sysroot-linux-x86_64-gnu7-llvm-1`.

When a cgo library is generated, the cgo tool converts this LD_FLAG value into an absolute path here: https://cs.opensource.google/go/go/+/master:src/cmd/cgo/main.go;l=353-357;drc=bdccd923e914ab61d77a8f23a3329cf1d5aaa7c1. For example, my sysroot ldflag might be converted to `/home/mike.danese/.cache/bazel/_bazel_mike.danese/71e455921cc87b7db5489570176e2cff/sandbox/linux-sandbox/400/execroot/universe/external/sysroot-linux-x86_64-gnu7-llvm-16`. These LD_FLAGs get backed into the object.

Now when it's time to link, `go tool link` probes the linker (in `linkerFlagSupported`) to determine if e.g. `-no-pie` should be configured: https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/ld/lib.go;l=1833-1834;drc=2ab9218c86ed625362df5060f64fcd59398a76f3.  It does this by compiling a trivial program. To construct the compile command, the ldflags from the cgo object are appended to the extdld flags: https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/ld/lib.go;l=2079-2080;drc=2ab9218c86ed625362df5060f64fcd59398a76f3. This results in a command like this:

```
external/llvm-toolchain-16-0-4-linux-x86_64/bin/clang \
  -m64 -fuse-ld=lld -Wl,--build-id=md5 -Wl,--hash-style=gnu -Wl,-z,relro,-z,now -Wl,-z,separate-code \
  --sysroot=external/sysroot-linux-x86_64-gnu7-llvm-16 -pthread -fuse-ld=lld \
  -Wl,--build-id=md5 -Wl,--hash-style=gnu -Wl,-z,relro,-z,now -Wl,-z,separate-code \
  --sysroot=/home/mike.danese/.cache/bazel/_bazel_mike.danese/71e455921cc87b7db5489570176e2cff/sandbox/linux-sandbox/423/execroot/universe/external/sysroot-linux-x86_64-gnu7-llvm-16 \
  -pthread -o /tmp/go-link-606451175/a.out -no-pie /tmp/go-link-606451175/trivial.c
```

`sysroot` arg is passed twice, once from --extld flags and once from the go_ldflags attribute in the cgo object so the second `--sysroot` wins. This is a problem because the second sysroot references a path in the no longer existant execution sandbox where the cgo object was compiled. Thus, the probe fails with a nonsensical error:

```
ld.lld: error: cannot open crt1.o: No such file or directory
ld.lld: error: cannot open crti.o: No such file or directory
ld.lld: error: cannot open crtbegin.o: No such file or directory
ld.lld: error: unable to find library -lgcc
ld.lld: error: unable to find library -lgcc_s
ld.lld: error: unable to find library -lpthread
ld.lld: error: unable to find library -lc
ld.lld: error: unable to find library -lgcc
ld.lld: error: unable to find library -lgcc_s
ld.lld: error: cannot open crtend.o: No such file or directory
ld.lld: error: cannot open crtn.o: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This error is swallowed silently, and `go link` assumes that the linker flag is unsupported. In the case of `-no-pie`, this causes rules_go to attempt to link PIE cgo objects into go binaries without relro (e.g. in normal exe buildmode on linux). This causes hard to debug runtime errors such as:

```
runtime: pcHeader: magic= 0xfffffff1 pad1= 0 pad2= 0 minLC= 1 ptrSize= 8 pcHeader.textStart= 0x1be100 text= 0x557164240100 pluginpath=
fatal error: invalid function symbol table
runtime: panic before malloc heap initialized

runtime stack:
runtime.throw({0x5571640f265a?, 0x0?})
        GOROOT/src/runtime/panic.go:1077 +0x5c fp=0x7fff932345a8 sp=0x7fff93234578 pc=0x55716427741c
runtime.moduledataverify1(0x8?)
        GOROOT/src/runtime/symtab.go:533 +0x816 fp=0x7fff932346c8 sp=0x7fff932345a8 pc=0x5571642958f6
runtime.moduledataverify(...)
        GOROOT/src/runtime/symtab.go:519
runtime.schedinit()
        GOROOT/src/runtime/proc.go:726 +0x4c fp=0x7fff93234710 sp=0x7fff932346c8 pc=0x55716427af6c
runtime.rt0_go()
        src/runtime/asm_amd64.s:349 +0x11c fp=0x7fff93234718 sp=0x7fff93234710 pc=0x5571642aa17c
```

There are ~10 other calls to `linkerFlagSupported` and none of them are correctly detecting flag support.

voxeljorge:
**What type of PR is this?**

> Bug fix

PR #4009 added a mechanism that uses a sentinel value as a placeholder for paths which need to be expanded to absolutes for compile operations. These paths end up embedded in `.a` files and come back to the linker during link operations. The previous PR replaces the `CC` tool with a pointer to `builder cc` which in turn substitutes the placeholder value for actual paths. Unfortunately this wasn't done for the linker, so the linker is receiving parameters like `--sysroot=__GO_BAZEL_CC_PLACEHOLDER__external/_main~_repo_rules~ubuntu_20_04_sysroot/` which is not valid. When a sysroot is configured this means that internally the linker won't be able to find the sysroot in some cases.

In almost comically bad luck, this *only* affects the `go tool link` calls to `linkerFlagSupported` because for some reason `linkerFlagSupported` puts the `-extldflags` at the *beginning* of the linker command, while the main host link command has them at the end. So this means the linker detects all flags as unsupported but then proceeds to run the linker "correctly" which results in really confusing linker errors. In my case this resulted in the linker trying to run with `-rdynamic` incorrectly.

Fixes #3886 

It could be reasonable to open a bug against `go tool link` describing the difference between the host link and the flag checks. I suspect it's not really intended behavior and was just sort of happenstance that things ended up that way.

I also really have no idea how to write a test for this. I'm willing to give it a shot but would really appreciate a suggestion as to where/how.

### DIFF
--- a/go/tools/builders/cc.go
+++ b/go/tools/builders/cc.go
@@ -13,12 +13,13 @@ import (
 func cc(args []string) error {
 	cc := os.Getenv("GO_CC")
 	if cc == "" {
-		errors.New("GO_CC environment variable not set")
+		return errors.New("GO_CC environment variable not set")
 	}
 	ccroot := os.Getenv("GO_CC_ROOT")
 	if ccroot == "" {
-		errors.New("GO_CC_ROOT environment variable not set")
+		return errors.New("GO_CC_ROOT environment variable not set")
 	}
+
 	normalized := []string{cc}
 	normalized = append(normalized, args...)
 	transformArgs(normalized, cgoAbsEnvFlags, func(s string) string {
@@ -26,10 +27,10 @@ func cc(args []string) error {
 			trimmed := strings.TrimPrefix(s, cgoAbsPlaceholder)
 			abspath := filepath.Join(ccroot, trimmed)
 			if _, err := os.Stat(abspath); err == nil {
-                                // Only return the abspath if it exists, otherwise it
-                                // means that either it won't have any effect or the original
-                                // value was not a relpath (e.g. a path with a XCODE placehold from
-                                // macos cc_wrapper)
+				// Only return the abspath if it exists, otherwise it
+				// means that either it won't have any effect or the original
+				// value was not a relpath (e.g. a path with a XCODE placehold from
+				// macos cc_wrapper)
 				return abspath
 			}
 			return trimmed
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -160,6 +161,86 @@ func (e *env) runCommandToFile(out, err io.Writer, args []string) error {
 	cmd.Stdout = out
 	cmd.Stderr = err
 	return runAndLogCommand(cmd, e.verbose)
+}
+
+func symlinkBuilderCC() (func(), string, error) {
+	absPath, err := filepath.Abs(os.Args[0])
+	if err != nil {
+		return nil, "", err
+	}
+
+	dirname, err := os.MkdirTemp("", "")
+	if err != nil {
+		return nil, "", err
+	}
+
+	// re-use the file extension if there is one (mostly for .exe)
+	builderPath := filepath.Join(dirname, "builder-cc"+filepath.Ext(os.Args[0]))
+
+	if err := os.Symlink(absPath, builderPath); err != nil {
+		_ = os.RemoveAll(dirname)
+		return nil, "", err
+	}
+
+	return func() { _ = os.RemoveAll(dirname) }, builderPath, nil
+}
+
+var symlinkBuilderName = regexp.MustCompile(`.*builder-(\w+)(\.exe)?$`)
+
+// verbFromName parses the builder verb from the builder name so
+// that the builder+verb can be invoked in a way that allows the verb
+// to be part of the program name, rather than requring it to be in the
+// arg list. this makes it easier to set a program like `builder cc` as
+// an environment variable in a way that is more compatible
+func verbFromName(arg0 string) string {
+	matches := symlinkBuilderName.FindAllStringSubmatch(arg0, 1)
+	if len(matches) != 1 {
+		return ""
+	}
+
+	if len(matches[0]) < 2 {
+		return ""
+	}
+
+	return matches[0][1]
+}
+
+// absCCLinker sets the target of the `-extld` to a name which `verbFromName`
+// can extract correctly, allowing go tool link to be able to use the `builder cc` wrapper
+func absCCLinker(argList []string) (func(), error) {
+	extldIndex := -1
+	for i, arg := range argList {
+		if arg == "-extld" && i+1 < len(argList) {
+			extldIndex = i + 1
+			break
+		}
+	}
+	if extldIndex < 0 {
+		// if we don't find extld just return
+		// a noop cleanup function
+		return func() {}, nil
+	}
+
+	cleanup, buildercc, err := symlinkBuilderCC()
+	if err != nil {
+		return nil, err
+	}
+
+	err = os.Setenv("GO_CC", argList[extldIndex])
+	if err != nil {
+		cleanup()
+		return nil, err
+	}
+
+	err = os.Setenv("GO_CC_ROOT", abs("."))
+	if err != nil {
+		cleanup()
+		return nil, err
+	}
+
+	argList[extldIndex] = buildercc
+
+	return cleanup, nil
 }
 
 // absCCCompiler modifies CGO flags to workaround relative paths.
--- a/go/tools/builders/link.go
+++ b/go/tools/builders/link.go
@@ -141,6 +141,19 @@ func link(args []string) error {
 	}
 	goargs = append(goargs, "-o", *outFile)
 
+	// substitute `builder cc` for the linker with a symlink to builder called `builder-cc`.
+	// unfortunately we can't just set an environment variable to `builder cc` because
+	// in `go tool link` the `linkerFlagSupported` [1][2] call sites used to determine
+	// if a linker supports various flags all appear to use the first arg after splitting
+	// so the `cc` would be left off of `builder cc`
+	//
+	//    [1]: https://cs.opensource.google/go/go/+/ad7f736d8f51ea03166b698256385c869968ae3e:src/cmd/link/internal/ld/lib.go;l=1739
+	//    [2]: https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/ld/lib.go;drc=c6531fae589cf3f9475f3567a5beffb4336fe1d6;l=1429?q=linkerFlagSupported&ss=go%2Fgo
+	linkerCleanup, err := absCCLinker(toolArgs)
+	if err != nil {
+		return err
+	}
+	defer linkerCleanup()
 	// add in the unprocess pass through options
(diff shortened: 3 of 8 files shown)


voxeljorge:
Clearly this naive approach won't work on windows, I guess we could detect the OS write an OS specific script? This all feels pretty messy so i'm hoping someone out there has ideas for a better solution.

go/tools/builders/cc.go
```diff
@@ -13,12 +13,13 @@ import (
 func cc(args []string) error {
 	cc := os.Getenv("GO_CC")
 	if cc == "" {
-		errors.New("GO_CC environment variable not set")
```
voxeljorge, line 16: this looks like just an error in the original code, `errors.New`here did nothing and was effectively just ignoring the error.

voxeljorge:
One idea for a different approach would be to use a symlink named something like `builder-cc` and have the builder inspect `os.Argv[0]` to set the command.

voxeljorge:
I've changed this to use a symlink approach instead of a script which seems to be more portable.

go/tools/builders/cc.go
fmeum: Nice catch. I think that you can just return the error here instead. I wonder why nogo didn't catch this.

go/tools/builders/link.go
```diff
@@ -141,6 +141,16 @@ func link(args []string) error {
 	}
 	goargs = append(goargs, "-o", *outFile)
 
+	// substitute `builder cc` for the linker with a script. unfortunately
+	// we can't just set an environment variable to `builder cc` because
+	// in `go tool link` the `linkerFlagSupported` call sites used to determine
```
fmeum, line 146: Could you add some references to the source code here? That would make it easier to look for alternative solutions.

fmeum (commented):
Thanks for picking this up, it looks genuinely tricky. I'll think about the approach.

Could you file an upstream issue and link it here? I agree that this is pretty clearly unintended behavior and having a fix in Go would allow us to drop complicated workarounds at some point.

fmeum:
Could you add a test for this under `tests/core/cgo`? It doesn't have to be a full end-to-end test, but perhaps you could set some `--copts` that are turned into absolute paths and assert on the build output with `-s`? That way you don't have to set up an actual sysroot in the test.

voxeljorge:
Sure, I will do both.

I honestly think even if an upstream issue is fixed there is still technically a bug here, we just won't hit the bug for sysroots because the last one will take precedence but it is conceivable that other parameters with paths wouldn't work the same way and we would end up with some new subtle bug.

go/tools/builders/link.go
voxeljorge: I'm not quite sure if you want references to the code in the comment above or just here in github but here's the relevant line:

https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/ld/lib.go;l=1732?q=hostlink

You can see the call to `linkerFlagSupported()` uses `argv[0]` as the `linker`. `argv` is set up here:

https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/ld/lib.go;drc=f2d118fd5f7e872804a5825ce29797f81a28b0fa;l=479?q=hostlink

This uses the global `flagExtLd` which is a `quoted.Flag`, a type defined here:

https://cs.opensource.google/go/go/+/master:src/cmd/internal/quoted/quoted.go;drc=f2d118fd5f7e872804a5825ce29797f81a28b0fa;l=107

This type is a `[]string` which means that `flag.Var` will parse this value into a slice of strings, resulting in something like `[]string{"/path/to/builder", "cc"}`.

voxeljorge:
It looks like someone else already filed an issue relating to this:

https://github.com/golang/go/issues/69350

go/tools/builders/cc.go
voxeljorge: fixed

voxeljorge:
@fmeum I've added a test. I think the test most likely only works on platforms with gcc-ish compilers (I have no idea if this bug really even exists on windows and have no way to test unfortunately) so I added a build constraint for !windows.

voxeljorge:
actually it seems the test passes without the build constraint and fails with it so I'll leave it alone for now. I'm assuming the windows test must still use gcc

go/tools/builders/env.go
```diff
@@ -162,6 +163,86 @@ func (e *env) runCommandToFile(out, err io.Writer, args []string) error {
...
+	if err != nil {
+		return func() {}, "", err
+	}
+
+	// re-use the file extension if there is one (mostly for .exe)
+	builderPath := filepath.Join(dirname, "builder-cc"+filepath.Ext(os.Args[0]))
+
+	if err := os.Symlink(absPath, builderPath); err != nil {
+		_ = os.RemoveAll(dirname)
+		return func() {}, "", err
```
fmeum, line 182: This cleanup func should remove the directory.

go/tools/builders/link.go
fmeum: Could you add the first two to the comment? Thanks for looking them up!

fmeum (approved):
Looks good, just two comments.

go/tools/builders/env.go
voxeljorge: what I was going for here was that the cleanup function is only needed if the function succeeds. the code is a little inconsistent in that it returns nil above for the cleanup function and non-nil here, but also returns a non-nil err.

what probably makes the most sense is just to make it consistent and return a nil cleanup function if err is non-nil. the directory should always be removed in case of errors otherwise the caller would have to *also* call the cleanup function on an error which doesn't feel right.

i'll make that change, let me know what you think

go/tools/builders/link.go
voxeljorge: np, done

voxeljorge:
I think the build failure is in an unrelated test, maybe a fluke? Either way I'm unable to trigger a retry without pushing an update.

voxeljorge:
@fmeum I believe this may be ready. I addressed one comment slightly differently than you asked so I let it unresolved but I think the issue is fixed. Let me know if you need me to do anything else on this one.

fmeum (approved):

fmeum:
Thanks for the very thorough execution!

PR #4150, opened 2024-10-21
Allow bazel test //... to run out-of-the-box

### LINKED ISSUE #4149: Missing instructions to build and test master branch
This ticket is specifically about developing and contributing to the bazel-contrib/rules_go repository.
Currently, the `master` branch of the repo is at 87e66bb

In the WORKSPACE file it doesn't import a specific version. I believe it's using the local implementation from the previously mentioned commit.

In WORKSPACE it says `0.34.0`

Bazelisk version: 1.22.0
Another command (pid=46832) is running. Waiting for it to complete on the server (server_pid=40423)...
Build label: 7.3.2
Build time: Tue Oct 1 17:48:21 2024 (1727804901)
Build timestamp: 1727804901
Build timestamp as int: 1727804901

I think that one of the two issues was fixed in bazel-gazelle 0.39.0
https://github.com/bazel-contrib/bazel-gazelle/issues/1858

But, the one about having to run the ./tests/core/cgo/generate_imported_dylib.sh script before running the tests should logically reproduce with any version.

macOS Sonoma 14.7
arm64 (Mac M3)

I don't think so.

### What did you do?

```
> git clone git@github.com:bazel-contrib/rules_go.git
> cd rules_go
> bazel test //...
```

All the tests pass because this is a super-healthy open-source project that uses Bazel to isolate the build from quirks in the local environment.

### What did you see instead?

First problem were some errors trying to download `go1.23.2`. This was because `go env GOTOOLCHAIN` was set to `go1.23.2` and gazelle was sensitive to that setting because of https://github.com/bazel-contrib/bazel-gazelle/issues/1858 which was still present in gazelle version `0.34.0` (which is being used on `master`). I needed to unset the `GOTOOLCHAIN` variable to get past that error.

Second problem was that there are some inputs to the test cases that need to be generated manually with `./tests/core/cgo/generate_imported_dylib.sh` but, that isn't documented in the `CONTRIBUTING.md` file, so I didn't know I needed to do it.

### Ideas

If it isn't possible to update gazelle to a later release that fixes: https://github.com/bazel-contrib/bazel-gazelle/issues/1858, then, at least document in `CONTRIBUTING.md` that users should be sure that `go env GOTOOLCHAIN` returns `local`

If it isn't possible to automate the generation of the inputs to those tests (maybe with a genrule that runs that script before the tests,) then, at least document the need to run the script in `CONTRIBUTING.md`.

Also, just as a general rule, it's probably a good idea to add test coverage that runs in github actions to ensure that the master branch is always building correctly on various different platforms and is not sensitive to some of the most common environment variables.

eljobe:
- **Update to gazelle 0.39.1**
- **Wrap generate_imported_dylib.sh in genrules**

**What type of PR is this?**

Bug fix

This PR fixes 3 issues which caused an invocation of
`bazel test //...` in a clean checkout of the repository to fail
depending on potential contributor's development environment.

1. Contributors needed to know to run `generate_imported_dylib.sh`
2. Contributors needed to have $JAVA_HOME set
3. Contributors needed NOT to have `go env GOTOOLCHAIN` set

For 1, this PR wraps the `generate_imported_dylib.sh` in genrules so
that the build will just automatically create the dynamic libraries
appropriate for the platform.

For 2, the PR adds a `--java_runtime_version=remotejdk_21` argument to
the one test in `lcov_coverage_test.go` that needs to be able to build
a Java rule.

For 3, the PR upgrades `gazellle` to version 0.39.1 and updates the
excludes lists in the `popular_repos.py` file to handle the latest
versions of those popular repos. This upgrade addresses the problem
because https://github.com/bazel-contrib/bazel-gazelle/issues/1858 is
fixed in that version of gazelle.

BONUS: Also, there were a few TODOs left in the `WORKSPACE` and
`.bazelrc` files which were addressed by upgrading to the latest
`gazelle` version. So, this PR cleans those up at the same time.

Fixes #4149

**Other notes for review**

I have tested this on the following two machines:

macos (M3 arm64)
linux (amd64)

In both cases, I can clone the repository into an fresh directory hop
onto the `build-fixes` branch and run:
```
❯ bazel test -j 8 //...
```

And, it actually has all 452 tests passing within 30 minutes or so.

On both machines, `go env GOTOOLCHAIN` is `go1.23.2` and `$JAVA_HOME`
is not set on either machine.


fmeum (approved):
Thank you for the thorough cleanup!

sluongng:
Amazing PR. Thanks so much 🙏

PR #4151, opened 2024-10-21
Support .syso files in the `srcs` lists for `go_library` and `go_test` rules.

### LINKED ISSUE #4148: Feature Request: Support syso files as source files in rules_go
0.50.1

0.39.1

Bazelisk version: 1.22.0
Build label: 7.1.0
Build time: Mon Mar 11 17:55:31 2024 (1710179731)
Build timestamp: 1710179731
Build timestamp as int: 1710179731

Yup.

GOARCH='amd64'
GOHOSTOS='linux'

I used `gcc` to compile the precompiled `.syso` in the example repository.

### What did you do?
I tried to use precompile `.syso` files as `srcs` for `go_library` and `go_test` rules.

I expected successful builds and tests.

### What did you see instead?
Check out the full description in the sample repo [eljobe/sysos](https://github.com/eljobe/sysos/blob/main/README.md)

eljobe:
**What type of PR is this?**
Feature

Essentially, `go build ./...` and `go test ./...` allow having `.syso` files in the packages and it includes them in the packages it creates so that functions defined therein can be called from Go code.

Bazel didn't allow them to be among the `srcs` in these rules until this PR.

Fixes #4148

**Other notes for review**

I also added a bunch of tests and ran them on both linux (amd64) and Mac OS (arm64).

It's quite possible that I need to add some more things to get these tests to work on other platforms which are in the CI system.


tests/core/go_library/BUILD.bazel
```diff
@@ -166,3 +166,45 @@ go_library(
     name = "no_srcs_lib",
     importpath = "github.com/bazelbuild/rules_go/tests/core/no_srcs_lib",
 )
+
+# Tests for go_library and go_source support for .syso srcs
+cc_library(
+    name = "imported",
+    srcs = ["imported.c"],
+    linkstatic = True,
```
sluongng, line 174: I suspect this attribute will be going away soon as `cc_static_library` matures over time. Not sure if it's worth adding a second test against `cc_static_library`.

sluongng (approved):
It seems like most of the groundwork has already been done in https://github.com/bazel-contrib/rules_go/pull/3763 which correctly instruments `go tool pack` to append the syso file(s) to the compiled archive. However, it only added support for `go_binary` and not `go_library` and `go_test`. This PR added support for the remaining rules.

LGTM

tests/core/go_library/BUILD.bazel
eljobe: I'd prefer to have one or the other for the "imported" library just because it's not the main use case we're trying to solve.  I don't want to distract from the way the `.syso` files are supported in "srcs".

fmeum: `linkstatic` on `cc_library` also has a different purpose as it allows local control over whether this particular library is linked statically into a consumer.

fmeum: Could you move the "this is a hack" comment to this target? We could replace it with `cc_static_library` once that's available in our minimum supported version of Bazel.

tests/core/go_binary/meaning2.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import(
+	"fmt"
+)
+
+//go:noescape
```
fmeum, line 7: What is this needed for? Calling convention?

fmeum (approved):

tests/core/go_binary/meaning2.go
eljobe: It declares that the go runtime can rest assured that the implementation of the method doesn't leak pointers which can "escape" the function onto the heap.
https://gcc.gnu.org/onlinedocs/gccgo/Compiler-Directives.html

fmeum: My question is more about why it is needed here. Would the test not work without it?

eljobe: Oh. Huh! How about that. I guess it's not a problem. Maybe the repo's where I've seen this technique did some linting or something in the past that wouldn't work without that. Removed in 3395669

PR #4164, opened 2024-11-01
fix: tests respect --strip like binaries (#4103)

### LINKED ISSUE #4103: Expose an option or flag to disable stripping of symbols on `go_test` targets without `--@io_bazel_rules_go//go/config:debug`
As of #2573, builds of `go_test` targets will always result in a binary with its symbols. In #2691, it's reiterated this is working as intended.

This behavior can be disabled with `--@io_bazel_rules_go//go/config:debug`, but that has the additional implication of disabling inlining. So, if you are trying to diagnose a problem or especially debug anything related to performance, this is not an option you realistically have as any investigation you will do will be on a generated binary that is probably substantively different from your production binary.

In the GitHub issues discussing this, it's claimed that this behavior was adopted for a performance benefit. Maybe that is true (I haven't profiled anything using recent versions of `go`, although it's possible the situation has changed in 4 years), but there should be some option that one can apply to disable this stripping where relevant without enabling any other command-line flags that would impact the performance or behavior of the generated binary.

I can see two obvious options:

1. Provide a new flag (e.g. `--@io_bazel-rules_go//go/config:no_strip_tests`). Update the logic [here](https://github.com/bazelbuild/rules_go/blob/ee74d01c799ee3069d59fb76d584e064b3cac852/go/private/rules/test.bzl#L144) to not add `-s -w` if either `debug` or `no_strip_tests` are set, instead of only checking `debug`.
2. Honor `--strip`, i.e. do not strip these binaries if `--strip=never` is set.

Alternatively this stripping behavior could be updated to be opt-in instead of mandatory. It could also be re-evaluated whether this optimization is needed on the latest versions of the Go SDK.

If it's agreed option 1 ("Provide a new flag...") is best, I can submit a PR to that effect.

ajwerner:
**What type of PR is this?**

> Bug fix

Fixes #4103

### DIFF
--- a/tests/core/strip/strip_test.go
+++ b/tests/core/strip/strip_test.go
@@ -29,11 +29,22 @@ func TestMain(m *testing.M) {
 	bazel_testing.TestMain(m, bazel_testing.Args{
 		Main: `
 -- BUILD.bazel --
-load("@io_bazel_rules_go//go:def.bzl", "go_binary")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
+
+go_library(
+    name = "strip_lib",
+    srcs = ["strip.go"],
+)
 
 go_binary(
     name = "strip",
-    srcs = ["strip.go"],
+    embed = [":strip_lib"],
+)
+
+go_test(
+    name = "strip_test",
+    srcs = ["strip_test.go"],
+    embed = [":strip_lib"],
 )
 -- strip.go --
 package main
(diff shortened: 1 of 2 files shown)


tests/core/strip/strip_test.go
```diff
@@ -259,10 +306,35 @@ var wantStackTrace = []string{
 	`^$`,
 }
 
-func embedWantedStackTraces() string {
+func embedWantedTestStackTrace() string {
+	return embedStringSliceVar("wantTestStackTrace", wantTestStackTrace)
+}
+
+var wantTestStackTrace = []string{
+	`^goroutine \d+ \[running\]:$`,
```
fmeum, line 314: This is very comprehensive, but also somewhat prone to breaking with future versions of Go. Could you pick out one or two lines that are essential for verifying the presence of debug information and drop the rest?

ajwerner: Thanks for the review! I changed the testing so that it reuses exactly the previous expectations and doesn't look at the tail that's due to the implementation of go's testing package. See if this pleases.

ajwerner:
@fmeum anything else you want here?

fmeum (approved):
@linzhp Any concerns from your side?

linzhp:
Fine with me. Sorry for late reply

PR #4167, opened 2024-11-08
Fix `go_tool_binary` non-hermeticity and Go 1.19 incompatibility

fmeum:
**What type of PR is this?**

Bug fix

`go_tool_binary` used to non-hermetically pick up rules_go's `go.mod` file when run unsandboxed, which resulted in it downloading a different toolchain and messing up version checks.

Also make `nogo_main.go` compatible with Go 1.19 and test with nogo (which requires fixing a bunch of violations).

Fixes #

**Other notes for review**

### DIFF
--- a/go/runfiles/manifest.go
+++ b/go/runfiles/manifest.go
@@ -138,7 +138,8 @@ func (m *manifest) open(name string) (fs.File, error) {
 			// name refers to an actual file or dir listed in the manifest. The
 			// basename of name may not match the basename of the underlying
 			// file (e.g. in the case of a root symlink), so patch it.
-			f, err := os.Open(r)
+			var f *os.File
+			f, err = os.Open(r)
 			if err != nil {
 				return nil, err
 			}
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,7 +1,7 @@
 workspace(name = "io_bazel_rules_go")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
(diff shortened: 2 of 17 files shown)


zecke:
Slightly related. Should the go.mod file be passed as dependency? For things like https://go.dev/blog/loopvar-preview the content of the go.mod file seems to determine the behavior.

From a rules_go users point of view. What is the capture behavior now when using Go 1.22+?

fmeum:
> From a rules_go users point of view. What is the capture behavior now when using Go 1.22+?

rules_go currently always uses the Go language version corresponding to the version of the Go SDK.

> Slightly related. Should the go.mod file be passed as dependency? For things like https://go.dev/blog/loopvar-preview the content of the go.mod file seems to determine the behavior.

I started working on this in https://github.com/bazel-contrib/rules_go/pull/4106, but haven't gotten around to finalizing the PR. If anyone wants to take a stab at this, please do.

go/runfiles/manifest.go
```diff
@@ -138,7 +138,8 @@ func (m *manifest) open(name string) (fs.File, error) {
 			// name refers to an actual file or dir listed in the manifest. The
 			// basename of name may not match the basename of the underlying
 			// file (e.g. in the case of a root symlink), so patch it.
-			f, err := os.Open(r)
+			var f *os.File
```
linzhp, line 141: accidental change?

linzhp (approved):

go/runfiles/manifest.go
fmeum: I needed this to have `nogo`'s `shadow` analyzer pass on the runfiles lib. Since we don't control the analyzers it is built with, I guess we just have to ensure that it's passing all the ones we know of.

PR #4189, opened 2024-12-03
Stop assuming deps of go_proto_library has importmap attribute

linzhp:
**What type of PR is this?**

Bug fix

`_go_proto_aspect` applies to all `deps` of `go_proto_library` transitively, following the `deps` and `embed` attributes. Meanwhile, the only requirement for `deps` of `go_proto_library` is to provide `GoInfo`. They don't have to be `go_proto_library` or `go_library`. We should not assume they also have `importmap`, `importpath_aliases` or `embed` attributes. They are likely to have `importpath`, otherwise Gazelle won't be able to resolve them. So we can keep the assumption of `importpath` for now until breakages are discovered.

cc @dzbarsky


proto/def.bzl
```diff
@@ -74,9 +74,6 @@ def _go_proto_aspect_impl(_target, ctx):
         attr,
         include_deprecated_properties = False,
         importpath = attr.importpath,
-        importmap = attr.importmap,
```
fmeum, line 77: I would have expected logic such as `importmap = getattr(attr, "importmap", None)`. Is it clear that we don't need this?

dzbarsky:
I didn't think people would add custom rule deps to go_proto_library. What's the use case?

proto/def.bzl
linzhp: `go_context` calls `getattr` automatically if importmap etc are not passed: https://github.com/bazel-contrib/rules_go/blob/702d10674ec48a679488a725c86bf35121420698/go/private/context.bzl#L581

fmeum (approved):

linzhp:
> I didn't think people would add custom rule deps to go_proto_library. What's the use case?

There could be custom protoc plugins that add dependency to custom generated code

dzbarsky:
> > I didn't think people would add custom rule deps to go_proto_library. What's the use case?
> 
> There could be custom protoc plugins that add dependency to custom generated code

I see, and the custom generated code isn't a normal go_library? Fair enough. Perhaps it would be nice to add a testcase for this usage to the repo to avoid it breaking in the future, since I suspect this is not a very common setup among rules_go users.

PR #4195, opened 2024-12-12
Fixing default includes of nogo

### LINKED ISSUE #4194: the default linter "composites" no longer works in the latest rules_go version
https://github.com/bazel-contrib/rules_go/releases/tag/v0.51.0-rc2

Build label: 7.3.1

yes

ubuntu, x86_64

### What did you see instead?

Take any example that previously failed the composites linter (which checks that we should use keyed field if the struct is defined by other packages), it will no longer alarm now.

linzhp:
Currently, nogo will only run within rules_go, not the main repo, by default. This PR also remove the setting of `includes` in tests by default to expose issues in the default `includes` of `go_register_nogo` 

Fixes #4194


fmeum:
@linzhp I fixed the issue on your branch, CI is passing now with the new test coverage.

fmeum (approved):

PR #4203, opened 2024-12-24
fix: Make context.mode always be GoConfigInfo

### LINKED ISSUE #4197: Archive mode does not match when pure = off

0.51.0

0.40.0

Bazel 8.0.0

Yes

Darwin/ARM

### What did you do?

When crosscompiling a `go_binary` with goos, goarch, and pure = "off" I'm getting a mode mismatch error.
This error happens on both our CI runner (linux/amd) and locally on a M1 macbook.

```python
    go_binary(
        name = "target",
        embed = embed,
        goarch = "arm64",
        goos = "darwin",
        pure = "off",
    )
```

The target should build without error. This worked in 0.50.1.

### What did you see instead?

```
Error in fail: Archive mode does not match @@//:target is darwin_arm64_debug expected darwin_arm64_debug
```

If I print `a.source.mode` and `go.mode` from [this line](https://github.com/bazel-contrib/rules_go/blob/54293b3eab11a2cdbc0b02fa9e8943b21880e04c/go/private/actions/archive.bzl#L77), I get the same struct values.

```
// Debugged from CI hence different arch/os/linkmode

// a.source.mode
struct(amd64 = "", arm = "", cover_format = "lcov", debug = False, gc_goopts = [], gc_linkopts = [], goarch = "amd64", goos = "linux", linkmode = "pie", msan = False, pgoprofile = None, pure = False, race = False, stamp = False, static = False, strip = True, tags = [])

// go.mode
struct(amd64 = "", arm = "", cover_format = "lcov", debug = False, gc_goopts = [], gc_linkopts = [], goarch = "amd64", goos = "linux", linkmode = "pie", msan = False, pgoprofile = None, pure = False, race = False, stamp = False, static = False, strip = True, tags = [])
```

dzbarsky:
**What type of PR is this?**
Bug fix 

Providers and structs with the same fields don't compare equal

Fixes #4197

I wasn't able to figure out a repro in rules_go, it seems like it may rely on some transitions? Not sure we should block landing on this, it seems like this should be the right fix

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -514,7 +514,7 @@ def go_context(
             if getattr(ctx.attr, "pure", None) == "off":
                 fail("{} has pure explicitly set to off, but no C++ toolchain could be found for its platform".format(ctx.label))
             mode_kwargs["pure"] = True
-        mode = struct(**mode_kwargs)
+        mode = GoConfigInfo(**mode_kwargs)
         validate_mode(mode)
 
     if stdlib:
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -503,7 +503,7 @@ def go_context(
 
     if goos == "auto" and goarch == "auto" and cgo_context_info:
         # Fast-path to reuse the GoConfigInfo as-is
-        mode = go_config_info
+        mode = go_config_info or default_go_config_info
     else:
         if go_config_info == None:
             go_config_info = default_go_config_info
--- a/tests/core/cross/BUILD.bazel
+++ b/tests/core/cross/BUILD.bazel
@@ -7,6 +7,16 @@ test_suite(
 )
 
 go_binary(
+    name = "non_pure_cross",
+    srcs = ["main.go"],
+    goarch = "amd64",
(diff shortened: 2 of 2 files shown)


fmeum (approved):

dzbarsky:
@fmeum I think this needs help to merge

fmeum:
I added a reproducer in https://github.com/bazel-contrib/rules_go/pull/4206 and then applied the fix. The history of CI checks leads me to believe that this is a genuine failure caused by this change, I just don't know why...

dzbarsky:
> I added a reproducer in #4206 and then applied the fix. The history of CI checks leads me to believe that this is a genuine failure caused by this change, I just don't know why...

Hmm your repro looks pretty similar to what I tried, I wonder why I didn't have success. Sadly I don't have a laptop with me to try it right now, but it seems good. Let's ship it!

fmeum:
We unfortunately can't ship since the fix legitimately breaks existing tests. I'll look into it.

fmeum (approved):
I think I found the fix

PR #4208, opened 2025-01-01
chore: remove Travis-specific install docs

alexeagle:
It was added eight years ago in #1014. Since then, Travis has become less popular, so it's not worth the burden of maintaining this section.
It's strange that we document this one particular CI platform and not others like CircleCI or GitHub Actions. Also, this documentation isn't really Go-specific.

**What type of PR is this?**

> New Year's resolution


fmeum (approved):
Didn't even know this existed

PR #4210, opened 2025-01-02
Undo API-breaking change to go_sdk experiments attribute

dzbarsky:
**What type of PR is this?**

Bug fix, see https://github.com/bazel-contrib/rules_go/pull/4022#issuecomment-2566548860

Reverts the behavior to the API while keeping the optimization

Fixes #

**Other notes for review**


dzbarsky:
@fmeum PTAL :)

fmeum:
This is failing with the same errors as the other PR. Is this based on current master?

fmeum (approved):

PR #4217, opened 2025-01-06
Put absolute paths into -extldflags

smertnik3sh:
Bug fix

The go build system stores information about build flags (and environment variables) in the "_go_.o" file (go package manifest). When you use cgo, data from "CGO_LDFLAGS" is added to this file as is and used during linking (in the "ld" command). If you add absolute path to "CGO_LDFLAGS" environment, it will be save inside manifest too. However, when using a distributed build system such as BuildBarn, different build agents not share all paths (such as "bazel sandboxing" paths). If the "_go_.o" file inside go library contains non-existent paths, linking with this module fails.

This patch puts all flags with absolute paths into `-extldflags`. In this case its not stored in the go package manifest.

Problem can be reproduced only on specific cases (we use custom sysroot and distributed build system) so I don't know how to write good "common" test.

### DIFF
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -33,7 +33,7 @@ import (
 
 var (
 	// cgoEnvVars is the list of all cgo environment variable
-	cgoEnvVars = []string{"CGO_CFLAGS", "CGO_CXXFLAGS", "CGO_CPPFLAGS", "CGO_LDFLAGS"}
+	cgoAbsEnvVars = []string{"CGO_CFLAGS", "CGO_CXXFLAGS", "CGO_CPPFLAGS"}
 	// cgoAbsEnvFlags are all the flags that need absolute path in cgoEnvVars
 	cgoAbsEnvFlags = []string{"-I", "-L", "-isysroot", "-isystem", "-iquote", "-include", "-gcc-toolchain", "--sysroot", "-resource-dir", "-fsanitize-blacklist", "-fsanitize-ignorelist"}
 	// cgoAbsPlaceholder is placed in front of flag values that must be absolutized
@@ -593,4 +593,41 @@ func useResponseFile(path string, argLen int) bool {
 		return true
 	}
 	return false
+}
+
+// The go build system collects the CGO_LDFLAGS from manifests (_go_.o) and uses them in the user module linking.
+// In some cases, such as when using the distributed Bazel executor, these paths may not exist (different build agents uses different sandboxing paths).
+// It is better to use flags with absolute paths in the -extldflags option (in this case, they will not be saved inside the manifest).
+func splitAbsEnvVar(originVar string, absArs []string) (string, string, error) {
+	var absolute []string
+	var nonAbsolute []string
+	absNext := false
+	splited, err := splitQuoted(originVar)
+	if err != nil {
+		return "", "", err
+	}
+	for i := range splited {
+		if absNext {
+			absolute = append(absolute, splited[i])
+			absNext = false
+			continue
+		}
+		isAbsolute := false
+		for _, f := range absArs {
+			if !strings.HasPrefix(splited[i], f) {
+				continue
+			}
+			isAbsolute = true
+			absolute = append(absolute, splited[i])
+			possibleValue := splited[i][len(f):]
+			if len(possibleValue) == 0 {
+				absNext = true
+			}
+			break
+		}
+		if !isAbsolute {
+			nonAbsolute = append(nonAbsolute, splited[i])
+		}
+	}
+	return strings.Join(absolute, " "), strings.Join(nonAbsolute, " "), nil
 }
--- a/go/tools/builders/stdlib.go
+++ b/go/tools/builders/stdlib.go
@@ -154,11 +154,23 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
 			break
 		}
 	}
+	absoluteLdFlags, nonAbsoluteLdFlags, err := splitAbsEnvVar(os.Getenv("CGO_LDFLAGS"), cgoAbsEnvFlags)
+	if err != nil {
+		return fmt.Errorf("error splitting cgo ldflags: %v", err)
+	}
+	os.Setenv("CGO_LDFLAGS", nonAbsoluteLdFlags)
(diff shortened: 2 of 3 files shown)


go/tools/builders/env.go
```diff
@@ -33,7 +33,7 @@ import (
 
 var (
 	// cgoEnvVars is the list of all cgo environment variable
-	cgoEnvVars = []string{"CGO_CFLAGS", "CGO_CXXFLAGS", "CGO_CPPFLAGS", "CGO_LDFLAGS"}
+	cgoAbsEnvVars = []string{"CGO_CFLAGS", "CGO_CXXFLAGS", "CGO_CPPFLAGS"}
```
fmeum, line 36: Please update the comment

go/tools/builders/env.go
```diff
@@ -594,3 +594,40 @@ func useResponseFile(path string, argLen int) bool {
 	}
 	return false
 }
+
+// The go build system collects the CGO_LDFLAGS from manifests (_go_.o) and uses them in the user module linking.
+// In some cases, such as when using the distributed Bazel executor, these paths may not exist (different build agents uses different sandboxing paths).
```
fmeum, line 599:
```suggestion
// In some cases, such as when using remote execution, these paths may not exist (different build agents uses different sandboxing paths).
```
Does this also come up with sandboxed builds? The absolute path prefixes should already be different in that case.

go/tools/builders/env.go
```diff
@@ -594,3 +594,40 @@ func useResponseFile(path string, argLen int) bool {
...
+// It is better to use flags with absolute paths in the -extldflags option (in this case, they will not be saved inside the manifest).
+func splitAbsEnvVar(originVar string, absArs []string) (string, string, error) {
+	var absolute []string
+	var nonAbsolute []string
+	absNext := false
+	splited, err := splitQuoted(originVar)
```
fmeum, line 605:
```suggestion
	split, err := splitQuoted(originVar)
```
and below

go/tools/builders/env.go
```diff
@@ -594,3 +594,40 @@ func useResponseFile(path string, argLen int) bool {
 	}
 	return false
 }
+
+// The go build system collects the CGO_LDFLAGS from manifests (_go_.o) and uses them in the user module linking.
+// In some cases, such as when using the distributed Bazel executor, these paths may not exist (different build agents uses different sandboxing paths).
+// It is better to use flags with absolute paths in the -extldflags option (in this case, they will not be saved inside the manifest).
+func splitAbsEnvVar(originVar string, absArs []string) (string, string, error) {
```
fmeum, line 601:
```suggestion
func splitAbsEnvVar(originVar string, absArgs []string) (string, string, error) {
```
and below.

go/tools/builders/env.go
```diff
@@ -594,3 +594,40 @@ func useResponseFile(path string, argLen int) bool {
 	}
 	return false
 }
+
+// The go build system collects the CGO_LDFLAGS from manifests (_go_.o) and uses them in the user module linking.
+// In some cases, such as when using the distributed Bazel executor, these paths may not exist (different build agents uses different sandboxing paths).
+// It is better to use flags with absolute paths in the -extldflags option (in this case, they will not be saved inside the manifest).
```
fmeum, line 600: Could we just put all flags into `-extldflags` instead? What's the advantage of splitting them?

go/tools/builders/stdlib.go
```diff
@@ -154,11 +154,23 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
 		}
 	}
+	absoluteLdFlags, nonAbsoluteLdFlags, err := splitAbsEnvVar(os.Getenv("CGO_LDFLAGS"), cgoAbsEnvFlags)
+	if err != nil {
+		return fmt.Errorf("error splitting cgo ldflags: %v", err)
+	}
+	os.Setenv("CGO_LDFLAGS", nonAbsoluteLdFlags)
+	if len(absoluteLdFlags) > 0 {
+		ldflags = append(ldflags, "-extldflags")
+		splitFlags := strings.Split(absoluteLdFlags, " ")
```
fmeum, line 164: We are splitting the flags we just joined. Instead, could we have the helper function return string slices?

fmeum:
Could you post examples of the paths that cause cache misses? I would like to understand why this hasn't come up before.

go/tools/builders/env.go
smertnik3sh: 2 reasons:
1. I usually prefer to fix only broken things)
2. I think it's a "go-way" for providing ldflags to dependencies (some kind of "transitive" flags). I'm afraid that changing this behavior car broke builds in some strange places. But it's a theory.

If you think that it's better to use `-extldflags` for all, I can implement it.

fmeum: It seems as if we somewhat lose the benefits of 2. as soon as we move any flag over, so I would be in favor of moving all of them. This aligns better with how Bazel works: We can track link flags of all transitive deps via providers, so there is no need to track them in executables.

The only potential breakage I can think of is for Go generated static archives that are considered shippable artifacts. That seems so rare that I don't think we should worry about it.

@linzhp @jayconrod Do you have opinions on this?

jayconrod:
Before reviewing this, I think we need a clearer description of the problem being solved. Please open an issue explaining that and link here.

What are the absolute paths coming in through cgo? Why is that happening?

I suspect that the Go builder and the linker may be the wrong place to fix this. Maybe we can avoid absolute paths in the first place.

smertnik3sh:
I was unable to quickly reproduce the issue using the actual opensource tools (in our repository we use self-built compiler and sysroot). The original issue in our build occurred on an older version of basel and a different configuration of the buildbarn server (few months ago). I was too lazy to creating pool request immediately, which I now regret)
Currently, when disabling the patch in our repository, the issue appears similar to https://github.com/bazel-contrib/rules_go/issues/4091.
Also we are still using the LLVM 16 infrastructure (Clang, LLD). Maybe that is the one of causes.
I will check if the issue persists on the latest version of rules_go. If it does, I will open the issue, send reproducer and create a new pull request. For now, this pull request can probably be closed. Thanks for your time.

dzbarsky:
Closing this out per the last message

PR #4231, opened 2025-01-15
Use same Go SDK as Gazelle for `go_bazel_test`

### LINKED ISSUE #4228: go/tools/bazel_testing downloads a second Go SDK
When `go/tools/bazel_testing` is imported, Bazel will download a second Go SDK, namely `@@rules_go~~go_sdk~go_default_sdk`, instead of the one registered from the root module. This is undesirable because users will have to download two Go SDKs during the build, and test with an SDK that they may not care about. In addition, some companies may download Go SDKs from internal mirrors and block all external downloads, causing the build to fail

fmeum:
**What type of PR is this?**

Bug fix

Fixes #4228

**Other notes for review**

### DIFF
--- a/go/tools/bazel_testing/BUILD.bazel
+++ b/go/tools/bazel_testing/BUILD.bazel
@@ -2,21 +2,19 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@go_host_compatible_sdk_label//:defs.bzl", "HOST_COMPATIBLE_SDK")
 load("//go:def.bzl", "go_library")
 
-_HOST_COMPATIBLE_SDK_ROOT = HOST_COMPATIBLE_SDK.same_package_label("ROOT")
-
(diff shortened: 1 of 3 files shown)


go/tools/bazel_testing/BUILD.bazel
```diff
@@ -1,17 +1,22 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@go_host_compatible_sdk_label//:defs.bzl", "HOST_COMPATIBLE_SDK")
 load("//go:def.bzl", "go_library")
 
+_HOST_COMPATIBLE_SDK_ROOT = HOST_COMPATIBLE_SDK.same_package_label("ROOT")
```
linzhp, line 5: Is `_HOST_COMPATIBLE_SDK_ROOT` the same as `HOST_COMPATIBLE_SDK`?

linzhp (approved):

go/tools/bazel_testing/BUILD.bazel
fmeum: Yes, simplified

PR #4232, opened 2025-01-16
Allow .so files to have more extensions

### LINKED ISSUE #4227: get_versioned_shared_lib_extension rejects some third-party libraries
For some reason https://github.com/awslabs/aws-c-s3 decided that correct soname for their library is `libaws-c-s3.so.0unstable`. As you can see, it has non-digits in its version part. If you import this library by its canonical soname and link it into your Go binary using rules_go:
* `NEEDED` entry will be created for `libaws-c-s3.so.0unstable`,
* `rpath` for linker _will not_ contain corresponding solib directory,
* `runfiles` _will not_ link corresponding solib.

As such, running such binaries will result in `error while loading shared libraries: libaws-c-s3.so.0unstable: cannot open shared object file: No such file or directory`.

It looks like the root cause of this issue are checks in the `get_versioned_shared_lib_extension` function: if version fails the `isdigit()` check, a dependency is not added into the rpath and runfiles.

matshch:
**What type of PR is this?**

Bug fix

This PR relaxes requirements on shared libraries' versions to be in line with bazelbuild/bazel#10148. Some prebuilt libraries have non-digit versions, and we need to allow those.

Fixes #4227.

**Other notes for review**

Bazel uses a regular expression to match version, which is not available in Starlark. I believe this implementation matches it exactly, but if you see some discrepancy — lets see what we can do.

I've also left extra tests that we already had here and added a case that checks library name mentioned in #4227. In result they are not sorted now, but Bazel's implementation does not sort its tests either 🤷

### DIFF
--- a/go/private/common.bzl
+++ b/go/private/common.bzl
@@ -123,15 +123,27 @@ def has_simple_shared_lib_extension(path):
     """
     return any([path.endswith(ext) for ext in SHARED_LIB_EXTENSIONS])
 
+def is_valid_shared_lib_version_part(part):
+    """
+    Checks if the version part is valid. Matches the same check from
+    //src/main/java/com/google/devtools/build/lib/rules/cpp:CppFileTypes.java
+    """
+    if not part[:1].isdigit():
+        return False
+    for c in part[1:].elems():
+        if c != "_" and not c.isalnum():
+            return False
+    return True
+
 def get_versioned_shared_lib_extension(path):
     """If appears to be an versioned .so or .dylib file, return the extension; otherwise empty"""
     parts = path.split("/")[-1].split(".")
-    if not parts[-1].isdigit():
+    if not is_valid_shared_lib_version_part(parts[-1]):
         return ""
 
     # only iterating to 1 because parts[0] has to be the lib name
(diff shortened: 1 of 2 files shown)


go/private/common.bzl
```diff
@@ -123,15 +123,27 @@ def has_simple_shared_lib_extension(path):
     """
     return any([path.endswith(ext) for ext in SHARED_LIB_EXTENSIONS])
 
+def is_valid_shared_lib_version_part(part):
+    """
+    Checks if the version part is valid. Matches the same check from
+    //src/main/java/com/google/devtools/build/lib/rules/cpp:CppFileTypes.java
+    """
+    if not part[:1].isdigit():
```
fmeum, line 131:
```suggestion
    if not part[0].isdigit():
```
Simpler if it works

fmeum (approved):

go/private/common.bzl
matshch: It will fail on empty strings, but I'll check if it's possible to get an empty string here.

fmeum: I see, then let's keep this as is to not risk a "crash"

PR #4236, opened 2025-01-17
Update protoc plugins to support Protobuf Editions

### LINKED ISSUE #4235: Support Protobuf editions
Protobuf editions is the successor of proto2 and proto3. See [documentation](https://protobuf.dev/editions/overview/).

When trying to compile a `go_proto_library()` with a proto file using editions, I get the following error:
```
Success!path/to/file.proto: is an editions file, but code generator protoc-gen-go-grpc hasn't been updated to support editions yet.  Please ask the owner of this code generator to add support or switch back to proto2/proto3.
```

mering:
**What type of PR is this?**

Feature

This adds support for compiling `go_proto_library` and `go_grpc_library` to support .proto files using Protobuf Editions.

Fixes #4235

**Other notes for review**


fmeum:
Thanks for working on this! CI is red with a reference to googleapis. Maybe we need to add a dep?

mering:
@fmeum I also had to update `google.golang.org/genproto` and also ran `bazel run @io_bazel_rules_go//go -- mod tidy`. CI is green now.

fmeum:
Do you happen to know whether these updates include any breaking changes (especially in protobuf)? I'm asking because an alternative approach that avoids a forceful update for users would be to document that contents of Go proto rules need to depend on a sufficiently recent protobuf version themselves.

mering:
> Do you happen to know whether these updates include any breaking changes (especially in protobuf)? I'm asking because an alternative approach that avoids a forceful update for users would be to document that contents of Go proto rules need to depend on a sufficiently recent protobuf version themselves.

I am not aware of any and Golang versioning requires breaking changes to become a new package (for example `/v2`).

fmeum:
Some of the packages are not yet at v1, but as users will have to update anyway at some point, I think we can risk it. If something breaks, we can always roll back.

PR #4238, opened 2025-01-21
sdk: remove logic for Bazel versions below 6.4.0

sluongng:
Before 6.4.0, there is a bug in Bazel when extracting archive with
special file names inside.

Since Bazel 6 is not the oldest supported LTS and 6.5.0 is already
released, we should be able to remove this logic safely.


sluongng:
Also relevant
```
# go/private/common.bzl:156
MINIMUM_BAZEL_VERSION = "6.5.0"
```

sluongng:
@fmeum could you hit the green button? 😄

PR #4244, opened 2025-01-29
Generate exports for stdlib

jgautier-dd:
**What type of PR is this?**

Feature

This PR adds exports to the standard library which is useful for the gopackagesdriver to avoid parsing files and various other tools that use https://pkg.go.dev/golang.org/x/tools/go/packages.

#3653 

**Other notes for review**
Basically a reimplementation of this PR: https://github.com/bazel-contrib/rules_go/pull/3653 but with the comment addressed about configuring it with `//go/config`. One additional thing added is the exports files are part of the `GoSDK` provider so that other custom rules can also gain access to the exported files.


yunjizhong (approved):

go/private/BUILD.sdk.bazel
```diff
@@ -16,6 +16,15 @@ filegroup(
     ),
 )
 
+filegroup(
+    name = "exports",
+    srcs = glob(
+        ["pkg/{goos}_{goarch}/**/*.x"],
+        allow_empty = True,
```
fmeum, line 23: Is this going to be empty on more revent Go SDKs that no longer ship precompiled stdlibs? If so, we might want to get rid of it entirely. If not, I would feel better about removing `allow_empty` so that we learn when this becomes unnecessary.

go/private/actions/stdlib.bzl
```diff
@@ -115,8 +117,10 @@ def _build_env(go):
     return env
 
 def _sdk_stdlib(go):
+    [list_json, cache_dir] = _build_stdlib_list_json(go)
```
fmeum, line 120:
```suggestion
    list_json, cache_dir = _build_stdlib_list_json(go)
```

go/private/actions/stdlib.bzl
```diff
@@ -115,8 +117,10 @@ def _build_env(go):
...
     return GoStdLib(
-        _list_json = _build_stdlib_list_json(go),
+        _list_json = list_json,
+        cache_dir = depset([cache_dir]),
```
fmeum, line 123: As this is a new property and `GoStdLib` is public API, let's prefix this with a `_` as well.

go/private/actions/stdlib.bzl
```diff
@@ -164,8 +168,10 @@ def _build_stdlib(go):
         toolchain = GO_TOOLCHAIN_LABEL,
         execution_requirements = SUPPORTS_PATH_MAPPING_REQUIREMENT,
     )
+    [list_json, cache_dir] = _build_stdlib_list_json(go)
```
fmeum, line 171:
```suggestion
    list_json, cache_dir = _build_stdlib_list_json(go)
```

go/private/actions/stdlib.bzl
```diff
@@ -164,8 +168,10 @@ def _build_stdlib(go):
...
     return GoStdLib(
-        _list_json = _build_stdlib_list_json(go),
+        _list_json = list_json,
         libs = depset([pkg]),
+        cache_dir = depset([cache_dir]),
```
fmeum, line 175: As above, prefix with `_`.

go/private/rules/sdk.bzl
```diff
@@ -69,6 +70,13 @@ go_sdk = rule(
             doc = ("Pre-compiled .a files for the standard library, " +
                    "built for the execution platform"),
         ),
+        "exports": attr.label_list(
+            # allow_files is not set to [".x"] because that wouldn't allow
```
fmeum, line 74: Which error are you seeing in this case? The only ones I can find shouldn't be shown when there are no files.

go/tools/builders/stdliblist.go
```diff
@@ -166,7 +166,7 @@ func flatPackageForStd(cloneBase string, pkg *goListPackage, pathReplaceFn func(
 		ID:              stdlibPackageID(pkg.ImportPath),
 		Name:            pkg.Name,
 		PkgPath:         pkg.ImportPath,
-		ExportFile:      outputBasePath(cloneBase, pkg.Target),
+		ExportFile:      pathReplaceFn(pkg.Export),
```
fmeum, line 169: Why is this necessary?

fmeum (approved):
Looks good, thanks for picking this up!

Could you add `Co-authored-by` lines referencing the author of #3653?

go/private/BUILD.sdk.bazel
jgautier-dd: at some point in my testing I saw .x files being written into the pkg directory. However after testing again today I did not see this happening so I just removed this and the other code related to having exports for the sdk provider.

jgautier-dd: i am currently testing with go 1.23.2

jgautier-dd: In my current testing the x files are getting written into the cachedir (but not with a .x extension)

go/tools/builders/stdliblist.go
jgautier-dd: This change makes the path to the `ExportFile` correct in the `stdlib.pkg.json` for example without this change the json for the `fmt` package looks like:
```
{
  "ID": "@@rules_go~//stdlib:fmt",
  "Name": "fmt",
  "PkgPath": "fmt",
  "Standard": true,
  "GoFiles": [
...
  ],
  "CompiledGoFiles": [
   ...
  ],
  "ExportFile": "__BAZEL_OUTPUT_BASE__",
  "Imports": {
...
  }
}
```

WIth this set it points to a file within the cache dir. for example
```
{
  "ID": "@@rules_go~//stdlib:fmt",
  "Name": "fmt",
  "PkgPath": "fmt",
  "Standard": true,
  "GoFiles": [
...
  ],
  "CompiledGoFiles": [
...
  ],
  "ExportFile": "__BAZEL_EXECROOT__/bazel-out/darwin_arm64-fastbuild/bin/external/rules_go~/stdlib_/gocache/51/512ffa63f575f10935211fc04e33f34fe3a81bfd37feb8d776803724cd8e809f-d",
  "Imports": {
...
  }
}
```

go/private/actions/stdlib.bzl
jgautier-dd: I added the prefix but I think it might make sense to make it part of the public API? I am currently using this in a custom rule that is using the `https://pkg.go.dev/golang.org/x/tools/go/packages` package loader and by accessing this cache_dir it avoids reparsing the stdlib files for every time the loader is called.

fmeum: If it's a conscious decision and useful then feel free to make it public. In that case, please add a line to the provider docs Markdown/RST.

fmeum (approved):

PR #4253, opened 2025-02-08
Capturing elapsed time of interrupted test cases due to timeout

linzhp:
**What type of PR is this?**
Feature

When a test case is interrupted at timeout, we don't have the pass/fail/skip event to report elapsed time. However, we can parse them from the error message at the end to figure it out. The error message looks like:

```
panic: test timed out after 8s
        running tests:
                TestReport (8s)
                TestReport/test_3 (2s)
```

Also, instead of reporting those test cases as "No pass/skip/fail event found for test", we can now report them as "Interrupted".

### DIFF
--- a/go/tools/bzltestutil/xml.go
+++ b/go/tools/bzltestutil/xml.go
@@ -73,6 +73,10 @@ type testCase struct {
 	duration *float64
 }
 
+const (
+	timeoutPanicPrefix = "panic: test timed out after "
+)
+
 // json2xml converts test2json's output into an xml output readable by Bazel.
 // http://windyroad.com.au/dl/Open%20Source/JUnit.xsd
 func json2xml(r io.Reader, pkgName string) ([]byte, error) {
@@ -89,6 +93,7 @@ func json2xml(r io.Reader, pkgName string) ([]byte, error) {
 	}
 
 	dec := json.NewDecoder(r)
+	var inTimeoutSection, inRunningTestSection bool
 	for {
 		var e jsonEvent
 		if err := dec.Decode(&e); err == io.EOF {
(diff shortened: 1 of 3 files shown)


go/tools/bzltestutil/xml.go
```diff
@@ -102,6 +107,38 @@ func json2xml(r io.Reader, pkgName string) ([]byte, error) {
...
+				}
+				continue
+			}
+			if inTimeoutSection && strings.HasPrefix(trimmedOutput, "running tests:") {
+				inRunningTestSection = true
+				continue
+			}
+			if inRunningTestSection {
+				// looking for something like "TestReport/test_3 (2s)"
+				parts := strings.Fields(e.Output)
```
fmeum, line 127: Can subtest names contain spaces? We could split on the last space (or multiple spaces) instead.

fmeum (approved):
Nice, I love contained hacks with a big impact on usability!

go/tools/bzltestutil/xml.go
linzhp: Spaces are allowed, but Go replaced them with `_`: https://github.com/golang/go/blob/215de81513286c010951624243c2923f7dc79675/src/testing/match.go#L282

In fact the test name here is called "test 3" in the code, but it became "test_3" from the report.

The error message is basically free-formed, so I am trying to be restrictive here, and avoid parsing wrong parts of the message. I set up 3 gates to trigger the logic:

1. look for "panic: test timed out after" first
2. look for "running tests" next
3. look for something like "TestReport/test_3 (2s)"

There is a lot of assumption on the implementation details in the testing package, which could change. But I would rather this parsing logic to stop working instead of updating wrong parts of the xml

PR #4268, opened 2025-02-13
Generate facts for targets not in nogo scope

### LINKED ISSUE #4241: Allow nogo to run against all dependencies
As discussed via Slack, I would like to have a way to get the old behavior back. I am using bzlmod and would like to tell nogo to run against all third_party code

fmeum:
**What type of PR is this?**

Bug fix

Nogo analyzers may need facts from their (transitive) dependencies to operate correctly. We thus need to run nogo on all targets, even those not in scope, but should only register the validation action for the targets that are in scope to have them fail the build on nogo findings.

Fixes #4241

**Other notes for review**


fmeum:
@fionera Could you test this?

fionera:
@fmeum Things like deprecations are now being correctly detected. Could you maybe also add a "magic package" alias to run nogo including failures it on all external dependencies?

linzhp (approved):
Tested in Uber by @peng3141 . It works

PR #4281, opened 2025-02-23
Calculating the test suite duraion from timestamps

linzhp:
**What type of PR is this?**
Bug fix

When test cases use `t.Parallel()`, the duration of a test suite is more than that of the test case with the same name. For example, when we run this test:

```go
func TestParallel(t *testing.T) {
	t.Run("Test 1", func(t *testing.T) {
		time.Sleep(1 * time.Second)
	})

	t.Run("Test 2", func(t *testing.T) {
		t.Parallel()
		time.Sleep(2 * time.Second)
	})
}
```

The log shows:

```
=== RUN   TestParallel
=== RUN   TestParallel/Test_1
=== RUN   TestParallel/Test_2
=== PAUSE TestParallel/Test_2
=== CONT  TestParallel/Test_2
--- PASS: TestParallel (1.00s)
    --- PASS: TestParallel/Test_1 (1.00s)
    --- PASS: TestParallel/Test_2 (2.00s)
PASS
```

`TestParallel/Test_2` took 2 seconds, but the test case `TestParallel` only took 1 second. Apparently, the test suite `TestParallel` should have taken more than 1 second.

This PR uses the timestamps from JSON events to determine end to end time of a test suite.

### DIFF
--- a/go/tools/bzltestutil/xml.go
+++ b/go/tools/bzltestutil/xml.go
@@ -72,7 +72,8 @@ type testCase struct {
 	state    string
 	output   strings.Builder
 	duration *float64
-	timestamp *time.Time
+	start *time.Time
+	end *time.Time
 }
 
 const (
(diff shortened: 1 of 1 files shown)


linzhp:
CI failed due to https://github.com/bazelbuild/continuous-integration/issues/2183

PR #4287, opened 2025-03-02
Fail when expected files are not produced by protoc

### LINKED ISSUE #3949: Proposal: Fail protoc code gen on missing expected output files
The go_proto_compile rule currently generates trivial files to substitute for expected files that are missing from the output of the protoc code generation action (ref: https://github.com/bazelbuild/rules_go/pull/1394) 

We've been facing some cache poisoning issues where under certain conditions, the go_proto_compile rule will not generate a file when it should. Currently, the action succeeds and the trivial file will be cached as the output of the Bazel action, which fails subsequent downstream actions because a trivial file has substituted for Go file contents that should have been generated.

The go_proto_compile rule should fail if an output that is expected is not generated instead of generating a trivial file. One benefit is that this prevents bad outputs from being cached. I believe it should be the onus of the code generator to do the extra work of generating predictable outputs instead of allowing for the go_proto_compile rule to fill the gaps.

Because go_proto_compile currently creates a trivial file when an expected file is not generated, we can't tell easily distinguish between situations where the code generator intentionally generates a blank file to comply with Bazel's restrictions on expected outputs or where the protoc action erronously didn't generate an expected file. 

I believe we should error when expected output files are not produced in the go protoc builder https://github.com/bazelbuild/rules_go/blob/master/go/tools/builders/protoc.go#L189

linzhp:
**What type of PR is this?**
Bug fix

This PR fail the action if a plugin does not produce a valid Go file that it is supposed to produce. This disallows people to add gRPC plugin to a `go_proto_library` target that has no service definition, but avoids the cache poisoning due to bugs in proto plugins. Because Gazelle is able to switch between proto and grpc plugins depending on the existence of service definition, existing repos managed by Gazelle will not be affected.

Fixes #3949

**Other notes for review**
Credits to @r-hang for coming up with the patch

### DIFF
--- a/go/tools/builders/protoc.go
+++ b/go/tools/builders/protoc.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"go/parser"
+	"go/token"
 	"io/ioutil"
 	"log"
 	"os"
@@ -36,7 +38,7 @@ type genFileInfo struct {
 	created    bool         // Whether the file was created by protoc
 	from       *genFileInfo // The actual file protoc produced if not Path
 	unique     bool         // True if this base name is unique in expected results
-	ambiguious bool         // True if there were more than one possible outputs that matched this file
+	ambiguous bool         // True if there were more than one possible outputs that matched this file
 }
 
 func run(args []string) error {
@@ -170,8 +172,8 @@ func run(args []string) error {
 		case !copyTo.unique:
 			// not unique, no copy allowed
 		case copyTo.from != nil:
-			copyTo.ambiguious = true
-			info.ambiguious = true
+			copyTo.ambiguous = true
+			info.ambiguous = true
 		default:
 			copyTo.from = info
 			copyTo.created = true
@@ -183,15 +185,9 @@ func run(args []string) error {
 	for _, f := range files {
 		switch {
 		case f.expected && !f.created:
-			// Some plugins only create output files if the proto source files have
-			// have relevant definitions (e.g., services for grpc_gateway). Create
-			// trivial files that the compiler will ignore for missing outputs.
-			data := []byte("// +build ignore\n\npackage ignore")
-			if err := ioutil.WriteFile(abs(f.path), data, 0644); err != nil {
-				return err
-			}
-		case f.expected && f.ambiguious:
-			fmt.Fprintf(buf, "Ambiguious output %v.\n", f.path)
+			return fmt.Errorf("file %q expected from plugin %q but not created", f.path, *plugin)
+		case f.expected && f.ambiguous:
+			fmt.Fprintf(buf, "Ambiguous output %v.\n", f.path)
 		case f.from != nil:
 			data, err := ioutil.ReadFile(f.from.path)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


fmeum (approved):

sluongng:
This is a bit too strict.

We have several use cases where a proto_library target would contain both message and service proto files inside `srcs` attribute. Then the go_proto_library target would wrap that proto_library target with both the regular compiler and the grpc compiler.

https://github.com/buildbuddy-io/buildbuddy/blob/87810234ab7ec1be168288d078a0d718d8e6ff25/proto/api/v1/BUILD#L36-L70

OR, a gazelle-generated BUILD like this in one of our dependencies

```
load("@rules_proto//proto:defs.bzl", "proto_library")

proto_library(
    name = "apis_proto",
    srcs = [
        "classification.proto",
        "get_model_metadata.proto",
        "inference.proto",
        "input.proto",
        "model.proto",
        "predict.proto",
        "prediction_service.proto", # service proto
        "regression.proto",
    ],
    deps = [
        "//tensorflow/core/example:example_proto",
        "//tensorflow/core/framework:framework_proto",
        "//tensorflow/core/protobuf:protobuf_proto",
        "@com_google_protobuf//:any_proto",
        "@com_google_protobuf//:wrappers_proto",
    ],
)

go_proto_library(
    name = "apis_go_proto",
    compilers = ["@io_bazel_rules_go//proto:go_grpc_v2"],
    importpath = "github.com/buildbuddy-io/tensorflow-proto/tensorflow_serving/apis",
    proto = ":apis_proto",
    deps = [
        "//tensorflow/core/example",
        "//tensorflow/core/framework",
        "//tensorflow/core/protobuf",
    ],
)

go_library(
    name = "apis",
    embed = [":apis_go_proto"],
    importpath = "github.com/buildbuddy-io/tensorflow-proto/tensorflow_serving/apis",
)

alias(
    name = "go_default_library",
    actual = ":apis",
)
```

Moersity:
> This is a bit too strict.
> 
> We have several use cases where a proto_library target would contain both message and service proto files inside `srcs` attribute. Then the go_proto_library target would wrap that proto_library target with both the regular compiler and the grpc compiler.
> 
> https://github.com/buildbuddy-io/buildbuddy/blob/87810234ab7ec1be168288d078a0d718d8e6ff25/proto/api/v1/BUILD#L36-L70
> 
> OR, a gazelle-generated BUILD like this in one of our dependencies
> 
> ```
> load("@io_bazel_rules_go//go:def.bzl", "go_library")
> load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
> load("@rules_proto//proto:defs.bzl", "proto_library")
> 
> proto_library(
>     name = "apis_proto",
>     srcs = [
>         "classification.proto",
>         "get_model_metadata.proto",
>         "inference.proto",
>         "input.proto",
>         "model.proto",
>         "predict.proto",
>         "prediction_service.proto", # service proto
>         "regression.proto",
>     ],
>     visibility = ["//visibility:public"],
>     deps = [
>         "//tensorflow/core/example:example_proto",
>         "//tensorflow/core/framework:framework_proto",
>         "//tensorflow/core/protobuf:protobuf_proto",
>         "@com_google_protobuf//:any_proto",
>         "@com_google_protobuf//:wrappers_proto",
>     ],
> )
> 
> go_proto_library(
>     name = "apis_go_proto",
>     compilers = ["@io_bazel_rules_go//proto:go_grpc_v2"],
>     importpath = "github.com/buildbuddy-io/tensorflow-proto/tensorflow_serving/apis",
>     proto = ":apis_proto",
>     visibility = ["//visibility:public"],
>     deps = [
>         "//tensorflow/core/example",
>         "//tensorflow/core/framework",
>         "//tensorflow/core/protobuf",
>     ],
> )
> 
> go_library(
>     name = "apis",
>     embed = [":apis_go_proto"],
>     importpath = "github.com/buildbuddy-io/tensorflow-proto/tensorflow_serving/apis",
>     visibility = ["//visibility:public"],
> )
> 
> alias(
>     name = "go_default_library",
>     actual = ":apis",
>     visibility = ["//visibility:public"],
> )
> ```

We encountered the same problem, so what should we do now? downgrage rules_go version seem likes a bad idea

linzhp:
We should revert this PR

PR #4295, opened 2025-03-19
Fix goos and goarch options in go_wrap_sdk

patrickmscott:
**What type of PR is this?**

Bug fix

Fixes #

**Other notes for review**

I am using `go_wrap_sdk` for a single toolchain override that only works on linux/amd64. Passing goos and goarch results in an error because _go_wrap_sdk does not have those attributes.

### DIFF
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -403,15 +403,15 @@ _go_wrap_sdk = repository_rule(
 )
 
 def go_wrap_sdk(name, register_toolchains = True, **kwargs):
-    _go_wrap_sdk(name = name, **kwargs)
     _go_toolchains(
         name = name + "_toolchains",
         sdk_repo = name,
         sdk_type = "remote",
         sdk_version = kwargs.get("version"),
-        goos = kwargs.get("goos"),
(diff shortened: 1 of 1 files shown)


go/private/sdk.bzl
```diff
@@ -403,15 +403,15 @@ _go_wrap_sdk = repository_rule(
...
 def go_wrap_sdk(name, register_toolchains = True, **kwargs):
-    _go_wrap_sdk(name = name, **kwargs)
     _go_toolchains(
         name = name + "_toolchains",
         sdk_repo = name,
         sdk_type = "remote",
         sdk_version = kwargs.get("version"),
-        goos = kwargs.get("goos"),
-        goarch = kwargs.get("goarch"),
+        goos = kwargs.pop("goos", None),
```
fmeum, line 411: This dependence on ordering is a bit too subtle for my taste. Could you introduce `goos`/`goarch` local variables that are assigned the result of `kwargs.pop`?

patrickmscott: Sure thing

fmeum (approved):

PR #4298, opened 2025-03-24
Pass headers along as transitive dependencies

### LINKED ISSUE #4154: Binary that depends on `cloudflare/circl` can't be built cross-platform
0.50.1

0.39.1

7.4.0

Yes

Darwin ARM64

This is related to the infamous `circl` dependency.

### What did you do?
I have a go_binary that through a long list of dependencies ends up depending on `circl`. I do get circl from the BCR: 

```
bazel_dep(
    name = "circl",
    version = "1.3.8",
)
```

My binary rule is defined as such:

```
go_binary(
    name = "api",
    embed = [":api_lib"],
    gotags = ["purego"],
    pure = "on",
)
```

I try to build it like:

```
bazel build //control-plane/api --platforms=@rules_go//go/toolchain:linux_arm64
```

My error:
```
ERROR: /private/var/tmp/_bazel_amir/3dbd0b78d662a8a6e641b2d6e1f7442e/external/bazel_tools/tools/cpp/BUILD:58:19: in cc_toolchain_alias rule @@bazel_tools//tools/cpp:current_cc_toolchain: 
        File "/virtual_builtins_bzl/common/cc/cc_toolchain_alias.bzl", line 26, column 48, in _impl
        File "/virtual_builtins_bzl/common/cc/cc_helper.bzl", line 225, column 17, in _find_cpp_toolchain
Error in fail: Unable to find a CC toolchain using toolchain resolution. Target: @@bazel_tools//tools/cpp:current_cc_toolchain, Platform: @@rules_go~//go/toolchain:linux_arm64, Exec platform: @@bazel_tools//tools:host_platform
ERROR: /private/var/tmp/_bazel_amir/3dbd0b78d662a8a6e641b2d6e1f7442e/external/bazel_tools/tools/cpp/BUILD:58:19: Analysis of target '@@bazel_tools//tools/cpp:current_cc_toolchain' failed
```

I expected this build to not need the C++ toolchain.

### What did you see instead?
An error for C++ toolchain.

patrickmscott:
This fixes github.com/cloudflare/circl relative imports by making those headers available at compile time.

**What type of PR is this?**

Feature

Expose headers in GoArchive so that compilepkg can make them visible. This will make it possible for relative imports to find header files.

Fixes #4154 

**Other notes for review**


go/private/actions/compilepkg.bzl
```diff
@@ -97,7 +97,8 @@ def emit_compilepkg(
     sdk = go.sdk
     inputs_direct = (sources + embedsrcs + [sdk.package_list] +
                      [archive.data.export_file for archive in archives])
-    inputs_transitive = [sdk.headers, sdk.tools, go.stdlib.libs]
+    inputs_transitive = ([sdk.headers, sdk.tools, go.stdlib.libs] +
+                         [archive.data._headers for archive in archives])
```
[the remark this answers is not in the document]
patrickmscott: These are header files from other archives.

go/private/actions/archive.bzl
```diff
@@ -196,6 +196,7 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         _validation_output = out_nogo_validation,
         _nogo_fix_output = out_nogo_fix,
         _cgo_deps = cgo_deps,
+        _headers = depset([f for f in source.srcs if f.path.split(".")[-1].lower().startswith("h")]),
```
fmeum, line 199: From a maintainability perspective, it would be better if these could be wrapped with `cc_common.create_compilation_context` and added into `cgo_deps` so that we only have a single place where headers are tracked and added from. 

But if that's too complicated or doesn't work, the current approach would be fine too

fmeum: Makes sense, forget my idea in that case :⁠-⁠)

Some non-Go files may be consumed locally (think `.s`), so that may add too much. We can start with `.h` and expand as long as we keep the attribute private, as you did.

@jayconrod Do you see a problem with this (or a better approach)?

jayconrod: So these .h files are Go assembly headers, not C/C++/regular assembly headers, right? I don't really remember how these work, but it seems sensible to include them in `GoArchive` regardless of whether cgo is enabled. We can't really distinguish them from C header files, but maybe that's fine.

Not sure I get why they should be transitive though. Is it because headers from dependencies need to include other headers from their dependencies? If so, the test should demonstrate that.

patrickmscott: Transitive was my first guess. I figured they are not "direct" in that the go package depends on the compiled go archive. I'm not sure what the implications are with direct vs transitive in this case.

jayconrod: For a given package, all the files in that package needed to produce the compiled archive are typically direct dependencies. If a file is needed to build other packages that import that package, then it could go into the transitive set.

Go typically avoids transitive file dependencies, since they can cause O(n^2) growth. The main use case is linking the final binary.

I'm not sure about assembly headers though. You mentioned `github.com/cloudflare/circl`. Is there a specific package within that module we should look at?

armfazh: In package [circl/dh/x25519](https://pkg.go.dev/github.com/cloudflare/circl@v1.6.0/dh/x25519), this [assembler file](https://github.com/cloudflare/circl/blob/main/dh/x25519/curve_amd64.s#L7) makes use of a header file defined in package [circl/math/fp25519](https://github.com/cloudflare/circl/blob/main/math/fp25519/fp_amd64.h).

albertocavalcante:
I don't want to deviate the PR discussion, just wondering, this change alone should fix `circl` issues or it would also require a change in `gazelle`?

go/private/actions/archive.bzl
patrickmscott: In my testing, this change alone fixed circl

patrickmscott:
> I don't want to deviate the PR discussion, just wondering, this change alone should fix `circl` issues or it would also require a change in `gazelle`?

In my testing, this change alone fixed `circl`.

go/private/actions/compilepkg.bzl
```diff
@@ -95,8 +95,10 @@ def emit_compilepkg(
         archives = archives + [go.coverdata]
 
     sdk = go.sdk
-    inputs_direct = (sources + embedsrcs + [sdk.package_list] +
-                     [archive.data.export_file for archive in archives])
+    inputs_direct = sources + embedsrcs + [sdk.package_list]
+    for archive in archives:
+        inputs_direct.append(archive.data.export_file)
+        inputs_direct.extend(archive.data._headers)
```
jayconrod, line 101: It looks like this will add headers from direct dependencies, but you also need headers from transitive dependencies right (in case a header includes another header)?

I think the right way to do this is:

1. Add a `_headers` field to `GoArchive` in `emit_archive`. It should be a `depset` of `File`. The `direct` field should be headers from the current package. The `transitive` field should be the `GoArchive._headers` depsets from dependencies.
2. Pass that set through to `emit_compilepkg` as a new `headers` parameter. It's a `depset`, so just add it to the `inputs_transitive` list there.

tests/core/go_library/BUILD.bazel
```diff
@@ -28,8 +28,10 @@ go_library(
     srcs = [
         "inc.go",
         "inc_asmhdr.s",
+        "inc_relative.s",
```
jayconrod, line 31: This tests including headers from direct dependencies with relative paths, but I think it's also important to test including headers from transitive dependencies. Let's add `tests/core/transitive_headers` with three packages each with a header file (`a.h`, `b.h`, and `c.h`). `a.h` should include `b.h`, and `b.h` should include `c.h`. The package for `a` should not directly depend on `c`, though the transitive header will be needed for compilation to succeed.

fmeum:
@patrickmscott Friendly ping, do you have capacity to address Jay's comments?

go/private/actions/compilepkg.bzl
patrickmscott: I _think_ I addressed this issue. I still added `_headers` to `GoArchiveData` in order to collect transitive headers.

go/private/actions/archive.bzl
```diff
@@ -196,6 +196,7 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         _validation_output = out_nogo_validation,
         _nogo_fix_output = out_nogo_fix,
         _cgo_deps = cgo_deps,
+        _headers = tuple([f for f in source.srcs if f.path.split(".")[-1].lower().startswith("h")]),
```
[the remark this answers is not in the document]
patrickmscott: I ended up using `GoArchiveData` to keep all transitive headers. Let me know if you still think it can be removed.

go/private/actions/compilepkg.bzl
jayconrod: Please add to `GoArchive` instead.

`GoArchiveData` holds information about one specific package. It shouldn't contain information about dependencies. Instead, `GoArchive` provides access to all transitively imported packages through the `transitive` field, which is a depset of `GoArchiveData`.

Normally I'd suggest reconstructing the headers depset in each package by iterating over all the `GoArchiveData` accessible from `GoArchive`, but that makes analysis at least O(n^2), maybe worse, even in the common case where there are no headers. There's probably no avoiding O(n^2) here, but perhaps we can avoid most of that by adding a `_headers` field that only contains the headers and is empty most of the time.

patrickmscott: Maybe I'm doing something wrong but adding to `GoArchive` is causing infinite recompiles of stdlib.

patrickmscott: I think I found my mistake

patrickmscott: Updated with `headers` added to `GoArchive`. Happy to rename to `_headers` if you feel strongly.

jayconrod: Please do rename to `_headers`. It communicates that it's a private field. I don't think other rules need to depend on it.

patrickmscott: done

jayconrod (approved):
Looks good, thanks for your patience. @fmeum any last thoughts on this one?

fmeum (approved):
No, looks good to me too!

armfazh:
I've created a minimal reproducible example showing (now solved) CIRCL [Bazel-Compilation](https://github.com/cloudflare/circl/wiki/Bazel-Compilation).

PR #4301, opened 2025-03-25
Add wrap extension tag to implement go_wrap_sdk

### LINKED ISSUE #4300: Add go_wrap_sdk to bzlmod extension
`go_wrap_sdk` is not available in bzlmod. WORKSPACE support is ending so it will be good to have it under bzlmod.

patrickmscott:
Add all wrapped sdks to the beginning of the toolchains so any specific platforms override the downloaded toolchain.

Add a test that transitions to a different sdk from the downloaded version. The test asserts the correct version.

**What type of PR is this?**

>
> Bug fix
Feature
> Documentation
> Other

Fixes #4300 

**Other notes for review**

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -183,7 +183,7 @@ def _go_sdk_impl(ctx):
             name = _default_go_sdk_name(
                 module = module,
                 multi_version = multi_version_module[module.name],
-                tag_type = "local",
+                tag_type = "wrap",
                 index = index,
             )
             go_wrap_sdk_rule(
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -14,7 +14,7 @@
 
 load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
 load("//go/private:nogo.bzl", "DEFAULT_NOGO", "NOGO_DEFAULT_EXCLUDES", "NOGO_DEFAULT_INCLUDES", "go_register_nogo")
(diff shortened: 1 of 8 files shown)


go/private/extensions.bzl
```diff
@@ -158,6 +177,30 @@ def _go_sdk_impl(ctx):
     host_detected_goos, host_detected_goarch = detect_host_platform(ctx)
     toolchains = []
     for module in ctx.modules:
+        # Apply wrapped toolchains first to override specific platforms from the
+        # default toolchain or any downloads.
+        for index, wrap_tag in enumerate(module.tags.wrap):
+            name = _default_go_sdk_name(
+                module = module,
+                multi_version = multi_version_module[module.name],
+                tag_type = "local",
```
fmeum, line 186: Let's call this `wrap` instead as you could conceivably wrap a remote SDK.

fmeum (approved):
Looks good, thanks!

I would have expected https://github.com/bazel-contrib/bazel-gazelle/blob/186298911d38850b47b198e8d933a93125ce7043/docs/BUILD.bazel#L17 to fail though. Could you check what happens if you run `//docs:update`?

patrickmscott:
> Looks good, thanks!
> 
> I would have expected https://github.com/bazel-contrib/bazel-gazelle/blob/186298911d38850b47b198e8d933a93125ce7043/docs/BUILD.bazel#L17 to fail though. Could you check what happens if you run `//docs:update`?

That is gazelle. Looks like rules_go _also_ has //docs:update but it was fine.

fmeum (approved):

PR #4316, opened 2025-04-11
deps: Get repositories.bzl in sync with go.mod

### LINKED ISSUE #4262: go.mod  out of sync with go/private/repositories.bzl
Even before 05ef4076cf853bd5faf8889a8218ad6a32219293, they were already out of sync a bit, but now moreso.
Compare,
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go.mod#L9-L25
vs. for example
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go/private/repositories.bzl#L297
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go/private/repositories.bzl#L225
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go/private/repositories.bzl#L269
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go/private/repositories.bzl#L131
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go/private/repositories.bzl#L186
https://github.com/bazel-contrib/rules_go/blob/4c47d52cefdf73a3097f5061e53421e10c35a314/go/private/repositories.bzl#L205

This may won't cause immediate problems for most people, but it would be nice to keep them in sync until such time as support for WORKSPACE is dropped entirely.

adam-azarchs:
Bump the versions of dependencies in repositories.bzl to match what's in
go.mod.

Specifically bumps:

  - golang.org/x/sys
  - google.golang.org/protobuf
  - google.golang.org/grpc/cmd/protoc-gen-go-grpc
  - google.golang.org/genproto
  - github.com/golang/mock

Additionally add features to the `releaser` tool for updating to a specific version, rather than latest, so we can avoid further modifications to `go.mod`.

Fixes #4262

### DIFF
--- a/go/private/repositories.bzl
+++ b/go/private/repositories.bzl
@@ -198,13 +198,13 @@ def go_rules_dependencies(force = False):
     wrapper(
         http_archive,
         name = "org_golang_google_grpc_cmd_protoc_gen_go_grpc",
-        sha256 = "6b0ebb2360fee569d930c1ff9e4d09de4ed982a305f81d157349e3c680765c74",
-        # cmd/protoc-gen-go-grpc/v1.5.1, from 2024-07-29
+        sha256 = "1e84df03c94d1cded8e94da7a2df162463f3be4c7a94289d85c0871f14c7b8e3",
+        # cmd/protoc-gen-go-grpc/v1.3.0, latest as of 2024-05-20
         urls = [
-            "https://mirror.bazel.build/github.com/grpc/grpc-go/archive/refs/tags/cmd/protoc-gen-go-grpc/v1.5.1.zip",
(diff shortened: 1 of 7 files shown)


adam-azarchs:
Some examples are failing to compile with the updated grpc.  Which is weird because they should have been failing already under with bzlmod, where they'd have been using what's in `go.mod` already.

adam-azarchs:
I guess maybe I'll just revert the grpc/protobuf changes for now and limit this to `golang.org/x/sys` and `github.com/golang/mock`, neither of which should contain any breaking changes.

fmeum (approved):

PR #4320, opened 2025-04-17
go_cross: allow transition on compilation_mode

sluongng:
**What type of PR is this?**

Feature

Let user change the compilation mode with go_cross_binary.
This is useful to force a opt mode for a deployment or packaging
purposes.

**Other notes for review**

### DIFF
--- a/docs/go/core/rules.md
+++ b/docs/go/core/rules.md
@@ -181,7 +181,7 @@ This builds an executable from a set of source files,
 ## go_cross_binary
 
 <pre>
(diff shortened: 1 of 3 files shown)


fmeum (approved):
I was about to reject this as `go_cross_binary` shouldn't grow too much (we should recommend extensions/with_cfg instead), but `compilation_mode` does meet the bar in terms of simplicity and usefulness.

Could you update the docs to fix CI?

sluongng:
docs should be updated now

fmeum (approved):

PR #4321, opened 2025-04-17
docs: added note about x/tools when using vet with nogo

emilrowland:
**What type of PR is this?**

Documentation

Makes it visible that there is a dependency on `golang.org/x/tools`.
If `golang.org/x/tools` is not in the go.mod file, then you get import errors on the standard library when using nogo.

**Other notes for review**


fmeum:
Hmm, this is certainly not the desired behavior. Could you provide steps to reproduce this as well as the error you are seeing?

PR #4322, opened 2025-04-19
Export new location of support files needed for using Go's WASM output.

malt3:
**What type of PR is this?**

Bug fix

In order to run code with GOOS=js GOARCH=wasm in the browser, you need to copy a `.js` file from the toolchain.
This file location was changed in Go 1.24, which is now also exported in this PR.

n/a

**Other notes for review**

See also this info from the Go wiki:

> For Go 1.23 and earlier, the wasm support files needed in this article are located in misc/wasm, and the path should be replaced when performing operations with files such as lib/wasm/wasm_exec.js.

[source](https://go.dev/wiki/WebAssembly)

### DIFF
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -117,7 +117,11 @@ filegroup(
 exports_files(
     glob([
         "lib/time/zoneinfo.zip",
-        "misc/wasm/**",
+        # wasm support files including wasm_exec.js
+        # for GOOS=js GOARCH=wasm
+        # located in misc/wasm/ (Go 1.23 and earlier)
+        # or lib/wasm/ (Go 1.24 and later)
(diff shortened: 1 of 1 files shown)


fmeum (approved):

go/private/BUILD.sdk.bazel
```diff
@@ -118,6 +118,7 @@ exports_files(
     glob([
         "lib/time/zoneinfo.zip",
         "misc/wasm/**",
+        "lib/wasm/**",
```
fmeum, line 121: Could you extract the two `wasm` glob patterns into a separate glob call that allows empty matches?

malt3:
I see, I'll need to update the glob to match the old and the new path in one pattern.

fmeum:
That's even better

PR #4332, opened 2025-04-29
Work around incompatibility between zlib and clang 17

fmeum:
Work around https://github.com/bazelbuild/continuous-integration/issues/2269


fmeum:
@linzhp @tyler-french FYI

PR #4334, opened 2025-05-06
Add stampsrcs attribute

jscissr:
**What type of PR is this?**

Feature

This change adds the stampsrcs attribute, which allows providing a list of files containing variable definitions, which are used in addition to the stable and volatile status files and have the same format.

This is needed for substituting values from files into `x_defs`. Currently, this is only possible for the workspace status.

https://github.com/bazel-contrib/rules_go/issues/3507

**Other notes for review**

This PR is only a draft implementation to show my proposal. If you decide to accept the feature, I can write documentation and tests. Maybe you also have a better idea for the name instead of stampsrcs.


jscissr:
Maintainers suggested using embed instead. I switched to that and no longer need stampsrcs. If someone else wants this feature, you can take my code and create a new PR.

PR #4336, opened 2025-05-06
Re-spawned test binary should not break PWD

### LINKED ISSUE #4335: bzltestutil changes the working directory for respawned test-executables
## Background

Some Go tests follow a pattern where they re-spawn the test executable
with a different name or different environment variables.
This usually acts as a means of mocking behavior of an external process
which the test does not want to spawn directly.

This usually takes the form:

```go
func TestMain(m *testing.M) {
	switch filepath.Base(os.Args[0]) {
	case "git":
	    os.Exit(fakeGit())
	case "bazel":
        os.Exit(fakeBazel())
    case "pwd":
        os.Exit(fakePwd())
    default:
        // Not a fake process, run the tests normally.
        os.Exit(m.Run())
	}
}
```

The test then uses `os.Executable()` to determine the name of the executable,
and then takes one of the following approaches to re-spawn the test executable:

- copy the binary to `TempDir()/$name`
  (where `$name` is the name of the fake process from the TestMain)
- symlink to `TempDir()/$name`
- change `os.Args[0]` to `$name` directly while still using the original binary

Regardless of the approach, the test re-spawns the test executable,
it runs through the [pre-TestMain logic][1], calls TestMain,
which then hijacks the process to emulate the external process and exit.

  [1]: https://github.com/golang/go/blob/30b2b767d6d902787b90476fd00eee4c9b3a3f15/src/cmd/go/internal/load/test.go#L843

This method also powers the [go-internal/testscript][2] package's
external process support.

  [2]: https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript

## Problem

bzltestutil, imported by rules_go's generated test binaries,
changes the working directory of the spawned executable upon entry
and there's no easy way to opt out of this behavior.

This means that the emulated process will switch to the incorrect directory
before running the logic that hijacks the binary.

For example, if `fakePwd` above were to return the output of `os.Getwd()`,
and the test did the equivalent of:

```go
cmd := exec.Command(testExe)
cmd.Args[0] = "pwd"
cmd.Dir = t.TempDir()

out, err := cmd.Output()
if err != nil {
    t.Fatal(err)
}
```

The output would not be `t.TempDir()`,
but the same directory as the test executable.

abhinav:
**What type of PR is this?**

Bug fix

This PR adds a test to reproduce the issue described in #4335,
and proposes one possible fix for it:
leveraging an environment variable to help the subprocess determine
whether it was invoked by Bazel or directly.

The test fails without the proposed fix,
but it succeeds fine with `go test`.

Fixes #4335

**Other notes for review**

The proposed fix is one possible way of doing this.
I welcome feedback from maintainers on other approaches here.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -192,7 +192,13 @@ def _go_test_impl(ctx):
         info_file = ctx.info_file,
     )
 
-    env = {}
+    env = {
+        # The test binary uses this to decide
+        # whether it was invoked by Bazel or directly.
+        # If invoked directly, it will not change its working directory
+        # to run_dir configured above.
+        "GO_TEST_RUN_FROM_BAZEL": "1",
+    }
     for k, v in ctx.attr.env.items():
         env[k] = ctx.expand_location(v, ctx.attr.data)
 
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -165,6 +165,15 @@ func testsInShard() []testing.InternalTest {
 }
 
 func main() {
+	// When the test process is originally spawned by Bazel,
+	// it should run in a test directory.
+	// However, if the test process is re-spawned by the test,
+	// it should run wherever that process is set to run.
+	//
+	// Clearing this environment variable will opt the child process
+	// out of the Chdir behavior.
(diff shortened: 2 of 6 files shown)


fmeum (approved):

fmeum:
Thanks!

PR #4338, opened 2025-05-12
go/tools/gopackagesdriver: Adding Cgo support

r-hang:
go/tools/gopackagesdriver: Filter cgo sources and add cgo processed files
    
The gopackagesdriver today fails to load Go programs
with cgo dependencies.
```
go-code/bazel-pkgdrv/external/com_github_mattn_go_sqlite3/backup.go:16:8: could not import C (no metadata for C)
```

These errors come from two sources:
- First, the gopackagesdriver includes cgo sources files in the
  CompiledGoFiles response.
- Second, the gopackagesdriver response and CompiledGoFiles do not
  contain cgo processed Go sources necessary to compile the program.

This change addresses both error points by
- Exposing the location of cgo generated artifacts in the rules_go archive object.
- Filtering out cgo source files from the CompiledGoFiles response, by
checking if the files import "C".
- Adding cgo processed files to the CompiledGoFiles response by exploiting
knowledge of how rules_go compiles cgo files.

I am not sure if there is a better solution to leaking build system
details to the gopackagesdriver as even 'go list' leaks where
the go build cache stores cgo processed files. Though in the case of 'go list'
the conventions for cgo processed files are a bit different, they don't
seem to have a file extension.

For example,
```
$ go list -json=ImportPath,CgoFiles,GoFiles,CompiledGoFiles -compiled -find github.com/mattn/go-sqlite3
 "CompiledGoFiles": [
                "convert.go",
                "doc.go",
                "sqlite3_func_crypt.go",
                "sqlite3_go18.go",
                "sqlite3_opt_preupdate.go",
                "sqlite3_opt_preupdate_omit.go",
                "/Users/rhang/Library/Caches/go-build/8f/8f8bfe0b75fcceeb093e990af3eaa4f05147147b13db050b2f706cb563280415-d",
                "/Users/rhang/Library/Caches/go-build/9c/9c0b42e0a548721eb17f0fe5cfc2b6f1c0f0f7fcfb0f0caf526229723b9d4927-d",
                "/Users/rhang/Library/Caches/go-build/f7/f7e25424abcdc8442dbe22670456cab79a881268995c4ed43715e6d56930bff6-d",
...
```

This PR resolves https://github.com/bazel-contrib/rules_go/issues/4337 additional
details are provided in the issue.

### DIFF
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -13,13 +13,15 @@
 # limitations under the License.
 
 load(
-    "@bazel_skylib//lib:paths.bzl",
-    "paths",
-)
-load(
     "//go/private:providers.bzl",
     "GoArchive",
     "GoStdLib",
+)
+load(
+    "//go/tools/gopackagesdriver/pkgjson:pkg_json.bzl",
+    "write_pkg_json",
+    file_path_lib = "file_path",
+    is_file_external_lib = "is_file_external",
 )
 
 GoPkgInfo = provider()
@@ -39,42 +41,14 @@ def bazel_supports_canonical_label_literals():
     return str(Label("//:bogus")).startswith("@@")
 
 def is_file_external(f):
-    return f.owner.workspace_root != ""
+    return is_file_external_lib(f)
 
 def file_path(f):
-    prefix = "__BAZEL_WORKSPACE__"
-    if not f.is_source:
-        prefix = "__BAZEL_EXECROOT__"
-    elif is_file_external(f):
-        prefix = "__BAZEL_OUTPUT_BASE__"
-    return paths.join(prefix, f.path)
+    return file_path_lib(f)
 
-def _go_archive_to_pkg(archive):
-    go_files = [
-        file_path(src)
-        for src in archive.data.srcs
-        if src.path.endswith(".go")
-    ]
-    return struct(
-        ID = str(archive.data.label),
-        PkgPath = archive.data.importpath,
-        ExportFile = file_path(archive.data.export_file),
-        GoFiles = go_files,
-        CompiledGoFiles = go_files,
-        OtherFiles = [
-            file_path(src)
-            for src in archive.data.srcs
-            if not src.path.endswith(".go")
-        ],
-        Imports = {
-            pkg.data.importpath: str(pkg.data.label)
-            for pkg in archive.direct
-        },
-    )
-
-def make_pkg_json(ctx, name, pkg_info):
+def make_pkg_json(ctx, pkg_json_tool, name, archive):
     pkg_json_file = ctx.actions.declare_file(name + ".pkg.json")
-    ctx.actions.write(pkg_json_file, content = json.encode(pkg_info))
+    write_pkg_json(ctx, pkg_json_tool, archive, pkg_json_file)
     return pkg_json_file
 
 def _go_pkg_info_aspect_impl(target, ctx):
@@ -110,16 +84,16 @@ def _go_pkg_info_aspect_impl(target, ctx):
     if GoArchive in target:
         archive = target[GoArchive]
         compiled_go_files.extend(archive.source.srcs)
+        if archive.data.cgo_out_dir:
+            compiled_go_files.append(archive.data.cgo_out_dir)
         export_files.append(archive.data.export_file)
-        pkg = _go_archive_to_pkg(archive)
-        pkg_json_files.append(make_pkg_json(ctx, archive.data.name, pkg))
+        pkg_json_files.append(make_pkg_json(ctx, ctx.executable._pkgjson, archive.data.name, archive))
 
         if ctx.rule.kind == "go_test":
             for dep_archive in archive.direct:
                 # find the archive containing the test sources
                 if archive.data.label == dep_archive.data.label:
-                    pkg = _go_archive_to_pkg(dep_archive)
-                    pkg_json_files.append(make_pkg_json(ctx, dep_archive.data.name, pkg))
+                    pkg_json_files.append(make_pkg_json(ctx, ctx.executable._pkgjson, dep_archive.data.name, dep_archive))
                     compiled_go_files.extend(dep_archive.source.srcs)
                     export_files.append(dep_archive.data.export_file)
                     break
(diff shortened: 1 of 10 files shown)


go/tools/gopackagesdriver/flatpackage.go
```diff
@@ -114,6 +115,90 @@ func FilterFilesForBuildTags(pak *packages.Package) {
...
+		}
+		var skip bool
+		for _, rawImport := range f.Imports {
+			imp, err := strconv.Unquote(rawImport.Path.Value)
+			if err != nil {
+				continue
+			}
+			if imp == "C" {
+				skip = true
+				continue
```
fmeum, line 158:
```suggestion
				break
```

go/tools/gopackagesdriver/flatpackage.go
```diff
@@ -114,6 +115,90 @@ func FilterFilesForBuildTags(pak *packages.Package) {
...
+	cgoGeneratedFiles = append(cgoGeneratedFiles, filepath.Join(cgoGeneratedDir, "_cgo_gotypes.go"))
+	cgoGeneratedFiles = append(cgoGeneratedFiles, filepath.Join(cgoGeneratedDir, "_cgo_imports.go"))
+	for _, csf := range cgoSourceFiles {
+		name := strings.TrimSuffix(filepath.Base(csf), ".go")
+		name = name + ".cgo1.go"
+		path := prf(filepath.Join(cgoGeneratedDir, name))
+		if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+			// Check if the cgo generated file exists. Due to situations
+			// like OS-specified build tags, we can't determine if a cgo
+			// is generate ahead of time.
```
fmeum, line 193:
```suggestion
			// is generated ahead of time.
```

fmeum (approved):
Thanks, looks great! Could you run buildifier to fix CI?

go/tools/gopackagesdriver/aspect.bzl
```diff
@@ -55,12 +55,16 @@ def _go_archive_to_pkg(archive):
...
     ]
+    cgo_out_dir = ""
+    if archive.data.cgo_out_dir:
+        cgo_out_dir = file_path(archive.data.cgo_out_dir)
     return struct(
         ID = str(archive.data.label),
         PkgPath = archive.data.importpath,
         ExportFile = file_path(archive.data.export_file),
         GoFiles = go_files,
         CompiledGoFiles = go_files,
```
linzhp, line 66: Do you think it makes more sense to remove Cgo source files from `CompiledGoFiles` and replace them with cgo generated files? So we don't need another attribute `CgoGeneratedDir`. Those who want the cgo source files can still get them from `GoFiles`

r-hang: Nice! I like this approach of moving the logic into the pkg.json generation logic. The earlier we process the Cgo files the better.

go/tools/gopackagesdriver/pkgjson/main.go
```diff
@@ -0,0 +1,228 @@
...
+	// Note: We check resolved file paths to get a path relative to the execroot
+	// but we add the unresolved file path to the package json as the gopackagesdriver
+	// uses the special bazel prefixes to resolve to absolute paths.
+	resolvedCgoGeneratedDir := resolvePath(cgoGeneratedDir)
+	path := filepath.Join(resolvedCgoGeneratedDir, "_cgo_gotypes.go")
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		return nil, nil
+	}
+	cgoGeneratedFiles = append(cgoGeneratedFiles, filepath.Join(cgoGeneratedDir, "_cgo_gotypes.go"))
+	cgoGeneratedFiles = append(cgoGeneratedFiles, filepath.Join(cgoGeneratedDir, "_cgo_imports.go"))
```
JamyDev, line 203:
```suggestion
	cgoGeneratedFiles = append(
		cgoGeneratedFiles,
		filepath.Join(cgoGeneratedDir, "_cgo_gotypes.go"),
		filepath.Join(cgoGeneratedDir, "_cgo_imports.go"),
  	)
```
nit: Single append

go/tools/gopackagesdriver/pkgjson/main.go
```diff
@@ -0,0 +1,228 @@
...
+	// present if the package contains cgo generated code.
+	//
+	// Note: We check resolved file paths to get a path relative to the execroot
+	// but we add the unresolved file path to the package json as the gopackagesdriver
+	// uses the special bazel prefixes to resolve to absolute paths.
+	resolvedCgoGeneratedDir := resolvePath(cgoGeneratedDir)
+	path := filepath.Join(resolvedCgoGeneratedDir, "_cgo_gotypes.go")
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		return nil, nil
+	}
```
JamyDev, line 201:
```suggestion
	cgotypes_path := filepath.Join(resolvedCgoGeneratedDir, "_cgo_gotypes.go")
	if _, err := os.Stat(cgotypes_path); err != nil && os.IsNotExist(err) {
		return nil, nil
	}
```
nit: `path` may be confused by the path import by readers.

go/tools/gopackagesdriver/pkgjson/main.go
```diff
@@ -0,0 +1,228 @@
...
+	cgoGeneratedFiles = append(cgoGeneratedFiles, filepath.Join(cgoGeneratedDir, "_cgo_gotypes.go"))
+	cgoGeneratedFiles = append(cgoGeneratedFiles, filepath.Join(cgoGeneratedDir, "_cgo_imports.go"))
+	for _, csf := range cgoSourceFiles {
+		name := strings.TrimSuffix(filepath.Base(csf), ".go")
+		name = name + ".cgo1.go"
+		resolvedPath := filepath.Join(resolvedCgoGeneratedDir, name)
+		unresolvedPath := filepath.Join(cgoGeneratedDir, name)
+		if _, err := os.Stat(resolvedPath); err != nil && os.IsNotExist(err) {
```
JamyDev, line 209: Seems like this combo of "Check file exists, but I only have the \"unresolved\" path" happens a few times. Can we have a utility function?

JamyDev (commented):
Also please add a test case for this functionality.

go/tools/gopackagesdriver/pkgjson/main.go
r-hang: This can't be easily placed in a utility function as what happens as a result of the check varies in each of the check cases (e.g. return nil, append to a slice, or append after checking cgo imports). Lines of code aren't really saved by making a utility function out of a check as from the check's perspective it doesn't know if the path is resolved or not.

JamyDev (approved):

PR #4339, opened 2025-05-13
Remove rules_proto

comius:
**What type of PR is this?**

> Cleanup

Replaces rules_proto uses with com_google_protobuf. The former is deprecated.

### DIFF
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -16,10 +16,7 @@ load(
     "@bazel_skylib//lib:paths.bzl",
     "paths",
 )
-load(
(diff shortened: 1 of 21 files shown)


proto/compiler.bzl
```diff
@@ -48,6 +49,30 @@ load(
 #  default.
 _PROTO_TOOLCHAIN_TYPE = "@rules_proto//proto:toolchain_type"
```
fmeum, line 50: Can this stay as is?

fmeum (approved):

fmeum (approved):

PR #4269, opened 2025-02-13
Only emit patch file if nogo has fixes

fmeum:
**What type of PR is this?**

Feature

**What does this PR do? Why is it needed?**

This makes it easier to download all relevant patches with `--remote_download_regex` without creating a bunch of empty patch files.

**Which issues(s) does this PR fix?**

**Other notes for review**

PR #4341, opened 2025-05-13
Rename `out` to `out_diagnostics`

fmeum:
**What type of PR is this?**

Refactor

Follow-up to #4269

**Other notes for review**


linzhp (approved):
nice!

PR #4344, opened 2025-05-15
Remove shadow analyzer from TOOLS_NOGO

### LINKED ISSUE #4340: Remove `shadow` analyzer from TOOLS_NOGO
`TOOLS_NOGO` is a convenient shortcut for enabling a large set of analyzers. This list currently includes the `shadow` analyzer.

However, the documentation for the [`shadow`](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/shadow) package mentions that it is not enabled by default because it generates too many false positives. In https://github.com/golang/go/issues/67440, one of the maintainers states

> The shadow analyzer has never been enabled by default in 'go vet' of gopls because it has always suffered from a high rate of false positives that no-one knows how to fix. We would rather not invest any more effort into it.

This package is known to be suboptimal by its maintainers and they have even stated that they are not interested in putting more effort into it.

For these reasons, does it make sense to remove (at least temporarily) `shadow` from the default list of analyzers in `TOOLS_NOGO`? The alternative for people that do not want to use `shadow` is to copy [the TOOLS_NOGO definition](https://github.com/bazel-contrib/rules_go/blob/fb90c467eeef04bed006546c692492a005ce0ba1/go/def.bzl#L81-L118) from the source and manually omit `shadow`, but this eliminates the usefulness of `rules_go` providing a convenient alias for the list of analyzers.

gpanders:
**What type of PR is this?**

Other

The shadow analyzer issues many false positives which are not in keeping with Go best practices. Even the upstream documentation for this analyzer admits this and claims that it is experimental. It should not be included in the default list of analyzers provided by nogo (users can still opt-in to it if they want it).

Fixes #4340

**Other notes for review**

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -103,7 +103,8 @@ _TOOLS_NOGO = [
     "@org_golang_x_tools//go/analysis/passes/nilness:go_default_library",
     "@org_golang_x_tools//go/analysis/passes/pkgfact:go_default_library",
     "@org_golang_x_tools//go/analysis/passes/printf:go_default_library",
-    "@org_golang_x_tools//go/analysis/passes/shadow:go_default_library",
+    # shadow analyzer is too noisy, see #4340
+    # "@org_golang_x_tools//go/analysis/passes/shadow:go_default_library",
     "@org_golang_x_tools//go/analysis/passes/shift:go_default_library",
     "@org_golang_x_tools//go/analysis/passes/sortslice:go_default_library",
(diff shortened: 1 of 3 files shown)


fmeum:
Looks like we are relying on this particular analyzer in tests. How about we migrate them to https://pkg.go.dev/golang.org/x/tools@v0.33.0/go/analysis/passes/assign?

fmeum (approved):

PR #4345, opened 2025-05-15
Run buildifier to add explicit loads

fmeum:
They will be required in Bazel 9.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -107,16 +107,6 @@ load("@rules_python//python:repositories.bzl", "py_repositories")
 
 py_repositories()
 
-http_archive(
-    name = "com_google_protobuf",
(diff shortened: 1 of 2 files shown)


fmeum:
Loading from `rules_cc` forces us to rely on a modern version of protobuf, which in turn we need to make work via our deps macros, but that fails since we can't run protobuf's own deps macros.

@jayconrod Do you have an idea how to make this work without breaking too much? Bazel 9 won't have WORKSPACE at all, but we'll be stuck supporting it for a while.

jayconrod:
I'm not sure I understand all the subtleties of what's happening. If protobuf becomes a mandatory dependency, we need to add its transitive dependencies to `go_rules_dependencies` right? I guess we can't call its deps macro because we're loading it there. We could flatten and inline the contents of that deps macro maybe.

Issue #2463, opened 2020-04-29
ArtifactPrefixConflictException using gazelle generated rules


robbertvanginkel:
_Not sure to file this on gazelle or on rules_go, but it seems like this is due to rules_go changes so filing here._

In case of using gazelle generated build rules and repated importpaths (like `example.com/foo/foo` where the first component `example.com/foo` points to a main package and the latter `example.com/foo/foo` to a library), can result in in `ArtifactPrefixConflictException` since https://github.com/bazelbuild/rules_go/commit/7a2ed944ae81f5c896e876e0c0975e9900832334:
```
ERROR: com.google.devtools.build.lib.actions.ArtifactPrefixConflictException: output path 'bazel-out/darwin-fastbuild/bin/foo/foo' (belonging to //foo:foo) is a prefix of output path 'bazel-out/darwin-fastbuild/bin/foo/foo/go_default_library.a' 
```

Here's steps to reproduce:
```
-- ./.bazelversion --
3.1.0
-- ./WORKSPACE --
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "io_bazel_rules_go",
    strip_prefix = "rules_go-7a2ed944ae81f5c896e876e0c0975e9900832334",
    urls = ["https://github.com/bazelbuild/rules_go/archive/7a2ed944ae81f5c896e876e0c0975e9900832334.zip"],
)
load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
go_rules_dependencies()
go_register_toolchains()
http_archive(
    name = "bazel_gazelle",
    urls = ["https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.20.0/bazel-gazelle-v0.20.0.tar.gz"],
)
load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
gazelle_dependencies()
-- ./foo/foo/foo.go --
package foo

type Foo struct{}
-- ./foo/main.go --
package main

func main() {}
-- ./BUILD.bazel --
load("@bazel_gazelle//:def.bzl", "gazelle")

# gazelle:prefix github.com/example/project
gazelle(name = "gazelle")
```

When generating build files with gazelle and attempting to build the repo:
```
$ bazel run //:gazelle
$ git diff HEAD
diff --git a/foo/BUILD.bazel b/foo/BUILD.bazel
new file mode 100644
index 0000000..08964e1
[...]

PR #4348, opened 2025-05-16
Always write the binary to the mode specific directory

mering:
**What type of PR is this?**

> Feature

Always write the binary to the mode specific directory and always create a symlink based on basename.

This approach has the following advantages:
- Reduce cache invalidation on mode changes
- Have a binary (symlink) available at basename instead of the unexpected "%s_/%s" % (name, name) subfolder

**Other notes for review**


jayconrod:
@mering Could you please create an issue describing what you want to do? It would be better to discuss the change in behavior before we get to the implementation.

Some concerns:

- Symbolic links aren't well supported on Windows. You need to be an administrator to enable them, and I'm not sure if Bazel will create a real symbolic link here or a junction. There are several kinds of link, and different software interprets them differently.
- Something that expects a real binary instead of a symbolic link may break. `go_binary` typically produces a single binary file, so it would be easy for some dependent rule to pick up the link without the binary it points to.
- I'm not clear on the cache invalidation concern. Currently, if you change `out` or `basename`, it triggers a relink for that target, but I wouldn't expect that to happen very often. What am I missing?
- I'd rather not introduce a new `out_auto` attribute. `out` and `basename` are already kind of redundant, so this makes the situation more confusing.

I wonder if a better solution for #2463 would have been for Gazelle to set `out` or `basename` explicitly when there's a subdirectory with the same name. Then `go_binary` by default could produce an executable with the same name as the target.

mering:
> @mering Could you please create an issue describing what you want to do? It would be better to discuss the change in behavior before we get to the implementation.

Created #4350.

> * Symbolic links aren't well supported on Windows. You need to be an administrator to enable them, and I'm not sure if Bazel will create a real symbolic link here or a junction. There are several kinds of link, and different software interprets them differently.

I know that for runfiles and for `pkg_tar`, `ctx.action.symlink` targets are resolved and the destination binary used directly.
So this should only change the behavior of `bazel-out` which at least on Linux already creates symlinks most of the time anyways.

> * Something that expects a real binary instead of a symbolic link may break. `go_binary` typically produces a single binary file, so it would be easy for some dependent rule to pick up the link without the binary it points to.

IIUC bazel rules should correctly resolve the symlinks.

> * I'm not clear on the cache invalidation concern. Currently, if you change `out` or `basename`, it triggers a relink for that target, but I wouldn't expect that to happen very often. What am I missing?

This is referring to the warning in the previous docstring of the `out` attribute.

> * I'd rather not introduce a new `out_auto` attribute. `out` and `basename` are already kind of redundant, so this makes the situation more confusing.

I would like to avoid having to specify the value of `name` twice for every rule. The `out_auto` could easily be applied to the whole code base via `buildozer` but copying values is not supported AFAIK.

> I wonder if a better solution for #2463 would have been for Gazelle to set `out` or `basename` explicitly when there's a subdirectory with the same name. Then `go_binary` by default could produce an executable with the same name as the target.

Yes, `go_binary` should by default produce an executable with the same name as the target. It makes much more sense to work around the edge cases where they appear instead of changing the default.

jayconrod:
Thanks for creating #4350, let's move discussion there.

PR #4353, opened 2025-05-20
Allow to exclude certain target from nogo facts collection

linzhp:
**What type of PR is this?**
Feature

After #4268, nogo collects facts from all targets. So `go_register_nogo`'s `includes` and `excludes` now overlap with the JSON config's `only_files` and `exclude_files`, which is to stop nogo from reporting errors in certain files/targets. Meanwhile, collecting facts can be slow for very large Go packages for some analyzers, slower than compilation. To address the performance issue, this PR revert `excludes` to its previous behavior: stop nogo from collecting facts from certain packages, while making `includes` of `go_register_nogo` a no-op with a warning message, asking people to use the JSON config instead. So nogo will continue to collect facts from all targets, which is the safe default, except those in `excludes`.

**Other notes for review**
To be backward-compatible, we can add a flag or attribute like `exclude_collect_facts`, but adding that on top of `excludes` and `exclude_files` makes hard to remember which is which.

### DIFF
--- a/go/config/BUILD.bazel
+++ b/go/config/BUILD.bazel
@@ -96,9 +96,3 @@ bool_flag(
     build_setting_default = False,
     visibility = ["//visibility:public"],
 )
-
-bool_flag(
-    name = "unsafe_not_generate_nogo_facts_for_excludes",
-    build_setting_default = False,
-    visibility = ["//visibility:public"],
-)
--- a/go/config/BUILD.bazel
+++ b/go/config/BUILD.bazel
@@ -96,3 +96,9 @@ bool_flag(
     build_setting_default = False,
     visibility = ["//visibility:public"],
 )
+
+bool_flag(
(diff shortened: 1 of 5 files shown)


go/tools/builders/nogo_main.go
```diff
@@ -542,6 +544,10 @@ func checkAnalysisResults(actions []*action, pkg *goPackage) ([]diagnosticEntry,
 		if baseConfig, ok := configs[nogoBaseConfigName]; ok {
 			currentConfig = baseConfig
 		}
+		if currentConfig.excludeFiles == nil {
+			// TODO(linzhp): add a test to verify the case of [].
+			currentConfig.excludeFiles = []*regexp.Regexp{_externalFilePattern}
```
linzhp, line 549: With this, most repos no longer need the JSON config unless they want to exclude certain paths in the repo. @fmeum

PR #4364, opened 2025-05-26
Allowing a go_proto_compiler to indicate whether it always generate files

### LINKED ISSUE #3949: Proposal: Fail protoc code gen on missing expected output files

linzhp:
**What type of PR is this?**
Feature

This PR allows a `go_proto_compiler` to  indicate whether it always generate files, if so, we can apply more strict check to the output of the proto compiler.

This is an alternative way to fix #3949

**Other notes for review**

### DIFF
--- a/go/tools/builders/protoc.go
+++ b/go/tools/builders/protoc.go
@@ -36,7 +36,7 @@ type genFileInfo struct {
 	created    bool         // Whether the file was created by protoc
 	from       *genFileInfo // The actual file protoc produced if not Path
 	unique     bool         // True if this base name is unique in expected results
-	ambiguous bool         // True if there were more than one possible outputs that matched this file
+	ambiguous  bool         // True if there were more than one possible outputs that matched this file
 }
 
 func run(args []string) error {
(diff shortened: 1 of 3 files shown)


linzhp:
@sluongng @Moersity can you test this?

PR #4368, opened 2025-06-03
Fix breakages with Bazel@HEAD and incompatible flags

fmeum:
**What type of PR is this?**

Adapt to a Bazel breaking change

The `local_config_platform` repo is no longer available with Bazel@HEAD, so the host constraints have to be loaded from `@platforms//host`.

Also prepare the BCR test repo for Bazel@HEAD with `--incompatible_autoload_externally=` by applying buildifier lint fixes.

**Other notes for review**

### DIFF
--- a/go/tools/releaser/boilerplate.go
+++ b/go/tools/releaser/boilerplate.go
@@ -41,7 +41,17 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 
 go_rules_dependencies()
 
-go_register_toolchains(version = "%[3]s")`, version, shasum, goVersion)
+go_register_toolchains(version = "%[3]s")
+
+# Create the host platform repository transitively required by rules_go.
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+load("@platforms//host:extension.bzl", "host_platform_repo")
+
+maybe(
+	host_platform_repo,
+	name = "host_platform",
+)
+`, version, shasum, goVersion)
 }
 
 func findLatestGoVersion() (v string, err error) {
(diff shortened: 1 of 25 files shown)


fmeum:
FYI @jayconrod @linzhp I'm going to merge this now to unblock dependent projects, but please take a look and let me know what I should change in a follow-up PR.

go/tools/releaser/boilerplate.go
```diff
@@ -41,7 +41,17 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 
 go_rules_dependencies()
 
-go_register_toolchains(version = "%[3]s")`, version, shasum, goVersion)
+go_register_toolchains(version = "%[3]s")
+
+# Create the host platform repository transitively required by rules_go.
```
jayconrod, line 46: Changing the boilerplate is a pretty substantial change that will cause friction for `WORKSPACE` users. Is there a way to roll this into `go_rules_dependencies`? I guess not if we need this `host_platform_repo`.

fmeum: I agree, but I don't see another way. We would need to add a new macro and at that point it's much more transparent to have the user instantiate the repo directly. 

I would consider this the cost of staying with a pure WORKSPACE setup - if you get at least rules_go via Bzlmod, this is not an issue for you.

jayconrod: Hmm, will this actually affect Bazel WORKSPACE users? If the Bazel authors indeed remove WORKSPACE support in 9, then WORKSPACE users will be on older versions, not affected by this change.

Maybe it's best to keep the boilerplate simple for them and disable the WORKSPACE-mode tests for newer versions of Bazel?

fmeum: The change in `go/private/rules/go_bin_for_host.bzl` does affect WORKSPACE users as it requires the new repo to be defined. We could indirect the `load` through a repo rule of our own in which we dynamically generate this label based on the Bazel version. 

But this is a dep of many other rulesets and users only have to call this once, so that additional complexity may very well not be worth it. It's not just the `load` after all, we would also need to make sure that stardoc deps are declared correctly for each variant.

PR #4371, opened 2025-06-04
go/tools/gopackagesdriver/pkgjson: Construct pkg json from file input

r-hang:
When testing the bazel driver, we ran into an error that flaged that the argument list as too long.

```
ERROR: ...json failed: (Exit -1): pkgjson failed: error executing Action command (from target //...) bazel-out/k8-opt-exec-ST-a828a81199fe/bin/external/io_bazel_rules_go/go/tools/gopackagesdriver/pkgjson/reset_pkgjson/pkgjson --id @//src/... --pkg_path ... (remaining 15 arguments skipped)

Action failed to execute: java.io.IOException: Cannot run program "/home/user/.cache/bazel/_bazel_rhang/install/7e4ce7b0d69e79cb6bd84c7f9dfefe6b/process-wrapper" (in directory "/home/user/.cache/pkgdrv/0ddf1c72b811bee41d29991c732306ef72553747/sandbox/processwrapper-sandbox/29913/execroot/__main__"): error=7, Argument list too long
ERROR: Build did NOT complete successfully
```

Digging into this issue, the cause is that the pkgjson command takes in all of the fields of package archive data as arguments.

To work around this, we should preserve the original approach of writing a pkg json, before  https://github.com/bazel-contrib/rules_go/pull/4338, which used Skylark builtins to write the package content directly to disk.

The pkgjson command is updated to parse ths json file directly and write out a transformed pkg json with the cgo related corrections in order to avoid limits regarding argument size.

Note:
    
This diff also makes changes to undo a breaking change (i.e. changing
the signature of the make_pkg_json function) that was made in
https://github.com/bazel-contrib/rules_go/pull/4338.

I revert changes to the functionalty of make_pkg_json and add a new
replacement function make_pkg_json_with_archive.

### DIFF
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -46,9 +46,20 @@ def is_file_external(f):
 def file_path(f):
     return file_path_lib(f)
 
-def make_pkg_json(ctx, pkg_json_tool, name, archive):
+# make_pkg_json_with_archive generates a pkg.json file from an archive
+# and supports cgo generated code.
+#
+# This function was created to avoid breaking the signature of make_pkg_json
+# and avoid adding an explicit field for cgo output files in the pkg.json.
+def make_pkg_json_with_archive(ctx, name, archive):
     pkg_json_file = ctx.actions.declare_file(name + ".pkg.json")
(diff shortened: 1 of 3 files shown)


fmeum:
@tyler-french Is this a reason for a 0.55.1?

fmeum (approved):

PR #4374, opened 2025-06-10
Gracefully handle a panicking analyzer

fmeum:
**What type of PR is this?**

Bug fix

We previously didn't show any error when an analyzer panicked and Bazel would show a "mandatory output not created" error for the nogo action.

**Other notes for review**


jayconrod (approved):

tvarney13 (commented):
❤️

PR #4379, opened 2025-06-18
Inherit `//go/config:static` and `//go/config:pure` as settings for a go_tool_transition.

### LINKED ISSUE #4378: nogo: compilation/link of `nogo_actual` only works in certain configurations
it seems that bazel is building `nogo_actual` at runtime. the binary is responsible for running analyzers against user code for linting.

https://github.com/bazel-contrib/rules_go/blob/b17435b816b36e9776e73879e973a148d15f82f5/go/private/rules/nogo.bzl#L113

the construction of this binary, among other things, drops whether or not the current execution has static linkage. this is a problem for builds which only work on static linkage.

for example: a common strategy for building against CGO is to cross-compile a statically linked binary with musl-c. my CI runner was happily doing this, until i tried to run `nogo`, at which point the resulting `nogo_actual` binary linked against a non-existent `/lib/ld-musl-x86_64.so.1`

discussion: https://bazelbuild.slack.com/archives/CDBP88Z0D/p1750086489848049

workaround:

```
diff --git a/go/private/rules/transition.bzl b/go/private/rules/transition.bzl
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -185,12 +185,10 @@
 )
 
 _common_reset_transition_dict = dict({
     "//go/private:request_nogo": False,
-    "//go/config:static": False,
     "//go/config:msan": False,
     "//go/config:race": False,
-    "//go/config:pure": False,
     "//go/config:debug": False,
     "//go/config:linkmode": LINKMODE_NORMAL,
     "//go/config:tags": [],
     "//go/config:pgoprofile": Label("//go/config:empty"),
```

```
single_version_override(
    module_name = "rules_go",
    patch_strip = 1,
    patches = [
        "//third_party/rules_go:foobar.patch",
    ],
    version = "0.55.1",
)
```

to repro, we may need 
```
bazel_dep(name = "toolchains_musl", version = "0.1.20", dev_dependency = True)
```

moribellamy:
**What type of PR is this?**

Bug fix

Enables NoGO to run in build chains where static linkage with MUSL-libc is specified

Discussed in the [bazel slack](https://bazelbuild.slack.com/archives/CDBP88Z0D/p1750086489848049)

Fixes #4378

**Other notes for review**

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -211,7 +211,7 @@ _stdlib_keep_keys = sorted([
     "//go/config:pgoprofile",
 ])
 
-_go_reset_transition_dict = dict(_common_reset_transition_dict)
+_go_reset_transition_dict = dict(_reset_transition_dict)
 _go_reset_transition_dict.pop("//go/config:static")
 _go_reset_transition_dict.pop("//go/config:pure")
 _go_reset_transition_keys = sorted(_go_reset_transition_dict.keys())
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -211,6 +211,11 @@ _stdlib_keep_keys = sorted([
     "//go/config:pgoprofile",
 ])
 
+_go_reset_transition_dict = dict(_common_reset_transition_dict)
+_go_reset_transition_dict.pop("//go/config:static")
+_go_reset_transition_dict.pop("//go/config:pure")
+_go_reset_transition_keys = sorted(_go_reset_transition_dict.keys())
+
 def _go_tool_transition_impl(settings, _attr):
     """Sets most Go settings to default values (use for external Go tools).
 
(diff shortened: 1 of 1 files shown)


moribellamy:
@fmeum not sure why the CI is failing

go/private/rules/transition.bzl
```diff
@@ -211,6 +211,11 @@ _stdlib_keep_keys = sorted([
     "//go/config:pgoprofile",
 ])
 
+_go_reset_transition_dict = dict(_common_reset_transition_dict)
```
fmeum:
```suggestion
_go_reset_transition_dict = dict(_reset_transition_dict)
```

fmeum:
Now that I've thought more about this, I think that we need a more complex solution: We should honor the `static` and `pure` flags for tools, but the `static` and `pure` attributes on rules should *not* propagate. I don't quite know how to do this yet.

PR #4389, opened 2025-07-01
Stop including deprecated properties in context by default

dzbarsky:
**What type of PR is this?**
Cleanup

We would like to drop support for these properties completely, but for now let's make them opt-in rather than opt-out. The core Go rules have not been using them for 9 months or so.

Fixes #

**Other notes for review**


dzbarsky:
We should land https://github.com/bazel-contrib/bazel-gazelle/pull/2135 first and have at least one gazelle release go out

PR #4390, opened 2025-07-02
Compute rpath correctly with nested bazel modules

dgoel:
**What type of PR is this?**

Bug fix

My company has a complicated, uncommon bazel set up with nested bazel modules. We also use cgo extensively and have many go tests and binaries that end up linking to shared C++ libraries generated via other bazel modules. When such go tests are run with rbe, they fail due to failure to locate the shared libraries within the runfiles directory.

On deeper investigation, I found that the existing `rpath` computation assumes that both the executable and the library are present within the working directory. In reality, the executable is present outside the working directory but within the runfiles directory.

Normalizing the paths with respect to the runfiles directory fixes the issue.

Relevant links:
- https://github.com/bazelbuild/bazel/issues/14307

**Other notes for review**

I added a test to verify that rpath computation works with nested modules. In fairness, this mainly guarantees that there's no regression in existing functionality as the test passes even without my fix. So far, I've not been able to produce a minimum example that showcases the problem.

Existing tests in tests/core/cgo also exercise rpath computation.

### DIFF
--- a/go/private/rpath.bzl
+++ b/go/private/rpath.bzl
@@ -17,7 +17,7 @@ load(
     "paths",
 )
 
-# TODO: Replace with the function from `bazel-skylib` when it's available.
+# TODO: Replace with a Bazel-provided function when it's available.
 # https://github.com/bazelbuild/bazel/issues/14307
 def _rlocation_path(ctx, file):
     """Returns the path relative to the runfiles directory."""
--- a/go/private/rpath.bzl
+++ b/go/private/rpath.bzl
@@ -17,6 +17,15 @@ load(
     "paths",
 )
 
+# TODO: Replace with the function from `bazel-skylib` when it's available.
+# https://github.com/bazelbuild/bazel-skylib/issues/303
+def _rlocation_path(ctx, file):
+    """Returns the path relative to the runfiles directory."""
+    if file.short_path.startswith("../"):
+        return file.short_path[3:]
+    else:
+        return ctx.workspace_name + "/" + file.short_path
+
 def _rpath(go, library, executable = None):
     """Returns the potential rpaths of a library, possibly relative to another file."""
     if not executable:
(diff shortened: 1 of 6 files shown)


go/private/rpath.bzl
```diff
@@ -17,6 +17,15 @@ load(
     "paths",
 )
 
+# TODO: Replace with the function from `bazel-skylib` when it's available.
```
fmeum: I just closed the linked issue as a duplicate, could you replace the reference?

go/private/rpath.bzl
```diff
@@ -29,12 +38,15 @@ def _rpath(go, library, executable = None):
 
     # 1. Where the executable is inside its own .runfiles directory.
     #  This is the case for generated libraries as well as remote builds.
-    #   a) go back to the workspace root from the executable file in .runfiles
-    depth = executable.short_path.count("/")
+    #   a) go back to the runfiles root from the executable file in .runfiles
+    executable_rloc = _rlocation_path(go._ctx, executable)
+    library_rloc = _rlocation_path(go._ctx, library)
```
fmeum: Move this below b) and update the comment to no longer mention "short path"

dgoel: Done!

go/private/rpath.bzl
fmeum:
```suggestion
# TODO: Replace with a Bazel-provided function when it's available.
```

go/private/rpath.bzl
```diff
@@ -29,12 +38,14 @@ def _rpath(go, library, executable = None):
 
     # 1. Where the executable is inside its own .runfiles directory.
     #  This is the case for generated libraries as well as remote builds.
-    #   a) go back to the workspace root from the executable file in .runfiles
-    depth = executable.short_path.count("/")
+    #   a) go back to the runfiles root from the executable file in .runfiles
+    executable_rloc = _rlocation_path(go._ctx, executable)
+    depth = executable_rloc.count("/")
```
fmeum:
```suggestion
    depth = _rlocation_path(go._ctx, executable).count("/")
```

go/private/rpath.bzl
```diff
@@ -29,12 +38,14 @@ def _rpath(go, library, executable = None):
...
     back_to_root = paths.join(*([".."] * depth))
 
-    #   b) then walk back to the library's short path
-    rpaths.append(paths.join(origin, back_to_root, paths.dirname(library.short_path)))
+    #   b) then walk back to the library's dir within runfiles dir.
+    library_rloc = _rlocation_path(go._ctx, library)
+    rpaths.append(paths.join(origin, back_to_root, paths.dirname(library_rloc)))
```
fmeum:
```suggestion
    rpaths.append(paths.join(origin, back_to_root, paths.dirname(_rlocation_path(go._ctx, library))))
```

PR #4391, opened 2025-07-02
Request stdlib cache dir in gopackagesdriver

jscissr:
**What type of PR is this?**

Bug fix

The stdlib_json_file contains paths of compiled cgo files, which are in the stdlib cache_dir. The gopackagesdriver reads those files, but this dependency was not explicitly declared previously.

Usually, it works despite the missing dependency. But when you have a Bazel cache configured, and then run `bazel clean` and run the gopackagesdriver again, Bazel will not restore the cache_dir from the Bazel cache. The gopackagesdriver then fails with an error like this:

error: unable to load JSON files: unable to resolve imports: open [...]/bin/external/rules_go+/stdlib_/gocache/a4/a4f91e8314b27a4bce0e8fbc01bb9736cb9ed10da747af722c5a4f2fcb1213a3-d: no such file or directory

This change adds a new output group containing the stdlib cache_dir, and requests this from the gopackagesdriver, fixing the problem.

### DIFF
--- a/go/tools/gopackagesdriver/aspect.bzl
+++ b/go/tools/gopackagesdriver/aspect.bzl
@@ -65,6 +65,7 @@ def make_pkg_json(ctx, name, pkg_info):
 def _go_pkg_info_aspect_impl(target, ctx):
     # Fetch the stdlib JSON file from the inner most target
     stdlib_json_file = None
+    stdlib_cache_dir = None
 
     transitive_json_files = []
     transitive_export_files = []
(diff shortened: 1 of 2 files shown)


blockjon-dd:
For what its worth, it appears that once we began using this version of rules_go, all of our builds started failing with this following sort of error:
```
...
fork/exec /home/builder/.cache/bazel/_bazel_builder/e77cf729f8701a609e6babb1491b4af8/external/rules_go++go_sdk+go_sdk_download/bin/go: no such file or directory
```
It seemed to taint each of our persistent CI agents where this new version was utilized. To remedy, we had to bounce our CI fleet bc the cache seems to have been corrupted in a way I dont understand yet.

PR #4393, opened 2025-07-08
go_sdk: store SDK filenames and hashes in lockfile facts

### LINKED ISSUE #3945: Calls to https://go.dev/dl/?mode=json are breaking airgapped builds - provide way to avoid these

0.47.1

0.36.0

7.2.0rc1

Yes

Windows

N/a

### What did you do?

Attempting to run bazel build from airgapped environment, the build fails unable to connect to https://go.dev/dl/?mode=json&include=all. We have created some mirrors for github.com, dl.google.com, etc on our internal artifactory, but artifactory will not mirror this URL as it uses a query.

I would like some way to input the required information into the use of rules_go in our MODULE.bazel file, and avoid the need for rules_go to make this request. I see in the code that a parameter named sdks can be used to avoid this request, but I can't see how to provide it in a module-based build.

An alternative would be to allow this URL to be overridden in the module file; then I could download this file and vendor it as versions.json

fmeum:
**What type of PR is this?**

Feature

This moves the download of the "all versions" JSON, which can't hit the repository cache, from each individual `go_download_sdk` into the module extension. If the current version of Bazel supports facts, this information will also be persisted in the lockfile, allowing for truly airgapped builds assuming an up-to-date download (formerly repository) cache.

See https://github.com/bazelbuild/bazel/pull/26198 for the PR that adds facts support to Bazel.

Fixes #3945 

**Other notes for review**

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -15,7 +15,7 @@
 load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
 load("//go/private:go_mod.bzl", "version_from_go_mod")
 load("//go/private:nogo.bzl", "DEFAULT_NOGO", "NOGO_DEFAULT_EXCLUDES", "NOGO_DEFAULT_INCLUDES", "go_register_nogo")
-load("//go/private:sdk.bzl", "detect_host_platform", "go_download_sdk_rule", "go_host_sdk_rule", "go_multiple_toolchains", "go_wrap_sdk_rule")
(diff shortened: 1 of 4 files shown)


jayconrod (approved):

go/private/extensions.bzl
```diff
@@ -370,7 +366,18 @@ def _go_sdk_impl(ctx):
...
+            "reproducible": True,
+        }
+
+        # See _download_sdk below for details on these facts.
+        if hasattr(ctx, "facts"):
+            kwargs["facts"] = {
+                version: sdk_info
+                for version, sdk_info in sdks_by_version.items()
+                if version in used_versions
+            }
```
aignas, line 379: Thanks for this example, LGTM.

PR #4397, opened 2025-07-23
Support integration test coverage system (coverageredesign)

r-hang:
go1.25 removes the current runtime code coverage system adopted by
rules_go (nocoverageredesign) https://go-review.googlesource.com/c/go/+/644997
which makes rules_go incompatible with the latest upcoming version of go.

The change completely adopts the new integration test friendly coverageredesign
system (https://go.dev/blog/integration-test-coverage) in rules_go.

Below I describe how the current rules_go coverage system currently works, what
changes in the new coverage redesign, and what needs to change as a result in
rules_go.

How nocoverageredesign currently works:

- rules_go invokes `go tool cover` to generate instrumented source files that
are then passed to the Go compiler (in compilepkg.go). Instrumented source
files update generated package variable counters.

- The rules_go generated test main (generate_test_main.go) calls
testing.RegisterCover to set a global variable that tracks references to
coverage generated package local variables that coverage instrumented code
modify (ref: https://github.com/golang/go/blob/go1.24.5/src/testing/cover.go#L72)
This replicates the behavior in go test's generated test main (cmd/go/internal/load/test.go)

- rules_go specially adds a call to coverdata.RegisterFile call to all coverage
instrumented files which connects each of the file coverage counter variables
with a global object and enables rules_go to customize the file name that
coverage variables are associated with.

- It's important to note that because rules_go accesses to this global coverage struct,
it's able to manipulate the file path name associated with coverage counter variables.
This is important because the lcov format that bazel uses expect exec root relative files.

- When a test exits, This global object is flushed in a call to `testing.coverReport()` to
generate a coverage report. This reports gets written a file configured in
test.coverprofile flag.

- Via a special SetPanicOnExit0 hook, the coverage file set in test.coverprofile
is coverted to a Bazel friendly Lcov format via logic in bzltestutil/lcov.go

What changes in coverageredesign:

- A new flexible system for configuring coverage behavior. Instead of a single
esting.RegisterCover function, there is a new testdeps package that exposes hooks
to configure behavior.

- The global variable (https://github.com/golang/go/blob/go1.24.5/src/testing/cover.go#L72)
that tracked all coverage data is completely removed. Instead, functions
in https://pkg.go.dev/internal/coverage/cfile that are called when
writing coverage reports on test exit are internally aware of and able
to access coverage counter variable data.

- Functionality to flush coverage counter variables write to an intermediate
coverage directory in a binary format. Hooks configured in the testdeps package
are responsible for translating this binary format into a human readable coverage
output file.

- 'go tool cover' takes a new package based configuration "pkgcfg" option that
enables this new coverage instrumentation logic in coverage source
files.

- Coverage counters can be flushed on demand via the APIs of "runtime/coverage"
This allows coverage to work outside of test situations where coverage
used to rely on exit hooks generated into test mains to flush and write
coverage data.

What needs to change in rules_go as a result:

- Use the new testdeps configuration system in the rules_go generated test main.
We can following the example of the go test generated test main.
ref: https://github.com/golang/go/blob/go1.24.5/src/cmd/go/internal/load/test.go#L986

- Invoke 'go tool cover' with the new -pkgcfg flag. We can follow the example of
how the go toolchain invokes coverage redesign instrumentation in
https://github.com/golang/go/blob/go1.24.5/src/cmd/go/internal/work/exec.go#L1932

- Not strictly necessary but removes technical debt, call bzltestutil.ConvertCoverToLcov
inside of testdeps.CoverProcessTestDirFunc hook wrapper  instead of the
bzltestutil.LcovTestDeps SetPanicOnExit0 hook.

- Adjust the output of lcov profile to make file path keys in coverage files
execution root relative.

Implementation Note:

Unfortunately, we still need github.com/bazelbuild/rules_go/go/tools/coverdata
dependencies in coverage source files because the lcov coverage format
expects file paths associated with coverage data to be execution root
relative. coverageredesign writes Go coverage file data in the format of
`importpath/file_name` which is not execution root (e.g. `src/`) relative.

During compilation is the only place we have the information to map
`importpath/file_name` to an execution root relative path so we have
to continue to use the trick of generating code that corrects the
file path coverage mapping at runtime.

This makes it hard for coverage to work outside of tests because depend on a special
github.com/bazelbuild/rules_go/go/tools/coverdata dependency that can't be
guaranteed because we don't control a generated test main in these situations.

If Go coverage filepaths can be configured during invocations of the `go
tool cover`, we can break this dependency. I've filed an upstream go ticket
to descrbe this issue https://github.com/golang/go/issues/74749.

ref https://github.com/bazel-contrib/rules_go/issues/3513

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -120,10 +120,9 @@ def _go_test_impl(ctx):
     arguments = go.builder_args(go, "gentestmain", use_path_mapping = True)
     arguments.add("-output", main_go)
     if go.coverage_enabled:
-        if go.mode.race:
-            arguments.add("-cover_mode", "atomic")
-        else:
-            arguments.add("-cover_mode", "set")
+        # Always use atomic mode as the "runtime/coverage" APIs require it
+        # and test behavior should follow non-test behavior.
+        arguments.add("-cover_mode", "atomic")
         arguments.add("-cover_format", go.mode.cover_format)
     arguments.add(
         # the l is the alias for the package under test, the l_test must be the
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"crypto/sha256"
 	"errors"
 	"flag"
 	"fmt"
@@ -243,7 +244,8 @@ func compileArchive(
 	goSrcsNogo := append([]string{}, goSrcs...)
 	cgoSrcsNogo := append([]string{}, cgoSrcs...)
 
-	// Instrument source files for coverage.
+	// Instrument source files in a package for coverage.
+	var coverageCfg string
 	if coverMode != "" {
 		relCoverPath := make(map[string]string)
 		for _, s := range coverSrcs {
@@ -254,6 +256,12 @@ func compileArchive(
 		if cgoEnabled {
 			combined = append(combined, cgoSrcs...)
 		}
+
+		var (
+			coverIn []string
+			coverOut []string
+			srcPathMapping = make(map[string]string)
+		)
 		for i, origSrc := range combined {
 			if _, ok := relCoverPath[origSrc]; !ok {
 				continue
@@ -269,7 +277,16 @@ func compileArchive(
 			case "lcov":
 				// Bazel merges lcov reports across languages and thus assumes
 				// that the source file paths are relative to the exec root.
+				//
+				// In the go's coverageredesign, rules_go no longer is able to
+				// set the filepath key generated to Go's coverage output files
+				// so we keep a mapping from importpath/filename format emitted
+				// by the go runtime in the coverageredesign to the exec root
+				// relative source path required by lcov. We will use this mapping
+				// to write the lcov file with the expected exec root relative source
+				// path.
 				srcName = relCoverPath[origSrc]
+				srcPathMapping[srcName] = path.Join(importPath, filepath.Base(srcName))
 			default:
 				return fmt.Errorf("invalid value for -cover_format: %q", coverFormat)
 			}
@@ -278,12 +295,10 @@ func compileArchive(
 			if ext := filepath.Ext(stem); ext != "" {
 				stem = stem[:len(stem)-len(ext)]
 			}
-			coverVar := fmt.Sprintf("Cover_%s_%d_%s", sanitizePathForIdentifier(importPath), i, sanitizePathForIdentifier(stem))
-			coverVar = strings.ReplaceAll(coverVar, "_", "Z")
 			coverSrc := filepath.Join(workDir, fmt.Sprintf("cover_%d.go", i))
-			if err := instrumentForCoverage(goenv, origSrc, srcName, coverVar, coverMode, coverSrc); err != nil {
-				return err
-			}
+
+			coverIn = append(coverIn, origSrc)
+			coverOut = append(coverOut, coverSrc)
 
 			if i < len(goSrcs) {
 				goSrcs[i] = coverSrc
@@ -292,6 +307,17 @@ func compileArchive(
 
 			cgoSrcs[i-len(goSrcs)] = coverSrc
 		}
+
+		// Modeled after go toolchain's coverage variables configuration.
+		// https://github.com/golang/go/blob/go1.24.5/src/cmd/go/internal/work/exec.go#L1932
+		sum := sha256.Sum256([]byte(importPath))
+		coverVar := fmt.Sprintf("goCover_%x_", sum[:6])
+		coverageCfg = workDir + "pkgcfg.txt"
+		coverOut, err := instrumentForCoverage(goenv, importPath, packageName, coverIn, coverVar, coverMode, coverOut, workDir, relCoverPath, srcPathMapping)
+		if err != nil {
+			return err
+		}
+		goSrcs = append(goSrcs, coverOut[0])
 	}
 
 	// If we have cgo, generate separate C and go files, and compile the
@@ -389,7 +415,7 @@ func compileArchive(
 	}
 
 	// Compile the filtered .go files.
-	if err := compileGo(goenv, goSrcs, packagePath, importcfgPath, embedcfgPath, asmHdrPath, symabisPath, gcFlags, pgoprofile, outLinkObj, outInterfacePath); err != nil {
+	if err := compileGo(goenv, goSrcs, packagePath, importcfgPath, embedcfgPath, asmHdrPath, symabisPath, gcFlags, pgoprofile, outLinkObj, outInterfacePath, coverageCfg); err != nil {
 		return err
 	}
 
@@ -465,6 +491,7 @@ func checkImportsAndBuildCfg(goenv *env, importPath string, srcs archiveSrcs, de
 			return "", errors.New("coverage requested but coverdata dependency not provided")
 		}
 		imports[coverdataPath] = coverdata
+		imports["runtime/coverage"] = nil
 	}
 
 	// Build an importcfg file for the compiler.
@@ -475,7 +502,7 @@ func checkImportsAndBuildCfg(goenv *env, importPath string, srcs archiveSrcs, de
 	return importcfgPath, nil
 }
 
-func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, embedcfgPath, asmHdrPath, symabisPath string, gcFlags []string, pgoprofile, outLinkobjPath, outInterfacePath string) error {
+func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, embedcfgPath, asmHdrPath, symabisPath string, gcFlags []string, pgoprofile, outLinkobjPath, outInterfacePath, coverageCfg string) error {
 	args := goenv.goTool("compile")
 	args = append(args, "-p", packagePath, "-importcfg", importcfgPath, "-pack")
 	if embedcfgPath != "" {
@@ -489,6 +516,9 @@ func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, embedcfgPa
 	}
 	if pgoprofile != "" {
 		args = append(args, "-pgoprofile", pgoprofile)
+	}
+	if coverageCfg != "" {
+		args = append(args, "-coveragecfg", coverageCfg)
 	}
 	args = append(args, gcFlags...)
 	args = append(args, "-o", outInterfacePath)
--- a/go/tools/builders/cover.go
+++ b/go/tools/builders/cover.go
@@ -26,6 +26,8 @@ import (
 	"strconv"
 )
 
+const writeFileMode = 0o666
+
 // instrumentForCoverage runs "go tool cover" on a source file to produce
 // a coverage-instrumented version of the file. It also registers the file
 // with the coverdata package.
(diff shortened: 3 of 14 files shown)


go/tools/builders/generate_test_main.go
```diff
@@ -228,20 +237,12 @@ func main() {
...
-		})
-
-		if coverageDat, ok := os.LookupEnv("COVERAGE_OUTPUT_FILE"); ok {
-			{{if eq .CoverFormat "lcov"}}
-			flag.Lookup("test.coverprofile").Value.Set(coverageDat+".cover")
-			{{else}}
-			flag.Lookup("test.coverprofile").Value.Set(coverageDat)
-			{{end}}
-		}
+	if coverageDat, ok := os.LookupEnv("COVERAGE_OUTPUT_FILE"); ok {
```
fmeum, line 238: The logic above depends on the ugly panic on exit 0 hook. If I remember correctly, the coverage redesign introduced a more principled exit hook approach. Do you happen to know how this works?

I am asking because using the new technique would potentially allow us to Instrument non-test binaries and capture coverage information for them. Is that something you are also interested in?

fmeum (commented):
This looks great, thanks for picking up this long overdue to-do!

jayconrod:
I'll review as soon as I can, but I probably won't have time until Friday, so sorry in advance for the delay!

go/tools/builders/compilepkg.go
```diff
@@ -292,6 +307,18 @@ func compileArchive(
...
 		}
+
+		// Modeled after the go compiler's coverage variables configuration.
+		// https://github.com/golang/go/blob/go1.24.5/src/cmd/go/internal/work/exec.go#L1932
+		sum := sha256.Sum256([]byte(importPath))
+		coverVar := fmt.Sprintf("goCover_%x_", sum[:6])
+		coverageCfg = workDir + "pkgcfg.txt"
+		coverOut, err := instrumentForCoverage(goenv, importPath, packageName, coverIn, coverVar, coverMode, coverOut, workDir, relCoverPath, srcPathMapping)
+
+		goSrcs = append(goSrcs, coverOut[0])
```
JacobOaks: should this go after the err check since it might be nil if `instrumentForCoverage` returns an error?

go/tools/builders/cover.go
```diff
@@ -16,31 +16,121 @@ package main
...
+		return nil, err
+	}
+	data = append(data, '\n')
+	if err := os.WriteFile(pkgcfg, data, 0666); err != nil {
+		return nil, err
+	}
+	var sb strings.Builder
+	for i := range outputFiles {
+		fmt.Fprintf(&sb, "%s\n", outputFiles[i])
+	}
```
JacobOaks, line 72: should we just use `strings.Join(outputFiles, "\n")` instead (do we need the newline at the end)?

go/tools/builders/cover.go
```diff
@@ -16,31 +16,121 @@ package main
...
+		PkgName:   pkgName,
+		Granularity: "perblock",
+		OutConfig: pkgcfg,
+		Local:     false,
+	}
+	data, err := json.Marshal(pcfg)
+	if err != nil {
+		return nil, err
+	}
+	data = append(data, '\n')
```
JacobOaks, line 65: just curious if you happen to know why we need a newline at the end?

go/tools/builders/compilepkg.go
r-hang: fixed!

go/tools/builders/cover.go
r-hang: I suspect it's not strictly needed but I was following what the go toolchain currently does
ref:  https://github.com/golang/go/blob/go1.24.5/src/cmd/go/internal/work/exec.go#L1954

go/tools/builders/generate_test_main.go
r-hang: I've updated the code to drop the dependency on the panic on exit hook. I've also filed https://github.com/golang/go/issues/74749 to see if we can remove all logic to augment coverage source files.

go/tools/builders/cover.go
r-hang: I think we do need the newline at the end since that is what delimits the output files. string.Join should also work but i'm just modeling what the go toolchain currently does.

go/tools/builders/compilepkg.go
```diff
@@ -269,7 +277,16 @@ func compileArchive(
...
+				//
+				// In the go's coverageredesign, rules_go no longer is able to
+				// set the filepath key generated to Go's coverage output files
+				// so we keep a mapping from importpath/filename format emitted
+				// by the go runtime in the coverageredesign to the exec root
+				// relative source path required by lcov. We will use this mapping
+				// to write the lcov file with the expected exec root relative source
+				// path.
 				srcName = relCoverPath[origSrc]
+				srcPathMapping[srcName] = filepath.Join(importPath, filepath.Base(srcName))
```
fmeum:
```suggestion
				srcPathMapping[srcName] = path.Join(importPath, filepath.Base(srcName))
```
since the value is a logical path, not a filesystem path (I think?)

go/tools/builders/cover.go
```diff
@@ -16,31 +16,123 @@ package main
...
+	if err := os.WriteFile(pkgcfg, data, 0666); err != nil {
```
fmeum: Does this permission match the other calls in the builder? We could probably restrict it further so that "others" can't write to the file.

go/tools/builders/cover.go
```diff
@@ -16,31 +16,123 @@ package main
...
+		return nil, err
+	}
+	var sb strings.Builder
```
fmeum, line 69: Could we write to the file directly instead of going through a `Builder`?

go/tools/builders/cover.go
fmeum: Use the helper below or drop it (also get rid of the ioutil usage)

go/tools/builders/generate_test_main.go
```diff
@@ -186,16 +190,25 @@ func main() {
...
+				return bzltestutil.ConvertCoverToLcov()
+			}
+		{{ else }}
+			testdeps.CoverProcessTestDirFunc = cfile.ProcessCoverTestDir
+		{{ end }}
+		testdeps.CoverMarkProfileEmittedFunc = cfile.MarkProfileEmitted
+	{{end}}
+
   {{if .Version "go1.18"}}
-	m := testing.MainStart(testDeps, testsInShard(), benchmarks, fuzzTargets, examples)
```
fmeum, line 196: We can drop the pre-1.18 case now.

go/tools/bzltestutil/lcov.go
```diff
@@ -173,16 +182,10 @@ type LcovTestDeps struct {
...
-}
-
-func lcovAtExitHook() {
-	if err := ConvertCoverToLcov(); err != nil {
-		log.Printf("Failed to collect coverage: %s", err)
-		os.Exit(TestWrapperAbnormalExit)
-	}
-}
+//
+// Note: This method is deprecated and no longer used now that go's coverageredesign
```
fmeum: If it's truly no longer used we can drop it, there are no compatibility guarantees for the internal package.

go/tools/builders/cover.go
r-hang: Updated the code to use to writeFile helper.

go/tools/builders/cover.go
r-hang: Yes we can but looking at the implementation of os.WriteFile, I think it's less error prone to let os.WriteFile handle the call to os.OpenFile and managing closing the file even if this means we need to manage a strings.Builder.

go/tools/builders/cover.go
r-hang: sounds good!

go/tools/builders/generate_test_main.go
r-hang: Could I add this removal in a separate commit?

go/tools/bzltestutil/lcov.go
r-hang: removed!

r-hang:
@fmeum, another commit i'll need to add on top is setting the covermode to atomic if we're building outside of tests as https://pkg.go.dev/runtime/coverage#WriteCounters only works for coverage generated in that mode and that's the only non-test hook that I think we have.

I was thinking of updating the go_context in the go_binary rule to mark that covered packages were non test related and use that info to set the `go tool cover` mode. Do you have another approach in mind?

fmeum:
Do you happen to know how go test gets around this requirement? Do you know how much slower atomic mode is?

r-hang:
@fmeum 
- Generated test main calls cfile.ProcessCoverTestDir which calls emitCounterDataToDirectory underneath.
- runtime/coverage does calls the same emitCounterDataToDirectory underneath however there is an explicit check against non-atomic coverage mode https://github.com/golang/go/blob/6fbad4be75e7746512bbe55794694ed788ea5c5b/src/internal/coverage/cfile/apis.go#L39 

ref:
https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/runtime/coverage/coverage.go;l=56
https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/internal/coverage/cfile/testsupport.go;l=30

I'm not sure of the performance differences but it seems like Go doesn't want this to happen for correctness concerns since non-test programs are likely to run concurrently.

Their concern is documented in the ClearCounters API.
https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/runtime/coverage/coverage.go;l=64

Update: I asked Gemini to generate me sample benchmark code that I built with atomic vs set covermode and the result doesn't seem widely different, maybe slightly (< 10%) more expensive for atomic but this is not very rigorous.

That said, existing uses of the covdata struct are probably unsafe as well and any user of non-test integration test coverage should assume their code will not be as performant.

fmeum:
I would say we go with atomic everywhere. Coverage runs are slow anyway and having them return accurate results is far more important than time savings. Different behavior for tests vs binaries would also be unexpected.

r-hang:
@fmeum, sounds good. I just pushed a commit to this branch that enables atomic mode for all coverage cases.

go/private/rules/test.bzl
```diff
@@ -123,7 +123,7 @@ def _go_test_impl(ctx):
         if go.mode.race:
             arguments.add("-cover_mode", "atomic")
         else:
-            arguments.add("-cover_mode", "set")
+            arguments.add("-cover_mode", "atomic")
```
fmeum: Could you fold away the if now that the branches are identical?

go/tools/builders/cover.go
```diff
@@ -16,31 +16,122 @@ package main
...
+		// test main exits and go coverage files are converted to lcov format.
+		if err := registerCoverage(outfile, importPathFile, srcName); err != nil {
+			return nil, err
+		}
 	}
+	return outputFiles, nil
+}
+
+func writeFile(path string, data []byte) error {
+	return os.WriteFile(path, data, 0o666)
```
fmeum: Let's just inline this and (if you want) extract the mode into a constant.

fmeum (approved):
Thanks a lot for figuring this out, I only left a few minor comments suggesting final cleanups before we can merge this.

r-hang:
@fmeum, @linzhp would it be possible to merge and release this by next week?

fmeum (approved):

fmeum:
@r-hang Could you remind me what's left for us to do so that go_binary can report coverage in non-Go test rules?

r-hang:
> @r-hang Could you remind me what's left for us to do so that go_binary can report coverage in non-Go test rules?

This should work already. Here's an example https://github.com/bazel-contrib/rules_go/issues/3513#issuecomment-3156229955

In non-test test code there isn't a test main to always hook into so I declared the coverage data emission manually.

PR #4401, opened 2025-07-29
Support go_path rule for targets that have a dir in embedsrcs

### LINKED ISSUE #4400: go_path rule does not support go_library using embedsrcs containing a directory
If you define a `go_library` that uses `embedsrcs` referring to a directory (i.e., by either containing the name of a local directory, or by referring to a label of a rule that creates a directory, like [`copy_to_directory`](https://github.com/bazel-contrib/bazel-lib/blob/main/docs/copy_to_directory.md)), then a `go_path` rule (transitively) referring to that library fails to build with an error from the `go_path.go` binary complaining about not being able to copy a directory.
This is because the `go_path.go` binary wrongly assumes that all paths in the manifest are files, which is not the case when specifying a directory like described above.

I have a PR ready with a proposed fix, but I am not a Bazel expert, so I am unsure whether my fix is correct.

cmfcmf:
**What type of PR is this?**

> Bug fix

This PR fixes `go_path` failing when a `go_library` rule has a directory in its `embedsrcs`. It does so by correctly copying folders (as well as symlinked folders).

Fixes #4400

**Other notes for review**

### DIFF
--- a/go/tools/builders/go_path.go
+++ b/go/tools/builders/go_path.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


go/tools/builders/go_path.go
```diff
@@ -112,6 +113,53 @@ func readManifest(path string) ([]manifestEntry, error) {
...
+func copySymlinkRecursively(src, dst string, writeFile func(src, dst string) error, createDir func(dst string) error) error {
+	target, err := os.Readlink(src)
+	if err != nil {
+		return err
+	}
+	if !filepath.IsAbs(target) {
+		target = filepath.Join(filepath.Dir(src), target)
+	}
+
+	return copyRecursively(target, dst, writeFile, createDir)
```
fmeum, line 125: Could this lead out us out of the sandbox? Would it work if we only copied the symlink as is without recurring on its target?

cmfcmf: > Could this lead out us out of the sandbox? 

I don't know enough about Bazel to answer this.

> Would it work if we only copied the symlink as is without recurring on its target?

I think that would only work for the `copyPath` function, but I don't think that would work for the ZIP archive created by `archivePath`, because the ZIP file cannot contain symlinks.

Issue #4327, opened 2025-04-22
nogo: improve cache friendliness


jayconrod:
### Context

See #4326. I'm collecting out-of-band discussion with @linzhp, @tyler-french, and @fmeum.

In short, Uber is using Nogo at scale, and a recent correctness fix has caused Nogo to be much slower. They have about 150 analyzers (a mix of the standard vet analyzers, staticcheck, and some custom packages). They also have a large `nogo.json` file that changes frequently.

We'd prefer not to revert the correctness fix (#4268), but we do want to find ways to ensure Nogo actions hit the cache more frequently.

### Background

- A Nogo analyzer is an implementation of [`analysis.Pass`](https://pkg.go.dev/golang.org/x/tools/go/analysis#Pass).
- All the analyzers for a repo get compiled into the same binary. This means changing the set of analyzers necessarily invalidates the cache.
- Nogo actions are run separately from Go compilation. Currently, a single action runs all analyzers on a single package.
- A nogo analyzer can produce facts, which are data about a package that can be consumed by analyzers run on importing package. For example, the `printf` analyzer records which functions defined in a package accept printf format strings.
- We currently collect facts from all analyzers into a single output file per package.
- Each Nogo action takes facts files from all transitively imported packages as inputs. This probably isn't necessary: we should take facts only from directly imported packages.
- `nogo.json` determines files that are included and excluded from analysis. This mainly records whether we report findings. We generally need to run Nogo on all transitively imported packages to generate facts, whether or not they're excluded.

### Changes proposed

These are all suggestions, not final recommendations. Each needs some investigation.

- Hoist `nogo.json` into Starlark, from execution to analysis. Instead of passing it as an input file to each action, we can pass appropriate arguments to the nogo binary. This should reduce cache invalidation when the configuration changes. If we change the set of excluded files in package A, that shouldn't invalidate results in package B.
- Only pass facts files for direct imports as inputs; don't include facts for transitive imports. This should reduce cache invalidation when an irrelevant transitively imported package changes. For example, say package A imports B, and B imports `fmt`, but A does not. A does not need facts from `fmt`. If B defines a Printf wrapper, that will be reflected in B's facts. If B does not define a Printf wrapper, A doesn't care what happens in `fmt`.
- Link analyzers into separate binaries; run analyzers as separate actions; write facts to separate files. This may reduce cache invalidation when the set of analyzers changes, or when facts change in a way that transitively affects a large subset of packages.
[...]

PR #4403, opened 2025-07-29
Allow targets to fully opt out of nogo

fmeum:
**What type of PR is this?**

Feature

By tagging a target with `"no-nogo"`, `nogo` isn't run at all, not even for fact generation.

Work towards #4327

**Other notes for review**


tests/core/nogo/custom/custom_test.go
```diff
@@ -145,6 +145,13 @@ go_library(
 	importpath = "panics",
 )
 
+go_library(
+    name = "panics_no_nogo",
+    srcs = ["panics.go"],
+	importpath = "panics_no_nogo",
```
jayconrod: Replace tab with spaces

jayconrod (approved):

PR #4405, opened 2025-07-30
fix the internal archive to include all of the original dependencies

jgautier-dd:
**What type of PR is this?**
Bug fix

I had an issue where the go packages driver was not loading all of the dependencies in a test resulting in unresolved imports. I tracked it down to the process of recompiling for internal/external in `go_test`. For the internal archive it is filtering out the dependencies that need to be recompiled, and recompiling them, but the final `internal_archive` does not include the recompiled dependencies. Later when the packages driver aspect lists the dependencies for the `go_test` target it is missing the dependencies which were recompiled.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -655,6 +655,7 @@ def _recompile_external_deps(go, external_go_info, internal_archive, library_lab
     # Pass internal dependencies that need to be recompiled down to the builder to check if the internal archive
     # tries to import any of the dependencies. If there is, that means that there is a dependency cycle.
     need_recompile_deps = []
+    original_internal_deps = internal_go_info.deps
     for archive in internal_go_info.deps:
         dep_data = archive.data
         if not need_recompile[dep_data.label]:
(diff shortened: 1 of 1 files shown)


jayconrod (commented):
Sorry to take a while to review. It's been a chaotic week. This code is also probably the most complicated logic in rules_go, so it takes a while to read back and understand.

1. Could you please ensure the tests cover this and reproduce the problem? Maybe add something to `gopackagesdriver_test.go`. That file creates a test workspace, then has test cases that run `gopackagesdriver` within that workspace in a few different ways.
2. For the actual code in `test.bzl`, I don't like that `internal_archive` is defined more than once, especially in a loop. `internal_archive` is also used a few lines below, and it's hard to reason about whether that happens before or after.

    Can this be fixed by changing some input to the `go.archive` call on L672? I think `attr["deps"]` is what eventually ends up as `internal_archive.direct`. So maybe `internal_deps` should include the recompiled version if available or the original version of each dependency? Currently it only includes the original versions.

jayconrod:
Superseded by #4606

PR #4414, opened 2025-08-02
fix: instrumentForCoverage for empty outfiles

bakjos:
**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #
https://github.com/bazel-contrib/rules_go/issues/4412

**Other notes for review**

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -258,8 +258,8 @@ func compileArchive(
 		}
 
 		var (
-			coverIn []string
-			coverOut []string
+			coverIn        []string
+			coverOut       []string
 			srcPathMapping = make(map[string]string)
 		)
 		for i, origSrc := range combined {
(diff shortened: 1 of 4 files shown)


go/tools/builders/compilepkg.go
```diff
@@ -312,12 +312,14 @@ func compileArchive(
...
-		coverOut, err := instrumentForCoverage(goenv, importPath, packageName, coverIn, coverVar, coverMode, coverOut, workDir, relCoverPath, srcPathMapping)
-		if err != nil {
-			return err
+		if len(coverOut) > 0 {
+			coverageCfg = workDir + "pkgcfg.txt"
+			coverOut, err := instrumentForCoverage(goenv, importPath, packageName, coverIn, coverVar, coverMode, coverOut, workDir, relCoverPath, srcPathMapping)
+			if err != nil {
+				return err
+			}
+			goSrcs = append(goSrcs, coverOut[0])
```
fmeum, line 321: @r-hang I know I reviewed this, but could you remind me why we don't merge in everything in `coverOut`?

bakjos:
@fmeum I have added the test, if you run it without my change, it will fail with the same error, it happens when `--instrument_test_targets` is used

fmeum:
Thanks!

go/tools/builders/compilepkg.go
r-hang: Because of https://github.com/bazel-contrib/rules_go/pull/4414/files/6942249454edec6f14e0003c5af04a65ab13ee7e#diff-83a1d12c8bed64ad8572a9eee4ea3e14d1dde0f3ddbd6b87299b33b444e773f4R304

goSrcs already contains all of the coverageSrc files with the exception of the cover variables file which is the first result in the list.

I think we could just set goSrcs to the entire result of coverOut but I didn't do that since goSrcs already contains almost all of the coverOut slice.

tests/core/coverage/issue4414_test.go
```diff
@@ -0,0 +1,61 @@
...
+go_test(
+    name = "issue_test",
+    srcs = ["issue_test.go"],
+    embed = [":issue"],
+)
+`,
+	})
+}
+
+func TestIssue4414(t *testing.T) {
```
fmeum, line 61: This test expectedly fails on Windows, could you exclude it in the `.bazelci` config?

PR #4416, opened 2025-08-05
Add host_compatible flag to extension

patrickmscott:
This controls whether the tag is considered for the host compatible toolchain. This is necessary if any of the tags point to a toolchain that should not be used to build host tools.

For example, we use a compiler that _requires_ cgo. For the code in the repository this works because cgo is enabled and we use a hermetic clang. But some tools like `builder` and `gazelle` either disable cgo or do not pass in the compiler.

**What type of PR is this?**

Feature

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -59,6 +59,7 @@ _download_tag = tag_class(
     doc = """Download a specific Go SDK at the optional GOOS, GOARCH, and version, from a customisable URL.  Optionally apply local customisations to the SDK by applying patches and setting experiments.""",
     attrs = _COMMON_TAG_ATTRS | {
         "version": attr.string(),
+        "host_compatible": attr.bool(default = True),
     },
 )
 
@@ -114,6 +115,7 @@ _wrap_tag = tag_class(
         ),
         "goos": attr.string(),
         "goarch": attr.string(),
+        "host_compatible": attr.bool(default = True),
     },
 )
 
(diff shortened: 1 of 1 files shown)


fmeum:
The "host compatible" toolchain should only be used to build Gazelle, none of the rules_go tools. Instead of adding a flag, do you think we could fix the underlying issue? Could you share the error you get? Maybe we can get rid of whatever is special about the Gazelle binary so that it does build with your hermetic toolchain SDK.

patrickmscott:
I'm digging in a bit more on my build. This change did not fix it even though the host-compatible toolchain was correctly set to an upstream go version.

The problem I see is that my wrapped toolchain is being used to build gazelle. Gazelle explicitly disables cgo but my compiler requires it. Let me switch this to a draft while I continue to debug.

patrickmscott:
https://github.com/bazel-contrib/rules_go/pull/4397

This seems to be the change that breaks the rules_go builder for me. I'm still trying to understand why.

fmeum:
@r-hang

patrickmscott:
Ugh, it's our toolchain being ridiculous.

The new coverage code pulls in `crypto` which our toolchain has patches to use boringcrypto - and it's super aggressive about it. I guess we have been getting lucky so far to not hit this code path yet.

patrickmscott:
There is still an issue with host vs target toolchains. I would like to use generic go for all host tools (gazelle, builder, etc) since those do not require our boringcrypto patches. It would be nice to be able to specify which toolchain to use for the execution platform.

patrickmscott:
I _think_ the alternative patch would be for both rules_go and gazelle to allow for arbitrary env variables somehow to get deep into go_tool_binary (rules_go) and go_repository_tools (gazelle).

patrickmscott:
I am going to patch these locally until I can think about a better solution. The toolchain mess is likely not something other companies are having to deal with.

PR #4417, opened 2025-08-05
switch root_files from string_dict to string_keyed_label_dict

novas0x2a:
**What type of PR is this?**

Bug fix

I'm trying to use an extension-generated repo as a wrapped sdk using `root_files`. I get a package load error similar to https://github.com/bazelbuild/bazel/issues/19301. 

I saw in the comments (specifically https://github.com/bazelbuild/bazel/issues/19301#issuecomment-1694588371) that this would probably work if `root_files` used a `string_keyed_label_dict` and, when I tried it locally it did.

**Other notes for review**

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -104,7 +104,7 @@ _wrap_tag = tag_class(
             mandatory = False,
             doc = "A file in the SDK root directory. Use to determine GOROOT.",
         ),
-        "root_files": attr.string_dict(
+        "root_files": attr.string_keyed_label_dict(
             mandatory = False,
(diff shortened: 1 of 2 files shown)


novas0x2a:
Oh. Unfortunately, `string_keyed_label_dict` was only added in bazel 8, and rules_go has bazel 7 compat.

novas0x2a:
Since the approach to fixing this got more complicated, I replaced this with an issue: https://github.com/bazel-contrib/rules_go/issues/4418

novas0x2a:
(i force-pushed my branch before reopened the MR, which broke this MR; work is now in #4419)

PR #4420, opened 2025-08-05
[go1.25 required] Install cmd/pack command if it's missing from pkg/tool

r-hang:
In go1.25 the "cmd/pack" command is no longer shipped with Go distributions, which results is build failures that look like
```
2025-06-09 00:44:46 PDT
compilepkg: error running subcommand external/go_sdk/pkg/tool/linux_amd64/pack: fork/exec external/go_sdk/pkg/tool/linux_amd64/pack: no such file or directory
2025-06-09 00:44:46 PDT
ERROR: /home/testuser/.cache/bazel-single-base/external/io_nhooyr_websocket/BUILD.bazel:3:11: GoCompilePkg external/io_nhooyr_websocket/websocket.a failed: (Exit 1): builder failed: error executing GoCompilePkg command (from target @@io_nhooyr_websocket//:websocket) bazel-out/k8-opt-exec-ST-a828a81199fe/bin/external/go_sdk/builder_reset/builder compilepkg -sdk external/go_sdk -goroot bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_ -installsuffix ... (remaining 63 arguments skipped)
```

Since the pack command is used to compile Go programs (in compilepkg.go) we need to build it from source.

ref: https://github.com/golang/go/issues/74080
fixes https://github.com/bazel-contrib/rules_go/issues/4398

### DIFF
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -200,6 +200,7 @@ def emit_compilepkg(
     go.actions.run(
         inputs = depset(inputs_direct, transitive = inputs_transitive),
         outputs = outputs,
+        tools = [go.toolchain.sdk.go],
         mnemonic = "GoCompilePkgExternal" if is_external_pkg else "GoCompilePkg",
         executable = go.toolchain._builder,
         arguments = ["compilepkg", shared_args, compile_args],
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -533,7 +533,7 @@ func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, embedcfgPa
 
 func appendToArchive(goenv *env, outPath string, objFiles []string) error {
 	// Use abs to work around long path issues on Windows.
-	args := goenv.goTool("pack", "r", abs(outPath))
+	args := goenv.goCmd("tool", "pack", "r", abs(outPath))
 	args = append(args, objFiles...)
 	return goenv.runCommand(args)
 }
--- a/go/tools/builders/stdlib.go
+++ b/go/tools/builders/stdlib.go
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"fmt"
 	"go/build"
+	"runtime"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -169,6 +170,22 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
 	installArgs = append(installArgs, packages...)
 	if err := goenv.runCommand(installArgs); err != nil {
 		return err
+	}
+
+	// Install the "cmd/pack" command from source if it's missing as it's no longer shipped with Go distributions
+	// after go1.25 https://github.com/golang/go/issues/74080
+	if _, err := os.Stat(filepath.Join(output, "pkg/tool", runtime.GOOS + "_" + runtime.GOARCH, "pack")); err != nil && os.IsNotExist(err) {
+		originalOS, originalARCH := os.Getenv("GOOS"), os.Getenv("GOARCH")
+		os.Setenv("GOOS", runtime.GOOS)
+		os.Setenv("GOARCH", runtime.GOARCH)
+		defer func() {
+			os.Setenv("GOOS", originalOS)
+			os.Setenv("GOARCH", originalARCH)
+		}()
+		toolArgs := goenv.goCmd("install", "-toolexec", abs(os.Args[0])+" filterbuildid", "cmd/pack")
+		if err := goenv.runCommand(toolArgs); err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/go/private/BUILD.sdk.bazel
+++ b/go/private/BUILD.sdk.bazel
@@ -28,7 +28,6 @@ filegroup(
         exclude = [
             "src/**/*_test.go",
             "src/**/testdata/**",
-            "src/cmd/**",
             # Only used by tests, cgo fails with linux before 3.17
             "src/crypto/internal/sysrand/internal/seccomp/**",
             "src/log/slog/internal/benchmarks/**",


go/tools/builders/stdlib.go
```diff
@@ -170,5 +171,21 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
 	if err := goenv.runCommand(installArgs); err != nil {
 		return err
 	}
+
+	// Install the "cmd/pack" command from source if it's missing as it's no longer shipped with Go distributions
+	// after go1.25 https://github.com/golang/go/issues/74080
+	if _, err := os.Stat(filepath.Join(output, "pkg/tool", runtime.GOOS + "_" + runtime.GOARCH, "pack")); err != nil && os.IsNotExist(err) {
```
fmeum, line 177: Let's fail if `err != nil && !os.IsNotExist(err)` rather than just skipping this logic.

go/private/actions/compilepkg.bzl
```diff
@@ -200,6 +200,7 @@ def emit_compilepkg(
     go.actions.run(
         inputs = depset(inputs_direct, transitive = inputs_transitive),
         outputs = outputs,
+        tools = [go.toolchain.sdk.go],
```
fmeum, line 203: Is this necessary? I would have expected this file to be in `sdk.tools`, which is already added to `inputs`. If not, I wonder how this worked before this change.

go/tools/builders/stdlib.go
```diff
@@ -170,5 +171,21 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
+		originalOS, originalARCH := os.Getenv("GOOS"), os.Getenv("GOARCH")
+		os.Setenv("GOOS", runtime.GOOS)
+		os.Setenv("GOARCH", runtime.GOARCH)
+		defer func() {
+			os.Setenv("GOOS", originalOS)
+			os.Setenv("GOARCH", originalARCH)
+		}()
+		toolArgs := goenv.goCmd("install", "-toolexec", abs(os.Args[0])+" filterbuildid", "cmd/pack")
```
fmeum, line 185: This may not work correctly in general: We build `pkg/tool` as part of the `stdlib` rule, but the execution platform of that rule may not match the execution platform of the Go rules that need run `pack`. 

Instead, I think that we need to duplicate (or extract) most of the `stdlib` rule (no need to split out `actions`) and make this new rule a `cfg = "exec"` dependency of each Go rule.

go/tools/builders/compilepkg.go
```diff
@@ -533,7 +533,7 @@ func compileGo(goenv *env, srcs []string, packagePath, importcfgPath, embedcfgPa
 
 func appendToArchive(goenv *env, outPath string, objFiles []string) error {
 	// Use abs to work around long path issues on Windows.
-	args := goenv.goTool("pack", "r", abs(outPath))
+	args := goenv.goCmd("tool", "pack", "r", abs(outPath))
```
sluongng, line 536: Make me wonder if we need to replace `goenv.goTool` entirely with `goenv.goCmdTool` in future versions. Because if we do, then we might as well make this change in `goTool` method for every tool.

matloob: So, at least for now, invoking a the tool directly will save you an extra process execution if you're using one of the core tools we still ship with the go distribution (compile, link, asm, etc.) which is probably worth it.

I don't think we can rule out the possibility that we may at some point require the compiler to be run with 'go tool compile' but it's not something we're anticipating.

go/tools/builders/stdlib.go
matloob: ah, good point.

r-hang: Yes, the environment variable set up and tear down was put there as I observed build failures in cases where executions platform of the rule changed. This change allowed us to build all targets but i'll take a look at the `cfg = exec` solution.

go/private/actions/compilepkg.bzl
r-hang: Yes removing that line yields
```
INFO: Analyzed 41 targets (2144 packages loaded, 24697 targets configured).
ERROR: /home/user/.cache/bazel/_bazel_rhang/b97476d719d716accead0f2d5b93104f/external/com_github_twmb_murmur3/BUILD.bazel:3:11: GoCompilePkg external/com_github_twmb_murmur3/murmur3.a failed: (Exit 1): builder failed: error executing GoCompilePkg command (from target @@com_github_twmb_murmur3//:murmur3) bazel-out/k8-opt-exec-ST-a828a81199fe/bin/external/go_sdk/builder_reset/builder compilepkg -sdk external/go_sdk -goroot bazel-out/k8-fastbuild/bin/external/io_bazel_rules_go/stdlib_ -installsuffix ... (remaining 35 arguments skipped)

compilepkg: error running subcommand external/go_sdk/bin/go: fork/exec external/go_sdk/bin/go: no such file or directory
```

Looking at compilepkg.go there were only references to goenv.goTool and not goenv.goCmd.

In the builders, such as stdlib, that do use goenv.goCmd
```
go/tools/builders/stdlib.go
125:    installArgs := goenv.goCmd("install", "-toolexec", abs(os.Args[0])+" filterbuildid")
178:            toolArgs := goenv.goCmd("install", "-toolexec", abs(os.Args[0])+" filterbuildid", "cmd/pack")
```

I see that stdlib.bzl has
```
    inputs_direct = [sdk.go]
    inputs_transitive = [sdk.headers, sdk.srcs, sdk.libs, sdk.tools]
    if not go.mode.pure:
        inputs_transitive.append(go.cc_toolchain_files)

    go.actions.run(
        inputs = depset(inputs_direct, transitive = inputs_transitive),
```

go/tools/builders/stdlib.go
r-hang: @fmeum, I tried pursuing the cfg = "exec" approach but I don't have enough Bazel experience to make good progress on this. Would you be able to sketch out a solution in more detail here? I can try again with more guidance.

I my approached centered around
```
diff --git a/go/private/rules/nogo.bzl b/go/private/rules/nogo.bzl
index 8fc736ed..0ed0dea1 100644
--- a/go/private/rules/nogo.bzl
+++ b/go/private/rules/nogo.bzl
@@ -104,7 +104,10 @@ _nogo = rule(
         "_cgo_context_data": attr.label(default = "//:cgo_context_data_proxy"),
         "_go_config": attr.label(default = "//:go_config"),
         "_go_difflib": attr.label(default = "@com_github_pmezard_go_difflib//difflib:go_default_library"),
-        "_stdlib": attr.label(default = "//:stdlib"),
+        "_stdlib": attr.label(
+            default = "//:stdlib",
+            cfg = "exec",
+        ),
         "_allowlist_function_transition": attr.label(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
```

but i'm not sure why forking the stdlib rule is needed.

fmeum: @matloob I'm looking into this now and it would be very convenient if there was a way to control where `go install` emits the binaries it produces. According to its docs, either `GOBIN` or `GOTOOLDIR` should do that, but I can't get that to work:

```
GOROOT=<path to GOROOT> GOTOOLDIR=/Users/fmeum/tmp/gotool GOBIN=/Users/fmeum/tmp/gotool go install cmd/pack
```
results in `~/tmp/gotool` still being empty.

matloob: @fmeum The internal build tools are a special case: the go command always wants them to be placed in `$GOROOT/pkg/tool/$GOOS_$GOARCH`. I apologize that we've poorly documented the GOTOOLDIR environment variable.. It's not settable, it just exists to report the location the tools are installed. We should document that. And GOBIN doesn't apply to tools.

But If I understand correctly, I think `go build` is better suited to this case. If you want to create an action that produces the pack binary built in the host configuration, could you run `go build -o <action output location> cmd/pack`, Setting the GOOS/GOARCH/etc. for the host configuration? The action wouldn't need to have the standard library `.a`'s pre compiled for the host configuration. As long as it has a "full" host config distribution with the sources and tools we ship it with it will build.

fmeum: Thanks, that works. I'll send out an updated PR soon, probably tomorrow.

fmeum: https://github.com/bazel-contrib/rules_go/pull/4425

r-hang:
Closing in favor of https://github.com/bazel-contrib/rules_go/pull/4425

PR #4402, opened 2025-07-29
Skip analyzers that don't emit facts when ignoring diagnostics

fmeum:
**What type of PR is this?**

Performance improvement

**What does this PR do? Why is it needed?**

Avoid unnecessary work by skipping analyzers that don't emit facts when reported diagnostics are known to be ignored.

**Which issues(s) does this PR fix?**

Work towards #4327

**Other notes for review**

PR #4422, opened 2025-08-11
Fix generation of facts file when diagnostics are ignored

lbcjbb:
**What type of PR is this?**

Bug fix

After #4402, it is no longer possible to use a custom analyzer without having to declare a new `Fact` type. The build of a Go library will fail if the library contains any external dependencies.

That's the case for executing gazelle by example (if we don't use a precompiled binary):

```
$ bazel run //:gazelle
ERROR: .../external/gazelle++go_deps+org_golang_x_sync/errgroup/BUILD.bazel:3:11: output 'external/gazelle++go_deps+org_golang_x_sync/errgroup/errgroup.facts' was not created
ERROR: .../external/gazelle++go_deps+org_golang_x_sync/errgroup/BUILD.bazel:3:11: Running nogo on @@gazelle++go_deps+org_golang_x_sync//errgroup:errgroup failed: not all outputs were created or valid
ERROR: .../external/gazelle++go_deps+com_github_bazelbuild_buildtools/labels/BUILD.bazel:3:11: output 'external/gazelle++go_deps+com_github_bazelbuild_buildtools/labels/labels.facts' was not created
ERROR: .../external/gazelle++go_deps+com_github_bazelbuild_buildtools/labels/BUILD.bazel:3:11: Running nogo on @@gazelle++go_deps+com_github_bazelbuild_buildtools//labels:labels failed: not all outputs were created or valid
ERROR: .../external/gazelle++go_deps+in_gopkg_yaml_v3/BUILD.bazel:5:11: output 'external/gazelle++go_deps+in_gopkg_yaml_v3/yaml_v3.facts' was not created
ERROR: .../external/gazelle++go_deps+in_gopkg_yaml_v3/BUILD.bazel:5:11: Running nogo on @@gazelle++go_deps+in_gopkg_yaml_v3//:yaml_v3 failed: not all outputs were created or valid
ERROR: .../external/gazelle++go_deps+org_golang_x_sys/execabs/BUILD.bazel:3:11: output 'external/gazelle++go_deps+org_golang_x_sys/execabs/execabs.facts' was not created
ERROR: .../external/gazelle++go_deps+org_golang_x_sys/execabs/BUILD.bazel:3:11: Running nogo on @@gazelle++go_deps+org_golang_x_sys//execabs:execabs failed: not all outputs were created or valid
...
```

**Other notes for review**

This review continue to skip analysers when ignoring diagnostics, but return a valid `pkg` to create the missing `.facts` file required by the validation action, and incidentally, fix a potential crash at line https://github.com/bazel-contrib/rules_go/blob/1f993522f463c142c5516f94852c3d6597f717c6/go/tools/builders/nogo_main.go#L98.

### DIFF
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -93,12 +93,26 @@ func run(args []string) (error, int) {
 	if err != nil {
 		return fmt.Errorf("error running analyzers: %v", err), nogoError
 	}
+
 	// Write the facts file for downstream consumers before failing due to diagnostics.
 	if *xPath != "" {
-		if err := os.WriteFile(abs(*xPath), pkg.facts.Encode(), 0o666); err != nil {
+		var factsContent []byte
+		if pkg != nil {
+			factsContent = pkg.facts.Encode()
+		}
+
+		if err := os.WriteFile(abs(*xPath), factsContent, 0o666); err != nil {
 			return fmt.Errorf("error writing facts: %v", err), nogoError
 		}
 	}
+
+	var fset *token.FileSet
+	if pkg != nil {
+		fset = pkg.fset
+	} else if len(diagnostics) > 0 {
+		return fmt.Errorf("pkg should not be nil with diagnostics"), nogoError
+	}
+
 	exitCode := nogoSuccess
 	var errMsg bytes.Buffer
 	if len(diagnostics) > 0 {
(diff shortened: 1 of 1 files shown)


go/tools/builders/nogo_main.go
```diff
@@ -275,6 +271,11 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath string, packageFil
 		return nil, nil, fmt.Errorf("error loading package: %v", err)
 	}
 
+	if len(roots) == 0 {
```
fmeum: It seems unnecessary to do the work above. Could we change nogo_main to write out an empty file if there are no facts? Could you share the crash stack trace?

lbcjbb: The stack trace is not visible, the nogo command seems to fail silently.

By adding a defer/recover to write the stack trace in a file, I can see the `runtime error: invalid memory address or nil pointer dereference` on the line `if err := os.WriteFile(abs(*xPath), pkg.facts.Encode(), 0o666); err != nil {` in the `nogo_main.go` file.

lbcjbb: > It seems unnecessary to do the work above. Could we change nogo_main to write out an empty file if there are no facts?

I updated my branch with this change.

fmeum (approved):

Issue #3513, opened 2023-04-02
Add support for "Coverage profiling support for integration tests"


sfc-gh-kleonhard:
In 1.20, Go introduced support for integration test profiling: https://go.dev/testing/coverage/#gomodselect. Can support for this be added to rules_go so that coverage is collected from Go binaries executed during `bazel coverage`?

My use case: I've got go_test rules that depend on go_binary artifacts via the `data` attribute. The tests run the binaries using exec.Command. I'd like to aggregate test coverage from my code included in the binary under test.

Can I tell bazel coverage to set the -cover flag on any go_binary rules invoked and have it aggregate coverage?

Here's an example I'd like to work with: `bazel coverage --combined_report=lcov //my_test`

```
# running coverage tests should include coverage
# from the invocation of my_binary_a from my_test
go_library(
    name = "my_binary_under_test_lib",
    srcs = ["my_binary_under_test.go"],
    importpath = "github.com/example/my_binary_under_test",
)
go_binary(
    name = "my_binary_under_test",
    embed = [":my_binary_under_test_lib"],
)

go_test(
    name = "my_test",
    srcs = [
       # Runs my_binary via exec.Command
        "my_test.go",
    ],
    # These binaries are executed by my_test.go
    data = [
        "//my_binary_under_test",
    ],
)
```

I'd expect the resulting coverage to include coverage for `my_binary_under_test.go` as it's used in my_binary_under_test which is invoked by my_test.

fmeum:
This is definitely something I would like to see, but it requires some work: Since we interface with Go on a lower level than `go build`, we need to figure out how coverage instrumentation is enabled on that level. We also need to call our cover to lcov conversion routine in the right places.

If anyone wants to work on this, I can offer my help. I am also in touch with the Go team and hope that they can provide some pointers.

sfc-gh-kleonhard:
Thanks for the quick response, Fabian! I hope some ambitious gopher takes you up on your offer.

ceejatec:
I'm very interested in this functionality as well. I'd like to add that for our purposes, we would like the ability to perform normal compilation steps with the equivalent of "go build -cover -coverpkg /...." available - we don't want this integrated with existing "bazel coverage" runs, because the whole point of this new feature is to allow running integration tests while gathering coverage information. Integration testing would not normally be driven by bazel. Hopefully that is something that will come out of this effort as well. Thanks!

sluongng:
[...]

PR #4424, opened 2025-08-13
coverageredesign: Map line directive file names to exec relative paths

r-hang:
coverageredesign (https://github.com/bazel-contrib/rules_go/pull/4397) honors line directives while the world before coverageredesign did not.

This is relevant in situations where you want coverage for generated code to map to source code in a user's workspace.

An example of this is a go_library rule which takes a dependency on a code generator which generates a replacement file for a user source file. The replacement file is run in the final binary but the user wants to see covarege for the source file that was an input to the generated replacement file.

```
go_library(
    name = "go_default_library",
    srcs = [
        "batch.go",
        "consumers.go",
        "edges.go",
        "token.go",
        ":codegen",
    ],
)

codegen(name = "codegen, ...)
```

The Go compiler, because of Bazel, only knows about the generated code but coverage is only relevant for the source file that the code generator replaces.

Given a generated file
`bazel-out/k8-fastbuild/bin/src/example.org/gen_/report_gen.go` nocoverageredesign would emit that file path in coverage files.

With coverageredesign, if a line directive in the generated code changes the file name to hello.go (e.g. //line hello.go:10) then the path in the runtime emitted coverage file would be `example.org/gen_/hello.go` which is not execution root relative and would be ignored by Bazel.

The prevent this code from breaking in coverage redesign, we need to map the line directive file name to the execution root relative filepath that Bazel respects to get coverage for these types of files.

This change extends the "srcPathMapping" trick used in https://github.com/bazel-contrib/rules_go/pull/4397/files to honor a mapping of line directive file names to map back to exec root relative paths that Bazel understands and requires for lcov mode.

ref #3513

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -98,13 +98,20 @@ Uses the same format as 'visibility', i.e., every entry must be a label that end
     },
 )
 
+# string_keyed_label_dict was added in 8.0.0
+_maybe_string_keyed_label_dict = getattr(
+    attr,
+    "string_keyed_label_dict",
+    attr.string_dict,
+)
+
 _wrap_tag = tag_class(
     attrs = {
         "root_file": attr.label(
             mandatory = False,
             doc = "A file in the SDK root directory. Use to determine GOROOT.",
         ),
-        "root_files": attr.string_dict(
+        "root_files": _maybe_string_keyed_label_dict(
             mandatory = False,
             doc = "A set of mappings from the host platform to a file in the SDK's root directory.",
         ),
@@ -234,7 +241,7 @@ def _go_sdk_impl(ctx):
                 if key not in ["go_mod"]
             }
             download_tag["version"] = version
-            additional_download_tags += [struct(**download_tag)]
+            additional_download_tags.append(struct(**download_tag))
 
         for index, download_tag in enumerate(module.tags.download + additional_download_tags):
             # SDKs without an explicit version are fetched even when not selected by toolchain
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -371,6 +371,12 @@ def _go_wrap_sdk_impl(ctx):
     _sdk_build_file(ctx, platform, version, ctx.attr.experiments)
     _local_sdk(ctx, goroot)
 
+# string_keyed_label_dict was added in 8.0.0
+_maybe_string_keyed_label_dict = getattr(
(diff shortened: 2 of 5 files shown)


go/tools/builders/cover.go
```diff
@@ -194,3 +195,120 @@ func init() {
...
+// importpath + the file base name. Line directives are honored but the location
+// path is still importpath relative.
+func coveragePath(src string, importPath string) (string, error) {
+	directiveName, err := findFirstLineDirectiveFilename(src)
+	if err != nil {
+		return "", err
+	}
+	filename := src
+	if directiveName != "" {
+		fmt.Println("directiveName: ", directiveName)
```
fmeum, line 210: Debug print?

go/tools/builders/cover.go
```diff
@@ -194,3 +195,120 @@ func init() {
...
+	directive = strings.TrimSpace(directive)
+
+	var content string
+
+	// Handle //line directive
+	if strings.HasPrefix(directive, "//line ") {
+		content = directive[7:] // Remove "//line "
+	} else if strings.HasPrefix(directive, "/*line ") && strings.HasSuffix(directive, "*/") {
+		// Handle /*line*/ directive
+		content = directive[7 : len(directive)-2] // Remove "/*line " and "*/"
```
fmeum, line 236: Replace constants with `len` calls and drop the comment

fmeum (approved):
Looks good to me after the comments have been resolved. 

@jayconrod Could you also take a look?

go/tools/builders/cover_test.go
```diff
@@ -113,3 +113,44 @@ func TestRegisterCoverage(t *testing.T) {
...
+
+		got, err := findFirstLineDirectiveFilename(testFile)
+		if err != nil {
+			t.Errorf("find first line directive filename: %v", err)
+		}
+		if got != test.want {
+			t.Errorf("find first line directive filename: got %q, want %q", got, test.want)
+		}
+	}
+}
```
sluongng, line 156: Nit: I don't think testing an unexported function like this is useful. It forces future maintainers/contributors to study the implementation details when doing refactoring.

I would prefer it if we could add a high-level test, say in `tests/core/coverage/`, that verifies the desired behavior. The setup could be as simple as running coverage over a `go_test` target with `genrule` created one of the source files. That way, we can check for regression in future releases.

PR #4425, opened 2025-08-14
Compile `cmd/pack` from source and update to Go 1.25

fmeum:
**What type of PR is this?**

Fix for compatibility with Go 1.25

The `pack` command, which is no longer shipped as a prebuilt binary with Go 1.25, is now compiled from source together with the `builder` binary.

Conditional logic is avoided by always compiling `pack` from source, even on Go versions before 1.25. By enabling `go build` caching in `go_tool_binary`, the overhead compared to building just the `builder` binary is negligible.

ref: https://github.com/golang/go/issues/74080
fixes https://github.com/bazel-contrib/rules_go/issues/4398

**Other notes for review**

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -497,7 +497,10 @@ set GOTOOLCHAIN=local
 set GO111MODULE=off
 set GOTELEMETRY=off
 set GOENV=off
-{go} build -o {out} -trimpath -ldflags \"-buildid='' {ldflags}\" {srcs}
+{go} build -trimpath -ldflags \"-buildid='' {ldflags}\" -o {out_pack} cmd/pack
+if %ERRORLEVEL% EQU 0 (
+  {go} build -trimpath -ldflags \"-buildid='' {ldflags}\" -o {out} {srcs}
+)
 set GO_EXIT_CODE=%ERRORLEVEL%
 RMDIR /S /Q "{gotmp}"
 MKDIR "{gotmp}"
@@ -506,6 +509,7 @@ exit /b %GO_EXIT_CODE%
             gotmp = gotmp.path.replace("/", "\\"),
             go = sdk.go.path.replace("/", "\\"),
             out = out.path,
+            out_pack = ctx.outputs.out_pack.path,
             srcs = " ".join([f.path for f in ctx.files.srcs]),
             ldflags = ctx.attr.ldflags,
         )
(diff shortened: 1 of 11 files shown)


fmeum:
CC @dzbarsky @r-hang 

@r-hang By building this together with the builder, we can be sure that it is built for a matching platform.

go/private/rules/binary.bzl
```diff
@@ -521,32 +525,35 @@ exit /b %GO_EXIT_CODE%
...
             toolchain = None,
-            outputs = [out, gotmp],
+            outputs = [out, ctx.outputs.out_pack, gotmp],
             mnemonic = "GoToolchainBinaryBuild",
         )
     else:
-        # -a flag instructs the compiler to not read from GOCACHE and force a rebuild.
-        # This provides extra safety in cases of unsandboxed execution.
+        # We do not use -a here as the cache drastically reduces the time spent
+        # on the second go build.
```
sluongng: hmm aren't we doing `go clean -cache` on exit, which should effectively wipe out the local cache for each action run?

fmeum: Yes, but we execute two separate `go build`s in this single action. With `-a`, the second one has to recompile most of the standard library.

go/private/rules/binary.bzl
```diff
@@ -521,32 +525,35 @@ exit /b %GO_EXIT_CODE%
...
         args = ctx.actions.args()
-        args.add("build")
-        args.add("-a")
-        args.add("-o", out)
-        args.add("-trimpath")
-        args.add("-ldflags", ctx.attr.ldflags, format = '-buildid="" %s')
+        args.add(ctx.outputs.out_pack)
+        args.add(out)
```
sluongng, line 536: This part is confusing to me.

Essentially we have `args = out_pack + out + srcs` and 2 `go build` commands:

- `go build -o "$1" cmd/pack` which essentially mean `-o out_pack cmd/pack` which makes sense

- `go build -o "$@"` which I think meant to do `-o out <srcs>` but will be turned into `-o out_pack out <srcs>` instead?

Could we just place `out_pack` and `out` through `format()` and `srcs` through `$@` here?

go/private/rules/binary.bzl
sluongng: ah ok. That makes sense.

go/private/rules/binary.bzl
fmeum: Updated the PR to carry through with what I was after with this and document it at the same time :-)

go/private/rules/binary.bzl
```diff
@@ -562,7 +570,8 @@ HOME={HOME} {go} "$@" """.format(
                 transitive = [sdk.headers, sdk.srcs, sdk.libs, sdk.tools],
             ),
             toolchain = None,
-            outputs = [out],
+            outputs = [out, ctx.outputs.out_pack],
+            execution_requirements = SUPPORTS_PATH_MAPPING_REQUIREMENT,
```
sluongng, line 574: Is this worth adding to the Windows action as well?

fmeum: Windows only supports path mapping with remote execution (not remote caching), so its pretty difficult to test in Bazel CI. I mostly just did this for Unix here since I had to rewrite the script anyway.

fmeum:
> In those cases the gosdk won't have a cmd/pack tool to use and would still end up rebuilding one for the host? 🤔

I don't feel bad for doing exactly what the vanilla Go SDK would do in that case :-D

go/private/rules/binary.bzl
```diff
@@ -521,32 +526,35 @@ exit /b %GO_EXIT_CODE%
...
+            # We place them in the execroot to avoid dependency on `mktemp` and because we don't know
+            # a safe scratch space on all systems. Note that HOME must be an absolute path, otherwise the
+            # Go toolchain will write some outputs to the wrong place and the result will be uncacheable.
+            # We include an output path of this action to prevent collisions with anything else,
+            # including differently configured versions of the same target, under an unsandboxed strategy.
             command = """
-trap "HOME={HOME} GOROOT={GOROOT} {go} clean -cache" EXIT;
-HOME={HOME} {go} "$@" """.format(
+set -eu
+export HOME="$PWD/_go_tool_binary-fake-home-${{1//\\//_}}"
```
dzbarsky, line 550: Did we lose our hardcoded uuid?

dzbarsky (approved):
Nice, doing this as part of the builder makes a lot of sense

go/private/rules/binary.bzl
fmeum: Yeah, I was worried about the path segment becoming too long. The output path is effectively globally (within a build) unique, which I think is good enough.

PR #4435, opened 2025-08-27
Fix trim path problems

jscissr:
**What type of PR is this?**

Bug fix

There were multiple problems with trim paths:

- The trim path was inconsistent between non-cgo and cgo. non-cgo trimmed the working directory, and cgo trimmed the parent of the working directory.
- For non-cgo, paths in external repos are not trimmed when --experimental_sibling_repository_layout is enabled.
- For cgo, the name of the working directory is included in the output, but this name is not consistent between local and remote builds, breaking reproducibility.

These are fixed here by creating the trim path in the same way for both non-cgo and cgo, and trimming the working directory as well as replacing the parent of the working directory with `..`. This means that the resulting paths are relative to the working directory.

Additionally, this fixes the existing feature to extend an existing trimpath. It used the wrong separator (`:` instead of `;`), and it applied `abs()` to the entire trimpath argument instead of the individual paths inside it. (See https://github.com/bazel-contrib/rules_go/pull/2994#issuecomment-1012730010 and https://github.com/bazel-contrib/rules_go/pull/2994#issuecomment-1012796895)

Fixes: https://github.com/bazel-contrib/rules_go/issues/4434
Fixes: https://github.com/bazel-contrib/rules_go/issues/4161

### DIFF
--- a/go/tools/builders/compilepkg.go
+++ b/go/tools/builders/compilepkg.go
@@ -353,14 +353,35 @@ func compileArchive(
 				return err
 			}
 		}
-		gcFlags = append(gcFlags, createTrimPath(gcFlags, srcDir))
+		gcFlags = append(gcFlags, "-trimpath="+srcDir)
 	} else {
 		if cgoExportHPath != "" {
 			if err := os.WriteFile(cgoExportHPath, nil, 0o666); err != nil {
 				return err
 			}
 		}
-		gcFlags = append(gcFlags, createTrimPath(gcFlags, "."))
+		trimPath, err := createTrimPath()
+		if err != nil {
+			return err
+		}
+		// Preserve an existing -trimpath argument, applying abs() to each prefix.
+		for i, flag := range gcFlags {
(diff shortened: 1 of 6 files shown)


fmeum (commented):
Love the test, thanks!

fmeum (approved):

go/tools/builders/compilepkg.go
```diff
@@ -353,14 +353,35 @@ func compileArchive(
...
 			}
 		}
-		gcFlags = append(gcFlags, createTrimPath(gcFlags, "."))
+		trimPath, err := createTrimPath()
+		if err != nil {
+			return err
+		}
+		// Preserve an existing -trimpath argument, applying abs() to each prefix.
+		for i, flag := range gcFlags {
+			if existing, ok := strings.CutPrefix(flag, "-trimpath="); ok {
```
fmeum: Could you replace CutPrefix so that the test verifying compatibility with older versions of Go passes?

PR #4437, opened 2025-09-02
Include blackbox tests in compilation_outputs group

be9:
**What type of PR is this?**

> Bug fix

The PR makes external test archive included in `compilation_outputs` output group. This is necessary because currently, e.g., `bazel build --output_groups=compilation_outputs //foo/bar:bar_test` won't build test files that have `package bar_test` in them.

**Other notes for review**

[Slack context](https://bazelbuild.slack.com/archives/CDBP88Z0D/p1756742089621919)


fmeum (approved):
Thanks!

PR #4438, opened 2025-09-02
Issue 1486: fix C++ static initialization when using cgo 

### LINKED ISSUE #1486: cgo: C++ static initialization not working

enscogitans:
**What type of PR is this?**

Bug fix

Fixes static linking against libraries that have `alwayslink = True` specified. Fixes C++'s static initialization when using cgo 

Fixes #1486

---

Related to https://github.com/bazel-contrib/rules_go/pull/2584 and https://github.com/bazel-contrib/rules_go/pull/2294


go/private/rules/cgo.bzl
```diff
@@ -144,7 +145,13 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
                         # libclntsh.dylib.12.1, users have to create a unversioned symbolic link,
                         # so it can be treated as a simple shared library too.
                         continue
-                lib_opts.append(lib.path)
+
+                if lib.basename.endswith(".lo") and lib.path not in seen_alwayslink_libs:
```
enscogitans, line 149: #### lib.basename.endswith(".lo")

From `bazelbuild/rules_cc` repo:

> If you specify <code>alwayslink=True</code>, you get the <code>.lo</code> file.
> [source](https://github.com/bazelbuild/rules_cc/blob/a0925f28bb4a70857ffdb0b855b9686eb4c0d0aa/cc/private/rules_impl/cc_library.bzl#L577-L578)

In the code:

<blockquote>
<pre><code>elif _matches(<b>[".pic.lo", ".lo", ".lo.lib"]</b>, artifact.basename):
    library_to_link = cc_common.create_library_to_link(
        actions = ctx.actions,
        feature_configuration = feature_configuration,
        cc_toolchain = cc_toolchain,
        static_library = artifact,
        <b>alwayslink = True</b>,
    )</code></pre>
<p><a href="https://github.com/bazelbuild/rules_cc/blob/a0925f28bb4a70857ffdb0b855b9686eb4c0d0aa/cc/private/rules_impl/cc_binary.bzl#L328-L335">source</a></p>
</blockquote>

Note that `.lo.lib` extension [is only used](https://github.com/bazelbuild/rules_cc/blob/db63fa98d8be4e7a1321103678503d0c6e8ca5f6/cc/private/toolchain/windows_cc_toolchain_config.bzl#L95-L110) for MSVC. And `rules_go` don't work with MSVC, so I didn't support this extension

enscogitans, line 149: #### seen_alwayslink_libs

Unless I deduplicate these libs, the `test_side_effect.go` produces 84 instead of 42 and fails (see `side_effect_wrapper` cc_library)
In worse cases such behavior causes a linker's "duplicate symbol" error

However, I don't understand, why it happens. I expected that the linker handles duplicated options well

I tried to trace what is going on with `clinkopts` further, and whether there already was some deduplication logic, which I could update. But I didn't find any. Neither could I find anything similar in the `bazelbuild/rules_cc` repository

Honestly, the presence of `seen_alwayslink_libs` seems suspicious to me. Do you have any ideas how to get rid of it? Or is it ok to leave it as it is?

go/private/rules/cgo.bzl
```diff
@@ -201,6 +208,11 @@ def _cc_libs_and_flags(target):
                 libs.append(library_to_link.dynamic_library)
     return libs, flags
 
+def _alwayslink_lib_opts(go, lib_path):
+    if go.mode.goos == "darwin":
+        return ["-Wl,-force_load,{}".format(lib_path)]
+    return ["-Wl,-whole-archive", lib_path, "-Wl,-no-whole-archive"]
```
enscogitans, line 214: ### darwin

Same flags are used in the `bazelbuild/rules_cc`
[Source](https://github.com/bazelbuild/rules_cc/blob/a0925f28bb4a70857ffdb0b855b9686eb4c0d0aa/tests/rule_based_toolchain/legacy_features_as_args/goldens/macos/libraries_to_link.textproto#L79-L87)

As far as I understand, the flags don't depend on a compiler, they only depend on an OS

### linux

Flags used in the `bazelbuild/rules_cc` are set [here](https://github.com/bazelbuild/rules_cc/blob/a0925f28bb4a70857ffdb0b855b9686eb4c0d0aa/tests/rule_based_toolchain/legacy_features_as_args/goldens/unix/libraries_to_link.textproto#L31-L37) and [here](https://github.com/bazelbuild/rules_cc/blob/a0925f28bb4a70857ffdb0b855b9686eb4c0d0aa/tests/rule_based_toolchain/legacy_features_as_args/goldens/unix/libraries_to_link.textproto#L84-L90)


### windows

MSVC uses [these flags](https://github.com/bazelbuild/rules_cc/blob/db63fa98d8be4e7a1321103678503d0c6e8ca5f6/cc/private/toolchain/windows_cc_toolchain_config.bzl#L529-L532)
mingw-gcc uses the same flags as the flags for linux above

enscogitans:
Hi, @fmeum!

Yes, I would need help with tests. I tried to fix them, but didn't succeed. Plus, I don't have debian and windows machines to debug tests locally

I've got 2 tests failing:

1. Windows
```
c:/tools/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -llib_285ff768: No such file or directory
collect2.exe: error: ld returned 1 exit status
```

Obviously, something is wrong with the configuration. Maybe `_alwayslink_lib_opts` function produces options incompatible with mingw. Or my deduplication (seen_alwayslink_libs) doesn't work well with it. Maybe the order of options is important

2. Debian 11 Bullseye (OpenJDK 17, gcc 10.2.1)

Strangely, my `//tests/core/cgo/cc_static_init:test_side_effect_cc` fails. This is a reference C++ test.
The Go test `//tests/core/cgo/cc_static_init:test_side_effect_go` works well to my surprise.

So, the problem is not in rules_go, actually. But something is wrong with the C++ part only

The test produced 0 instead of 42. My theory was that it happened because the `GetValue()` from `main.cpp` executed faster than the one in `side_effect.cpp`. So, I added a sleep in this [commit](https://github.com/bazel-contrib/rules_go/pull/4438/commits/058a0ff183767ccfd8866973d894e0f5fb47b079), but it didn't help. Maybe the problem is that alwayslink=True doesn't work well on Debian? It's unlikely though.
Maybe I should get rid of the cc test at all? Maybe there are some other rules_cc effects that lead to such a behavior

*upd* Let's see if my next commit fixes this error

enscogitans:
Finally, I got everything fixed, as you can see

fmeum:
Thanks and sorry for the late reply!

Avoiding the manual deduplication would require a very different approach to the entire cgo code. Since your change fixes the bug with minimal additional complexity, I think that it's fine as is.

PR #4439, opened 2025-09-03
Resolve Go and C++ toolchains for each Go target

### LINKED ISSUE #4127: Go rules run cpp toolchain from cgo_context_data on an incompatible execution platform

fmeum:
**What type of PR is this?**

Bug fix

Before this change, rules_go would resolve the C++ toolchain for a fixed helper target and use it in Go targets via a provider. This results in problems for multi-platform builds since the execution platform of the C++ toolchain needs to match that of the Go toolchain, which can only be enforced if both toolchain requirements are declared on the same rule.

Fixes #4127 
Work towards https://github.com/bazel-contrib/rules_go/pull/3740#issuecomment-1904816340

**Other notes for review**


fmeum:
@dzbarsky @jesses-canva @anguslees FYI, this could use some testing :-)

jayconrod (approved):
If this passes all the tests, then it's probably good. I don't know if there's a way to test that the original problem is fixed (C++ and Go toolchains resolved for different platforms). Probably not without a very contrived, complicated example.

If I remember correctly, when toolchains were introduced, there were no optional toolchains, so an optional dependency on a fake target was the only way to get that. It seems like `config_common.toolchain_type(mandatory = False)` is the missing piece.

go/private/context.bzl
```diff
@@ -75,6 +75,19 @@ load(
     "get_source",
 )
 
+CPP_TOOLCHAIN_TYPE = Label("@bazel_tools//tools/cpp:toolchain_type")
+CGO_ATTRS = {
```
dzbarsky, line 79: should we expose these to `defs.bzl`? (Is the intention for extending rules that use `go_context` to use these instead of the previous mechanism?)

go/private/context.bzl
```diff
@@ -479,27 +492,28 @@ def go_context(
...
-        if CgoContextInfo in go_context_data:
-            cgo_context_info = go_context_data[CgoContextInfo]
         go_config_info = go_context_data[GoConfigInfo]
         stdlib = go_context_data[GoStdLib]
         go_context_info = go_context_data[GoContextInfo]
 
-    if goos == "auto" and goarch == "auto" and cgo_context_info:
+    if getattr(attr, "_cc_toolchain", None) and CPP_TOOLCHAIN_TYPE in ctx.toolchains:
+        cgo_context_info = cgo_context_data_impl(ctx)
+    elif go_context_data and CgoContextInfo in go_context_data:
```
dzbarsky, line 509: should these fallbacks go away eventually?

go/private/rules/library.bzl
```diff
@@ -191,8 +194,9 @@ go_library = rule(
         "_allowlist_function_transition": attr.label(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
-    },
-    toolchains = [GO_TOOLCHAIN],
+    } | CGO_ATTRS,
+    fragments = CGO_FRAGMENTS,
```
dzbarsky, line 198: probably want to update https://github.com/bazel-contrib/rules_go/blob/36d0bcbdcbba5875a828e87ff37e5de74d95667e/go/toolchains.rst#writing-new-go-rules to reflect the new recommendations?

dzbarsky (approved):
I don't have access to any complicated go repos at the moment, but this seems reasonable!

go/private/context.bzl
fmeum: That is certainly the intention mid- to long-term, but I would like to be sure that the new way of doing resolution doesn't regress anything before we bake it into public API.

go/private/context.bzl
fmeum: Yes, but given that this would be very breaking, it will likely not happen in the foreseeable future.

PR #4440, opened 2025-09-04
rules/test.bzl: add missing test main archive to compilation_outputs

sluongng:
**What type of PR is this?**

Feature

There can be up to 3 compile actions for each go_test target.
Add the 3rd missing archive, similar to #4437.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -219,6 +219,7 @@ def _go_test_impl(ctx):
             compilation_outputs = [
                 internal_archive.data.file,
                 external_archive.data.file,
+                test_archive.data.file,
             ],
             nogo_fix = nogo_diagnosticss,
             _validation = validation_outputs,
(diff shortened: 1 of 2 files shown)


go/private/rules/test.bzl
```diff
@@ -219,6 +219,7 @@ def _go_test_impl(ctx):
             compilation_outputs = [
                 internal_archive.data.file,
                 external_archive.data.file,
+                test_archive.data.file,
```
fmeum, line 222: Is this just the generated test main? The aim of `compilation_outputs` is to do the minimum work required to ensure that all user-provided code compiles - perhaps it's better to omit this?

sluongng: It depends on whether the user wants to validate GenTestMain result.

Since `generate_test_main.go` is pretty simple right now, I guess there is no need to validate it outside of rules_go development? 🤔

fmeum: Yeah, that's what I'm thinking: wouldn't we really just verify that rules_go doesn't have bugs? 😅

sluongng:
closing this for now.

PR #4461, opened 2025-10-01
Coverage hacks

dzbarsky:
**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**


.bazelrc
```diff
@@ -1,5 +1,6 @@
 startup --host_jvm_args=-DBAZEL_TRACK_SOURCE_DIRECTORIES=1
 
+common --disk_cache=~/disk_cache
```
fmeum: Drop this?

go/private/coverage_shim.go
```diff
@@ -0,0 +1,7 @@
+package main
```
fmeum, line 1: Explain why this doesn't have a go extension (or add one)

go/private/rules/binary.bzl
```diff
@@ -130,10 +132,27 @@ def _go_binary_impl(ctx):
         goarch = ctx.attr.goarch,
     )
 
+    generated_srcs = []
+    data = ctx.attr.data
+    deps = ctx.attr.deps
+
+    if go.coverage_enabled and go.coverage_instrumented:
```
fmeum:
```suggestion
    if go.coverage_enabled:
```
The binary itself may not be covered by the filter, but have transitive deps that do

go/tools/bzltestutil/lcov.go
```diff
@@ -39,7 +44,46 @@ var testFlags = flag.CommandLine
...
+			if err := coverage.WriteCountersDir(coverageDir); err != nil {
+				fmt.Printf("write coverage data: %v", err)
+				return
+			}
+
+			out := filepath.Join(coverageDir, "/go_coverage."+strconv.Itoa(os.Getpid())+".dat")
+
+			// go_binary adds this data dependency in coverage mode.
+			goBin := filepath.Join(os.Getenv("RUNFILES_DIR"), "go_sdk/bin/go")
+			cmd := exec.Command(goBin, "tool", "covdata", "textfmt", "-i", coverageDir, "-o", out)
```
fmeum: Is there an internal package we can use for this? Forking out to a subprocess isn't great if we can avoid it

foo/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
```
fmeum: Drop

go/private/rules/binary.bzl
dzbarsky: yep, good point. Also another (preexisting) issue - we only propagate for .go files across data, so binaries calling across languages also won't propagate coverage info. We may want to expand that to other files, perhaps in a separate change

go/tools/bzltestutil/lcov.go
dzbarsky: I supposed we can try using https://github.com/golang/go/blob/633dd1d475e7346b43d87abc987a8c7f256e827d/src/cmd/covdata/covdata.go#L229-L234 directly, it's internal but we're already abusing internal exithook... :)

.bazelrc
dzbarsky: this was really handy for testing `bazel coverage` because IDEs like to run commands with different flags; I think it would be nice to have something like this for maintainer quality of life. We can change the path though or do it in a separate PR

go/private/coverage_shim.go
dzbarsky: I wasn't sure if anyone runs gazelle manually on the repo and didn't want it creating a dummy target. if we don't care, I can give it the extension

foo/BUILD.bazel
dzbarsky: yep will do once everything else is in a good state (or if we get the integration test working)

go/tools/bzltestutil/lcov.go
darkrift: Doesn't seem like we have access to that internal package, unless we need to explicitely add it somewhere.

```
ERROR: /Users/rlavoie/rlavoie/dzbarsky/rules_go/go/tools/bzltestutil/BUILD.bazel:3:16: GoCompilePkg go/tools/bzltestutil/bzltestutil.a failed: (Exit 1): builder failed: error executing GoCompilePkg command (from target //go/tools/bzltestutil:bzltestutil) bazel-out/darwin_arm64-opt-exec-ST-a828a81199fe/bin/external/go_sdk/builder_reset/builder compilepkg -sdk external/go_sdk -goroot bazel-out/darwin_arm64-fastbuild/bin/stdlib_ -installsuffix ... (remaining 35 arguments skipped)

go/tools/bzltestutil/lcov.go:20:2: could not import cmd/internal/cov (open bazel-out/darwin_arm64-fastbuild/bin/stdlib_/pkg/darwin_arm64/cmd/internal/cov.a: no such file or directory)
```

fmeum: @matloob Can you help us here? We would like to postprocess the Go coverage data to something we can convert to lcov, ideally inprocess without shelling out to `go`. We are okay with hacks and promise to take on the responsibility.to keep this working with future versions of Go (with a fallback to a separate process) 🙂

fmeum: I can successfully import `cmd/internal/cov` if I add `-packages` flags for `cmd/internal/cov` and `cmd/internal/bio` to `_build_stdlib`.

dzbarsky: Yep I came to the same conclusion, thanks for double checking:)

go/private/rules/binary.bzl
fmeum: What is it that we don't propagate? `instrumented_files_info` adds `data` as a `dependency_attribute`.

fmeum:
Not sure about macOS yet, but I could get the Linux integration test to pass with these changes:

```patch
diff --git a/tests/core/coverage/coverage_test.go b/tests/core/coverage/coverage_test.go
index 22009b3c..7cc6e54b 100644
--- a/tests/core/coverage/coverage_test.go
+++ b/tests/core/coverage/coverage_test.go
@@ -50,8 +50,14 @@ go_library(
     name = "a",
     srcs = ["a.go"],
     importpath = "example.com/coverage/a",
-    deps = [":b"],
+    deps = [
+        ":b",
+        "@io_bazel_rules_go//go/runfiles",
+    ],
 	data = [":a_binary"],
+    x_defs = {
+        "aBinaryRlocationPath": "$(rlocationpath :a_binary)",
+    },
 )
 
 go_binary(
@@ -107,28 +113,25 @@ func main() {
 -- a_test.go --
 package a
 
-import (
-    "path/filepath"
-	"os"
-	"os/exec"
-	"testing"
-)
+import "testing"
 
 func TestA(t *testing.T) {
 	ALive()
 }
 
 func TestBinary(t *testing.T) {
-	cmd := exec.Command(filepath.Join(os.Getenv("RUNFILES_DIR"), "bazel_testing", "a_binary"))
-	err := cmd.Run()
-	if err != nil {
-		t.Fatal(err)
-	}
+    RunBinary()
 }
 -- a.go --
 package a
 
-import "example.com/coverage/b"
+import (
+	"os/exec"
+    "example.com/coverage/b"
+    "github.com/bazelbuild/rules_go/go/runfiles"
+)
+
+var aBinaryRlocationPath string
 
 func ALive() int {
 	return b.BLive()
@@ -138,6 +141,18 @@ func ADead() int {
 	return b.BDead()
 }
 
+func RunBinary() {
+    p, err := runfiles.Rlocation(aBinaryRlocationPath)
+	if err != nil {
+		panic(err)
+	}
+	cmd := exec.Command(p)
+	err = cmd.Run()
+	if err != nil {
+		panic(err)
+	}
+}
+
 -- b.go --
 package b
 
@@ -219,6 +234,8 @@ func testCoverage(t *testing.T, expectedCoverMode string, extraArgs ...string) {
 		extraArgs,
 		"--instrumentation_filter=-//:b",
 		"--@io_bazel_rules_go//go/config:cover_format=go_cover",
+		"--test_output=errors",
+		"--test_env=VERBOSE_COVERAGE=1",
 		":a_test",
 	)...)
 
```

fmeum:
For macOS, you could try adding `apple_support` to the `MODULE.bazel` file used by `bazel_testing`. The default Unix toolchain is completely unmaintained on macOS.

go/tools/bzltestutil/lcov.go
dzbarsky: ok no more shelling out now. I had to copy-paste + simplify the visitor but it's not too too bad

dzbarsky:
> For macOS, you could try adding `apple_support` to the `MODULE.bazel` file used by `bazel_testing`. The default Unix toolchain is completely unmaintained on macOS.

thanks, i'll give it a shot. I'm confused why this wasn't an issue before when building the go_test, only with the go_binary...

go/tools/bzltestutil/bincov/coverage_visitor.go
```diff
@@ -0,0 +1,179 @@
+package bzltestutil
+
+// This is a stripped-down version of https://github.com/golang/go/blob/633dd1d475e7346b43d87abc987a8c7f256e827d/src/cmd/covdata/dump.go
+// which supports only the `textfmt` sub-command.
```
fmeum, line 4: Could you add a TODO here to merge this with the lcov conversion routine? If we already copy this (which I think is very reasonable), we might as well emit lcov directly instead of going through the textual format.

go/private/actions/stdlib.bzl
```diff
@@ -143,6 +143,10 @@ def _build_stdlib(go):
     if not go.mode.pure:
         args.add("-package", "runtime/cgo")
 
+    # For bzltestutil's coverage support.
```
fmeum, line 146: We could add these only `if go.coverage_enabled`

dzbarsky: if we did that we'd need to split the coverage bits out of `bzltestutil`, since it's used in non-coverage case as well. Also (we don't do this yet) but I figured we may want to transition stdlib to build without coverage if possible since coverage data there is not useful, to save the rebuilds. WDYT?

go/tools/bzltestutil/bincov/coverage_visitor.go
dzbarsky: yeah good point, will add

go/private/actions/stdlib.bzl
fmeum: Good point. Yeah, let's just always build this here and normalize `--collect_code_coverage` in the stdlib transition.

go/private/rules/binary.bzl
dzbarsky: Yes, but we also pass `extensions = ["go"]`, which means non-go dependencies won't get coverage propagated, right? It doesn't matter too much on this side but if we invoke a python binary it won't get instrumented. And if that python binary invokes a go binary in a child, it also won't be instrumented, right?

So really if we want child-process coverage to work reliably across rulesets, they all need to avoid this filtering

dzbarsky:
> For macOS, you could try adding `apple_support` to the `MODULE.bazel` file used by `bazel_testing`. The default Unix toolchain is completely unmaintained on macOS.

I came back this morning and couldn't reproduce the issue :(

I guess this should be good to go once we resolve the last issue from https://github.com/golang/go/commit/40b3c0e58a0ae8dec4684a009bf3806769e0fc41

```
external/rules_go~/go/tools/bzltestutil/coverage_visitor.go:176:39: too many arguments in call to d.format.EmitTextual
	have (nil, io.Writer)
	want (io.Writer)
	```

File: go/private/rules/binary.bzl (hunk 3 above)
Comment (fmeum) [reply go/private/rules/binary.bzl]: As far as I remember `extensions` only matters for the source attributes, instrumented files from dep attributes are always forwarded. Cross-language coverage should work with this.

Comment (dzbarsky) [reply go/private/rules/binary.bzl]: oh ok, perfect

File: go/tools/bzltestutil/lcov.go (hunk 4 above)
Comment (matloob) [reply go/tools/bzltestutil/lcov.go]: This seems reasonable.

File: go/tools/bzltestutil/bincov/exit_hook.go [hunk 8]
```diff
@@ -0,0 +1,69 @@
...
+
+	os.Setenv("GOCOVERDIR", coverageDir)
+
+	exithook.Add(exithook.Hook{
+		F: func() {
+			if err := coverage.WriteMetaDir(coverageDir); err != nil {
+				fmt.Fprintf(os.Stderr, "write meta: %v\n", err)
+				return
+			}
+			if err := coverage.WriteCountersDir(coverageDir); err != nil {
```
fmeum: Is this safe even if this binary is run multiple times concurrently? The metadata file apparently uses a hash of the file and is static, but I'm not completely sure about that either.

go/private/rules/binary.bzl
```diff
@@ -130,10 +131,24 @@ def _go_binary_impl(ctx):
...
+    generated_srcs = []
+    deps = ctx.attr.deps
+
+    if go.coverage_enabled:
+        coverage_shim = ctx.actions.declare_file(ctx.attr.name + "_coverage_shim.go")
+        ctx.actions.symlink(
+            output = coverage_shim,
+            target_file = ctx.file._coverage_shim,
+        )
+        generated_srcs.append(coverage_shim)
```
fmeum, line 143: Could we just add the source file to the regular sources?

tests/core/coverage/coverage_test.go
```diff
@@ -173,19 +222,18 @@ func TestPanic(t *testing.T) {
...
+		testCoverage(t, "--@io_bazel_rules_go//go/config:race")
 	})
 }
 
-func testCoverage(t *testing.T, expectedCoverMode string, extraArgs ...string) {
+func testCoverage(t *testing.T, extraArgs ...string) {
 	args := append([]string{"coverage"}, append(
 		extraArgs,
 		"--instrumentation_filter=-//:b",
-		"--@io_bazel_rules_go//go/config:cover_format=go_cover",
```
fmeum, line 188: Could we keep this test case and add the lcov test as a new one?

go/tools/bzltestutil/bincov/exit_hook.go
dzbarsky: Yeah, great question. [These writes are atomic](https://github.com/golang/go/blob/2a71af11fc7e45903be9ab84e59c668bb051f528/src/internal/coverage/cfile/emit.go#L41-L58)

dzbarsky: The files look like so:
```
covcounters.018e32252cedf32b4170bdfa58cdf42b.60747.1759602683829898000 - covmeta.018e32252cedf32b4170bdfa58cdf42b
```

i.e. counters have the pid and the timestamp in there, and are also [written atomically](https://github.com/golang/go/blob/2a71af11fc7e45903be9ab84e59c668bb051f528/src/internal/coverage/cfile/emit.go#L325-L329) so the (extremely unlikely) case of a collision is a bit of data loss vs corruption. Seems fine.

go/private/rules/binary.bzl
dzbarsky: `generated_srcs` does end up in the `srcs` [field on the provider](https://github.com/bazel-contrib/rules_go/blob/7ef8fb1bc91d2ea70a53fe123ea0556acb3bf87b/go/private/context.bzl#L315-L316). The only difference is it doesn't end up in `cover`, which seems fine, instrumenting the shim isn't useful. (BTW this is also how the test_main works)

tests/core/coverage/coverage_test.go
dzbarsky: Yeah good call. It only works with lcov, but we can leave the `go_cover` test as-is and maybe fix it in the future if anyone feels passionately about it

tests/core/coverage/coverage_test.go
```diff
@@ -96,15 +117,40 @@ import "testing"
...
+	RunBinary()
+}
 -- a.go --
 package a
 
-import "example.com/coverage/b"
+import (
+	"os"
+	"os/exec"
+    "example.com/coverage/b"
```
fmeum: These lines mix spaces and tabs.

go/tools/bzltestutil/bincov/exit_hook.go
fmeum: Is it obvious that the coverage reader created below only reads back the files we add here vs. counter files corresponding to a different execution of the same binary, which may have different coverage?

In any case, could you add a comment here describing this and the other relevant properties (atomic writes, globally unique or content described names) and why they hold? This seems pretty subtle and if we get it wrong, coverage may be non-deterministic.

We could also emit the files into a unique temp subdirectory instead, which I think is obviously safe.

tests/core/coverage/coverage_test.go
dzbarsky: cleaned this up a bit but it turns out the previous code already mixed it, which is why it's inconsistent. gofmt doesn't handle go code inside strings :)

go/tools/bzltestutil/bincov/exit_hook.go
dzbarsky: yeah I also considered the unique tempdir and hesitated, but let's just do it. it's more obvious

dzbarsky: (done)

tests/core/coverage/coverage_test.go
```diff
@@ -216,6 +266,36 @@ func testCoverage(t *testing.T, expectedCoverMode string, extraArgs ...string) {
...
+	if err := bazel_testing.RunBazel(args...); err != nil {
+		t.Fatal(err)
+	}
+
+	coveragePath := filepath.FromSlash("bazel-testlogs/a_test/coverage.dat")
+	coverageData, err := os.ReadFile(coveragePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !bytes.Contains(coverageData, []byte("SF:main.go")) {
```
fmeum: Could we also test for non-zero coverage on some line?

go/tools/bzltestutil/bincov/exit_hook.go
```diff
@@ -0,0 +1,75 @@
...
+// Lock in the COVERAGE_DIR during test setup in case the test uses e.g. os.Clearenv.
+var coverageDir = os.Getenv("COVERAGE_DIR")
+
+func AddExitHook() {
+	if coverageDir == "" {
+		log.Printf("Not collecting coverage: COVERAGE_DIR is not set")
+		return
+	}
+
+	os.Setenv("GOCOVERDIR", coverageDir)
```
darkrift: This should be the dir created by mkdirTemp

fmeum: Does this have to be set at all if we manually write out data?

dzbarsky: Richard was saying he needed this to avoid an early-return in coverage code, but I don't understand how it's being used, in go upstream that var doesn't seem to be checked with the codepath above. Is it perhaps an earlier Go version?

darkrift: This is most likely coming from https://github.com/golang/go/blob/e74b224b7c7b7511fe37686d81a6e33d40fdeb17/src/internal/coverage/cfile/hooks.go#L36

I was led in a wrong direction, however, we'll need to document the behavior and explain to the user that it can ignore it otherwise someone will come back and raise an issue with this log.

dzbarsky: We can make a other tempdit to dump it into and ignore

darkrift: Just for completeness, the log that is sent when this field is unset is this : https://github.com/golang/go/blob/e74b224b7c7b7511fe37686d81a6e33d40fdeb17/src/internal/coverage/cfile/emit.go#L133-L137

fmeum: Another tempdir would also avoid bloating the coverage dir, which is persisted in remote caches.

tests/core/coverage/coverage_test.go
dzbarsky: done, thanks Richard!

go/tools/bzltestutil/bincov/exit_hook.go
dzbarsky: yep, done.

dzbarsky:
I believe all feedback is addressed now and this should be good to go!

PR #4464, opened 2025-10-02
Allow environment variables to override `go_env` in `@rules_go//go`

### LINKED ISSUE #4462: Environment variables in `go_env` are inconsistently overrideable
Background: we're setting `GOPROXY` using `go_env` in the `go_deps` extension. For obscure reasons, we need to set the value of `GOPROXY` to a different value when Bazel is running on developer workstations vs CI. The approach we've taken is to default to the developer workstation value in MODULE.bazel, and override the value of `GOPROXY` using an environment variable set on the CI runners. In our case, we do this by adding `common --repo_env="GOPROXY=https://override_value/"` to our .bazelrc, but just setting the environment variable also works.

Today, I wanted to add a CI check to ensure that `bazel run @rules_go//go -- mod tidy` is a no-op in CI and noticed that the value for `GOPROXY` is not overriddeable when invoking `bazel run @rules_go//go`.

You can reproduce this as follows:

1) Set `GOPROXY` to `foo.com` in `go_env`
2) Run `GOPROXY=bar.com bazel build //...` and notice that "bar.com" is used to resolve Go modules.
3) Run `GOPROXY=bar.com bazel run @rules_go//go -- env GOPROXY` and see that it prints `foo.com`

Is there another way I can override `go_env` settings from the command line for `bazel run` invocations of the Go toolchain?

rafikk:
**What type of PR is this?**

Bug fix

This PR fixes an inconsistency in how environment variables interact with `go_env`. Prior to this PR, environment variables could override settings in `go_env` within `go_repository`, but could not override `go_env` when invoking `@rules_go//go`.

Fixes #4462

**Other notes for review**

### DIFF
--- a/go/tools/go_bin_runner/main.go
+++ b/go/tools/go_bin_runner/main.go
@@ -128,10 +128,11 @@ func parseConfig() (Config, error) {
 }
 
 func getGoEnv(goBin string, cfg Config) ([]string, error) {
-	env := os.Environ()
+	var env []string
 	for k, v := range cfg.GoEnv {
 		env = append(env, k+"="+v)
 	}
+	env = append(env, os.Environ()...)
 
 	// The go binary lies at $GOROOT/bin/go.
 	goRoot, err := filepath.Abs(filepath.Dir(filepath.Dir(goBin)))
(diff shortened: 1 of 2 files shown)


jayconrod:
Thanks for fixing this!

PR #4466, opened 2025-10-03
Cleanup some cgo handling

dzbarsky:
**What type of PR is this?**
Code cleanup

Improve cgo_deps depset construction and simplifies cgo_export handling

Fixes #

**Other notes for review**


fmeum:
I like where this is going, but CI is red

dzbarsky:
> I like where this is going, but CI is red

Got a bit too impatient with local tests. Should be green now

go/private/actions/archive.bzl
```diff
@@ -204,26 +204,21 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
         source = source,
         data = data,
         direct = direct,
         libs = depset(direct = [out_lib], transitive = [a.libs for a in direct]),
         transitive = depset([data], transitive = [a.transitive for a in direct]),
         x_defs = x_defs,
-        cgo_deps = depset(transitive = [cgo_deps] + [a.cgo_deps for a in direct]),
-        cgo_exports = cgo_exports,
+        cgo_deps = depset(cgo_deps, transitive = [a.cgo_deps for a in direct]),
+        cgo_export = out_cgo_export_h,
```
fmeum, line 221: I guess this is technically a breaking change since `GoArchive` is public API? Could we keep this backwards compatible by making this a one-element `depset`?

dzbarsky: I'm not a huge fan of 1-element depsets and code comments that say "this is small so flattening is ok, trust me", they tend to get stale :) I suppose we could add an additional field for it just for backwards-compat, although it feels like anyone consuming it should be able to adjust prety trivially

Really the `out_cgo_export_h` could have been a plain return from the function, it's just a bit annoying to wire up because the functions are invoked via `go.binary` calling `go.archive`.

fmeum: @jayconrod Do you think this breakage is minor enough that we can pull it off?

I would want to fix the regressions introduced by the latest minor release before we merge a change that's intentionally breaking though.

jayconrod: I'd lean toward preserving backward compatibility.

I'm not sure I really understand this change though. `cgo_exports` has a single direct element, but it also contains files from transitive dependencies. You need all of those to include the top-level file. Why drop those? I'm sure it's possible to reassemble the same set, but it seems harder.

PR #4468, opened 2025-10-04
A bit of starlark micro-optimization

dzbarsky:
**What type of PR is this?**
Cleanup/perf improvement

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -316,6 +316,7 @@ def new_go_info(
     srcs = attr_srcs + generated_srcs
     embedsrcs = [f for t in getattr(attr, "embedsrcs", []) for f in as_iterable(t.files)]
     deps = [get_archive(dep) for dep in getattr(attr, "deps", [])]
+    data = getattr(attr, "data", [])
 
     go_info = {
         "name": go.label.name if not name else name,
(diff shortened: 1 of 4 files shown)


go/private/rules/test.bzl
```diff
@@ -202,7 +201,9 @@ def _go_test_impl(ctx):
         "GO_TEST_RUN_FROM_BAZEL": "1",
     }
     for k, v in ctx.attr.env.items():
-        env[k] = ctx.expand_location(v, ctx.attr.data)
+        if "$" in v:
+            v = ctx.expand_location(v, ctx.attr.data)
+        env[k] = v
```
fmeum: Ternary if would be cleaner here and in the other if, I think

go/private/context.bzl
```diff
@@ -351,7 +352,8 @@ def new_go_info(
     x_defs = go_info["x_defs"]
 
     for k, v in getattr(attr, "x_defs", {}).items():
-        v = _expand_location(go, attr, v)
+        if "$" in v:
+            v = go._ctx.expand_location(v, data)
```
fmeum, line 356: This is good as is since k is also mutated below

fmeum (approved):

go/private/rules/test.bzl
dzbarsky: sure, swapped

PR #4469, opened 2025-10-05
Get rid of `as_iterable`

fmeum:
**What type of PR is this?**

Cleanup

Fixes #

**Other notes for review**


fmeum:
Nvm, this can be both a list and a depset. Not great, but probably hard to fix at this point.

PR #4470, opened 2025-10-05
Fix pure mode detection for special rules

### LINKED ISSUE #4447: Regression: mixed pure/non‑pure modes across deps after PR-4439 causes archive mode mismatch when using --host_platform bazel flag
### Summary
Hi team. First of all thank you for all what you're doing. I recently stumbled upon an issue which looks like regression. After [PR 4439](https://github.com/bazel-contrib/rules_go/pull/4439) landed, some Go targets in the our configuration end up in different “pure” modes depending on whether a C/C++ toolchain is resolvable for that specific target. This leads to a mode mismatch during `go_proto_library` analysis. I would appreciate if you could take a look on it

### Use case
I'm not sure exactly how to reproduce the issue, but roughly our usecase look like this:
```bzl
proto_library(
    name = "example_proto",
    srcs = ["example.proto"],
)

go_proto_library(
    name = "example_go_proto",
    importpath = "example.org/project/example",
    proto = ":example_proto",
)

go_binary(
    name = "binary",
    srcs = ["main.go"],
    deps = [":example_go_proto"],
)
```

### Environment
- **rules_go version**: master (commit after [PR 4439](https://github.com/bazel-contrib/rules_go/pull/4439))
- **Host/Execution platform**: macOS
- **Target platform**: linux/arm64

### Error
```text
ERROR: /.../BUILD.bazel:11:17: in go_proto_library rule //golang/proto:example_go_proto: 
        File "/private/var/tmp/_bazel_dbieliaiev/8c62e38a32f8dc70f07523f1ae120a02/external/io_bazel_rules_go/proto/def.bzl", line 166, column 29, in _go_proto_library_impl
                archive = go.archive(go, go_info)
        File "/private/var/tmp/_bazel_dbieliaiev/8c62e38a32f8dc70f07523f1ae120a02/external/io_bazel_rules_go/go/private/actions/archive.bzl", line 86, column 17, in emit_archive
                fail("Archive mode does not match {} is {} expected {}".format(a.data.label, mode_string(a.source.mode), mode_string(go.mode)))
Error in fail: Archive mode does not match @com_github_golang_protobuf//proto:proto is linux_arm64_stripped expected linux_arm64_pure_stripped
ERROR: /.../BUILD.bazel:11:17: Analysis of target '//golang/proto:example_go_proto' failed
```

The failure is reproducible on master.

### Why this looks like a regression introduced by [PR 4439](https://github.com/bazel-contrib/rules_go/pull/4439)
The behavior change appears tied to `go/private/context.bzl` logic introduced/modified in PR 4439. Specifically this branch:

- [go/private/context.bzl:L507–L508](https://github.com/bazel-contrib/rules_go/blob/0683029b61d4e52dae7e7016e29143d764b02a79/go/private/context.bzl#L507-L508)

If I remove this lines:
```python
if getattr(attr, "_cc_toolchain", None) and CPP_TOOLCHAIN_TYPE in ctx.toolchains:
    cgo_context_info = cgo_context_data_impl(ctx)
```

The build succeeds.

### Additional context: why we are on master

- We require features from `>= 0.56.0`, specifically [PR 4397](https://github.com/bazel-contrib/rules_go/pull/4397).
- All released versions versions `>= 0.56.0` still exhibits [issue 3945](https://github.com/bazel-contrib/rules_go/issues/3945).
- The fix landed in [PR 4432](https://github.com/bazel-contrib/rules_go/pull/4432), which has not been released yet.
- Therefore I pinned to a master commit but found out this regression.

fmeum:
**What type of PR is this?**

Bug fix

`go_proto_library`, `nogo`, and `stdlib` may use cgo and thus need the relevant attributes to detect the availability of a C++ toolchain.

Fixes #4447

**Other notes for review**

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -16,10 +16,14 @@ http_archive(
 # The non-polyfill version of this is needed by rules_proto below.
 http_archive(
     name = "bazel_features",
-    sha256 = "d7787da289a7fb497352211ad200ec9f698822a9e0757a4976fd9f713ff372b3",
-    strip_prefix = "bazel_features-1.9.1",
-    url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.9.1/bazel_features-v1.9.1.tar.gz",
+    sha256 = "9390b391a68d3b24aef7966bce8556d28003fe3f022a5008efc7807e8acaaf1a",
+    strip_prefix = "bazel_features-1.36.0",
+    url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.36.0/bazel_features-v1.36.0.tar.gz",
 )
+
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
+bazel_features_deps()
 
 # Required by protobuf and for //go/private:context.
 http_archive(
@@ -44,10 +48,6 @@ load("@rules_java//java:repositories.bzl", "rules_java_dependencies", "rules_jav
 rules_java_dependencies()
 
 rules_java_toolchains()
-
-load("@bazel_features//:deps.bzl", "bazel_features_deps")
-
-bazel_features_deps()
 
 go_rules_dependencies()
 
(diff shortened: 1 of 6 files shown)


WORKSPACE
```diff
@@ -345,3 +345,10 @@ zig_toolchains(
     },
     version = "0.11.0-dev.3886+0c1bfe271",
 )
+
+http_archive(
```
jayconrod, line 350: Maybe add a quick comment here on how this is used?

I guess we only run tests in WORKSPACE mode, so we don't need to add a dev dependency to MODULE.bazel. That'll be a problem some day, but not for this PR.

fmeum:
@jayconrod Looks like my latest change fixes the bug as per https://github.com/bazel-contrib/rules_go/issues/4447#issuecomment-3435825167. I haven't been able to produce a test for this, the one I added also passes on master. I'm inclined to merge this anyway and add a test later if someone manages to create a reproducer.

PR #4474, opened 2025-10-08
gopackagesdriver: Add additional labels

r-hang:
When using the Bazel gopackagesdriver to analyze a rules_go project it's possible to run into missing packages errors that look like
```
/home/user/go/bazel-pkgdrv/execroot/__main__/bazel-out/
k8-fastbuild/example.go:8:7: could not import example.org/mock (missing package: "@//src/example.org/mock:go_default_library")
```

The root cause is that the gopackagesdriver aspect only traverses target's deps field to populate the package information in the driver registry. In cases where a target depends on another target that is not declared in the "deps" field this packages will be missing from the gopackagesdriver response.

ref: https://github.com/bazel-contrib/rules_go/blob/master/go/tools/gopackagesdriver/aspect.bzl#L30

An example of this edge case is on-the-fly generated go code produced by a bazel rule where the bazel rule provides a static dependency in the rule implementation instead of relying soley on the deps attribute. In this situation, this package is still required for the package analysis to complete but it is missing from the gopackagesdriver response.

This change adds a mechanism to the gopackagesdriver to specify packages that should always be added.

ref https://github.com/bazel-contrib/rules_go/issues/4473

**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/main.go
+++ b/go/tools/gopackagesdriver/main.go
@@ -66,6 +66,7 @@ var (
 	buildWorkingDirectory = os.Getenv("BUILD_WORKING_DIRECTORY")
 	additionalAspects     = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_ADDTL_ASPECTS"))
 	additionalKinds       = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_KINDS"))
+	additionalLabels      = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_ADDTL_LABELS"))
 	emptyResponse         = &driverResponse{
 		NotHandled: true,
 		Compiler:   "gc",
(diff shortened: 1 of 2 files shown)


go/tools/gopackagesdriver/main.go
```diff
@@ -66,6 +66,7 @@ var (
 	buildWorkingDirectory = os.Getenv("BUILD_WORKING_DIRECTORY")
 	additionalAspects     = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_ADDTL_ASPECTS"))
 	additionalKinds       = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_KINDS"))
+	additionalLabels      = strings.Fields(os.Getenv("GOPACKAGESDRIVER_BAZEL_ADDTL_LABELS"))
```
fmeum, line 69: Let's write this out, the abbreviation isn't that commonly used

PR #4481, opened 2025-10-21
Default link mode to PIE on supported platforms

### LINKED ISSUE #4478: Automatically select PIE when building for Android
Currently, I need to explicitly set `linkmode` to `pie` when building for Android, otherwise I get this error:

> runtime.gcdata: missing Go type information for global symbol .dynsym: size 24

However, `go build` does that automatically when `GOOS` is `android`. It feels that `rules_go` should behave the same.

Related to #2371.

alextercete:
**What type of PR is this?**
Feature

We now choose `pie` over `normal` when `linkmode` is omitted (or explicitly set to `auto`) and target supports it. This mirrors the behaviour of `go build`.

Fixes #4478

### DIFF
--- a/tests/extras/gomock/reflective/BUILD.bazel
+++ b/tests/extras/gomock/reflective/BUILD.bazel
@@ -20,6 +20,10 @@ gomock(
     interfaces = ["Client"],
     library = ":client",
     package = "client",
+    target_compatible_with = select({
+        "@platforms//os:windows": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
     visibility = ["//visibility:public"],
 )
 
@@ -30,5 +34,9 @@ go_test(
         "client_test.go",
     ],
     embed = [":client"],
+    target_compatible_with = select({
+        "@platforms//os:windows": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    }),
     deps = ["@com_github_golang_mock//gomock"],
 )


alextercete:
@fmeum I've managed to fix the tests by adding `pure = "on"` to my `go_binary` test target.

dzbarsky:
Doesn't have to be in this PR, but should we try to flip the default to PIE globally? I think that would match the vanilla go toolchain?

fmeum:
> Doesn't have to be in this PR, but should we try to flip the default to PIE globally? I think that would match the vanilla go toolchain?

If that's actually the case then let's do it here so that we don't have to introduce the Android special case at all.

dzbarsky:
> If that's actually the case then let's do it here so that we don't have to introduce the Android special case at all.

Looks like it's on by default in androis/ios/macos and windows without race, but not linux/bsds.

See https://github.com/golang/go/blob/694182d77b1a0e3676214ad0e361bdbdafde33a1/src/internal/platform/supported.go#L235-L253

fmeum:
@alextercete Could you update the condition to match what native Go does?

fmeum (approved):

fmeum:
@dzbarsky The failures look like `cmd/internal/cov` isn't available for all platforms - we may have to limit where we request it.

dzbarsky:
> @dzbarsky The failures look like `cmd/internal/cov` isn't available for all platforms - we may have to limit where we request it.

So there's a few things going on here:
1) `cmd/internal/cov` only exists in go 1.20+, but some of our tests indicate `go 1.16` in their go.mod. I'm not sure if we have an actual minimum version here, but we might as well condition the inclusion of those packages on 1.20+

This patch can do it:
```
--- a/go/private/actions/stdlib.bzl
+++ b/go/private/actions/stdlib.bzl
@@ -143,9 +143,11 @@ def _build_stdlib(go):
     if not go.mode.pure:
         args.add("-package", "runtime/cgo")
 
-    # For bzltestutil's coverage support.
-    args.add("-package", "cmd/internal/cov")
-    args.add("-package", "cmd/internal/bio")
+    version = parse_version(go.sdk.version)
+    if version and version[0] >= 1 and version[1] >= 20:
+        # For bzltestutil's coverage support.
+        args.add("-package", "cmd/internal/cov")
+        args.add("-package", "cmd/internal/bio")
```
(Pulled this out into https://github.com/bazel-contrib/rules_go/pull/4489)

2) The reason this PR triggers this at all is because of the following check, which make us think we can't use the prebuilt SDK and need to build from source instead:
https://github.com/bazel-contrib/rules_go/blob/e3ec0294d47f8d5d71f2e9cac57582d8be59e8dc/go/private/actions/stdlib.bzl#L63
I think that line should be changed to `go.mode.linkmode in (LINKMODE_NORMAL, LINKMODE_PIE)` since the prebuilt SDK should be fine to use for PIE linking.

Hope that helps!

go/private/context.bzl
```diff
@@ -535,6 +544,15 @@ def go_context(
         mode = GoConfigInfo(**mode_kwargs)
         validate_mode(mode)
 
+    # Mirror Go's logic by defaulting to PIE on supported platforms
```
dzbarsky: Oh, one more thing - I'd prefer if `_defaults_to_pie` took `goos` and `race` directly, and then we correctly initalized the `GoConfigInfo` in https://github.com/bazel-contrib/rules_go/blob/e3ec0294d47f8d5d71f2e9cac57582d8be59e8dc/go/private/context.bzl#L997C22-L997C34 and a few lines above.

That way the common case of all build actions could reuse the fastpath (the existing `go_config_info` singleton provider) instead of doing the work below on every build action's analysis

alextercete:
@dzbarsky I tried your suggestion to fix the CI failures on Windows, but they didn't work. Any other ideas?

fmeum:
cc @linzhp for mocks

linzhp:
We can disable that target on Windows if it's failing. We are moving the mock support to https://github.com/uber-go/mock

alextercete:
I'm not sure if there are any targets I can disable. The new ones I've added are all tagged as `manual`, so shouldn't affect the build.

I've rebased onto the latest `master` and tried a few different ideas, but I still can't make CI pass on Windows.

alextercete:
@fmeum @dzbarsky Any advice on how to proceed?

dzbarsky:
@alextercete can you try marking [these targets](https://github.com/bazel-contrib/rules_go/blob/3c293b09f65f8f6b60eec620f471f75d992fd9f0/tests/extras/gomock/reflective/BUILD.bazel#L16-L34) as
```
target_compatible_with = select({
    "@platforms//os:windows": ["@platforms//:incompatible"],
    "//conditions:default": [],
})
```

alextercete:
It turns out there was a legitimate failure on Windows related to testing for stripped portable executables. I've adjusted the test and now CI is green.

go/private/context.bzl
```diff
@@ -995,6 +1004,11 @@ def _go_config_impl(ctx):
 
     toolchain = ctx.toolchains[GO_TOOLCHAIN]
 
+    # Mirror Go's logic by defaulting to PIE on supported platforms
+    linkmode = ctx.attr.linkmode[BuildSettingInfo].value
+    if linkmode == LINKMODE_NORMAL and _defaults_to_pie(toolchain.default_goos, race):
```
fmeum: Shouldn't we do this only if the `linkmode` flag is set to `auto`? Is there a way to force this back to `normal` after this PR?

tests/core/strip/strip_test.go
```diff
@@ -167,12 +167,15 @@ func isStrippedPE(f io.ReaderAt) (bool, error) {
...
+	// pointer and the symbol count in the PE header. That mirrors what Go's
+	// toolchain emits for PIE binaries.
+	if pe.FileHeader.PointerToSymbolTable == 0 || pe.FileHeader.NumberOfSymbols == 0 {
+		return true, nil
 	}
-	emptySymtab := (symtab.VirtualSize <= 4) && (symtab.Size <= 512)
-	return emptySymtab, nil
+	// Some toolchains leave the header populated but drop the table contents.
+	// Fall back to checking whether the parsed table still has entries.
+	return len(pe.COFFSymbols) == 0, nil
```
fmeum, line 178: Thanks for figuring this out, looks quite subtle!

go/private/context.bzl
alextercete: That's a good point. I accounted for that in my previous approach, but now we can't distinguish the default `normal` from the explicitly set `normal` in the _flag_ when the _attribute_ is `auto`.

We could remove the `normal` default from the flag (make it empty or `auto`, to match the attribute). Then if unset we use the logic to choose between `normal` or `pie`. Otherwise, we respect the flag. What do you think?

fmeum: That sounds better to me, but let's ask @jayconrod for his thoughts on this.

jayconrod: @alextercete Your suggestion sounds good to me. The flag can default to `auto`, and we can pick either `normal` or `pie` depending on the platform.

dzbarsky:
@alextercete thanks for powering through this one!

PR #4484, opened 2025-10-22
Commit MODULE.bazel.lock files

### LINKED ISSUE #4483: Commit MODULE.bazel.lock files
This ensures rules_go developers always see the same bits when downloading
files. We should see a security error if a module version changes upstream.

jayconrod:
**What type of PR is this?**

> Other

This ensures rules_go developers always see the same bits when downloading
files. We should see a security error if a module version changes upstream.

Fixes #4483

**Other notes for review**

### DIFF
--- a/tests/bcr/MODULE.bazel
+++ b/tests/bcr/MODULE.bazel
@@ -18,8 +18,8 @@ local_path_override(
     path = "../..",
 )
 
-bazel_dep(name = "gazelle", version = "0.33.0")
-bazel_dep(name = "protobuf", version = "3.19.6")
+bazel_dep(name = "gazelle", version = "0.36.0")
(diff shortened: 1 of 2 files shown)


jayconrod:
Should `tests/bcr/MODULE.bazel.lock` be committed? I'm not sure how that's used exactly.

Should `examples/*/MODULE.bazel.lock` be committed? I think examples ought to have lock files. I added `.bazelversion` files there too so the lock files stay consistent.

Does it matter if `MODULE.bazel.lock` changes with different Bazel versions? Probably not. If so, we could add explicit dependencies on Bazel's implicit module dependencies so that we select the same versions across known Bazel versions. I don't know if that would be enough though.

PR #4490, opened 2025-10-24
.bazelci/presubmit.yml: run BCR tests with multiple Bazel versions

jayconrod:
**What type of PR is this?**

> Test

On presubmit, run the BCR test with multiple Bazel versions. This helps us
catch issues before a release.

n/a

**Other notes for review**

### DIFF
--- a/.bazelci/bcr.yml
+++ b/.bazelci/bcr.yml
@@ -1,1 +1,0 @@
-../.bcr/presubmit.yml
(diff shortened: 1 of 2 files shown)


jayconrod:
Got CI to be green at last. Unsure of the syntax I tried a few different things. There's an `imports` feature, but it looks like we can't import `.bcr/presubmit.yml` because the BCR container image makes some assumptions about what directories exist.

`matrix` seems to work though. It only expands tasks if they're actually used, so setting a `bazel` list there only affects tasks with `bazel: ${{ bazel }}`.

Also added `ubuntu2004_arm64` as one of the matrix platforms because the BCR tests cover that.

@fmeum Mind taking another look? I changed this after you reviewed.

PR #4493, opened 2025-10-30
Reland "go_sdk: store SDK filenames and hashes in lockfile facts"

### LINKED ISSUE #3945: Calls to https://go.dev/dl/?mode=json are breaking airgapped builds - provide way to avoid these

fmeum:
**What type of PR is this?**

Feature

This moves the download of the "all versions" JSON, which can't hit the repository cache, from each individual `go_download_sdk` into the module extension. If the current version of Bazel supports facts, this information will also be persisted in the lockfile, allowing for truly airgapped builds assuming an up-to-date download (formerly repository) cache.

This is a reland of #4393 with the following improvements:
* Adapted to the more limited facts API that actually got merged.
* Prefetching of SDK hashes is now performed on a best-effort basis so that otherwise airgapped builds that provide SDK hashes do not result in failures.

See https://github.com/bazelbuild/bazel/pull/26198 for the PR that added facts support to Bazel.

Fixes #3945 

**Other notes for review**

You can verify that this works by running `bazel mod show_repo @go_default_sdk` in the BCR test repo.

### DIFF
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -217,7 +217,9 @@ def _go_sdk_impl(ctx):
             # resolution. We must not break those builds.
             if not all_sdks_by_version:
                 all_sdks_by_version.clear()
-                all_sdks_by_version.update(fetch_sdks_by_version(ctx, allow_fail = True))
+                all_sdks_by_version.update(fetch_sdks_by_version(ctx, allow_fail = True) or {
+                    "fetch_failed_but_should_not_fetch_again_sentinel": [],
+                })
             sdks = all_sdks_by_version.get(version)
         if sdks == None:
             # This is either caused by an invalid version or because we are in an airgapped build
--- a/go/private/sdk.bzl
+++ b/go/private/sdk.bzl
@@ -580,7 +580,7 @@ def fetch_sdks_by_version(ctx, allow_fail = False):
         allow_fail = allow_fail,
     )
     if not result.success:
-        return {}
+        return None
     data = ctx.read("versions.json")
 
     # If the download is redirected through a proxy such as Artifactory, it may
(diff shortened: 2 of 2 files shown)


tests/bcr/MODULE.bazel.lock
```diff
@@ -0,0 +1,3161 @@
...
+            "python_3_9_host",
+            "rules_python++python+python_3_9_host"
+          ]
+        ]
+      }
+    }
+  },
+  "facts": {
+    "@@rules_go+//go:extensions.bzl%go_sdk": {
+      "1.23.1": {
```
fmeum, line 2831: Please review these facts - they should be as pretty and mergeable as possible.

dzbarsky: won't they merge better if it's a flat list? The artifact names already have both the platform and the version in their name, do we need the nesting?

jayconrod (approved):

tests/bcr/MODULE.bazel.lock
fmeum: So just a dict with `"go1.23.1.aix-ppc64.tar.gz": "f17f2791717c15728ec63213a014e244c35f9c8846fb29f5a1b63d0c0556f756"` entries? That could work, but I don't know how reliably we can parse the version and platform out of the filenames. The fact that you get them from the JSON makes me think that they may not be so stable.

dzbarsky: Fair enough, ok we can try it as you have it. Luckily this isn't too awful to change later if we figure out a better pattern

PR #4498, opened 2025-11-02
remove .vscode and gitignore it

### LINKED ISSUE #4497: checked in .vscode directory causing problems during local dev
It would be nice to not check in the .vscode directory into the repo and instead have some documentation in CONTRIBUTING on how to set it up for this repo.

I'm currently trying to develop in this repo with a fresh copy with VSCode. As soon as I open it in vscode, there are edits made automatically to the `.vscode` directory by vscode. 

Any changes a developer makes to what extensions they've installed for the workspace, and any changes they make to any workspace-wide (that is, repo-wide) settings will be reflected in the .vscode directory.

For instance, here's the diff to .vscode/settings.json that happens as soon as I open this repo in vscode on my macos machine. Every developer will have to remember to remove these kinds of changes every time they submit a new PR to this repo.

```patch
diff --git a/.vscode/settings.json b/.vscode/settings.json
index d81fc004..a79f975e 100644
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -23,10 +23,42 @@
       "tidy": false,
       "upgrade_dependency": false,
       "vendor": false
-    },
+    }
   },
   "go.useLanguageServer": true,
   "go.buildOnSave": "off",
   "go.lintOnSave": "off",
   "go.vetOnSave": "off",
+  "bazel.projectview.notification": false,
+  "bazel.projectview.open": false,
+  "files.exclude": {
+    "**/.git": true,
+    "**/.svn": true,
+    "**/.hg": true,
+    "**/.DS_Store": true,
+    "**/Thumbs.db": true,
+    "**/.classpath": true,
+    "**/.project": true,
+    "**/.settings": true,
+    "**/.factorypath": true,
+    ".bazelci": false,
+    ".bcr": false,
+    ".eclipse": false,
+    ".git": false,
+    ".github": false,
+    ".vscode": false,
+    "docs": false,
+    "examples": false,
+    "extras": false,
+    "go": false,
+    "internal": false,
+    "proto": false,
+    "tests": false,
+    "third_party": false,
+    "tools": false,
+    "bazel-bin": false,
+    "bazel-rules_go": false,
+    "bazel-testlogs": false,
+    "bazel-out": false
+  }
 }

```

(There's also some number of processes from vscode extensions that decided to fire up automatically from vscode that stalled out my bazel 8.4.2 server, but perhaps unrelated. I've not dug into that)

Are y'all comfortable with a PR removing .vscode, adding it to .gitignore, and add some docs to CONTRIBUTING about it?

jmhodges:
**What type of PR is this?**

Other

The .vscode directory isn't meant to be checked in like this, and causes
churn that vscode users have to remove before sending PRs. We remove it,
add it to .gitignore so future folks don't accidentally check it in.
Folks can check out the more general Editor setup wiki page for
instructions.

Fixes #4497.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -139,6 +139,14 @@ http_archive(
     urls = ["https://github.com/bazelbuild/buildtools/archive/refs/tags/v6.4.0.tar.gz"],
 )
 
+# Used for both testing objc interop and building on Apple platforms. Must be
+# above the llvm_toolchain declaration while it's still at 8.0.0.
+http_archive(
+    name = "build_bazel_apple_support",
+    sha256 = "85a7dc13e370f355bf00381238d1cba56450d3e598566b8c52d90ddf301c5dfb",
+    url = "https://github.com/bazelbuild/apple_support/releases/download/1.24.3/apple_support.1.24.3.tar.gz",
+)
+
 # For manual testing against an LLVM toolchain.
 # Use --extra_toolchains=@llvm_toolchain//:cc-toolchain-linux,@llvm_toolchain//:cc-toolchain-darwin
 http_archive(
@@ -292,13 +300,6 @@ http_archive(
 load("@io_bazel_stardoc//:setup.bzl", "stardoc_repositories")
 
 stardoc_repositories()
-
-# For testing objc_library interop, users should not need to install it
-http_archive(
-    name = "build_bazel_apple_support",
-    sha256 = "100d12617a84ebc7ee7a10ecf3b3e2fdadaebc167ad93a21f820a6cb60158ead",
-    url = "https://github.com/bazelbuild/apple_support/releases/download/1.12.0/apple_support.1.12.0.tar.gz",
-)
 
 load(
     "@build_bazel_apple_support//lib:repositories.bzl",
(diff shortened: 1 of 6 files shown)


fmeum (approved):
@jayconrod

CONTRIBUTING.md
```diff
@@ -37,3 +39,59 @@ the one above, the
    this point, if you haven't done so already. Our continuous integration bots
    will test your change automatically on supported platforms. Once everything
    looks good, your change will be merged.
+
+### Editor setup
```
jayconrod: I don't think this belongs in `CONTRIBUTING.md`. The instructions for rules_go itself shouldn't be any different from the general purpose instructions, so let's link from the main `README.md` file instead.

I think we can drop the VSCode instructions or move them to the editor setup wiki page.

jmhodges: Very fair. Happy to drop all of that. One thought: How about a link out to the editor setup wiki from here? It's been pretty hard to google up that doc (even as recently as this past week) and a lil juice might help.

jmhodges: Pushed an update with your feedback in it!

jayconrod: Yes! I'll do that today actually.

I don't know why it is, but Google seems to downrank GitHub wiki pages pretty hard. We should move any relevant pages into the main repo.

PR #4503, opened 2025-11-04
Fix BCR test module for Bazel 9

fmeum:
**What type of PR is this?**

Bug fix

Also run all tests against the latest release candidate for each major release to catch issues early.

Work towards https://github.com/bazelbuild/bazel-central-registry/issues/6402

**Other notes for review**


jayconrod (approved):

fmeum:
The update seems to drop the facts. I'll have to investigate that.

fmeum:
Found the bug: https://github.com/bazelbuild/bazel/pull/27522

We can still proceed with this PR, I will add the tests manually.

PR #4504, opened 2025-11-04
docs: copy editor setup instructions from wiki to docs/editors.md

jayconrod:
**What type of PR is this?**

> Documentation

And link from README.rst.

GitHub wikis seem to be downranked pretty hard by Google, so these instructions haven't been very visible. Linking from the README should help.

n/a

**Other notes for review**

For some reason, the vim instructions were in docs/go/editors/vim.md instead of in the wiki. I've squashed them into a new section here.


jayconrod:
Hat tip to @jmhodges for suggesting we do this. We really should have done this a while ago.

PR #4509, opened 2025-11-07
Add buildInfo metadata support

### LINKED ISSUE #3090: Go 1.18 BuildInfo support

v0.30.0

6bbfc47f1b0a27ee1efeddcc6671f3e4e03235dc

```
Build label: 5.0.0
Build time: Wed Jan 19 14:08:54 2022 (1642601334)
Build timestamp: 1642601334
Build timestamp as int: 1642601334
```

Yes.

```
% uname -sm
Linux x86_64
```

and

```
% uname -sm
Darwin arm64
```

### What did you do?

Wrote a program that tries to access build information via [`runtime/debug.ReadBuildInfo`](https://pkg.go.dev/runtime/debug#ReadBuildInfo) and built it inside a Git repository.

```go
package main

import (
	"fmt"
	"runtime/debug"
)

func main() {
	info, ok := debug.ReadBuildInfo()
	fmt.Println(info, ok)
}
```

Information about the build environment.

### What did you see instead?

```
<nil> false
```

### LINKED ISSUE #4159: how to get the dependencies and their versions for go executables built with bazel
Hi,

Me and https://github.com/RonFed are working on [OpenTelemetry auto instrumentation module](https://github.com/open-telemetry/opentelemetry-go-instrumentation). This issue is regarding the one [here](https://github.com/open-telemetry/opentelemetry-go-instrumentation/issues/1212)

In order to correctly instrument a go binary, we need to extract the dependencies and their versions from the executable file.

We use the [`debug/buildinfo` from the standard library](https://pkg.go.dev/debug/buildinfo), doing something similar to this:

```go
import  (
  "debug/buildinfo"
)

	f, err := os.Open("/path/to/go/executable")
	if err != nil {
		fmt.Println("Error:", err)
		return
	}
	defer f.Close()

	bi, err := buildinfo.Read(f)
	if err != nil {
		fmt.Println("Error:", err)
		return
	}

	fmt.Println("deps:")
	fmt.Println(bi.Deps)
```

When running this inspection on a binary built with `go build` and `go.mod`, we are able to see the deps in the executable. However, when inspecting an executable built with bazel, the deps are not there.

Example sections from our build configuration:

**BUILD.bazel**

```
go_library(
    name = "rolldice_lib",
    srcs = [
        "main.go",
        "rolldice.go",
    ],
    importpath = "github.com/tendermint/tendermint/rolldice",
    deps = [
        "@com_github_gin_gonic_gin//:go_default_library"
    ],
)
```

**deps.bzl**

```
...
    go_repository(
        name = "com_github_gin_gonic_gin",
        importpath = "github.com/gin-gonic/gin",
        sum = "h1:nTuyha1TYqgedzytsKYqna+DfLos46nTv2ygFy86HFU=",
        version = "v1.10.0",
    )
...
```

How can we get this info from a binary built with bazel? Any help will be appreciated

jsharpe:
Implements Go 1.18+ runtime/debug.BuildInfo embedding in Bazel-built binaries with accurate dependency versions collected from Gazelle-generated package_info targets.

This change introduces a buildinfo aspect that traverses the dependency graph and collects version metadata via the package_metadata common attribute (set by go_repository's default_package_metadata in REPO.bazel files). The aspect follows the bazel-contrib/supply-chain gather_metadata pattern.

Features:
- Collects real module versions (e.g., v1.2.3) from package_info metadata
- Replaces (devel) sentinels with actual versions from package_info
- Supports both package_metadata (Bazel 5.4+) and applicable_licenses (legacy)
- Filters out internal monorepo packages to avoid invalid version strings
- Embeds build settings (GOOS, GOARCH, CGO_ENABLED, etc.) in BuildInfo
- Makes dependency information accessible via runtime/debug.ReadBuildInfo()

Implementation details:
- New buildinfo_aspect.bzl traverses all dependencies using attr_aspects=["*"]
- Version map file aggregates importpath→version tuples and passes to linker
- Link builder merges version map with buildinfo dependency list
- Follows supply-chain memory optimization pattern for depset handling
- Includes modinfo.go and buildinfo.go from Go stdlib for metadata generation

Limitations:
- Module checksums are not included as they are not available in package_metadata

This enables tools like `go version -m` and OpenTelemetry instrumentation to properly identify dependencies in Bazel-built binaries.

Fixes #3090
Fixes #4159

🤖 Generated with [Claude Code](https://claude.com/claude-code)


go/private/actions/archive.bzl
```diff
@@ -167,6 +167,29 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
+            # Use (devel) as sentinel - invalid version that will be replaced
+            # with real version from PackageInfo or filtered out if internal
+            version = "(devel)"
+            dep_buildinfo.append((d.data.importpath, version))
+            dep_buildinfo_set[dep_key] = None
+
+        # Add transitive dependencies
+        for dep_path, dep_version in getattr(d.data, "_buildinfo_deps", ()):
+            if dep_path not in dep_buildinfo_set:
+                dep_buildinfo.append((dep_path, dep_version))
```
fmeum: Wouldn't this result in potentially quadratic total memory usage for deeper dependency trees? Would it be possible to instead track separate `depset`s of importpaths and `depset`s of package metadata providers and doing the matching at execution time in top-level rules only?

go/private/actions/link.bzl
```diff
@@ -44,14 +44,37 @@ def emit_link(
...
+            buildinfo_content += "path\t{}\n".format(archive.data.importpath)
+
+        # Add dependencies in sorted order for determinism
+        deps_list = sorted(archive.data._buildinfo_deps, key = lambda x: x[0])
+        for dep_path, dep_version in deps_list:
+            # Format: dep\t<importpath>\t<version>
+            buildinfo_content += "dep\t{}\t{}\n".format(dep_path, dep_version)
+
+        go.actions.write(
+            output = buildinfo_file,
```
fmeum, line 107: If we track info as `depset`s (see comment above), we could materialize that into a file at execution time using `TemplateDict` or `Args`. I can take over that part if you want.

go/private/aspects/buildinfo_aspect.bzl
```diff
@@ -0,0 +1,103 @@
...
+    doc = "Provides dependency version metadata for buildInfo",
+    fields = {
+        "version_map": "Depset of (importpath, version) tuples",
+    },
+)
+
+def _buildinfo_aspect_impl(target, ctx):
+    """Collects package_info metadata from dependencies.
+
+    Following bazel-contrib/supply-chain gather_metadata pattern, this checks:
```
fmeum: @Yannic I haven't followed this repo lately, are there any bits that rulesets can reuse rather than adapt?

go/private/aspects/buildinfo_aspect.bzl
```diff
@@ -0,0 +1,103 @@
...
+    Args:
+        target: The target being visited
+        ctx: The aspect context
+
+    Returns:
+        List containing BuildInfoMetadata provider
+    """
+    direct_versions = []
+
+    # Approach 1: Check package_metadata common attribute (Bazel 5.4+)
```
fmeum: We only support Bazel 6 and even that will be dropped soon. We could skip the fallback.

go/private/aspects/buildinfo_aspect.bzl
```diff
@@ -0,0 +1,103 @@
...
+    # Approach 2: Check applicable_licenses (legacy compatibility)
+    if not package_metadata and hasattr(ctx.rule.attr, "applicable_licenses"):
+        attr_value = ctx.rule.attr.applicable_licenses
+        if attr_value:
+            package_metadata = attr_value if type(attr_value) == "list" else [attr_value]
+
+    # Collect metadata from transitive dependencies (supply-chain pattern)
+    transitive_depsets = []
+
+    # Traverse all attributes (supply-chain uses attr_aspects = ["*"])
```
fmeum: Since we only want to collect info on non-exec Go deps, we should limit collection to `deps` and `embed` (everything that accepts `GoInfo`). Restricting in this way would greatly simplify the logic below.

go/private/aspects/buildinfo_aspect.bzl
```diff
@@ -0,0 +1,103 @@
...
+                continue
+
+            # Collect transitive BuildInfoMetadata
+            if BuildInfoMetadata in dep:
+                transitive_depsets.append(dep[BuildInfoMetadata].version_map)
+
+    # Return using supply-chain memory optimization pattern
+    if not direct_versions and not transitive_depsets:
+        # No metadata at all, return empty provider
+        return [BuildInfoMetadata(version_map = depset([]))]
```
fmeum: Extract this instance into a shared constant to further reduce memory usage.

go/private/aspects/buildinfo_aspect.bzl
```diff
@@ -0,0 +1,103 @@
...
+
+    if not direct_versions and len(transitive_depsets) == 1:
```
fmeum: This optimization is unnecessary, the `depset` constructor already does this (and more)

go/private/aspects/buildinfo_aspect.bzl
```diff
@@ -0,0 +1,103 @@
...
+        ),
+    )]
+
+buildinfo_aspect = aspect(
+    doc = "Collects package_info metadata for Go buildInfo",
+    implementation = _buildinfo_aspect_impl,
+    attr_aspects = ["*"],  # Traverse all attributes (supply-chain pattern)
+    provides = [BuildInfoMetadata],
+    # Apply to generated targets including package_info
+    apply_to_generating_rules = True,
```
fmeum, line 99: I would think that we don't want this (e.g. don't traverse code generators that produce .go files in sources), but I may be missing a use case.

go/private/rules/binary.bzl
```diff
@@ -155,6 +160,42 @@ def _go_binary_impl(ctx):
...
+        version_map_file = ctx.actions.declare_file(ctx.label.name + "_versions.txt")
+
+        # Sort and deduplicate
+        version_dict = {}
+        for importpath, version in version_tuples:
+            if importpath not in version_dict:
+                version_dict[importpath] = version
+
+        # Write tab-separated file
+        lines = ["{}\t{}".format(imp, ver) for imp, ver in sorted(version_dict.items())]
```
fmeum: We can also do this lazily.

go/tools/builders/importcfg.go
```diff
@@ -196,6 +219,49 @@ package with this path is linked.`,
...
+		buildSettingAppend := func(key, value string) {
+			if value != "" {
+				buildInfo.Settings = append(buildInfo.Settings, BuildSetting{Key: key, Value: value})
+			}
+		}
+
+		buildSettingAppend("-compiler", cfg.compiler)
+		buildSettingAppend("-buildmode", cfg.buildMode)
+		if cfg.pgoProfilePath != "" {
+			buildSettingAppend("-pgo", cfg.pgoProfilePath)
```
fmeum, line 242: We need to check that these paths aren't absolute. It would be very easy to accidentally destroy hermeticity that way. This also applies to other flags below.

go/tools/builders/modinfo.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Magic numbers for the buildinfo (apparently). First added in
+// https://go-review.googlesource.com/c/go/+/123576/4/src/cmd/go/internal/modload/build.go
+var (
+	infoStart, _ = hex.DecodeString("3077af0c9274080241e1c107e6d618e6")
+	infoEnd, _   = hex.DecodeString("f932433186182072008242104116d8f2")
+)
+
+func ModInfoData(info string) []byte {
+	return []byte(string(infoStart) + info + string(infoEnd))
```
fmeum, line 33: There's a lot of back and forth conversion here and at the call site. Are the magic numbers UTF-8 strings? We should either choose `[]byte` and `append` or `string` and `+` and convert at most once.

go/private/actions/archive.bzl
jsharpe: Yes, have changed it to track via depsets.

go/private/aspects/buildinfo_aspect.bzl
jsharpe: Since we need to support WORKSPACE, I don't think we can yet depend on the supply-chain tools as that is currently bzlmod only which is why I have duplicated this logic in here.

go/private/aspects/buildinfo_aspect.bzl
jsharpe: Done

go/private/aspects/buildinfo_aspect.bzl
jsharpe: Removed

fmeum:
@jsharpe Could you edit the commit messages to remove the `fmeum` mention? Otherwise I get an email on every push 😅

darkrift:
How does that compare with #4595 ?

fmeum:
Superseded by #4595

PR #4524, opened 2025-11-30
Exclude cgotest from stdlib build

dzbarsky:
It didn't build for me with a zig toolchain, but anyway we shouldn't need to build this for user code

```
# net/internal/cgotest
bazel-out/darwin_arm64-opt-exec-ST-6c94b0e707c0/bin/external/rules_go+/stdlib_/src/net/internal/cgotest/resstate.go:10:10: fatal error: 'resolv.h' file not found
   10 | #include <resolv.h>
```

**What type of PR is this?**
Perf tweak?


linzhp:
This only happen on macOS, right? Zig toolchain cannot build for macOS right now. Can we only exclude this tests for macOS builds?

fmeum (approved):

PR #4526, opened 2025-12-02
Start migrating rules_go workspace to bzlmod

dzbarsky:
**What type of PR is this?**

>
> Bug fix
> Feature
> Documentation
> Other

Fixes #

**Other notes for review**

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -59,7 +59,9 @@ use_repo(
 
 bazel_dep(name = "stardoc", version = "0.8.0", dev_dependency = True)
 
-go_sdk.download(
+dev_go_sdk = use_extension("//go:extensions.bzl", "go_sdk", dev_dependency = True)
+
+dev_go_sdk.download(
     name = "rules_go_internal_compatibility_sdk",
     version = "1.22.12",
 )
--- a/go/private/extensions.bzl
+++ b/go/private/extensions.bzl
@@ -275,6 +275,8 @@ def _go_sdk_impl(ctx):
         # over extra SDKs specified with `download`. That way the `from_file` toolchains are registered
         # with higher precedence and become default, while `download`'ed toolchains can still be
         # requested explicitly.
+        # TODO(zbarsky/fmeum): This is still not the ideal ordering. We should respect the order that tags are
+        # specified in, but Bzlmod currently doesn't provide this information across tag classes.
         for index, download_tag in enumerate(additional_download_tags + module.tags.download):
             # SDKs without an explicit version are fetched even when not selected by toolchain
             # resolution. This is acceptable if brought in by the root module, but transitive
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -14,7 +14,7 @@ matrix:
 tasks:
   debian11_bazel6:
     platform: debian11
-    bazel: 6.5.0 # test minimum supported version of bazel
+    bazel: 7.1.0 # test minimum supported version of bazel that works with our bzlmod setup
     build_targets:
       - "//..."
       - "-//tests/core/go_binary:goos_pure_bin"
--- a/go/tools/gopackagesdriver/BUILD.bazel
+++ b/go/tools/gopackagesdriver/BUILD.bazel
@@ -33,20 +33,12 @@ go_binary(
     },
 )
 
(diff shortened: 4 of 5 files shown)


jayconrod:
Thanks for working on this, Bzlmod migrations always seem to be a bit harder than expected, and this is no exception.

This is a large change, but all looks good. `//go/tools/gopackagesdriver:gopackagesdriver_test` is failing on a few platforms, but as soon as that's green, I think this is ready to go.

Replying to a couple things from Slack:

> Flipped the order of toolchain registration to prefer from_file over download tags. Otherwise our old compatibility SDK would be used for go mod download which is no good. This feels good in general, use the go.mod toolchain by default unless user explicitly asks for a different version

This may cause a user-visible change in functionality, but I expect most users want exactly one version of Go, so it probably won't be an issue in practice.

> We got some minor version bumps in go.mod and grew a few extra minor deps. Given we already exported some big stuff like grpc I think it's not a huge deal, but it would be nice in the future if we used isolated = True to avoid polluting the overall closure with our deps from //tests

The changes here look fine, this shouldn't have much impact on downstream users.

>  we still test with bazel version 6.5, shall we bump minimum to 7.x? (Do we want to wait for 9 to actually ship first?)

Bumping to 7.x is probably a good idea at this point. The 9.x release should be very soon, and 6.x module functionality was pretty new at the time.

MODULE.bazel
```diff
@@ -50,4 +50,19 @@ use_repo(
...
+    "org_golang_google_genproto_googleapis_bytestream",
+    "org_golang_google_genproto_googleapis_rpc",
+    "org_golang_x_crypto",
 )
+
+### Dev dependencies
+
+bazel_dep(name = "stardoc", version = "0.8.0", dev_dependency = True)
+
+go_sdk.download(
```
fmeum: Could you add another use_extension for this with `dev_dependency = True` so that downstream doesn't pick this up?

go/private/extensions.bzl
```diff
@@ -271,7 +271,11 @@ def _go_sdk_impl(ctx):
             download_tag["version"] = version
             additional_download_tags.append(struct(**download_tag))
 
-        for index, download_tag in enumerate(module.tags.download + additional_download_tags):
+        # We handle the `additional_download_tags` first so that `from_file` takes precedence
+        # over extra SDKs specified with `download`. That way the `from_file` toolchains are registered
+        # with higher precedence and become default, while `download`'ed toolchains can still be
+        # requested explicitly.
```
fmeum, line 277: Could you leave a TODO here that this is still not the ideal ordering? We should respect the order that tags are specified in, but Bzlmod currently doesn't provide this information across tag classes. I think I know how to add this and will try to get it into Bazel 9.

tests/core/stdlib/BUILD.bazel
```diff
@@ -4,7 +4,10 @@ load(":stdlib_files.bzl", "stdlib_files")
 go_test(
     name = "buildid_test",
     srcs = ["buildid_test.go"],
-    data = [":stdlib_files"],
+    data = [
+        "@go_sdk//:files",
```
fmeum, line 8: The reference to `../go_sdk` won't work since the path would need to mention the canonical name. Could you depend on `@rules_go//go` instead and run `go env` with it to learn the path?

fmeum (approved):

MODULE.bazel
dzbarsky: yep! I actually did that originally but I wasn't sure if it actually had the effect we wanted - seems like it will get treated the same and all the toolchains get registered together anyway?

tests/core/stdlib/BUILD.bazel
dzbarsky: I didn't follow which reference you're worried about, I only see that in `stdliblist_test.go`?

MODULE.bazel
fmeum: Yes, the order of tags should not be affected, just whether they are seen by the extension.

tests/core/stdlib/BUILD.bazel
fmeum: Ah, that's right. That's the one I meant can't work, buildid_test correctly uses the runfiles library

MODULE.bazel
dzbarsky: oh gotcha, makes sense

go/private/extensions.bzl
fmeum: https://github.com/bazelbuild/bazel/pull/27883

PR #4531, opened 2025-12-05
refactor: Remove next batch of WORKSPACE content

dzbarsky:
**What type of PR is this?**
Modernize the repo

Fixes #

**Other notes for review**


MODULE.bazel
```diff
@@ -60,11 +60,46 @@ use_repo(
...
+use_repo(protoc, "protoc_toolchains")
+
+register_toolchains(
+    "@protoc_toolchains//...",
+    dev_dependency = True,
+)
+
 # Used to transition binaries in rules_go's test suite to different configurations.
 bazel_dep(name = "with_cfg.bzl", version = "0.14.1", dev_dependency = True)
+bazel_dep(name = "runfiles_remote_test", version = "0.0.0", dev_dependency = True)
```
fmeum, line 84: This could probably be a regular repo backed by `use_repo_rule` and `local_repository`.

dzbarsky: Oh I wasn't sure if use_repo_rule allows dev dep but looks like it becomes an extra kwarg on the generated rule , I'll try

dzbarsky:
```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.NullPointerException: Cannot invoke "com.google.devtools.build.lib.skyframe.PackageValue.getPackage()" because the return value of "com.google.devtools.build.skyframe.SkyframeLookupResult.getOrThrow(com.google.devtools.build.skyframe.SkyKey, java.lang.Class)" is null
	at com.google.devtools.build.lib.skyframe.TestExpansionFunction.getPrerequisites(TestExpansionFunction.java:169)
	at com.google.devtools.build.lib.skyframe.TestExpansionFunction.computeExpandedTests(TestExpansionFunction.java:85)
```

:(

fmeum: Can you reproduce this? The code is clearly buggy, but I would like to confirm that there is just one bug :-)

PR #4538, opened 2026-01-25
chore(docs): add Quick Start section to README

alexeagle:
Added a 'Quick Start' section to the README for easier project setup.

Note, this is the same link reachable from https://bazel.build/start

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -164,6 +164,10 @@ Note: Since version v0.51.0, rules_go requires Bazel ≥ 6.5.0 to work.
 
 The ``master`` branch is only guaranteed to work with the latest version of Bazel.
 
+Quick Start
+-----------
+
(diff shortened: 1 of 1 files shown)


fmeum:
Doesn't https://bazel.build/start/go reference a bazel-examples subdirectory? Is that the same thing?

jayconrod:
I'd rather not add this. https://github.com/bazel-starters org seems Aspect-specific. The template repo itself also has Aspect stuff. 

It's important for a Quick Start example to be minimal. I don't think it should endorse any company or require any tech that's not strictly needed.

jayconrod:
Declining this.

- It's an Aspect-specific repo, though it doesn't openly declare that.
- It's not open source. No license.

PR #4539, opened 2026-01-29
Disable all experiments during bootstrap

patrickmscott:
We are using a custom toolchain that enables boringcrypto by default. This requires a cgo toolchain that is not available during bootstrap.

**What type of PR is this?**

Feature

Disable all go experiments during bootstrapping of the builder.

Fixes #

**Other notes for review**

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -539,6 +539,7 @@ set GOTOOLCHAIN=local
 set GO111MODULE=off
 set GOTELEMETRY=off
 set GOENV=off
+set GOEXPERIMENT=none
 {go} build -trimpath -ldflags \"-buildid='' {ldflags}\" -o {out_pack} cmd/pack
 if %ERRORLEVEL% EQU 0 (
   {go} build -trimpath -ldflags \"-buildid='' {ldflags}\" -o {out} {srcs}
@@ -601,6 +602,7 @@ exit /b %GO_EXIT_CODE%
                 "GO111MODULE": "off",
                 "GOTELEMETRY": "off",
                 "GOENV": "off",
+                "GOEXPERIMENT": "none",
                 "GO_BINARY": sdk.go.path,
                 "LD_FLAGS": ctx.attr.ldflags,
             },


jayconrod (approved):
Seems reasonable to disable `GOEXPERIMENT` for bootstrapping. Seems difficult to test, but be aware that without a test, it's quite possible this will regress in the future if we change how bootstrapping works.

HDYA:
Removing all experiment configs may break other toolchain, can we add a parameter for this so both cases can work?
Such as what is proposed by #4547 
@fmeum thank you.

xuxife:
Hi rules_go maintainers, we're a team in Microsoft leveraging Bazel and rules_go, and this change may potentially prevent us upgrading our Go to 1.25.
[Microsoft build of Go 1.25 crypto backend changes](https://devblogs.microsoft.com/go/microsoft-go-defaults-to-system-crypto/)
We're using Microsoft fork of Go (yes, for the sake of FIPS), and since 1.25, it by-default enable "systemcrypto" which requires CGO.
Since bootstrap toolchain is CGO-disabled, our proposal #4547 is intended to expose GOEXPERIMENT.
I think our proposal is more general.
We can use `toolchain_experiments="nosystemcrypto"` for Microsoft-Go.
@patrickmscott can use `toolchain_experiments="none"`

patrickmscott:
> Hi rules_go maintainers, we're a team in Microsoft leveraging Bazel and rules_go, and this change may potentially prevent us upgrading our Go to 1.25. [Microsoft build of Go 1.25 crypto backend changes](https://devblogs.microsoft.com/go/microsoft-go-defaults-to-system-crypto/) We're using Microsoft fork of Go (yes, for the sake of FIPS), and since 1.25, it by-default enable "systemcrypto" which requires CGO. Since bootstrap toolchain is CGO-disabled, our proposal #4547 is intended to expose GOEXPERIMENT. I think our proposal is more general. We can use `toolchain_experiments="nosystemcrypto"` for Microsoft-Go. @patrickmscott can use `toolchain_experiments="none"`

I was able to solve this another way. I used a custom repository rule that modifies `go.env` in the toolchain to set different defaults. Then I configure the toolchain within rules_go for what I want. The bootstrap process uses the defaults (which disables experiments for me) and then everything else is configured by `go_wrap_sdk`.

PR #4540, opened 2026-01-30
Remove invalid double html break

katre:
Looks like the conversion of this file from HTML to markdown left in several instances of `<br><br>`, at least some of which were making the triple-backtick syntax invalid and causing rendering issues in the docs for `go_test`.

I've left the ones inside the tables: I don't know a better way offhand to have a linebreak inside a table.

We should go back over this and fix other html-isms, I also noted several uses of `<ul><li>` instead of an actual markdown list.

**What type of PR is this?**

Documentation

Cleans up docs formatting.


katre:
@dzbarsky

katre:
I edited the wrong file(s): closing this and trying again.

PR #4541, opened 2026-01-30
Replace html linebreaks with newlines.

katre:
The html linebreaks, coupled with the markdown style triple-backtick, was breaking rendering in some docs, specifically for `go_test`.

**What type of PR is this?**

Documentation


katre:
@dzbarsky Take two!

PR #4542, opened 2026-01-30
Replace HTML with Markdown in docs

katre:
Changes:
- Convert html lists to Markdown lists.
  - Except in attributes, which breaks rendering
- Remove html `code` tag in favor of Markdown
- Remove html `strong` tag in favor of markdown

**What type of PR is this?**

Documentation


katre:
More cleanups! @dzbarsky

PR #4546, opened 2026-02-03
fix(cgo): select static or dynamic libraries based on link mode

ma-oli:
In cgo, when picking between the dynamic and the static version of a library to link against, we're implicitly assuming that each library is present only once in the list of LibraryToLink in the CcInfo provider.

In cgo.bzl, we're therefore going through each LibraryToLink object, and picking one of the libraries, always favoring static (if available).

This logic has two issues that I'm going to outline below.

First, when the cgo library is defined as a cc_import(), we have the following problem. A picture is worth a thousand words:
```
        A
       / \
      D   S
       \ /
        X
```
Let's say that we have the diamond dependency below, where A depends on
D (that is defined as a cc_import() with only a shared_library) and S
(that is defined as a cc_import() with a shared and a static library).
X is available as both dynamic and a static library.
The current implementation will link D dynamically, S statically, X statically, but
at test time, D will require libX.so, which will be missing.

Second, when the cgo library is defined as a cc_library(), the following happens. cc_library() works differently than cc_import() in a sense that it will create one LibraryToLink object per object file. So there will be one for the .so file, and another one for the .a file. The current implementation will therefore assume we need to link the library twice, once statically, and once dynamically. Therefore leading to requiring the .so file at runtime if `static = "off"`, and erroring out if `static = "on"`.

In this change, we're changing the logic by taking into account the mode when trying to extract the cc libs and flags. If `static = "on"`, we'll pick only the .a file. Otherwise, we'll only pick the dynamic libraries. This means that when cc_library() is used to define these libraries, "static" will be respected and the binary will be linked accordingly.

### DIFF
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -103,7 +103,7 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
         if CcInfo in d:
             cc_transitive_headers = d[CcInfo].compilation_context.headers
             inputs_transitive.append(cc_transitive_headers)
-            cc_libs, cc_link_flags = _cc_libs_and_flags(d)
+            cc_libs, cc_link_flags = _cc_libs_and_flags(d, go.mode)
             inputs_direct.extend(cc_libs)
             deps_direct.extend(cc_libs)
             cc_defines = d[CcInfo].compilation_context.defines.to_list()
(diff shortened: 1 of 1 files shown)


go/private/rules/cgo.bzl
```diff
@@ -191,22 +191,24 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
...
         for library_to_link in li.libraries:
-            if library_to_link.static_library != None:
-                libs.append(library_to_link.static_library)
-            elif library_to_link.pic_static_library != None:
-                libs.append(library_to_link.pic_static_library)
-            elif library_to_link.interface_library != None:
-                libs.append(library_to_link.interface_library)
-            elif library_to_link.dynamic_library != None:
-                libs.append(library_to_link.dynamic_library)
+            if mode.static:
```
fmeum, line 202: While it seems obvious to me that we should only link static libraries in `static` mode, wouldn't static libraries be fine to link in non-`static` mode?

It would be great if you could add some integration that we can use to reason about the effects.

PR #4553, opened 2026-02-11
Allow experimental ability to bootstrap go toolchain from source

dzbarsky:
This allow users to apply targeted patches on top of the Go compiler in an easy way. The motivating use case is https://github.com/cerisier/toolchains_llvm_bootstrapped requiring a small compiler patch that landed upstream but did not make the Go 1.26 release, but it can also be used to easily test Go compiler changes against a Bazel codebase


sluongng:
This is very nice. On top of my head, there are several Bazel users who are shipping bleeding edges Go SDK: cockroachdb, datadog(?), bytedance, …

this should help them experiment with go sdk much easier through rules_go

jayconrod (dismissed):
I said this on Slack, but I'll echo this here: I'm sure this is useful for some folks, but implementing this correctly and supporting it will be *complicated*, and I don't think we maintainers of rules_go should take on that complication.

For this type of thing, I would actually want to understand what every line does and why it's done that way, but there are parts of this PR I don't understand at all. I would not want to have to fix this when a new version of Go comes out that builds slightly differently.

As alternatives, I'd suggest a couple approaches:

1. Customers needing forked SDKs can build a separate binary archive outsize of Bazel then download it with the `go_sdk` module extension instead of the first party downloads from `go.dev`. We can make this work more smoothly if needed, but I'd expect this to work today.
2. Customers needing forked SDKs could provide a separate module extension, toolchain, and rule that is compatible with `go_toolchain`, but it builds from source using another Go toolchain. So this uses Bazel, but it can be in a different module than rules_go and can be maintained by a different set of people willing to take on the maintenance cost.

sluongng:
> implementing this correctly and supporting it will be complicated, and I don't think we maintainers of rules_go should take on that complication

I think we should add more tests to ease that pain.
The more detailed the test is, the easier it will be to maintain over time.

> there are parts of this PR I don't understand at all

I would love to hear which part and learn together.
It's totally possible that the implementation could be trimmed down further.

Personally, I would like to see this PR broken down into smaller PRs:
- A separate one for refactoring existing stuff to `sdk_build_defs.bzl`.
- 

> I would not want to have to fix this when a new version of Go comes out that builds slightly differently.

I actually talked to @dzbarsky a little bit about this.
I would prefer if we do the build in Bazel instead of using the official `make.{bash,bat}` fat action.
But the current implementation of `_experimental_bootstrap_go_sdk` simply creates one big action and runs the upstream `make.bash` inside, which should make it a lot easier to maintain through multiple versions of Go.

> As alternatives

I think (1) is already happening today. The problem with it is the lack of transparency and the increased cost of maintenance pushed down to the end user. Instead of just reviewing a set of patches applied to the SDK, the user have to (a) create and apply those patches, (b) maintain a separate build and release process for the new sdk distribution, (c) storage, checksum of the released sdk archive.

(2) is essentially the same as using the current code, but moving it outside of rules_go: fragmentation, harder to discover, and potential breakages over time, as the API inside rules_go may change.

---

I think your concern is valid, but the tradeoffs of the alternative options are not worth that much. And I suspect the maintenance cost can be lowered once we add tests to cover this logic.

The current implementation has already moved this to an `experimental_build_compiler_from_source` attribute to warn users about the feature status, so I'm in favor of having this inside rules_go for a while before we decide to yank it.

At the very least, we know that there are real use-cases this helps unlock (i.e., changing how the go linker works with the new cc toolchain)

go/private/BUILD.sdk.bazel
```diff
@@ -54,6 +35,20 @@ filegroup(
...
+        [
+            "pkg/tool/{goos}_{goarch}/asm*",
+            "pkg/tool/{goos}_{goarch}/cgo*",
+            "pkg/tool/{goos}_{goarch}/compile*",
+            "pkg/tool/{goos}_{goarch}/cover*",
+            "pkg/tool/{goos}_{goarch}/link*",
+        ],
+        allow_empty = True,
+    ),
+)
```
sluongng: I don't think this is used in the current PR?

Perhaps remove and re-add when it's actually used, or add a comment explaining the use case.

go/private/BUILD.sdk.bazel
```diff
@@ -1,9 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_sdk")
-load("@io_bazel_rules_go//go/private:common.bzl", "RULES_GO_STDLIB_PREFIX")
-load("@io_bazel_rules_go//go/private:go_toolchain.bzl", "declare_go_toolchains")
-load("@io_bazel_rules_go//go/private/rules:binary.bzl", "go_tool_binary")
-load("@io_bazel_rules_go//go/private/rules:sdk.bzl", "package_list")
-load("@io_bazel_rules_go//go/private/rules:transition.bzl", "non_go_reset_target")
+load("@io_bazel_rules_go//go/private:sdk_build_defs.bzl", "STDLIB_SRCS_EXCLUDE", "define_sdk_repository_targets")
```
sluongng, line 1: For easier reviewing, I think this refactoring could be done in a smaller, separate PR.

go/private/experimental/bootstrap_rules.bzl
```diff
@@ -0,0 +1,280 @@
...
+    CGO_ENABLED=0 \
+    GO111MODULE=off \
+    GOENV=off \
+    GOTELEMETRY=off \
+    GOTOOLCHAIN=local \
+    GOROOT_BOOTSTRAP="$BOOTSTRAP_ROOT" \
+    GOEXPERIMENT="$EXPERIMENTS" \
+    ./make.bash
+  fi
+)
```
sluongng, line 149: I think this could be out of date

See: golang.org/s/go15bootstrap and https://github.com/golang/go/issues/54265

The new official way to bootstrap Go is by using https://github.com/golang/go/tree/master/src/cmd/dist

This is OK for experimental usage, but I do think trying to do this correctly with `cmd/dist` would be better.

go/private/extensions.bzl
```diff
@@ -255,28 +256,22 @@ def _go_sdk_impl(ctx):
                 sdk_repo = name,
                 sdk_type = "remote",
                 sdk_version = wrap_tag.version,
+                sdk_source = "prebuilt",
```
sluongng: These "prebuilt" and "bootstrapped" values seem quite important to be left as magic strings? Move to a set of constants, perhaps (or struct fields)?

.bazelrc
```diff
@@ -33,6 +33,8 @@ build:ci --spawn_strategy=standalone
 build:ci --genrule_strategy=standalone
 test:ci --test_strategy=standalone
 
+common:bootstrap --//go/toolchain:source=bootstrapped
```
sluongng: Let's create a dedicated test for this flag instead of setting it in the default .bazelrc.

MODULE.bazel
```diff
@@ -64,6 +64,11 @@ dev_go_sdk.download(
     name = "rules_go_internal_compatibility_sdk",
     version = "1.22.12",
 )
+dev_go_sdk.from_file(
+    name = "go_bootstrap_sdk",
+    go_mod = "//:go.mod",
+    experimental_build_compiler_from_source = True,
+)
```
sluongng: Ditto: we should just move this into a test.

go/toolchain/BUILD.bazel
```diff
@@ -19,6 +19,29 @@ string_flag(
...
 )
 
+string_flag(
+    name = "source",
+    build_setting_default = "prebuilt",
+    values = [
+        "bootstrapped",
+        "prebuilt",
+    ],
+)
```
sluongng: Might need an additional document for the flag.

We should also rename it to add experimental_ prefix to avoid any confusion about support status.

dzbarsky:
I'm happy to spend some cycles cleaning up the PR a bit, breaking it up, etc, if we get alignment from rules_go maintainers to land, with the following caveat: clearly labeled experimental, no commitment to maintain it in future (though I personally likely will). I agree with @sluongng that it is better for the community to have a centralized way to do this than the same capabilities hiding fragmented in private repos :)

If we don't see a future where this gets merged, then there is no point breaking this up, I'll just keep this PR up as-is so we have an easy `archive_override` stanza for folks to copy when they want to use our new LLVM toolchain, which is quite soon going to ship the 1.0. I think it's a little goofy to have our README point at an "official rules_go override to make cgo work" but it's not the end of the world

jayconrod:
> I think (1) is already happening today. The problem with it is the lack of transparency and the increased cost of maintenance pushed down to the end user. Instead of just reviewing a set of patches applied to the SDK, the user have to (a) create and apply those patches, (b) maintain a separate build and release process for the new sdk distribution, (c) storage, checksum of the released sdk archive.
>
> (2) is essentially the same as using the current code, but moving it outside of rules_go: fragmentation, harder to discover, and potential breakages over time, as the API inside rules_go may change.
>
> I think your concern is valid, but the tradeoffs of the alternative options are not worth that much. And I suspect the maintenance cost can be lowered once we add tests to cover this logic.
> 
> The current implementation has already moved this to an experimental_build_compiler_from_source attribute to warn users about the feature status, so I'm in favor of having this inside rules_go for a while before we decide to yank it.

Fair points. If you guys are willing to maintain this, I won't block it. But I expect it will be a lot of work.

However we end up doing this, it'd be best if it's decoupled from the normal way of building as much as possible.

> @sluongng  The current implementation has already moved this to an experimental_build_compiler_from_source attribute to warn users about the feature status, so I'm in favor of having this inside rules_go for a while before we decide to yank it.
>
> At the very least, we know that there are real use-cases this helps unlock (i.e., changing how the go linker works with the new cc toolchain)

> @dzbarsky If we don't see a future where this gets merged, then there is no point breaking this up, I'll just keep this PR up as-is so we have an easy archive_override stanza for folks to copy when they want to use our new LLVM toolchain, which is quite soon going to ship the 1.0. I think it's a little goofy to have our README point at an "official rules_go override to make cgo work" but it's not the end of the world

I feel like I'm missing some context. Do we expect a lot of people to need to maintain long-lived forks of Go? Are there changes needed to make cgo work that can't be upstreamed? @dzbarsky I remember your linker flag change that was slowed down because of the freeze, but that should eventually merge, right?

dzbarsky:
> Fair points. If you guys are willing to maintain this, I won't block it. But I expect it will be a lot of work.

I'm happy to sign up to either keep it working or kill it and take the heat :) @sluongng you down to help?

> However we end up doing this, it'd be best if it's decoupled from the normal way of building as much as possible.
> 
> > @sluongng  The current implementation has already moved this to an experimental_build_compiler_from_source attribute to warn users about the feature status, so I'm in favor of having this inside rules_go for a while before we decide to yank it.
> > At the very least, we know that there are real use-cases this helps unlock (i.e., changing how the go linker works with the new cc toolchain)
> 
> > @dzbarsky If we don't see a future where this gets merged, then there is no point breaking this up, I'll just keep this PR up as-is so we have an easy archive_override stanza for folks to copy when they want to use our new LLVM toolchain, which is quite soon going to ship the 1.0. I think it's a little goofy to have our README point at an "official rules_go override to make cgo work" but it's not the end of the world
> 
> I feel like I'm missing some context. Do we expect a lot of people to need to maintain long-lived forks of Go? Are there changes needed to make cgo work that can't be upstreamed? @dzbarsky I remember your linker flag change that was slowed down because of the freeze, but that should eventually merge, right?

Yeah it will be released in Go 1.27 in 6 months. We would like to ship a 1.0 of the new LLVM toolchain way before that and would like the story to be "You can make it work with cgo via the following MODULE.bazel snippet" vs "You need to manually rebuild go compiler from source or use one of our prebuilts", it's way cleaner that way. But even beyond this specific use-case, I think it's really nice to be able to easily pull in compiler patches from upstream Go and test them on your own codebase, all within a single bazel invocation. Even something as simple as validating that an upstream fix works for you, etc.

sluongng:
Yeah I don't mind helping out. I was able to read and review the PR so I think maintaining it is ok.

Though as I mentioned in the review. The current bootstrapping process should switch to cmd/dist instead of relying on the bash/bat script. And I am not entirely sure what's the exact way to build cmd/dist to do that just yet. 

I think as long as we stay as consistent as possible with how upstream handle the bootstrap process and document the deltas, add tests, then we can keep the maintenance cost to minimal.

go/private/experimental/bootstrap_rules.bzl
sluongng: Ok it seems like the official doc is pointing to `make.bash` and `make.bat` still

https://go.dev/doc/install/source

Underneath these scripts are cmd/dist, so we should be safe relying on the script instead of the tool directly.

I added a test to check for the reproducibility of these scripts just to check for regression.

go/toolchains.rst
```diff
@@ -210,6 +215,61 @@ temporary limitation that will be removed in the future.
...
+
+    # MODULE.bazel
+    go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
+
+    go_sdk.download(
+        name = "go_sdk",
+        version = "1.23.5",
+        patches = ["//:my_go_patch.diff"],
+        patch_strip = 1,
+        experimental_build_compiler_from_source = True,
```
fmeum, line 265: Could we also use this longer attribute name for `go_download_sdk`? It seems weird to have different names for Bzlmod/non-Bzlmod builds.

go/toolchains.rst
```diff
@@ -210,6 +215,61 @@ temporary limitation that will be removed in the future.
...
+    )
+
+    use_repo(go_sdk, "go_sdk")
+
+Build with
+``--@io_bazel_rules_go//go/toolchain:experimental_source=bootstrapped`` to use
```
fmeum: Same as above, let's use the same name for the build setting.

On second thought, do we even need the build setting or could we just use the new attribute on the SDK to toggle this?

go/private/experimental/bootstrap_rules.bzl
```diff
@@ -0,0 +1,291 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def _experimental_bootstrap_go_sdk_impl(ctx):
+    windows_constraint = ctx.attr._windows_constraint[platform_common.ConstraintValueInfo]
+    is_windows_exec = ctx.target_platform_has_constraint(windows_constraint)
```
fmeum: This seems wrong unless we somehow enforce that target == exec.

go/private/experimental/bootstrap_rules.bzl
```diff
@@ -0,0 +1,291 @@
...
+    sh_toolchain = ctx.toolchains["@rules_shell//shell:toolchain_type"]
+    if not sh_toolchain or not sh_toolchain.path:
+        fail("Go bootstrap SDK requires @rules_shell//shell:toolchain_type with a configured shell path")
+
+    root_file = ctx.actions.declare_file(ctx.label.name + "/ROOT")
+    version = ctx.actions.declare_file(ctx.label.name + "/VERSION")
+    go_env = ctx.actions.declare_file(ctx.label.name + "/go.env")
+    go = ctx.actions.declare_file(ctx.label.name + "/bin/go" + (".exe" if ctx.attr.goos == "windows" else ""))
```
fmeum: It's a bit confusing that we match logic based on `ctx.attr.goos == "windows"` with logic based on `windows_constraint`. If we want to support cross-bootstrapping, we probably need to be more careful, if we don't want to do that, we can most likely get away with a single "Windows indicator".

go/private/experimental/bootstrap_rules.bzl
```diff
@@ -0,0 +1,291 @@
...
+    args.add(version)
+    args.add(go_env)
+    args.add(go)
+    args.add(gofmt)
+    args.add_all([srcs], expand_directories = False)
+    args.add_all([libs], expand_directories = False)
+    args.add_all([headers], expand_directories = False)
+    args.add_all([tools], expand_directories = False)
+    args.add_all([lib_misc], expand_directories = False)
+    args.add(",".join(ctx.attr.experiments))
```
fmeum: nit:
```suggestion
    args.add_joined(ctx.attr.experiments, join_with = ",")
```

go/private/experimental/bootstrap_rules.bzl
```diff
@@ -0,0 +1,291 @@
...
+    args.add(ctx.attr.goos + "_" + ctx.attr.goarch)
+    args.add("1" if is_windows_exec else "0")
+
+    ctx.actions.write(
+        output = bootstrap_script,
+        content = """set -euo pipefail
+
+MAKE_BASH="$1"
```
fmeum, line 71: Is this fully original or can we cite sources here?

dzbarsky: cited above

go/private/experimental/bootstrap_rules.bzl
dzbarsky: also needed an `omit_if_empty = False` :)

go/private/experimental/bootstrap_rules.bzl
dzbarsky: want to take a look at what we swapped to?

go/toolchains.rst
dzbarsky: yeah, you're right, we can simplify

go/private/experimental/bootstrap_rules.bzl
dzbarsky: yeah you're right, i'll pin this via exec_compatible_with. Cross-bootstrapping it doesn't seem that helpful since you're typically building for the platform you'll be using for the majority of execution anyway

fmeum (approved):

PR #4305, opened 2025-03-29
Add `go_sdk.from_file` to read the SDK version from `go.mod`.

jaqx0r:
**What type of PR is this?**

Feature

**What does this PR do? Why is it needed?**

Adds a `go_sdk.from_file` rule to the extension that allows modules to say that their SDK version should be read from their `go.mod` file.  This allows callers to avoid repeating themselves and stay within the Go tool environment for version specification, allowing better integration with other tools.

**Which issues(s) does this PR fix?**

Fixes #4292

**Other notes for review**

The toolchain version is selected based on the rules in https://go.dev/doc/toolchain

This change copies parts of the `go_mod.bzl` parser from the Gazelle repository:
https://github.com/bazel-contrib/bazel-gazelle/blob/master/internal/bzlmod/go_mod.bzl

A test based on the `go_download_sdk` test sets up a module-under-test with a `go.mod` and
checks that the version specified in it is used to run the test in that repo-under-test.

PR #4555, opened 2026-02-11
Also allow to read the SDK version from `go.work`

### LINKED ISSUE #4554: Add support for `go.work` to `go_sdk.from_file`
tl;dr: in a multi-module monorepo where `go.work` defines the prescribed development SDK, Bazel should be provided a way to derive the Go SDK version from `go.work` to ensure toolchain consistency across build systems and developer environments.

Context: `bazel`ifying a large monorepo, where:
- multiple `go.mod` files exist across the tree, most of them declare an older `go` version for compatibility reasons,
- developers use different IDEs and workflows,
- the repository already uses `go.work` to prescribe the Go SDK version for workspace operations.

In this setup, `go.work` effectively acts as the repository-level development contract: it defines the Go SDK that developers use locally and that some IDEs detect automatically.

`rules_go` currently allows to select the Go SDK based on a specific `go.mod`, but this creates a mismatch in the above context:
- the Go CLI (via `go.work`) may use Go `1.25.6`,
- Bazel (via `go.mod`'s) selects Go `1.25.0`,
- developers and CI may unknowingly build with different SDK versions,
- reproducibility and alignment across tools are weakened.

To avoid this divergence, it would be nice `rules_go` supports selecting the Go SDK version from `go.work`.

Proposed implementation:
- #4555.

rdesgroppes:
**What type of PR is this?**

Feature

Extends `go_sdk.from_file` (#4305) to accept a `go_work` attribute alongside the existing `go_mod` attribute. This allows modules to read their SDK version from a `go.work` file, providing better integration for workspace-based Go projects and maintaining consistency with the existing `go.mod` support.

Version extraction follows the same rules as `go.mod`: the `toolchain` directive takes precedence over the `go` directive.

Fixes #4554.

**Other notes for review**

When the go line is omitted, `go.mod` has an implicit `go 1.16` line while `go.work` has an implicit `go 1.18` line, as per the Go toolchain documentation at https://go.dev/doc/toolchain#config.

The implementation reuses the existing parser by extracting common logic into a shared `_version_from_go_file` function that both `version_from_go_mod` and the new `version_from_go_work` call with different default versions.

Tests follow the same pattern as `from_go_mod_file_test`, covering toolchain precedence, minor version support, go-only directive, and missing directives.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -34,6 +34,7 @@ tasks:
       # TODO: Bzlmod setup requires at least Bazel 7.
       - "-//tests/core/cross:proto_test"
       - "-//tests/core/from_go_mod_file:from_go_mod_file_test"
+      - "-//tests/core/from_go_work_file:from_go_work_file_test"
       - "-//tests/core/transition:hermeticity_test"
       - "-//tests/integration/gazelle:gazelle_test"
       # Requires https://github.com/bazelbuild/bazel/commit/ceddfb1ece1f8ed7ff81558fa1751e6526df031b.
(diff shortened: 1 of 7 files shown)


rdesgroppes:
Gern geschehen!

_(you're welcome)_

PR #4562, opened 2026-03-01
Allow go_binary to override go_library x_defs

### LINKED ISSUE #4561: xdefs: allow go_binary override of go_library values
Per the [docs on stamping](https://github.com/bazel-contrib/rules_go/blob/master/docs/go/core/defines_and_stamping.md):

> You can also override stamp values from libraries using x_defs on the go_binary rule if needed.

This behavior makes sense to me but does not match the [current implementation](https://github.com/bazel-contrib/rules_go/blob/master/go/private/actions/archive.bzl#L209) which instead first uses the x_defs from the go_binary and then (potentially) overwrites those with x_defs from the go_library dependencies.

```
    x_defs = dict(source.x_defs)
    for a in direct:
        x_defs.update(a.x_defs)
```

Fixing this seems easy enough (if this behavior is truly desired by upstream):

```
    x_defs = {}
    for a in direct:
        x_defs.update(a.x_defs)
    x_defs.update(source.x_defs)
```

Can you please clarify the intended behavior here?

hochhaus:
**What type of PR is this?**

Bug fix

Fixes #4561

**Other notes for review**

x_defs from dependencies were incorrectly applied after the source's x_defs, causing library values to overwrite binary values. Reverse the order so that source x_defs are applied last. This behavior matches the documentation which states:

  https://github.com/bazel-contrib/rules_go/blob/master/docs/go/core/defines_and_stamping.md

  > You can also override stamp values from libraries using x_defs on
  > the go_binary rule if needed.

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -206,9 +206,10 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         _nogo_diagnostics = out_diagnostics,
         _cgo_deps = cgo_deps,
     )
-    x_defs = dict(source.x_defs)
+    x_defs = {}
     for a in direct:
         x_defs.update(a.x_defs)
+    x_defs.update(source.x_defs)
 
     # Ensure that the _cgo_export.h of the current target comes first when cgo_exports is iterated
     # by prepending it and specifying the order explicitly. This is required as the CcInfo attached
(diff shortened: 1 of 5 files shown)


fmeum:
Thanks!

hochhaus:
Thanks @fmeum.

It looks like the build is waiting on a windows agent for ~20 hours.  I'm not too familiar with Buildkite. Does anything manual need to happen or will the Windows builds complete eventually on their own?

https://buildkite.com/bazel/rules-go-golang/builds/8571/steps/canvas

PR #4564, opened 2026-03-12
nogo: plumb SDK Go version into go/types

sluongng:
Forward the selected SDK Go version from compilepkg into nogo and initialize the corresponding go/types version plumbing during type checking. This keeps nogo aligned with the active toolchain version without keying the action on unrelated gcflags.

Keep gc_goopts -lang overrides unsupported for now. RunNogo still forwards only the SDK version, and compilepkg now documents that -lang parsing may be added later once nogo has proper gc flag parsing.

Add focused nogo coverage that checks both the RunNogo action

### DIFF
--- a/go/tools/builders/generate_test_main.go
+++ b/go/tools/builders/generate_test_main.go
@@ -436,6 +436,9 @@ func genTestMain(args []string) error {
 		cases.Imports = append(cases.Imports, importMap[name])
 	}
 	sort.Slice(cases.Imports, func(i, j int) bool {
+		if cases.Imports[i].Name == cases.Imports[j].Name {
+			return cases.Imports[i].Path < cases.Imports[j].Path
+		}
 		return cases.Imports[i].Name < cases.Imports[j].Name
 	})
 	tpl := template.Must(template.New("source").Parse(testMainTpl))


sluongng:
I played around with supporting `-lang` for a bit and found the effort to be not worth it (need to filter gc_goopts for -lang values in starlark, then passed it down, etc...)

If somebody found a use case for it in the future, feel free to add.

go/private/actions/compilepkg.bzl
```diff
@@ -216,6 +221,7 @@ def emit_compilepkg(
             sources = sources,
             cgo_go_srcs = cgo_go_srcs,
             archives = archives,
+            go_version = _sdk_go_version(go.sdk.version),
```
fmeum: This retains a new string per action. Either store the new string in the SDK providers or make the transformation lazy via `Args`.

sluongng: We can probably just pass down the value as-is and append the prefix with nogo main

sluongng:
@fmeum AI code review turned out to be pretty useful here. It called out for some edge cases that is required to support earlier Go versions, as well as some cases involve rc releases, or dev releases built from a commit hash. I think the attached tests act as pretty good documentation for those cases though.

PTAL

fmeum:
@sluongng I don't see those changes, did you push them?

sluongng:
Duh i pushed to the wrong branch. Give me a bit

jayconrod (approved):

Issue #4565, opened 2026-03-12
gopackagesdriver: packages.Load error: unexpected end of JSON input


jayconrod:
vscode-go reports the error `packages.Load error: unexpected end of JSON input`. This can happen for several reasons, and I'm trying to track at least a few of them down. Today, I saw this with a package that was incompatible with the current platform (according `target_compatible_with`). I've also seen this when editing a new file that hasn't been mentioned in the `BUILD` file yet.

It's really hard to diagnose the cause of these errors, given the inscrutable message. I added some logging, which I'll send in a PR.

The root cause seems to be Bazel failing on analysis with exit code 7. This is `ANALYSIS_FAILURE` in [ExitCode.java](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/ExitCode.java).

We should expect some Bazel errors since gopackagesdriver is often invoked while the user is editing code. The build is not expected to succeed. In [bazel.go](https://github.com/bazel-contrib/rules_go/blob/master/go/tools/gopackagesdriver/bazel.go#L121), we actually ignore exit code 1 and return partial results. We should ignore 7 as well.

PR #4566, opened 2026-03-12
gopackagesdriver: log events using slog

jayconrod:
**What type of PR is this?**

> Feature

When GOPACKAGESDRIVER_LOG is set, gopackagesdriver now logs events by appending JSON objects to the file. This is useful for finding arguments gopackagesdriver was invoked with, bazel commands it ran, bazel exit codes returned, and bazel stderr.

For #4565

**Other notes for review**

### DIFF
--- a/go/tools/gopackagesdriver/bazel.go
+++ b/go/tools/gopackagesdriver/bazel.go
@@ -113,7 +113,7 @@ func (b *Bazel) run(ctx context.Context, command string, args ...string) (string
 			exitCode = -1
 		}
 	}
-	slog.Info("bazel", "command", command, "args", cmd.Args, "exit_code", exitCode, "stderr", stderrBuf.String())
+	slog.Info("bazel", "command", command, "args", cmd.Args, "exit_code", exitCode, "stderr", stderrBuf)
 
 	return string(output), err
 }
--- a/go/tools/gopackagesdriver/main.go
+++ b/go/tools/gopackagesdriver/main.go
@@ -85,7 +85,7 @@ func run(ctx context.Context, in io.Reader, out io.Writer, args []string) (retEr
 		defer f.Close()
 		slog.SetDefault(slog.New(slog.NewJSONHandler(f, nil)).With("pid", os.Getpid()))
 	} else {
-		slog.SetDefault(slog.New(slog.NewJSONHandler(io.Discard, nil)))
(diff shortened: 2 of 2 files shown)


go/tools/gopackagesdriver/bazel.go
```diff
@@ -98,8 +100,21 @@ func (b *Bazel) run(ctx context.Context, command string, args ...string) (string
...
+
+	exitCode := 0
+	if err != nil {
+		if exitErr := (*exec.ExitError)(nil); errors.As(err, &exitErr) {
+			exitCode = exitErr.ExitCode()
+		} else {
+			exitCode = -1
+		}
+	}
+	slog.Info("bazel", "command", command, "args", cmd.Args, "exit_code", exitCode, "stderr", stderrBuf.String())
```
fmeum: It probably doesn't matter much here, but is there a pattern that avoids the `stderrBuf.String()` call when logging at info level isn't enabled (e.g. using a func).

go/tools/gopackagesdriver/main.go
```diff
@@ -73,14 +75,33 @@ var (
...
+	if logPath != "" {
+		f, err := os.OpenFile(logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error opening log file: %v\n", err)
+			os.Exit(1)
+		}
+		defer f.Close()
+		slog.SetDefault(slog.New(slog.NewJSONHandler(f, nil)).With("pid", os.Getpid()))
+	} else {
+		slog.SetDefault(slog.New(slog.NewJSONHandler(io.Discard, nil)))
```
fmeum: Is `io.Discard` special cased to make the log calls true noops? I would have expected a simpler way to disable this.

go/tools/gopackagesdriver/bazel.go
jayconrod: Yes, I should just drop the `.String()` call, done. The arguments are evaluated eagerly, but converted to strings lazily.

go/tools/gopackagesdriver/main.go
jayconrod: Wasn't obvious to me from the docs, but I found the proposal that added `slog.DiscardHandler`, which seems like what we want. It ignores output and is disabled at all levels, so values won't be converted to strings unnecessarily.

PR #4574, opened 2026-03-18
fix: avoid generating -Wl for dynamic libraries with an interface

thesayyn:
**What type of PR is this?**

> Bug fix

_cc_libs_and_flags only collected files from LibraryToLink.libraries and ignored
linker_input.additional_inputs, so any files provided via that field were never added to the GoCompilePkg or
GoLink sandboxes. Additionally, when a LibraryToLink has both interface_library and dynamic_library set, adding
the dynamic_library to libs caused it to flow into cgo_deps, which the GoLink rpath generation loop iterated
over — with many such libraries this generates enough -Wl,-rpath flags to exceed the OS argument limit. The fix
collects additional_inputs from each linker input and makes them available to the compile sandbox, while
dynamic_library from interface+dynamic pairs is tracked separately as link_inputs and threaded through GoArchive
to GoLink as plain sandbox inputs — bypassing rpath generation entirely.

Follow up to https://github.com/bazel-contrib/rules_go/pull/4569 which it incorrectly fixed. 

**Other notes for review**


fmeum (approved):

tests/core/cgo/BUILD.bazel
```diff
@@ -545,6 +545,12 @@ go_test(
     pure = "on",
 )
 
+go_bazel_test(
```
jayconrod: Could this instead be a set of targets in a separate directory? `go_bazel_test` is pretty expensive to run. It's useful for verifying failures, but it looks like both builds here are expected to succeed.

jayconrod (commented):
One small comment on the test, LGTM though.

tests/core/cgo/BUILD.bazel
thesayyn: fixed

thesayyn:
This is now ready.

fmeum (approved):

PR #4579, opened 2026-03-22
Use rules_cc to load cc symbols

dzbarsky:
**What type of PR is this?**
Refactor/cleanup

These should disappear from @bazel_tools eventually

Fixes #

**Other notes for review**


fmeum (approved):

wrayste:
I hit what looks like the same Bazel 9 breakage downstream in a repo that pulled `rules_go@0.50.1` transitively through the `aws_sdk` BCR module.

What fixed it locally for me was exactly the two pieces reflected here:

1. load `CcInfo` from `@rules_cc//cc/common:cc_info.bzl` instead of the old location
2. add an explicit `bazel_dep(name = "rules_cc", version = "0.2.18")` in `rules_go`'s `MODULE.bazel`

In my local patch, I had to touch these files:

- `MODULE.bazel`
- `go/private/context.bzl`
- `go/private/rules/binary.bzl`
- `go/private/rules/cgo.bzl`
- `go/private/rules/cross.bzl`

Without that patch set, analysis failed under Bazel 9; with it, the build progressed normally and let the rest of the dependency graph resolve.

So from my side, this PR looks directly relevant, and I can confirm that the `rules_cc` module dependency is part of the fix, not just the `CcInfo` load-site updates.

Environment for the repro:
- Bazel `9.1.0`
- `rules_go` `0.50.1`
- failure surfaced from a normal `bazel build //...` after introducing the `aws_sdk` BCR module

If helpful, I can also post the exact local patch diff I used.

dzbarsky:
I can try bumping the module but the failures only occur on bazel 7.7 so I'm not sure if it will help

PR #4583, opened 2026-04-05
fix(cgo): Handle headers in external repositories from cc_imports

blorente:
**What type of PR is this?**

Bug fix

When depending on `cc_import` targets from external repositories, their `includes` attribute doesn't get resolved to execroot-relative paths. This means that we end up having `-Imy/includes/dir` instead of `-Iexternal/external_repo_name/my/includes/dir`, which leads to missing headers during compilation.

As a consequence, the `-dev` packages generated by `rules_distroless` were not being imported correctly, and we had to manually insert `-isystem` flags with canonical repository names.

Fixes #

**Other notes for review**
 This PR doesn't really depend on #4574, but both are needed for the project I'm working on and I'd rather that one merges first if possible, as it doesn't have a workaround.

### DIFF
--- a/tests/core/cgo/BUILD.bazel
+++ b/tests/core/cgo/BUILD.bazel
@@ -546,6 +546,12 @@ go_test(
 )
 
 go_bazel_test(
+    name = "cc_header_inputs_test",
+    srcs = ["cc_header_inputs_test.go"],
+    target_compatible_with = ["@platforms//os:linux"],
+)
+
+go_bazel_test(
     name = "wrapped_cgo_test",
     srcs = ["wrapped_cgo_test.go"],
     target_compatible_with = select({
@@ -556,3 +562,4 @@ go_bazel_test(
         "//conditions:default": ["@platforms//:incompatible"],
     }),
 )
+
--- a/go/private/rules/cgo.bzl
+++ b/go/private/rules/cgo.bzl
@@ -120,14 +120,44 @@ def cgo_configure(go, srcs, cdeps, cppopts, copts, cxxopts, clinkopts):
             cc_defines = d[CcInfo].compilation_context.defines.to_list()
             cppopts.extend(["-D" + define for define in cc_defines])
             cc_includes = d[CcInfo].compilation_context.includes.to_list()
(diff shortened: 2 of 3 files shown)


tests/core/cgo/BUILD.bazel
```diff
@@ -545,6 +545,12 @@ go_test(
     pure = "on",
 )
 
+go_bazel_test(
```
blorente, line 548: I'm happy to move this somewhere else, but I'm not quite sure where the best place would be. Any suggestions?

fmeum:
@keith Do you see a less manual way to produce the correct paths?

keith:
fwiw i fixed the cc_import not resolving includes like cc_library in bazel 9.x https://github.com/bazelbuild/bazel/commit/d948e4dd24445c6240efb793b6e6f381eda2401b

blorente:
@keith that is so cool! I think that change makes this one unnecessary for Bazel 9.x because the external includes will be resolved correctly. Since it's also in `rules_cc`, I guess we should be able to just update rules_cc and consume it? I'll do some experimenting. 

However, that would mean bumping the `rules_cc` requirement in `rules_go` to at least `0.2.0`, I think ([link](https://github.com/bazelbuild/rules_cc/commit/8e0a527eab2645a7ccd4d585b747779caacbc0b6)). @fmeum is that something we're comfortable doing in this repository, or are we trying to not to impose dependency versions if we can avoid it? I think we can always "vendor in" the change if we need to as an alternative.

fmeum:
No, we usually just bump dependencies when a feature in them could be useful.

blorente:
Sorry for the radio silence! I finally got some time to experiment with this.

I've tried upgrading rules_cc (but leaving the Bazel version the same), and the tests unfortunately do fail even on versions with @keith 's fixes.

Since the `.bazelversion` on the repo points to Bazel 7.7.0, I'm assuming we want to keep compat with older Bazels. While these problems would go away with Bazel 9, would it make sense to merge this for now, until we can remove support for older Bazels?

I'm also happy to reflect this in the comments somewhere, or leave an issue open in the repo so that we don't forget. 

I've rebased the changes on `master` and CI is green, please let me know if there's anything else you'd like me to do.

PR #4586, opened 2026-04-13
bzltestutil: fix re-exec failure on Windows when runfiles path exceeds MAX_PATH

chouquette:
**What type of PR is this?**

Bug fix

Fix `go_test` failures when the path containing the test executable is longer than `MAX_PATH`.
The issue stems from `CreateProcess` which will systematically reject any working directory (whether passed explicitly or to be inherited) longer than `MAX_PATH`, regardless of longpath prefix (`\\?\`), long path setting being enabled in the registry, or `longPathAware` being present in the binary's manifest.

AFAIK, there is no open issue.

I initially reported this to go's tracker as https://github.com/golang/go/issues/78601, but in the context of `rules_go` I don't think there is one.

**Other notes for review**

I wasn't sure what to do about testing. AFAICS testing `Wrap` might quickly become over complicated, and testing the new condition in isolation doesn't bring much. Please let me know if you have something in mind here!

### DIFF
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -136,20 +136,18 @@ func Wrap(pkg string) error {
 	// will be killed by Bazel after the grace period (15s) expires.
 	signal.Ignore(syscall.SIGTERM)
 
-	wd, wdErr := os.Getwd()
-
 	cmd := exec.Command(exePath, args...)
 	cmd.Env = append(os.Environ(), "GO_TEST_WRAP=0")
 	// On Windows, any current directory value longer than MAX_PATH(260 chars)
 	// will cause CreateProcess to fail, regardless of LongPathsEnabled=1 or a
-	// longPathAware PE manifest.
-	// Inheriting the value from the parent process (by passing NULL as
-	// lpCurrentDirectory) will also fail.
-	// Setting cmd.Dir to a short path bypasses this.
-	// The child's chdir package init() will call os.Chdir to restore the correct
-	// runfiles directory after launch: os.Chdir (SetCurrentDirectoryW) respects
-	// Go's runtime PEB long-path bit
-	if runtime.GOOS == "windows" && wdErr == nil && len(wd) >= 260 {
+	// longPathAware PE manifest. Inheriting the value from the parent process
+	// (by passing NULL as lpCurrentDirectory) will also fail.
+	// Always set cmd.Dir to a short path so CreateProcess succeeds.
+	// Re-add GO_TEST_RUN_FROM_BAZEL so the child's chdir package init() will
+	// call os.Chdir to restore the correct runfiles directory after launch.
+	// os.Chdir (SetCurrentDirectoryW) respects Go's runtime PEB long-path bit,
+	// so it handles paths longer than MAX_PATH.
+	if runtime.GOOS == "windows" {
 		cmd.Dir = os.TempDir()
 		cmd.Env = append(cmd.Env, "GO_TEST_RUN_FROM_BAZEL=1")
 	}
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -151,6 +151,7 @@ func Wrap(pkg string) error {
 	// Go's runtime PEB long-path bit
 	if runtime.GOOS == "windows" && wdErr == nil && len(wd) >= 260 {
 		cmd.Dir = os.TempDir()
+		cmd.Env = append(cmd.Env, "GO_TEST_RUN_FROM_BAZEL=1")
 	}
 	cmd.Stderr = io.MultiWriter(os.Stderr, streamMerger.ErrW)
 	cmd.Stdout = io.MultiWriter(os.Stdout, streamMerger.OutW)
--- a/go/tools/bzltestutil/wrap.go
+++ b/go/tools/bzltestutil/wrap.go
@@ -25,6 +25,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


go/tools/bzltestutil/wrap.go
```diff
@@ -135,8 +136,22 @@ func Wrap(pkg string) error {
 	// will be killed by Bazel after the grace period (15s) expires.
 	signal.Ignore(syscall.SIGTERM)
 
+	wd, wdErr := os.Getwd()
+
 	cmd := exec.Command(exePath, args...)
 	cmd.Env = append(os.Environ(), "GO_TEST_WRAP=0")
+	// On Windows, any current directory value longer than MAX_PATH(260 chars)
+	// will cau/ CreateProcess to fail, regardless of LongPathsEnabled=1 or a
```
JSGette:
```suggestion
	// will cause CreateProcess to fail, regardless of LongPathsEnabled=1 or a
```

go/tools/bzltestutil/wrap.go
```diff
@@ -135,8 +136,22 @@ func Wrap(pkg string) error {
...
+	// On Windows, any current directory value longer than MAX_PATH(260 chars)
+	// will cause CreateProcess to fail, regardless of LongPathsEnabled=1 or a
+	// longPathAware PE manifest.
+	// Inheriting the value from the parent process (by passing NULL as
+	// lpCurrentDirectory) will also fail.
+	// Setting cmd.Dir to a short path bypasses this.
+	// The child's chdir package init() will call os.Chdir to restore the correct
+	// runfiles directory after launch: os.Chdir (SetCurrentDirectoryW) respects
+	// Go's runtime PEB long-path bit
+	if runtime.GOOS == "windows" && wdErr == nil && len(wd) >= 260 {
```
fmeum: Would we lose anything if we dropped the last condition? That would make it more likely that bugs in this area are found.

chouquette: No, all in all the code guarded by the condition is a no-op AFAICT, so it could just be executed unconditionally

chouquette: I think I spoke too fast. The child process didn't always run `os.Chdir`, that was guarded by the `GO_TEST_RUN_FROM_BAZEL`  env variable.

I explicitly added that variable to the child env to ensure it will change to the expected working directory, and removed the conditional assignment of `cmd.Dir`

Please let me know if that matches what you had in mind!

jayconrod (commented):
I'm not sure I understand this change, and there are a couple things that make me nervous about it.

First, how does it work? This code is in a wrapper that's linked into the same binary. If the executable path or the working directory path is too long for the child process, how could the current process have been started successfully with the same executable and working directory?

Second, this introduces unpredictable behavior. This causes the test code to execute in a different directory, and it may not be at all obvious to the user when that's happening or why. This would break runfiles when `--enable_runfiles` is used (which I think it always is for remote execution). I'm not sure if it breaks coverage recording or additional test log files (those found in `bazel-testlogs/.../test.outputs`).

I think it would be better to fail with a clear message, explaining what the user can do to fix it. Generally that means setting `--output_base` to something short like `C:\b`, or shortening directory paths within the repo. Windows has terrible error messages here: it often implies the file doesn't exist.

You mentioned in https://github.com/golang/go/issues/78601 that no change was needed when the mingw-x64 external linker was used. Should that be used here too?

chouquette:
> I'm not sure I understand this change, and there are a couple things that make me nervous about it.
> 
> First, how does it work? This code is in a wrapper that's linked into the same binary. If the executable path or the working directory path is too long for the child process, how could the current process have been started successfully with the same executable and working directory?

I don't think I can fully answer that question with certitude, I don't know how bazel spawns its test processes.

 What I do know for sure is that `CreateProcess` can't handle long path for the CWD. Passing `NULL` to inherit the value will fail, passing a long folder explicitly will fail, and that is true regardless of the binary manifest enabling `longPathAware`, the `PEB` flag, or long path being enabled through the registry.

Since the wrapper process is able to start, it must mean that it was started in a shorter path and only then moved to another directory.

> 
> Second, this introduces unpredictable behavior. This causes the test code to execute in a different directory, and it may not be at all obvious to the user when that's happening or why. This would break runfiles when `--enable_runfiles` is used (which I think it always is for remote execution). I'm not sure if it breaks coverage recording or additional test log files (those found in `bazel-testlogs/.../test.outputs`).
> 

I think that's only true for the very first instructions, with `GO_TEST_RUN_FROM_BAZEL=1` being passed unconditionally, one of the very first things the program will do is to invoke `chdir.init()` which will in turn call `SetCurrentDirectory` to move to the expected working directory.

I agree there's a window in which the program runs in an unexpected directory though. I'm not familiar enough with the code to swear that's innocuous, but it felt like an acceptable trade off: no test code outside of the `init()` functions will run in the "incorrect but short" directory.

> I think it would be better to fail with a clear message, explaining what the user can do to fix it. 

I disagree, improving the error message would be beneficial, but in this specific case, we just can't run our tests, which is not a viable outcome IMHO.

> Generally that means setting `--output_base` to something short like `C:\b`, or shortening directory paths within the repo. Windows has terrible error messages here: it often implies the file doesn't exist.
> 

Shortening the output_base might work but only temporarily, it's fairly easy to exceed the 260 chars limit with nested packages. If the build includes some transitions, an additional suffix will be added to the CWD which increases the likelyhood of a failure.
TL;DR I don't think shortening the path on the project side can work in the long term.


> You mentioned in [golang/go#78601](https://github.com/golang/go/issues/78601) that no change was needed when the mingw-x64 external linker was used. Should that be used here too?

That was for added context, but as it turns out, the manifest doesn't impact CreateProcess handling of long paths, so the linker can be ignored altogether in this context.

jayconrod (approved):
Hmm, you're right, I'd forgotten about that call to `chdir.init()`. That happens early, I think it's meant to happen before any user code runs, or else tests would notice how different Bazel is already from Go. The risk should be much lower than I was thinking.

Let's try this. I'd still love to know if Bazel has a trick for executing programs with long executable or working directory paths.

@fmeum Any other comments? I think we can merge this.

fmeum:
I'm still surprised that there's no way to reliably launch a long path executable, but the fix looks reasonable to me.

fmeum:
@chouquette Have you tried the usual tricks of 1. prepending `\\?\` to the cwd or 2. using the Unicode version `CreateProcessW`?

chouquette:
> @chouquette Have you tried the usual tricks of 1. prepending `\\?\` to the cwd or 2. using the Unicode version `CreateProcessW`?

For the `\\?\` prefix, I did. Using it changes the error code from `ERROR_INVALID_PARAMETER` to `ERROR_INVALID_NAME`, but the process creation still fails.

About `CreateProcessW` my understanding is that go already uses the unicode variant of the WinAPI, but for the sake of checking, I tried to use it directly through `kernel32.NewProc("CreateProcessW")` and the results are identical.

I'm including the test sample I generated along with the exploration, it contains all the variant tested along with their results (named as .txt because github refuses to upload a .go files AFAICS)
[main.go.txt](https://github.com/user-attachments/files/26819191/main.go.txt)

fmeum (approved):

PR #4590, opened 2026-04-16
go/private: optimize _filter_options to avoid O(n²) scan

jsharpe:
Split the denylist into exact-match entries (dict lookup, O(1)) and prefix-match patterns (those ending in "=", e.g. "-fmax-errors="). 

Only the rare prefix patterns require a linear scan, eliminating the previous O(options × denylist) _match_option loop for the common case.

This provides a significant reduction in CPU time on large repos:
| | Before | After |
|--|--|--|
| `_filter_options` CPU | 123s | 0.004s |
| `_match_option` CPU | 56s | **0s** |

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -147,17 +147,16 @@ _UNSUPPORTED_FEATURES = [
     "rules_go_unsupported_feature",
 ]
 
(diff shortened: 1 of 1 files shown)


fmeum (approved):
Very nice!

PR #4591, opened 2026-04-16
context: prefer pre-computed CgoContextInfo from go_context_data

jsharpe:
Reorder the CgoContextInfo resolution in go_context() so that the pre-computed value from go_context_data is checked before falling back to cgo_context_data_impl(). 

This avoids re-running the expensive cgo_context_data_impl for every go_library target when a shared CgoContextInfo is already available from the cgo_context_data rule.

Perf improvements:
| | Before | After |
|--|--|--|
| `cgo_context_data_impl` CPU | 132s | **0.02s** |

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -552,10 +552,15 @@ def go_context(
         stdlib = go_context_data[GoStdLib]
         go_context_info = go_context_data[GoContextInfo]
 
-    if getattr(attr, "_cc_toolchain", None) and CPP_TOOLCHAIN_TYPE in ctx.toolchains:
-        cgo_context_info = cgo_context_data_impl(ctx)
-    elif go_context_data and CgoContextInfo in go_context_data:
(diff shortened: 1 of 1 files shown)


dzbarsky (approved):

fmeum:
I should have left a comment back then, but I think that this is an incorrect optimization: The Go toolchain and the C++ toolchain have to be resolved in the same exec group and the rule attributes for a `go_library` (most importantly `exex_compatible_with`) should affect the C++ toolchain selection, neither of which is the case with `cgo_context_info`.

dzbarsky:
> I should have left a comment back then, but I think that this is an incorrect optimization: The Go toolchain and the C++ toolchain have to be resolved in the same exec group and the rule attributes for a `go_library` (most importantly `exex_compatible_with`) should affect the C++ toolchain selection, neither of which is the case with `cgo_context_info`.

Ah yes that's right. Lets undo it and delete those bits?

PR #4593, opened 2026-04-18
perf: Only resolve cc toolchain for rules that will need it

dzbarsky:
This PR is a re-write of https://github.com/bazel-contrib/rules_go/pull/4591 which was incorrect - the cc toolchain must be resolved by the rule owning the action that uses it, it cannot be resolved by a dependent rule (CgoContextData) as that will not work correctly when there are multiple execution platforms. See https://bazelbuild.slack.com/archives/CDBP88Z0D/p1776380536174449 for discussion

### DIFF
--- a/go/private/context.bzl
+++ b/go/private/context.bzl
@@ -480,6 +480,31 @@ def _matches_scopes(label, scopes):
             return True
     return False
 
+def _go_infos_use_cgo(go_infos):
+    for go_info in go_infos:
+        if GoInfo in go_info and go_info[GoInfo].cgo:
+            return True
+    return False
+
+def _sources_use_cgo(attr, go_infos):
+    """Returns whether attr's sources may need cgo processing."""
+    if getattr(attr, "cgo", False):
+        return True
+    return _go_infos_use_cgo(getattr(attr, "embed", [])) or _go_infos_use_cgo(go_infos)
+
+def maybe_needs_cc_toolchain(attr, go_infos = []):
+    """Returns whether this rule's own sources may use the C/C++ toolchain."""
+    return _sources_use_cgo(attr, go_infos)
+
+def _precomputed_cgo_context_info(attr, go_context_data):
+    if go_context_data and CgoContextInfo in go_context_data:
+        return go_context_data[CgoContextInfo]
+    if getattr(attr, "_cgo_context_data", None) and CgoContextInfo in attr._cgo_context_data:
+        return attr._cgo_context_data[CgoContextInfo]
+    if getattr(attr, "cgo_context_data", None) and CgoContextInfo in attr.cgo_context_data:
+        return attr.cgo_context_data[CgoContextInfo]
+    return None
+
 def validate_nogo(go):
     """Whether nogo should be run as a validation action rather than just to generate fact files for the current
     target."""
@@ -522,6 +547,7 @@ def go_context(
         embed = None,
         importpath_aliases = None,
         go_context_data = None,
+        maybe_needs_cc_toolchain = True,
         goos = "auto",
         goarch = "auto"):
     """Returns an API used to build Go code.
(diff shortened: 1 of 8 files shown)


go/private/context.bzl
```diff
@@ -480,6 +480,31 @@ def _matches_scopes(label, scopes):
             return True
     return False
 
+def _go_infos_use_cgo(go_infos):
+    for go_info in go_infos:
+        if GoInfo in go_info and go_info[GoInfo].cgo:
+            return True
+    return False
+
+def _source_uses_cgo(attr, go_infos):
```
fmeum: Nit: `_sources_use_cgo` seems more fitting?

go/private/rules/library.bzl
```diff
@@ -27,6 +27,7 @@ load(
     "CGO_TOOLCHAINS",
     "go_context",
     "new_go_info",
+    rule_maybe_needs_cc_toolchain = "maybe_needs_cc_toolchain",
```
fmeum: Do we need this renaming?

fmeum (approved):
Nice, please mark as ready for review and add a link to the slack convo 

@jayconrod Could you also take a look?

go/private/rules/library.bzl
dzbarsky: ah i think an earlier version also has an intermediate var named `maybe_needs_cc_toolchain` so there was a collision, thanks, will clean

go/private/context.bzl
```diff
@@ -522,6 +547,7 @@ def go_context(
         embed = None,
         importpath_aliases = None,
         go_context_data = None,
+        maybe_needs_cc_toolchain = True,
```
jayconrod, line 550: I don't really understand what this is. I think it's saying whether the calling rule could depend on the C++ toolchain in order to compile C++ source. Could we infer that from `ctx` instead? It seems awkward to make so many rules specify this.

go/private/context.bzl
```diff
@@ -551,16 +577,26 @@ def go_context(
...
+            cgo_context_info = cgo_context_data_impl(ctx)
+        else:
+            cgo_context_info = _precomputed_cgo_context_info(attr, go_context_data)
+
+    if maybe_needs_cc_toolchain:
+        cgo_available = cgo_context_info != None
+    else:
+        # Preserve cgo build-mode/tag behavior for rules that won't use the
+        # C/C++ toolchain in their own actions.
+        cgo_available = not cgo_disabled and _precomputed_cgo_context_info(attr, go_context_data) != None
```
jayconrod, line 597: I'm kind of confused by all the branches and variables here. `cgo_disabled`, `maybe_needs_cc_toolchain`, and `cgo_available` all seem like they have overlapping purposes to me. Is there a way to consolidate this?

jayconrod (commented):
Couple small readability comments.

go/private/context.bzl
dzbarsky: yeah, agree it's a little awkward. I think it would be nice to split all the cc toolchain stuff out of this context and just invoke a separate `configure_cc_toolchain` function when needed, that would be a lot clearer in my mind, and that's generally what other rulesets that interop with cc (like Rust) do.

I'm not sure how to infer it from `ctx` - ultimately the rule is the only place that really knows what actions it intends to create and whether cgo will be involved. But perhaps I'm missing some idea you're thinking about?

go/private/context.bzl
dzbarsky: Agreed, I think we can take a stab at cleaning this as part of https://github.com/bazel-contrib/rules_go/pull/4599 since the precomputed cgo stuff is incorrect and should be removed anyway, are you cool with doing it as a followup?

go/private/context.bzl
jayconrod: So for example, could we infer that cgo is desired based on the presence of a C++ toolchain and a `pure` attribute (on the target or platform)? A rule like `go_proto_library` might never be able to use cgo, but `go_binary` would.

(Apologies if this doesn't make sense, it's been a while since I was really active in this code, and it's changed a lot. A lot of rules_go code was written before platforms and toolchains were introduced, certainly before optional toolchains, so there may be some elaborate workarounds that no longer make sense in modern versions of Bazel.)

go/private/context.bzl
jayconrod: Followup is good, whatever cleanup makes sense.

go/private/context.bzl
dzbarsky: something like that could potentially work. it's a little more complicated, you can also skip cc toolchain for a non-pure library as long as it doesn't have any `cgo` sources or `cdeps` (including embedded) but the generic `go_library` rule still needs a c++ toolchain on it. That case is the majority of the win this PR unblocks so we need to be a bit careful. But yes lets merge this, do further cleanups, and then once the bad codepaths are one figure out the shape that makes most sense. Definitely some room for simplifications nowadays I think!

PR #4594, opened 2026-04-19
Add C++ runtime libs to cgo action inputs

dzbarsky:
With @llvm toolchain, libc++ is provided via cc_toolchain's `static_runtime_library`/`dynamic_runtime_library`. These groups should be added to link actions otherwise cgo code that attempts to link libc++ will not work properly. It seems there were previous attempts to drop libc++ - I'm not sure the best way to harmonize these, some tweaking of this PR may be necessary. @jayconrod @fmeum wdyt?


go/private/rules/cgo.bzl
```diff
@@ -23,8 +23,19 @@ load(
...
 
+_CXX_SOURCE_EXTENSIONS = [
+    ".cc",
+    ".cpp",
+    ".cxx",
+    ".mm",
+]
+
+def _has_cxx_sources(srcs):
+    return any([src.path.endswith(ext) for src in srcs for ext in _CXX_SOURCE_EXTENSIONS])
```
fmeum: `File` has an `extension` field

go/private/context.bzl
```diff
@@ -505,6 +508,14 @@ default_go_config_info = GoConfigInfo(
     export_stdlib = False,
 )
 
+def _cc_runtime_libs_for_mode(mode, cgo_context_info):
+    if mode.linkmode in (LINKMODE_SHARED, LINKMODE_PLUGIN, LINKMODE_C_SHARED):
+        return cgo_context_info.cc_toolchain_dynamic_runtime_lib
+    return cgo_context_info.cc_toolchain_static_runtime_lib
+
+def _file_linkopts(files):
```
fmeum: Where is this used? For future path mapping efforts, it would be better to keep the `File`s.

fmeum (approved):

go/private/context.bzl
dzbarsky: reused ctx.actions.args

tyler-french (approved):

PR #4595, opened 2026-04-20
link: add external deps to Go build info

sluongng:
Binaries built with rules_go omit Go module metadata from `runtime/debug.ReadBuildInfo().Deps` because package metadata never reaches the link action. Dependency inspection is therefore incomplete for external and vendored packages.

Thread Gazelle-generated `PackageMetadataInfo` artifacts through `GoInfo` and `GoArchiveData`, add them as link inputs, and parse their `pkg:golang` PURLs in the builder. Emit the resulting modules as dep-only modinfo for supported build modes.

Package metadata is an action output, so parsing it at link time avoids trying to expose its JSON contents through Starlark providers. Read metadata from both `package_metadata` and `applicable_licenses` for Bazel compatibility, ignore non-Go PURLs, normalize unprefixed semantic versions, preserve opaque versions such as commit hashes, and represent versionless modules as `(devel)`, matching `cmd/go` behavior. Metadata propagation also tolerates exported `GoArchive` values created without the new private field.

Build information follows the declared Bazel dependency closure because that graph is available during analysis. It may therefore include metadata for a declared dependency whose Go package is not imported. Package metadata also need not describe one MVS-resolved graph, so distinct versions for the same module path are preserved rather than silently selecting one.

Target-owned metadata is not reported as a dependency for either `go_binary` or `go_test`; test binaries retain metadata from their declared dependency closure. Coverage and non-coverage links use the same metadata traversal while preserving the existing coverage archive deduplication required to avoid duplicate coverdata packages.

Rebase onto current `master`, including the Go 1.20 minimum-version change. Use Go 1.20-compatible `os` and `io` APIs throughout the touched code and keep metadata lookup compatible with Bazel 7 and Bazel 9. Replace the direct `rules_license` dependency and Stardoc patch with `package_metadata` for both Bzlmod and WORKSPACE users.

Validation:
- `bazel --ignore_all_rc_files test --test_output=errors //tests:buildifier_test //docs:docs_go_core_rules-difftest //tests/core/starlark:package_metadata_test //tests/core/starlark:applicable_licenses_test //tests/core/go_binary:buildinfo_test`
- `USE_BAZEL_VERSION=7.2.1 bazel --ignore_all_rc_files test --lockfile_mode=off --test_output=errors //tests/core/go_binary:buildinfo_test //tests/core/starlark:package_metadata_test //tests/core/starlark:applicable_licenses_test`
- `USE_BAZEL_VERSION=7.2.1 bazel --ignore_all_rc_files test --lockfile_mode=off --@io_bazel_rules_go//go/toolchain:sdk_version=1.20.14 //go/tools/builders:buildinfo_test`
- `bazel build //docs/... && bazel test //docs/...`
- From `tests/bcr`: `bazel build --allow_yanked_versions=all //... @go_default_sdk//... && bazel test --allow_yanked_versions=all //...`


darkrift:
@sluongng would you mind rebasing this on master and switch to [@package_metadata](https://github.com/bazel-contrib/supply-chain/tree/main/metadata) instead ? rules_license is being replaced altogether with this other module

go/tools/builders/buildinfo.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		if ok {
+			modules = append(modules, module)
+		}
+	}
+	return modules, nil
+}
+
+func moduleFromPURL(purl string) (moduleInfo, bool, error) {
+	const prefix = "pkg:golang/"
```
darkrift, line 114: why not using https://github.com/package-url/packageurl-go ?

sluongng: Because the “builder” binary is bootstrapped from vendored source. And i think maintaining 1 file like this will be much simpler than the to be vendored code

sluongng:
I also rebased the whole stack to this https://github.com/bazel-contrib/rules_go/compare/master...sluongng:rules_go:sluongng/vcs-buildinfo-stamping

we dont have stack pr setup so i will send the other commits separately after this one is merged

docs/BUILD.bazel
```diff
@@ -3,11 +3,13 @@ load("//docs:doc_helpers.bzl", "stardoc_with_diff_test", "update_docs")
 # For each doc file, generate MD from bzl_library, then perform diff test
 stardoc_with_diff_test(
     bzl_library_target = "//docs/go/extras:extras",
+    extra_deps = ["@package_metadata//providers:package_metadata_info.bzl"],
```
fmeum: This shouldn't be necessary if we add the correct deps to the `//docs` targets in the first place.

go/private/actions/link.bzl
```diff
@@ -125,7 +130,17 @@ def emit_link(
     else:
         arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])
 
-    builder_args.add_all(arcs, before_each = "-arc", map_each = _format_archive)
+    arc_list = arcs.to_list() if type(arcs) == "depset" else arcs
```
fmeum: This is unfortunately pretty likely to regress rules_go's peak memory footprint and analysis time performance. Could we use an aspect to collect the set of files in a `depset` or just do this in the same way archives are collected? 

I hope to add lazy depset transformations to Bazel in the near future, but until we have them, there doesn't seem to be any other (performant) way.

sluongng: what do you think about https://github.com/bazel-contrib/rules_go/pull/4595/changes/dea3ce4ec1ce6832cdb6d06fbd0ad00261167ed7 ?

docs/BUILD.bazel
sluongng: fixed

go/private/actions/archive.bzl
```diff
@@ -231,4 +237,6 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         cgo_exports = cgo_exports,
         runfiles = runfiles,
         _headers = headers,
+        _package_metadata = getattr(data, "_package_metadata", None),
```
fmeum: Do we even need this field anymore now that we track the files separately?

go/private/context.bzl
```diff
@@ -250,6 +251,19 @@ def _tool_args(go):
     args.use_param_file("-param=%s")
     return args
 
+def package_metadata_file_from_metadata(label, package_metadata = (), applicable_licenses = ()):
+    if not label.repo_name and not label.workspace_root:
```
fmeum: Wouldn't `not label.repo_name` be sufficient to exclude the main repo?

I'm also wondering whether that's the right condition. If folks vendor code in their tree, they would probably still want to include metadata for it that they add under third_party (gazelle doesn't have to be the only producer).

go/tools/builders/buildinfo.go
```diff
@@ -0,0 +1,204 @@
...
+	return deps
+}
+
+func modInfoData(modules []moduleInfo) string {
+	deps := buildInfoDeps(modules)
+	if len(deps) == 0 {
+		return buildInfoStart + buildInfoEnd
+	}
+
+	// debug.BuildInfo.String was added after Go 1.17. Emit the dep-only
```
fmeum: We don't need to support Go versions before 1.18. let's use the types if they make this easier to follow

go/tools/builders/buildinfo.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	major, err := strconv.Atoi(parts[0])
+	if err != nil {
+		return true
+	}
+	minorVersion, err := strconv.Atoi(minor)
+	if err != nil {
+		return true
+	}
+	return major > 1 || (major == 1 && minorVersion >= 18)
```
fmeum: Likewise, let's fold this to `true`

rdesgroppes:
> we dont have stack pr setup so i will send the other commits separately after this one is merged

Thanks for pushing this forward, @sluongng!
The [DataDog agent](https://github.com/DataDog/datadog-agent) has exactly this gap in its radar: `go_binary`-built binaries omitting module info that security scanners expect from `ReadBuildInfo()`.

The rebase onto `package_metadata` looks right given `rules_license`'s [deprecation](https://github.com/bazelbuild/rules_license#rules_license).

Happy to help (reviewing, contributing, testing against our setup) if that'd help get it over the line.

go/tools/builders/buildinfo.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type packageMetadata struct {
+	PURL string `json:"purl"`
+}
+
+func modulesFromPackageMetadataFiles(paths []string) ([]moduleInfo, error) {
+	modules := make([]moduleInfo, 0, len(paths))
+	for _, path := range paths {
+		data, err := ioutil.ReadFile(path)
```
rdesgroppes: nit: since an earlier thread above already established `rules_go` doesn't support Go < 1.18, we may consider using `os.ReadFile` (since 1.16) instead (the test file already uses `os.WriteFile`).

go/tools/builders/buildinfo.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		seen[module] = struct{}{}
+		unique = append(unique, module)
+	}
+
+	sort.Slice(unique, func(i, j int) bool {
+		if unique[i].path != unique[j].path {
+			return unique[i].path < unique[j].path
+		}
+		return unique[i].version < unique[j].version
```
rdesgroppes, line 179: Since dedup here is keyed on the full (path, version) pair, is it worth considering what happens if two `package_metadata` inputs end up reporting different versions for the same module path (e.g. a stale hand-authored metadata target)?
Reason why I ask is that `cmd/go`'s own build info never has two entries for the same path since MVS resolves to one version first, so I wasn't sure if that's an intentional simplification here or something worth guarding against.

go/private/actions/link.bzl
sluongng: Addressed by propagating package metadata as lazy depsets alongside `GoArchive` dependencies. Normal links no longer materialize the transitive archive closure; only the existing coverage compatibility path materializes archives. The final focused suite and reproducibility test pass on the rebased branch.

go/tools/builders/buildinfo.go
sluongng: Updated this to `os.ReadFile`. Since current `master` now requires Go 1.20, I also replaced direct `io/ioutil` uses in the other PR-touched Go files with their `os` or `io` equivalents and verified the builder test with Go 1.20.14.

go/tools/builders/buildinfo.go
sluongng: This is intentional because Bazel package metadata can be independently authored and is not guaranteed to describe one MVS-resolved graph. The builder now has an explicit comment documenting that choice, and `TestBuildInfoDepsSortAndDedup` covers two distinct versions of the same module path.

fmeum (commented):
A few comments on the `link.bzl` plumbing, this time coming from a Claude review that I filtered down to what seems actually relevant

go/private/actions/link.bzl
```diff
@@ -122,10 +122,30 @@ def emit_link(
                 continue
             importmaps[arc.importmap] = True
             arcs.append(arc)
+        package_metadata_files = depset([
```
fmeum: The coverage branch derives `package_metadata_files` from the importmap-deduplicated `arcs` list, while the non-coverage branch derives it from the un-deduplicated `_package_metadata_files` depsets. That lets `--collect_code_coverage` silently change the set of modules embedded in the binary: the dedup loop above keeps only the first archive per importmap, so two distinct `GoArchiveData` that share an importmap but carry different `_package_metadata` lose the second one's metadata under coverage and keep it otherwise.

Since we can drop pre-7.2.0 support, the cleanest fix is to apply the TODO above and delete the workaround outright — then there is a single branch and no divergence to keep in sync:

```starlark
    if go.coverage_enabled and go.coverdata:
        test_archives = list(test_archives) + [go.coverdata.data]
    arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])

    # This is the declared Bazel dependency closure, which is available
    # during analysis. It can include metadata for a declared dependency
    # whose Go package is not imported by the binary.
    package_metadata_files = depset(
        direct = [
            metadata
            for archive_data in test_archives
            for metadata in [getattr(archive_data, "_package_metadata", None)]
            if metadata
        ],
        transitive = [d._package_metadata_files for d in archive.direct],
    )
```

That also wants `MINIMUM_BAZEL_VERSION` in `go/private/common.bzl` bumped from `6.5.0` to `7.2.0`.

One note on the swap itself: the importmap loop collapses *distinct* `GoArchiveData` that happen to share an importmap, whereas depset dedup only collapses *value-identical* entries. For coverdata those coincide — it is literally the same object appearing twice — which is what makes the TODO safe, but it is the thing to double-check if anything else was relying on importmap-level collapsing in `arcs`.

go/private/actions/link.bzl
```diff
@@ -122,10 +122,30 @@ def emit_link(
...
     else:
         arcs = depset(test_archives, transitive = [d.transitive for d in archive.direct])
 
+        # This is the declared Bazel dependency closure, which is available
+        # during analysis. It can include metadata for a declared dependency
+        # whose Go package is not imported by the binary.
+        package_metadata_files = depset(
+            direct = [
+                metadata
+                for archive_data in test_archives
```
fmeum: `test.bzl:190` passes `test_archives = [internal_archive.data]`, so for `go_test` the library under test contributes its own module metadata as a `dep` entry. For `go_binary` the target's own metadata is excluded — `tests/core/go_binary/buildinfo_test.go` asserts that `example.com/main` is absent from `Deps` — so the same library reports differently depending on whether it is linked into a binary or into a test. Nothing covers the `go_test` case.

Worth deciding explicitly which of the two is intended. cmd/go's model is a third option: the target's own module becomes `BuildInfo.Main` and `path` rather than being either dropped or listed as a dep. That would resolve the asymmetry and also give `Main` a value — right now it is always the zero `Module`, which is what makes `debug.ReadBuildInfo()` return `ok=true` with an empty `Main` for consumers like `go version -m` and prometheus' `NewBuildInfoCollector`.

go/private/actions/link.bzl
```diff
@@ -122,10 +122,30 @@ def emit_link(
...
+                for metadata in [getattr(archive_data, "_package_metadata", None)]
+                if metadata
+            ],
+            transitive = [d._package_metadata_files for d in archive.direct],
```
fmeum: `d._package_metadata_files` is a direct field access, whereas every other new access in this PR uses `getattr(..., None)` — including `getattr(arc, "_package_metadata", None)` two lines above.

`GoArchive` is an exported schemaless provider and `go.link`/`go.archive` are public toolchain functions, so downstream rulesets do construct `GoArchive` by hand. Passing such an archive as `archive.direct` here (or as a `source.deps` entry in `archive.bzl`, or as a recompiled test dep in `test.bzl`) now fails analysis with `'GoArchive' value has no field or method '_package_metadata_files'`.

`getattr(d, "_package_metadata_files", depset())` would keep it consistent.

sluongng:
Rebased onto current `master` and addressed the remaining review threads. [Buildkite #9038](https://buildkite.com/bazel/rules-go-golang/builds/9038) is green with 29/29 steps passed; the one pre-test `curl` connection reset was retried and passed. The PR body has the focused Go 1.20 and Bazel 7/9 compatibility validation. @fmeum @rdesgroppes, could you take another look?

go/tools/builders/buildinfo.go
rdesgroppes: Thank you!

go/tools/builders/buildinfo.go
rdesgroppes: Perfect!

go/private/actions/link.bzl
sluongng: Addressed in 891564e9. I tried the suggested TODO cleanup first, but the new coverage regression test exposed that it is not safe even with Bazel 9.2.0: the link failed with `github.com/bazelbuild/rules_go/go/tools/coverdata: multiple copies of package passed to linker`. I therefore retained the existing importmap deduplication, but moved metadata collection completely outside the coverage/non-coverage branches so both modes traverse the same original declared dependency closure. `TestGoTestBuildInfoDepsWithCoverage` now covers this path, and the focused test passes under Bazel 9.2.0 and 7.2.1.

go/private/actions/link.bzl
sluongng: Addressed in 891564e9. I kept this PR on the existing dep-only model rather than expanding it to populate `BuildInfo.Main`: target-owned metadata is now excluded consistently for both `go_binary` and `go_test`, while declared dependency metadata remains in `Deps`. The rule docs state that choice explicitly, and the new regular and coverage `go_test` cases assert that the target metadata is absent and the external dependency metadata is present.

go/private/actions/link.bzl
sluongng: Addressed in 891564e9. All three propagation sites now use `getattr(..., "_package_metadata_files", depset())`: archive construction, link input collection, and test dependency reconstruction. Hand-constructed exported `GoArchive` values that predate this private field therefore contribute no metadata instead of failing analysis.

go/private/actions/link.bzl
```diff
@@ -122,23 +122,21 @@ def emit_link(
...
-                for archive_data in test_archives
-                for metadata in [getattr(archive_data, "_package_metadata", None)]
-                if metadata
-            ],
-            transitive = [d._package_metadata_files for d in archive.direct],
-        )
+
+    # This is the declared Bazel dependency closure, which is available
+    # during analysis. It can include metadata for a declared dependency
+    # whose Go package is not imported by the binary.
```
fmeum: Drop

go/private/rules/binary.bzl
```diff
@@ -506,6 +506,13 @@ def _go_binary_kwargs(go_cc_aspects = []):
         which must all be in the `main` package. You can run the binary with
         `bazel run`, or you can build it with `bazel build` and run it directly.
 
+        Module package metadata in the declared Bazel dependency closure is
+        included in the binary's Go build information. Since this closure is
+        determined during analysis, the build information may include metadata
+        for a declared dependency whose Go package is not imported.
+        Metadata attached to the binary target itself is not reported as a
+        dependency.
+
```
fmeum: Drop, this is very niche for such a prime spot in the docs

go/private/rules/test.bzl
```diff
@@ -528,6 +529,10 @@ _go_test_kwargs = {
     `GO_TEST_WRAP_TESTV=1` in the test environment; this will result in the
     `XML_OUTPUT_FILE` containing more granular data.
 
+    Module package metadata in the declared Bazel dependency closure is
+    included in the test binary's Go build information. Metadata attached to
+    the test target itself is not reported as a dependency.
+
```
fmeum: Drop

fmeum (approved):
Looks good to me after removing comment slop :-)

sluongng:
All current review feedback is addressed on `891564e9`, including coverage/non-coverage metadata parity, consistent target-owned metadata handling for `go_test`, and compatibility with legacy hand-constructed `GoArchive` values. Focused Bazel 9.2.0 and 7.2.1 tests are green, and the authoritative Buildkite run passed all 29 steps with zero failures: https://buildkite.com/bazel/rules-go-golang/builds/9041

@fmeum @rdesgroppes could you please re-review when convenient?

go/tools/builders/buildinfo.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return moduleInfo{}, false, fmt.Errorf("unescaping Go module path: %w", err)
+	}
+	if moduleVersion != "(devel)" {
+		moduleVersion, err = url.PathUnescape(moduleVersion)
+		if err != nil {
+			return moduleInfo{}, false, fmt.Errorf("unescaping Go module version: %w", err)
+		}
+		if !strings.HasPrefix(moduleVersion, "v") {
+			moduleVersion = "v" + moduleVersion
```
rdesgroppes, line 151: Unless I'm mistaken, this adds a `v` to anything that does not already start with one, so a commit hash gets mangled into something that is neither a version nor a commit:
- pkg:golang/example.com/mod@1.2.3 -> v1.2.3
... is fine, but:
- pkg:golang/github.com/gorilla/context@234fd47e07d1004f0aed9c -> v234fd47e07d1004f0aed9c
- pkg:golang/github.com/gorilla/context@234fd47e07d1004f0aed9c#api -> v234fd47e07d1004f0aed9c

go/private/actions/link.bzl
sluongng: Dropped in 5a45616c.

go/private/rules/binary.bzl
sluongng: Dropped in 5a45616c.

go/private/rules/test.bzl
sluongng: Dropped in 5a45616c.

go/tools/builders/buildinfo.go
rdesgroppes: It doesn't seem Gazelle itself would hit this code path, though: `internal/go_repository.bzl` already builds the PURL from a `v`-prefixed version and local-path modules get no version at all, which correctly becomes `(devel)`.

So the only producers this would affect are hand-written or third-party ones using a commit hash.

sluongng: Fixed in ad26419c. The parser now adds `v` only when the decoded value satisfies Go semver syntax; opaque values such as commit hashes are preserved unchanged. The tests cover your hash-with-fragment example, a valid unprefixed prerelease/build version, invalid numeric prerelease leading zeroes, and invalid suffixes on Go semver shorthand forms. The updated unit test passes with Bazel 9.2.0 and the Go 1.20.14 SDK.

rdesgroppes: Thanks!

fmeum (approved):

PR #4598, opened 2026-04-24
Fix invalid comparison

keith:
`go.mode` is a struct, so it was never equal to `LINKMODE_NORMAL`. I'm
not sure what visible effect this had but I noticed it when debugging
another issue.


fmeum (approved):
@jayconrod

jayconrod (approved):

PR #4599, opened 2026-04-26
Remove cgo_context_data

dzbarsky:
It's tempting to use a single target to centralize cc toolchain lookup, but ultimately it is incorrect. Toolchain resolution must be performed by the rules using that toolchain, in order to properly match on the exec platforms. The other reason for cgo_context_data was that previously optional toolchains were not possible, but now they are. Cleanup this incorrect rule to thwart the temptations.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -4,8 +4,6 @@ load(
 )
 load(
     "//go/private:context.bzl",
-    "cgo_context_data",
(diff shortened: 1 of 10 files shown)


fmeum (approved):

fmeum:
So glad to see this go away!

jayconrod (approved):

PR #4601, opened 2026-04-29
Remove cc toolchain attr

dzbarsky:
This is a legacy thing from before dedicated toolchains. We can drop it now.


fmeum (commented):
CI is failing with what looks like a legitimate failure on all platforms.

PR #4602, opened 2026-05-01
fix(nogo): generate accurate patches

blampe:
**What type of PR is this?**

> Bug fix

The unmaintained [difflib](https://github.com/pmezard/go-difflib) library we're currently using to generate patches doesn't correctly handle whitespace. There is some discussion of this on the original implementation #4102.

In particular, difflib always appends a phantom `\n` character when splitting lines, which the differ then treats as common context. Patches which remove trailing newlines (i.e. gofmt output) are incorrect as a result. Several failing unit tests are included which demonstrate this and other problems with the library.

As a fix, I've replaced the outdated difflib dependency with [go-udiff](https://github.com/aymanbagabas/go-udiff), which just exposes the internal [diff package](https://github.com/golang/tools/tree/master/internal/diff) used by gopls.

Fixes #

**Other notes for review**


fmeum (approved):
@jayconrod Do you know the new diff library?

jayconrod (approved):
I don't know the new one, but I don't have a particular reason to like the old one either. Seems fine to replace, especially if the new one is a clone of `golang.org/x/tools/internal/diff`.

fmeum:
@Wyverald rules_go CI's RBE build is failing with: 

```
 (22:09:46) ERROR: error loading package '@@buildkite_config//cc': at /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/buildkite_config/cc/cc_toolchain_config.bzl:35:6: cannot load '@@rules_cc~//cc/toolchains:feature_injection.bzl': no such file
```

Is this a known issue? We haven't changed anything.

Wyverald:
I ran into the same issue as well at https://github.com/bazelbuild/bazel/pull/29535. It looks like the new RBE configs are generated expecting a new `rules_cc` version? This line in the logs also doesn't inspire confidence: `DEBUG: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/bazel_ci_rules~/rbe_repo.bzl:35:14: rbe_preconfig: Unsupported version '7.7.0' (Values: '9.1.0', '9.0.0', '8.0.1', '7.0.2', '6.3.2', '5.2.0', '5.1.0', '5.0.0', '4.2.2', '4.2.1', '4.2.0', '4.1.0', '4.0.0'), using '9.1.0'.` Should we have actually been generating a set of RBE configs for literally every single Bazel version?

cc @coeuvre

coeuvre:
> Should we have actually been generating a set of RBE configs for literally every single Bazel version?

That's unfortunately true.

I think there are some breaking changes for the toolchain between 9.1.0 and 9.0.0. Previously, the pipeline just picked up the toolchain for 9.0.0 and worked fine. Now, the toolchain for 9.1.0 can no-longer be used for Bazel 7.7.0.

I have generated a new toolchain for Bazel 7.7.0. CI is green now.

Issue #2216, opened 2019-09-10
pure+static cgo does not link symbols from "runtime/cgo"


siddharthab:
### What version of rules_go are you using?

0.19.4 and HEAD (c44823f9d83597fd22cfed75f063aefc5de429eb).

### What version of gazelle are you using?

N/A

### What version of Bazel are you using?

0.28.0

### Does this issue reproduce with the latest releases of all the above?

Yes.

### What operating system and processor architecture are you using?

Linux, x86_64

### Any other potentially useful information about your toolchain?

Issue present with both gcc and llvm with lld linker.

### What did you do?

```
bazel build --features=static //tests/core/cgo:pure
```

### What did you expect to see?

A successful build.

### What did you see instead?

```
external/go_sdk/pkg/tool/linux_amd64/link: running /usr/bin/gcc failed: exit status 1
/tmp/go-link-689103257/go.o:go.go:_cgo_callers: error: undefined reference to 'x_cgo_callers'
/tmp/go-link-689103257/go.o:go.go:_cgo_init: error: undefined reference to 'x_cgo_init'
/tmp/go-link-689103257/go.o:go.go:_cgo_mmap: error: undefined reference to 'x_cgo_mmap'
/tmp/go-link-689103257/go.o:go.go:_cgo_munmap: error: undefined reference to 'x_cgo_munmap'
/tmp/go-link-689103257/go.o:go.go:_cgo_notify_runtime_init_done: error: undefined reference to 'x_cgo_notify_runtime_init_done'
/tmp/go-link-689103257/go.o:go.go:_cgo_sigaction: error: undefined reference to 'x_cgo_sigaction'
/tmp/go-link-689103257/go.o:go.go:_cgo_thread_start: error: undefined reference to 'x_cgo_thread_start'
/tmp/go-link-689103257/go.o:go.go:runtime._cgo_setenv: error: undefined reference to 'x_cgo_setenv'
/tmp/go-link-689103257/go.o:go.go:runtime._cgo_unsetenv: error: undefined reference to 'x_cgo_unsetenv'
/tmp/go-link-689103257/go.o:go.go:runtime.cgo_yield: error: undefined reference to '_cgo_yield'
```

hectorgimenez:
Same error with rules version 0.18.10 and Bazel 0.24.1 and 0.29. Using latest gazelle 0.18.2.

gebn:
This appears to be a regression in 0.19.4 - reverting to 0.19.3 fixes it for me.

Issue #2168, opened 2019-08-04
--features=static doesn't produce static binary


ashi009:
### What version of rules_go are you using?

0.19.1

### What version of gazelle are you using?

0.18.1

### What version of Bazel are you using?

```
Build label: 0.28.1
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Jul 19 15:19:51 2019 (1563549591)
Build timestamp: 1563549591
Build timestamp as int: 1563549591
```

### Does this issue reproduce with the latest releases of all the above?

yes

### What operating system and processor architecture are you using?

Ubuntu amd64

### Any other potentially useful information about your toolchain?

no

### What did you do?

```
$ git clone https://github.com/ashi009/rules_go_static.git
$ cd rules_go_static/
$ bazel build -s //:*
Loading: 
Loading: 0 packages loaded
Analyzing: 5 targets (1 packages loaded, 0 targets configured)
Analyzing: 5 targets (10 packages loaded, 18 targets configured)
Analyzing: 5 targets (19 packages loaded, 317 targets configured)
Analyzing: 5 targets (19 packages loaded, 1974 targets configured)
Analyzing: 5 targets (28 packages loaded, 6797 targets configured)
INFO: Analyzed 5 targets (28 packages loaded, 6799 targets configured).
INFO: Found 5 targets...
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
SUBCOMMAND: # @go_sdk//:builder [action 'GoToolchainBinary external/go_sdk/builder [for host]']
(cd /home/ashi009/.cache/bazel/_bazel_ashi009/9f4af616c9f429671d7a4338e4256eca/execroot/__main__ && \
  exec env - \
    GOROOT=external/go_sdk \
[...]

PR #4603, opened 2026-05-02
Static mode internal linker

jgautier-dd:
**What type of PR is this?**

Bug fix/Feature

When static linking is enabled in a project that contains a mix of CGo and non-CGo binaries, rules_go currently forces `-linkmode external` for all of them. This means pure Go binaries are unnecessarily routed through the external C linker even though Go's internal linker can handle them just fine.

This matters because the internal linker is significantly faster than invoking an external C linker. In a large monorepo with hundreds of Go binaries where only a fraction use CGo, the wasted link time adds up.

The go tooling already handles this correctly, when you run go build with CGO_ENABLED=1, it only uses the external linker for binaries that actually import runtime/cgo (directly or transitively). Pure Go binaries get the fast internal linker regardless of the global CGO_ENABLED setting. This patch brings rules_go in line with that behavior.

How it works:

The linker wrapper (link.go) gains a -force-external-if-cgo flag, passed by link.bzl only when static=on. At link time, the wrapper performs a BFS walk of the import graph by reading Autolib entries from each Go archive's goobj header. If runtime/cgo appears anywhere in the transitive closure, -linkmode external is injected. Otherwise, the Go linker's default "auto" mode is left alone, and it picks the internal linker.

This preserves the fix for #2168 and #2216 (static builds that transitively import runtime/cgo via packages like net still get external linking and a proper static binary), while avoiding it for binaries that don't need it.

Other conditions that force external linking (race builds, non-normal link modes, Windows msan) are left unchanged.

### DIFF
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -79,6 +79,22 @@ def emit_link(
     else:
         extld = extld_from_cc_toolchain(go)
         tool_args.add_all(extld)
+        if extld and (go.mode.race or
+                      go.mode.linkmode != LINKMODE_NORMAL or
+                      go.mode.goos == "windows" and go.mode.msan):
+            # Force external linking for the following conditions:
+            # * Non-normal build mode: may not be strictly necessary, especially
+            #   for modes like "pie".
+            # * Race or msan build for Windows: Go linker has pairwise
+            #   incompatibilities with mingw, and we get link errors in race mode.
+            #   Using the C linker avoids that. Race and msan always require a
+            #   a C toolchain. See #2614.
+            # * Linux race builds: we get linker errors during build with Go's
+            #   internal linker. For example, when using zig cc v0.10
+            #   (clang-15.0.3):
+            #
+            #       runtime/cgo(.text): relocation target memset not defined
+            tool_args.add("-linkmode", "external")
         if go.mode.static:
             # When static linking is requested, the Go linker's auto mode may
             # not pick external linking for packages like net that implicitly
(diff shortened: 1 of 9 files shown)


dzbarsky:
https://github.com/bazel-contrib/rules_go/commit/feb90eb90cc0902e20848cdf3a7a0e1c288a712f added go_infos_use_cgo and propagates the has-cgo bit up the providers. Can we reuse that at the binary level in starlark and avoid the complexity of the parallel wall you're doing inside the builder action?

jgautier-dd:
I agree that using something in starlark would be better. The one case that I don't think can be determined with starlark though is when a stdlib package uses cgo. Since there is no `go_library` target for the stdlib packages there is no cgo attribute that can be used to determine if it needs the C/C++ linker or not.

fmeum:
@jayconrod Do you see a good way to avoid the object file introspection while still handling stdlib packages correctly?

jayconrod:
Ooh, really don't like object file introspection. 😬 

If we need to track which std packages use cgo, we could probably do that with a `go list` command, saving the results from `GoStdlib`. I haven't look in a while, but when I worked on Go, there were two: `net` and `os/user`. I think on macOS, some of the crypto libraries need to link with system frameworks, too.

Building without `cgo` changes behavior in uncommon cases for both packages. `net` uses a builtin DNS resolver instead of `getaddrinfo`. I'm not quite sure how it differs though.

Does this come up often? Every non-trivial program uses `net`. I don't quite remember what cases the internal linker can handle though, and it's probably changed.

PR #4605, opened 2026-05-07
fix: put GoProtocGen in its own exec group

### LINKED ISSUE #4604: GoProtocGen cannot execute binary file - go-protoc-bin built for the wrong exec platform
## Summary

When using multiple exec platforms (e.g. via `--host_platform` and `--extra_execution_platforms`), it is possible for `go-protoc-bin` to be built for the wrong platform, leading to a "cannot execute binary file" or "exec format error" error during the GoProtocGen action.

## Detailed Explanations

The issue is due to two inconsistent platform resolutions:

`go-protoc-bin` is built with `cfg = "exec"` which, barring any `exec_compatible_with` constraint, selects the first available exec platform (e.g. the first one from `--extra_execution_platforms`).

Separately, the `go_proto_library` rule depends on toolchains (GO_TOOLCHAIN and CGO_TOOLCHAINS) which may constrain which exec platform the library can be built on (e.g. a darwin exec platform if the target platform is darwin). This makes sense for the compilation of the go library, but the `GoProtocGen` action does not need that constraint (the go source generation can be run on any platform, e.g. linux, even if the library is compiled for another platform e.g. darwin). When `--incompatible_auto_exec_groups` is not explicitly set or no `exec_group` are set, all the actions under a rule (here, `go_proto_library`) resolve to the same exec platform (e.g. darwin).

We may hence end up with e.g. a linux binary (`go-protoc-bin`) running on e.g. a darwin exec platform (the `GoProtocGen` action), causing a "cannot execute binary file" error.

Related to #4571

sitaktif:
When building go_proto_library with two registered execution platforms (e.g. linux/x86_64 RBE registered first, darwin/arm64 local fallback), the GoProtocGen action would fail with "cannot execute binary file":

- go-protoc-bin is built via `cfg = "exec"` in _go_proto_compiler, which independently selects the first registered exec platform (e.g. linux ELF binary).
- Without --incompatible_auto_exec_groups, go_proto_library's Go toolchain declaration may have additional constraints (e.g. darwin arm64) and forces ALL rule actions onto darwin, including GoProtocGen.

This may lead to e.g. a linux binary executing on a darwin platform, leading to a "exec format error" error.

Fix: declare an unconstrained `go_proto_gen` exec group in go_proto_library and assign GoProtocGen to it via `exec_group = "go_proto_gen"`. The group has no toolchain constraints, making it equivalent to `cfg = "exec"`, so Bazel picks the first compatible exec platform — the same one that built go-protoc-bin via `cfg = "exec"`.

go.env already carries GOOS/GOARCH/GOROOT as concrete strings from go_context(), so the generated .pb.go files remain correctly targeted to the host platform. The generator itself does not need to run on the target platform.

Fixes #4604 

-- 

I pushed a way to reproduce the issue in https://github.com/sitaktif/rules_go/tree/rules-go-exec-format-error-repro, but I don't have an actual test in this PR to prevent future regressions because I'm not sure how to test that without an overly complicated and slow setup. Please share any suggestions on how to do that :-)

### DIFF
--- a/proto/def.bzl
+++ b/proto/def.bzl
@@ -215,6 +215,18 @@ go_proto_library = rule(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
     } | CGO_ATTRS,
+    # go-protoc-bin is built via cfg = "exec" and lands on the first registered
+    # exec platform, while the Go toolchain may have constraints on the exec
+    # platform. Because of the possible differences in platform resolution,
+    # this may end up with a go-protoc-bin that is incompatible with the
+    # selected exec platform (when --incompatible_auto_exec_groups is unset).
+    # Assign it an explicit exec group with no toolchain constraints so it
+    # resolves to the same platform as the one resolved for the binary.
+    # Note that this exec platform matching is still fragile; the proper fix
+    # is to create proper toolchains for the protoc compiler.
+    exec_groups = {
+        "internal_use_only_go_proto_gen": exec_group(),
+    },
     fragments = CGO_FRAGMENTS,
     toolchains = [GO_TOOLCHAIN] + CGO_TOOLCHAINS,
 )
--- a/proto/compiler.bzl
+++ b/proto/compiler.bzl
@@ -29,7 +29,6 @@ load(
 load(
     "//go/private:common.bzl",
     "GO_TOOLCHAIN",
-    "GO_TOOLCHAIN_LABEL",
 )
 load(
     "//go/private:context.bzl",
@@ -177,7 +176,7 @@ def go_proto_compile(go, compiler, protos, imports, importpath):
         progress_message = "Generating into %s" % go_srcs[0].dirname,
         mnemonic = "GoProtocGen",
         executable = compiler.internal.go_protoc,
-        toolchain = GO_TOOLCHAIN_LABEL,
+        exec_group = "internal_use_only_go_proto_gen",
         tools = [compiler.internal.protoc, compiler.internal.plugin],
         arguments = [args],
         env = go.env,


sitaktif:
@fmeum you suggested a different approach in https://github.com/bazel-contrib/rules_go/issues/4571#issuecomment-4382463685, but it looks that simply setting an empty exec group does the job as that should match the `cfg = "exec"` from the tool. I might have missed something important though so it would be great if you could vet the approach.

fmeum:
This will work in more cases, but it still has issues in edge cases. For example, `exex_compatible_with` on a rule will influence toolchain resolution for that rule only, so `cfg = "exec"` can behave differently on two different rules. 

Turning the proto compiler into a toolchain would fix this cleanly as we could then add the toolchain requirement right next to the Go toolchain.

sitaktif:
Yeah your comment makes complete sense. In this commit, we're basically saying "set a plain exec config for this exec group... and we happen to know that this matches the `cfg = 'exec'` on the other side... at least until that's not true anymore."

Agreed re: making the proto compiler into a toolchain being a holistic fix, but that's a change I don't quite have enough confidence to make at this point. 

Given the small footprint of this change, and given that it is a small incremental improvement, would you consider accepting it?

fmeum:
Yes, it's definitely an improvement. But please add a comment clearly stating that this is a hack and that the correct setup would be based on joint resolution of toolchain types.

sitaktif:
@fmeum thanks, done 👍 

(also rebased the branch)

proto/def.bzl
```diff
@@ -215,6 +215,18 @@ go_proto_library = rule(
...
+    # exec platform, while the Go toolchain may have constraints on the exec
+    # platform. Because of the possible differences in platform resolution,
+    # this may end up with a go-protoc-bin that is incompatible with the
+    # selected exec platform (when --incompatible_auto_exec_groups is unset).
+    # Assign it an explicit exec group with no toolchain constraints so it
+    # resolves to the same platform as the one resolved for the binary.
+    # Note that this exec platform matching is still fragile; the proper fix
+    # is to create proper toolchains for the protoc compiler.
+    exec_groups = {
+        "go_proto_gen": exec_group(),
```
fmeum: Please give this a really ugly name such as `internal_use_only_go_proto_gen` as exec groups become public API and we would want to get rid of this when we fix it properly.

sitaktif: Name is now ugly 👍 🙂

sitaktif:
@fmeum addressed your latest comment!

fmeum:
@sitaktif looks like you forgot to commit the change? I rebased into master, but the name is still "pretty".

sitaktif:
@fmeum sorry, wrong remote, fixed it (and rebased) 🙇

fmeum (approved):

PR #4606, opened 2026-05-08
fix: restore internal_archive.direct after cycle-break recompile

### LINKED ISSUE #3981: gopackagesdriver: dependencies missing from Imports of _xtest targets
v0.48.1

v0.35.0

7.2.1

Yes.

Darwin aarch64

No.

### What did you do?

Create two packages, foo and bar, such that bar depends on foo. 

`foo/foo.go`
```go
package foo

func Foo() {}
```

`bar/bar.go`
```go
package bar

import "github.com/dt/externdemo/foo"

func Bar() { foo.Foo() }
```

Now create a test in `foo_test` xtest package that depends on the `bar` package, e.g. `foo/foo_external_test.go`:
```go
package foo_test

import "github.com/dt/externdemo/bar"

func FooTest() {bar.Bar()}
```

Construct BUILD files accordingly, for example with `bazel run :gazelle`. Specifically, in `foo/BUILD.bazel`: 

```
go_library(
    name = "foo",
    srcs = ["foo.go"],
    importpath = "github.com/dt/externdemo/foo",
)

go_test(
    name = "foo_test",
    srcs = [
        "foo_external_test.go",
    ],
    embed = [":foo"],
    deps = ["//pkg/barlib"],
)
```

Finally, ask `gopackagesdriver` about `foo_external_test.go`, e.g. 
```
echo "{}" | bazel run -- @io_bazel_rules_go//go/tools/gopackagesdriver file=./pkg/keys/printer_test.go
```

In the `Imports` section for the entry in `Packages` returned by gopackagesdriver, for package foo_xtest, I expected to see `//bar`. In my editor, I expected normal completions and type highlighting for bar.

### What did you see instead?

 `bar` is is not included in foo_test's deps:
```
"@@//foo:foo_test_xtest"
{
  "fmt": "@@io_bazel_rules_go//stdlib:fmt",
  "github.com/dt/externdemo/foo": "@@//foo:foo_test",
  "os": "@@io_bazel_rules_go//stdlib:os",
  "testing": "@@io_bazel_rules_go//stdlib:testing"
}
```

In my editor (vscode) this then manifests as failing to resolve types or completions on `bar` and the import being shown with an error:
```
could not import github.com/dt/externdemo/bar (missing metadata for import of "github.com/dt/externdemo/bar")compiler[BrokenImport](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#BrokenImport)
```

findleyr:
When a go_test embeds a library and an external test (x_test) source imports a package that transitively reaches the embedded library, _recompile_external_deps recompiles the internal archive with a filtered deps list to break the cycle. (We call these intermediate test variants in cmd/go).

However, when the go/packages driver queries package information for an x_test source, the go_pkg_info_aspect looks up imports in the first archive it encounters. Prior to this PR, that could be the internal archive, which only contained imports that were filtered in the middle of cycle-breaking.

This change restores the recompiled imports back onto the internal archive, so that the driver can reconstruct accurate import information for the x_test package.

Credit to @jgautier-dd for #4405, on which this was based.

Fixes #3981

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -665,6 +665,10 @@ def _recompile_external_deps(go, external_go_info, internal_archive, library_lab
     # can't import anything that imports itself.
     internal_go_info = internal_archive.source
 
+    # Store the original labels so that we can map dependencies back onto
+    # internal_archive.direct after recompilation (see below).
+    original_internal_dep_labels = [dep.data.label for dep in internal_go_info.deps]
+
     internal_deps = []
 
     # Pass internal dependencies that need to be recompiled down to the builder to check if the internal archive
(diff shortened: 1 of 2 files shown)


jayconrod (approved):
Thanks for fixing this!

PR #4610, opened 2026-05-18
Add loongarch64 support

zhaixiaojuan:
What type of PR is this?

Bug fix

It matches GOARCH loong64 to something so that rules_go would work on LoongArch.

https://github.com/bazel-contrib/rules_go/issues/4609


fmeum (approved):

go/private/platforms.bzl
```diff
@@ -35,6 +35,7 @@ BAZEL_GOARCH_CONSTRAINTS = {
     "ppc64le": "@platforms//cpu:ppc64le",
     "s390x": "@platforms//cpu:s390x",
     "riscv64": "@platforms//cpu:riscv64",
+    "loong64": "@platforms//cpu:loongarch64",
```
fmeum, line 38: CI makes it look like you need to update the dependency on `platforms`.

zhaixiaojuan: > CI makes it look like you need to update the dependency on `platforms`.

@fmeum 
Hi， version 1.1.0 of platforms added support for loongarch64, so I updated platforms to this version.

fmeum: You also have to update the WORKSPACE deps macro

PR #4611, opened 2026-05-20
Set direct nogo attribute to prevent building nogo for the wrong platform when multiple exec platforms are set

sitaktif:
**What type of PR is this?**

Bug fix

This PR fixes an issue whereby, when two or more exec platforms are set (via `--host_platform` and/or `--extra_execution_platforms`), it was possible for Nogo to be build for one exec platform A, but be executed on exec platform B, leading to an "exec format error" / "cannot execute binary file" error when the action is run.

This PR moves the nogo dep directly to the relevant rules instead of via `go_context_data`, removing the possibility for a discrepancy for the two exec configs.

Fixes #4571

**Other notes for review**

I wasn't sure about the handling of "legacy host mode" mentioned in the comments so I added a test that covers the case... if that's what the comments meant by "host mode". Please someone check whether that's actually testing that (and/or if we still care about host mode).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -65,7 +65,6 @@ go_context_data(
     name = "go_context_data",
     coverdata = "//go/tools/coverdata",
     go_config = ":go_config",
-    nogo = "@io_bazel_rules_nogo//:nogo",
     stdlib = ":stdlib",
     visibility = ["//visibility:public"],
 )
--- a/go/private/BUILD.bazel
+++ b/go/private/BUILD.bazel
@@ -168,25 +168,19 @@ bool_setting(
     visibility = ["//visibility:public"],
 )
 
-# Usually true. The entire toolchain gets nogo to use in builds via the
-# go_context_data rule, which has an incoming transition that flips this flag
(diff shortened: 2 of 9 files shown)


fmeum:
`host` is no longer supported in Bazel. Could you drop all associated logic and the test?

sitaktif:
@fmeum I removed the refs to legacy host mode from my change (only the new test was referring to it)

Separately, CI is failing; I wonder if it's due to some cache evictions: https://buildkite.com/bazel/rules-go-golang/builds/8780/list?tab=output&jid=019e49f5-4339-422c-ac1b-ead95d73feb3#L1189

sitaktif:
Looking good after a rebase and a re-build! @fmeum PTAL! :-)

fmeum (approved):
This is great, thanks for a fix that's also a cleanup!

sitaktif:
FYI I identified another place in the code that refers to the legacy host mode; I will open another PR.

sitaktif:
Host mode cleanup: #4614

PR #4615, opened 2026-05-27
fix: produce facts of required analyzers in nogo facts-only mode

### LINKED ISSUE #4616: nogo: fact-only mode skips analyzers whose facts come from required sub-analyzers
**What version of rules_go are you using?**

0.60.0; the relevant code is unchanged on master.

**What did you do?**

Registered a nogo analyzer that declares no fact types of its own but `Requires` a fact-exporting analyzer — the same shape as staticcheck's SA1019, which consumes the result of a required `deprecated` analyzer, which itself computes deprecation facts. A package outside the nogo `includes` (or in an external repository) declares the symbol the fact describes; a package inside the `includes` uses it.

**What did you expect to see?**

The dependency's facts-only compile produces the fact, and the analyzer reports a diagnostic on the use in the included package. The comment at the `-facts_only` call site says "all analyzers that don't generate facts can be skipped", which is the behavior I expected: skip only analyzers whose requirement closure produces no facts.

**What did you see instead?**

No diagnostic. In facts-only mode, `go/tools/builders/nogo_main.go` keeps an analyzer only if it declares fact types itself:

```go
if !factsOnly || len(a.FactTypes) > 0 {
```

If the analyzer producing facts is only present in the *requirements* of the root analyzers, it's skipped, and the root analyzers using the results of those requirements are broken.

findleyr:
**What type of PR is this?**

Bug fix

In facts-only compiles (packages outside the nogo `includes`, which includes every external repository), nogo keeps an analyzer only if it declares `FactTypes` itself:

```go
if !factsOnly || len(a.FactTypes) > 0 {
```

Analyzers whose facts are produced by a *required* sub-analyzer are dropped entirely, so their requirement closure never runs and downstream packages never see those facts. The common shape is a checker that requires a fact-exporting analyzer: staticcheck's SA1019 (use of deprecated identifiers) requires its deprecation fact analyzer and declares no fact types of its own, so deprecation facts are never produced for external repositories and uses of deprecated third-party APIs in the main repository are silently missed. Analyzers like `printf` are unaffected only because they happen to declare their fact type directly.

This PR keeps an analyzer in facts-only mode when anything in its requirement closure declares fact types, which matches the intent of the existing comment ("all analyzers that don't generated facts can be skipped").

Verified against a large monorepo by patching its rules_go: with the fix, SA1019 immediately reported uses of deprecated identifiers declared in external module dependencies that were silently missed before, while external packages themselves still produced no diagnostics.

No existing issue that I could find; happy to file one if preferred.

**Other notes for review**

The new `tests/core/nogo/bzlmod/facts_only_test.go` pins the behavior with a pair of custom analyzers mirroring the deprecation shape: the fact producer is only required, the nogo-registered analyzer declares no fact types, and the marked function is declared in a package outside the nogo `includes`. The first test fails without the `nogo_main.go` change and passes with it; the second checks that facts-only compiles still report no diagnostics of their own.

Fixes #4616

### DIFF
--- a/go/private/actions/compilepkg.bzl
+++ b/go/private/actions/compilepkg.bzl
@@ -253,7 +253,7 @@ def _run_nogo(
 
     nogo_args.add_all(archives, before_each = "-facts", map_each = _facts)
     if not out_validation:
-        # Since diagnostics are ignored, all analyzers that don't generated facts can be skipped.
+        # Since diagnostics are ignored, analyzers that don't generate facts can be skipped.
         nogo_args.add("-facts_only")
     nogo_args.add("-out_facts", out_facts)
     nogo_args.add_all("-out", [out_diagnostics], expand_directories = False)
--- a/go/tools/builders/nogo_main.go
+++ b/go/tools/builders/nogo_main.go
@@ -274,9 +274,19 @@ func checkPackage(analyzers []*analysis.Analyzer, packagePath, goVersion string,
 		}
 	}
 
+	// In facts-only mode diagnostics are discarded, so we only need to run
+	// analyzers that produce facts.
+	//
+	// Note that this set may be disjoint from the initial set of analyzers: root
+	// analyzers may consume results from required analyzers which themselves use
+	// facts.
 	roots := make([]*action, 0, len(analyzers))
-	for _, a := range analyzers {
-		if !factsOnly || len(a.FactTypes) > 0 {
+	if factsOnly {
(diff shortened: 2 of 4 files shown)


findleyr:
CC @jayconrod

jayconrod (approved):
LGTM. @fmeum Could you take a look too?

fmeum (approved):

PR #4617, opened 2026-05-29
Revert "fix: restore internal_archive.direct after cycle-break recompile"

### LINKED ISSUE #4607: gopackagesdriver: test-vendored deps resolve to production targets, violating `go/packages` PkgPath uniqueness

v0.60.0 (also reproducible at HEAD).

v0.50.0.

8.5.0.

Yes.

linux amd64.

The bug is in `go_pkg_info_aspect`'s output, so it reproduces with any toolchain.

### What did you do?

This is distinct from #3981 (missing imports). Both bugs share the same trigger — a `go_test` with an external test that imports a package that transitively reaches the embedded library — but #3981 is a *missing* edge in the driver response, and this is a *wrong* edge that survives even after #4606 restores the missing one.

This is the source of over two years of confusion in golang/go#63822.
CC @jayconrod @adonovan

**Repro:**

Set up a `go_test` whose external test (`package foo_test`) imports a package `helper` that itself imports `foo`:

```
go_library(name = "foo", importpath = "example.com/foo", srcs = ["foo.go"])
go_library(name = "helper", importpath = "example.com/helper", srcs = ["helper.go"], deps = [":foo"])
go_test(
    name = "foo_test",
    srcs = ["foo_internal_test.go", "foo_external_test.go"],  # package foo + package foo_test
    embed = [":foo"],
    deps = [":helper"],
)
```

Then load `foo_external_test.go` through the gopackagesdriver (e.g. open it in gopls, or run `go vet` with `GOPACKAGESDRIVER` set) and inspect the `DriverResponse`.

`go_test` has to recompile `helper` here, because `helper` was compiled against the production `foo` but the test embeds a *different* `foo` — the production sources plus `foo_internal_test.go`. If the external test got `helper`'s view of `foo` from one compilation and its own view from another, type identity would break: `helper` would return a `*foo.T` and the test would hold a different `*foo.T`. So `_recompile_external_deps` recompiles the chain between the external test and the embedded library against the test variant — the same thing `cmd/go` does, and what `go list -compiled -test` reports as `helper [example.com/foo.test]`: a separate package record with the same import path but a distinct identity.

The build does this correctly. The driver's *description* of the build does not.

`go_pkg_info_aspect` emits one `pkg.json` per Bazel target. The recompiled `helper [foo.test]` is not a target — it is an archive `_recompile_external_deps` constructs internally. So the aspect never emits a record for it. When the aspect builds the `Imports` map for the test record, the only `helper` it can point at is the production `:helper` target. Production `helper` imports production `:foo`. And the external test also reaches the test variant of `foo` directly, by embedding it.

Walking the transitive `Imports` graph from the external test then reaches **two** records with `PkgPath = "example.com/foo"`: the test variant (the merged `:foo_test` record) and the production `:foo` (through production `:helper`).

The `go/packages` driver protocol requires that, from any package, the transitive `Imports` graph reaches at most one record per import path. That is the invariant `go/types` consumers rely on — `go/types` keys its package universe on `PkgPath`, and two `*types.Package` reachable from the same root with the same `Path` means every type in `foo` exists twice.

`go list -test` upholds the invariant by giving the recompiled chain distinct IDs (`helper [example.com/foo.test]`) with `Imports` that point at the test variant, not the production package. The gopackagesdriver should do the same.

### What did you see instead?

A driver response with two records for `example.com/foo` reachable from the external test. Downstream `go/types` consumers see two `*types.Package` with the same `Path` and produce errors like:

```
cannot use h.Get() (value of type *foo.T) as *foo.T value in assignment
```

— two types that print identically and are defined in the same source file but are not the same `*types.Named`.

In gopls this is intermittent and hard to attribute: gopls deduplicates packages by `Metadata.ID` per query and reuses results across queries via the filecache, so whether the duplicate is observed depends on load order, what else is in the cache, and which file the user opens first. The visible symptoms are flaky type-identity diagnostics that go away on a cache rebuild, and duplicate `PkgPath` entries in the shallow export-data manifest. The latter is the long-standing telemetry signal under golang/go#63822, which has lacked a known cause; this is one.

Interestingly, an alternative fix I considered for #3981 was to have the aspect query by (label, name) rather than just name, and that is probably the shape of the solution here: the aspect *cannot* fix this by querying for more targets; the recompiled archives are not targets. It has to walk `_recompile_external_deps`'s output through the `GoArchive` provider on the `go_test` target, which carries the recompiled archives as `direct` deps of the internal/external test archives.

I will send a fix for this issue.

findleyr:
Reverts bazel-contrib/rules_go#4606

I think this contributes another case to #4607 -- per my investigation not the *only* path to duplicate import paths, but it does make it possible.

I'm still investigating, but I think we should revert this patch and "do it properly", ensuring coherency of direct imports.

I will send a follow-up, more principled fix that also fixes #4607.

### DIFF
--- a/go/private/rules/test.bzl
+++ b/go/private/rules/test.bzl
@@ -669,10 +669,6 @@ def _recompile_external_deps(go, external_go_info, internal_archive, library_lab
     # can't import anything that imports itself.
     internal_go_info = internal_archive.source
 
-    # Store the original labels so that we can map dependencies back onto
-    # internal_archive.direct after recompilation (see below).
-    original_internal_dep_labels = [dep.data.label for dep in internal_go_info.deps]
-
     internal_deps = []
 
     # Pass internal dependencies that need to be recompiled down to the builder to check if the internal archive
@@ -759,20 +755,6 @@ def _recompile_external_deps(go, external_go_info, internal_archive, library_lab
                 _headers = internal_archive._headers,
             )
         label_to_archive[label] = archive
-
-    # Fix up internal_archive.direct for gopackagesdriver, which queries the
-    # archive by label to build the go/packages response map for a test file,
-    # which could be part of an x_test package. Because the aspect uses the
-    # first archive it finds for that label, we need direct to hold the
-    # recompiled external deps (=intermediate test variants), even for the
-    # internal archive. See #3981.
-    attrs = structs.to_dict(internal_archive)
-    attrs["direct"] = [
-        label_to_archive[label]
-        for label in original_internal_dep_labels
-    ]
-    internal_archive = GoArchive(**attrs)
-    label_to_archive[internal_archive.data.label] = internal_archive
 
     # Finally, we need to replace external_go_info.deps with the recompiled
     # archives.
--- a/go/tools/gopackagesdriver/gopackagesdriver_test.go
+++ b/go/tools/gopackagesdriver/gopackagesdriver_test.go
@@ -40,20 +40,6 @@ go_test(
         "hello_external_test.go",
     ],
     embed = [":hello"],
-    deps = [":hellohelper"],
-)
-
-# hellohelper imports hello, so a go_test that embeds hello and depends on
-# hellohelper from an external (xtest) source forces _recompile_external_deps
-# to run. This is the trigger for #3981: the recompiled internal archive's
-# direct deps used to drop the cycle-forming dependency, so the
-# gopackagesdriver reported the xtest with a hole in its Imports.
-go_library(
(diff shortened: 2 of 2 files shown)


findleyr:
CC @jayconrod @fmeum -- I think we should revert this workaround and implement a more principled fix, that also avoids duplicate import paths. I'm not confident that #4606 was a pure improvement, since it introduces a new path to having both the production package and test variant in the import graph.

Sorry for the churn.

jayconrod (approved):

fmeum (approved):

PR #4631, opened 2026-06-08
WIP: FIPS 140 v1.0.0 support

adnanbhanji:
Go 1.24 added [GOFIPS140](https://go.dev/doc/security/fips140), which builds the
standard library and all crypto against a frozen, CMVP-validated snapshot of the
Go Cryptographic Module for FIPS 140-3 compliance. rules_go currently has no way
to enable it, so projects that need a FIPS 140-3 validated build cannot use a
rules_go toolchain.

This adds a `gofips140` attribute to `go_download_sdk`, `go_host_sdk`,
`go_local_sdk` and `go_wrap_sdk` (and to the `go_sdk` module extension), which
sets GOFIPS140 for the toolchain so the standard library and all downstream Go
code are compiled against the selected FIPS module snapshot, e.g. `v1.0.0`.

## How GOFIPS140 works in the Go toolchain

- `GOFIPS140=<version>` selects a frozen snapshot of the Go Cryptographic
  Module. The snapshots ship inside the SDK under `lib/fips140/<version>.zip` —
  there is no separate download.
- With a snapshot selected, the `go` command compiles the versioned FIPS
  packages (the `crypto/internal/fips140/<version>/...` tree) from the frozen
  zip instead of the in-tree `src/`, so all crypto routes through the validated
  module.
- `off`/`""` disable FIPS mode and `latest` builds from `GOROOT/src`; every
  other value (a version like `v1.0.0`, or an alias like `certified`/
  `inprocess` that resolves via `lib/fips140/<value>.txt`) pulls the frozen
  snapshot.

## What this requires from rules_go

- Thread a `gofips140` value from the SDK rules through the generated SDK BUILD
  file, the `go_sdk` rule and the `GoSDK` provider, and export it into every Go
  compile action's environment as `GOFIPS140` (in `go_context`).
- Include `lib/fips140/**` in the SDK `srcs`/filegroups so the snapshot zips are
  available as action inputs.
- Rebuild the standard library when a snapshot is selected. The prebuilt stdlib
  shipped in the SDK is compiled with GOFIPS140 unset, so
  `_should_use_sdk_stdlib` falls through to a rebuild whenever `gofips140` is
  set.
- Default the binary to `GODEBUG=fips140=on` at link time, so a FIPS-built
  binary actually activates FIPS mode.
- Enumerate the versioned FIPS packages that must be materialized into `pkg/`.

## Enumerating the snapshot

The package set has to come out of the snapshot zip, which Starlark cannot read
during rule analysis. An earlier revision of this PR shelled out to `unzip` from
a build action, which @r-hang and @blampe correctly pointed out is neither
hermetic nor portable.

`repository_ctx` *can* read an archive, so enumeration now happens at SDK-fetch
time in `_fips_snapshot_packages` (`go/private/sdk.bzl`): resolve any alias via
`ctx.read`, `ctx.extract` the snapshot to a temp dir, walk it with `readdir`,
delete the temp dir, and bake the resulting sorted import paths into the
generated BUILD file. `package_list` then writes both `packages.txt` (for the
linker's importcfg) and `fips_packages.txt` (read verbatim by the stdlib
builder) with plain `ctx.actions.write`. One walk feeds both outputs, so there
is a single enumeration and no re-derivation in Go, and the build itself needs
no shell action or external tool.

The walk returns `.go` file paths and a separate pure function maps them to
import paths, which makes the mapping unit-testable from `tests/core/starlark`.

`_sdk_build_file` takes the GOROOT to enumerate from, because `go_host_sdk`,
`go_local_sdk` and `go_wrap_sdk` all call it *before* `_local_sdk` has symlinked
GOROOT into the repository; only `go_download_sdk` extracts in place beforehand.

Enumeration fails loudly when the snapshot zip is missing, the resolved version
directory is absent, the snapshot holds no Go sources, or the bounded walk does
not drain — rather than quietly producing an empty package list and a non-FIPS
build for a caller who asked for a validated one. For the same reason,
`gofips140` combined with `experimental_build_compiler_from_source` is refused:
the bootstrap SDK build file does not propagate GOFIPS140.

## Implementation note

With `GOFIPS140=<version>`, `go install std` compiles the versioned snapshot
packages but, due to golang/go#76225, leaves their archives in the build cache
rather than in `$GOROOT/pkg/<platform>/`, so the linker cannot find them. To
place them without recompiling, the builder runs
`go build -o $GOROOT/pkg/<platform>/<pkg>.a <pkg>` for each snapshot package,
reusing the exact flags `go install std` used. Because the flags match, each
`go build` is a build-cache hit on the archive `go install` already produced
(verified with `go build -x`: it copies the cached `.a`, no recompile), so the
placed archives are byte-identical to the rest of `pkg/`. This is marked with a
TODO to remove once #76225 is fixed upstream.

Snapshot builds also unpack into a module cache, so the builder points
`GOMODCACHE` at a temporary directory for the duration of the `std` build;
`latest` builds from `GOROOT/src` and need none.

## Platform support

GOFIPS140 builds are exercised on Linux. The package-list step no longer needs a
POSIX shell, so the earlier Windows restriction is gone, but Windows is untested
rather than known to work — the `gofips140` docs say so explicitly.

## Testing

- `tests/core/starlark/sdk_tests.bzl` unit-tests the snapshot path-to-import-path
  mapping.
- `tests/core/go_download_sdk` builds a real SDK with `gofips140 = "v1.0.0"`,
  asserts `crypto/fips140.Enabled()` in a test binary, and checks the generated
  package list; a negative case asserts that requesting a snapshot from a
  pre-1.24 SDK fails.
- Verified manually against Go 1.24.0 and, for alias resolution, Go 1.26.3,
  across `go_download_sdk`, `go_host_sdk`, `go_wrap_sdk` and
  `go_sdk.from_file` — 27 snapshot packages each, an empty list for a non-FIPS
  SDK, and working binaries under both WORKSPACE and bzlmod.

ref golang/go#76225


go/tools/builders/importcfg.go
```diff
@@ -167,16 +167,42 @@ func buildImportcfgFileForLink(archives []archive, stdPackageListPath, installSu
...
 		fmt.Fprintf(buf, "packagefile %s=%s.a\n", line, filepath.Join(prefix, filepath.FromSlash(line)))
 	}
 	if err := scanner.Err(); err != nil {
 		return "", err
 	}
+	// When GOFIPS140=v1.0.0 is set, the stdlib builder copies versioned FIPS
+	// snapshot .a files into pkg/<platform>/ that are not in the static SDK
+	// package list. Scan the pkg directory for any extra .a files and add them
+	// to the importcfg so the linker can resolve them.
+	if _, statErr := os.Stat(prefix); statErr == nil {
```
r-hang: This traversal and installFIPSSnapshotArchives are performing the same logic in slightly different ways. They both seek to get a list of standard library packages that account for the GOFIPS140 snapshot content to prepare them for linking.

I think these can be unified. I think the best approach this problem would be to replace the current logic that generates the packages.txt, which powers the data source for this buildImportcfgFileForLink function to consider the GOFIPS140 snapshot packages (right now it just uses pure starlark to iterate over the directories. We would turn the bazel logic that generates a packages.txt into a go program / builder that can read lib/fips140 archives for the expected packages.

We could then use a single source of truth in the packages.txt to power build importcfg and also work around the issue that `installFIPSSnapshotArchives` works around by running `go build -o` on the GOFIP140 packages we want to place in the package tree instead of calling `go list` to get the stdlib packages.

go/tools/builders/stdlib.go
```diff
@@ -57,8 +59,14 @@ func stdlib(args []string) error {
...
 	}
 
-	// Link in the bare minimum needed to the new GOROOT
-	if err := replicate(goroot, output, replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
+	// Link in the bare minimum needed to the new GOROOT. The lib/ directory holds
+	// the FIPS snapshot zips and only exists when GOFIPS140 is set, so only
+	// replicate it in that case to avoid failing on SDKs that lack it.
+	replicateDirs := []string{"src", "pkg/tool", "pkg/include"}
+	if v := os.Getenv("GOFIPS140"); v != "" && v != "off" {
+		replicateDirs = append(replicateDirs, "lib")
```
r-hang: instead of `lib` we should be more narrow and specify to `libflips140` so that we only replicate the minimal set of files.

go/tools/builders/stdlib.go
```diff
@@ -81,6 +89,15 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
 	// modules on in "auto" mode.
 	os.Setenv("GO111MODULE", "off")
 
+	// GOFIPS140=v1.0.0 triggers a module-like download of the FIPS snapshot.
+	// Set GOMODCACHE to a temporary directory to avoid "module cache not found" errors.
+	if gofips140 := os.Getenv("GOFIPS140"); gofips140 != "" && gofips140 != "off" {
```
r-hang: Is this needed in non-snapshot settings? If it's not let's update the logic to make that clear.

go/tools/builders/stdlib.go
```diff
@@ -170,5 +187,67 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
+	//
+	// Workaround: after `go install std`, run `go list -json -export std` to
+	// locate the cached .a files for the versioned snapshot packages and copy
+	// them into the expected pkg/ directory so the linker can find them.
+	if gofips140 := os.Getenv("GOFIPS140"); gofips140 != "" && gofips140 != "off" && gofips140 != "latest" {
+		if err := installFIPSSnapshotArchives(goenv, output); err != nil {
+			return fmt.Errorf("failed to install FIPS snapshot archives: %v", err)
+		}
+	}
+
```
r-hang: nit: remove space

go/tools/builders/stdlib.go
```diff
@@ -170,5 +187,67 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
+	// GOFIPS140=v1.0.0 places the versioned FIPS .a files into the Go build
+	// cache (~/.cache/go-build/) instead of $GOROOT/pkg/<platform>/. This is an
+	// upstream Go bug: https://github.com/golang/go/issues/76225.
+	//
+	// Workaround: after `go install std`, run `go list -json -export std` to
+	// locate the cached .a files for the versioned snapshot packages and copy
+	// them into the expected pkg/ directory so the linker can find them.
+	if gofips140 := os.Getenv("GOFIPS140"); gofips140 != "" && gofips140 != "off" && gofips140 != "latest" {
+		if err := installFIPSSnapshotArchives(goenv, output); err != nil {
+			return fmt.Errorf("failed to install FIPS snapshot archives: %v", err)
```
r-hang: nit: you can remove the "failed to" and just use `"install FIPS snapshot archives: %w"` as the error message.

go/tools/builders/stdlib.go
```diff
@@ -170,5 +187,67 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
+			continue
+		}
+		dst := filepath.Join(pkgDir, filepath.FromSlash(pkg.ImportPath)+".a")
+		if _, err := os.Stat(dst); err == nil {
+			continue
+		}
+		if err := os.MkdirAll(filepath.Dir(dst), 0o777); err != nil {
+			return err
+		}
+		if err := copyFile(pkg.Export, dst); err != nil {
```
r-hang: as mentioned above in https://github.com/bazel-contrib/rules_go/pull/4631/changes#r3377550090, I think we can get rid of this `go list` call or at the very least share the result of this go list call with the linker configuration logic above.

go/tools/builders/stdliblist.go
```diff
@@ -239,7 +239,11 @@ func stdliblist(args []string) error {
 	defer func() { cleanup() }()
 
 	newGoRoot := filepath.Join(cloneBase, goenv.sdk)
-	if err := replicate(abs(goenv.sdk), abs(newGoRoot), replicatePaths("src", "pkg/tool", "pkg/include")); err != nil {
+	replicateDirs := []string{"src", "pkg/tool", "pkg/include"}
+	if v := os.Getenv("GOFIPS140"); v != "" && v != "off" {
+		replicateDirs = append(replicateDirs, "lib")
```
r-hang: same comment above on "lib/"

go/tools/builders/importcfg.go
adnanbhanji: Agreed on the direction, to make packages.txt the single source of truth for the GOFIPS140 snapshot packages, dropping both the go list -export copy and the importcfg walk.

Enumeration's feasible and hermetic: the version resolves from `lib/fips140/<value>.txt (v1.0.0 → v1.0.0-c2097c7c)` and the package set comes straight from the snapshot zip layout, gives exactly the 27 packages matching the binary's symbols, no go list.

I tried wiring it but hit a circular dependency:

`go_sdk → :package_list → :builder → :go_sdk`

Converting package_list into a compiled builder is the snag: I had it run the existing :builder, but :builder is compiled with sdk = ":go_sdk", and go_sdk already depends on :package_list, so it loops. The old rule avoided this only by being pure Starlark (no compiled tool). So the blocker is bootstrap: any compiled Go tool needs a compiler, and the only one here is the SDK we're still defining.

Any thoughts on how you'd approach this?

r-hang: I see, if that's the case could we execute a bash program to unzip and list the directory contents as replacement logic for the current package list builder?

adnanbhanji: Yes, that worked. I used a checked-in bash script (`generate_fips_package_list.sh`) run as a shell action, no compiled tool, so no cycle which let me do the unification:

- `packages.txt` now includes the snapshot packages, so it's the single source the importcfg reads from, dropped the `filepath.Walk` fallback in `importcfg.go`.
- `installFIPSSnapshotArchives` no longer calls `go list`; it runs `go build -o` on each snapshot package with the same flags `go install std` uses.

The builder needs to know the package set before `packages.txt` exists (consuming it would reintroduce the cycle), so it enumerates the zip via a small Go helper. To avoid that drifting from the script, `fips_test.go` runs both over the same fixture and fails if they differ.

go/tools/builders/fips.go
```diff
@@ -0,0 +1,70 @@
...
+// alias (e.g. v1.0.0 -> v1.0.0-c2097c7c), then the matching zip is read. Entries
+// look like golang.org/fips140@<ver>/fips140/<v>/<pkg>/<file>.go and map to the
+// import path crypto/internal/fips140/<v>/<pkg>.
+//
+// IMPORTANT: this MUST produce the same package set as
+// go/private/rules/generate_fips_package_list.sh, which performs the same
+// enumeration to build packages.txt for the linker's importcfg. If the two
+// diverge, the importcfg (from packages.txt) and the on-disk archives (built
+// here) will mismatch at link time. fips_test.go pins the two together.
+func fipsSnapshotPackages(libDir, gofips140 string) ([]string, error) {
```
r-hang: Is this function and the test code even needed? If we know at the packages.txt generation step all of the FIPS140 packages we need to account for and the mapping patterns is known to generate a list of packages we need to call `go build` for can't we generate that in the pacakges.txt step as well? Is there any logic that can't be done in a bash script?

I was thinking the step that calls packages.txt can emit two files, one a packages.txt and a list of GOFIPS140 packages that need to be install installed that the stdlib builder can read. The 2nd part can be removed once go upstream adds GOFIPS140 related packages to the package tree instead of the build cache, we should leave a note for that in the logic.

go/tools/builders/stdlib.go
```diff
@@ -158,17 +175,69 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
+// `go install std` so the archives stay compatible with the rest of pkg/.
+func installFIPSSnapshotArchives(goenv *env, goroot string, buildFlags []string) error {
+	goos := os.Getenv("GOOS")
+	goarch := os.Getenv("GOARCH")
+	if goos == "" || goarch == "" {
+		return fmt.Errorf("GOOS or GOARCH not set")
+	}
+	pkgDir := filepath.Join(goroot, "pkg", goos+"_"+goarch)
+
+	pkgs, err := fipsSnapshotPackages(filepath.Join(goroot, "lib", "fips140"), os.Getenv("GOFIPS140"))
```
r-hang: As mentioned above, I think this could reference the contents of a generated file instead of being computed in fipsSnapshotPacakges.

go/tools/builders/stdlib.go
```diff
@@ -158,17 +175,69 @@ You may need to use the flags --cpu=x64_windows --compiler=mingw-gcc.`)
...
+	}
+	for _, imp := range pkgs {
+		dst := filepath.Join(pkgDir, filepath.FromSlash(imp)+".a")
+		if _, err := os.Stat(dst); err == nil {
+			continue
+		}
+		if err := os.MkdirAll(filepath.Dir(dst), 0o777); err != nil {
+			return err
+		}
+		buildArgs := goenv.goCmd("build")
```
r-hang, line 247: Could we mention the reason for using go build? the alternative would be to use 'go list' to find all of the go installed archives and copy them over to the package tree. Can we also reference the upstream ticket and explain why the 'go install' already done here is not sufficient?

Could we also test that go build fetches from the existing go build cache. If it does not, it might just be better to go back to the go list approach.

go/private/rules/generate_fips_package_list.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Generates packages.txt for a GOFIPS140 snapshot build: the normal standard
+# library package list plus the versioned FIPS module packages, listed from the
+# snapshot zip under lib/fips140 (which Starlark cannot read).
+#
+# IMPORTANT: the FIPS enumeration below MUST produce the same package set as
```
r-hang: As commented below, if the enumeration must be the same let's not replicate the logic but just produce all the outputs we need here.

go/private/rules/generate_fips_package_list.sh
```diff
@@ -0,0 +1,34 @@
...
+# fipsSnapshotPackages in go/tools/builders/fips.go, which builds the matching
+# archives. If the two diverge, the linker's importcfg (from this packages.txt)
+# and the on-disk archives will mismatch at link time. fips_test.go pins them
```
r-hang: can't the fips_test.go go replaced by a test that runs this bash script against a lightweight file tree of the lib/fips140 shipped in go distributions? I think this input / output test would be minimal.

go/tools/builders/fips.go
adnanbhanji: Done, dropped the Go enumeration entirely. The package-list step now emits two files: packages.txt + a dedicated FIPS list that the stdlib builder reads. Also left a note that the second file can go away once upstream lands the GOFIPS140 packages in the package tree.

go/tools/builders/stdlib.go
adnanbhanji: Added a comment with the reasoning + golang/go#76225 (the go install std we already do leaves these in the build cache, not pkg/). And I checked the cache concern: with the same flags, go build is a pure cache hit, go build -x just copies the cached .a, no recompile, so it reuses exactly what go install built.

go/private/rules/generate_fips_package_list.sh
adnanbhanji: Done, fips_test.go is now an input/output test: it builds a small lib/fips140 fixture, runs the script, and checks both emitted files.

go/private/actions/stdlib.bzl
```diff
@@ -61,6 +61,7 @@ def _should_use_sdk_stdlib(go):
             not go.mode.msan and
             not go.mode.pure and
             not go.mode.gc_goopts and
+            not go.sdk.gofips140 and  # FIPS stdlib must be rebuilt from source
```
r-hang, line 64: nit: let's explain why this is true.

go/private/rules/sdk.bzl
```diff
@@ -127,6 +165,20 @@ package_list = rule(
             allow_single_file = True,
             doc = "A file in the SDK root directory. Used to determine GOROOT.",
         ),
+        "gofips140": attr.string(
+            default = "",
+            doc = "GOFIPS140 version. For a snapshot version, the FIPS snapshot " +
```
r-hang, line 170: nit: provide example values in the docstring

go/private/context.bzl
```diff
@@ -34,11 +39,6 @@ load(
     "@rules_cc//cc:find_cc_toolchain.bzl",
     "find_cc_toolchain",
 )
-load(
```
r-hang, line 37: nit: is this change needed? let's keep things strictly to what we need for fips140 support

go/tools/builders/fips_test.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// TestGenerateFIPSPackageListScript is an input/output test for
+// generate_fips_package_list.sh: it builds a lightweight lib/fips140 tree (a
+// snapshot zip + alias file) mirroring a Go distribution, runs the script, and
+// checks both emitted files. Skipped when the script or the shell tools it needs
+// are unavailable (e.g. `go test` without Bazel runfiles).
+func TestGenerateFIPSPackageListScript(t *testing.T) {
+	// generate_fips_package_list.sh is a POSIX shell script that only runs in the
+	// SDK's package-list step on Linux/macOS. On Windows the msys2 tools may be on
```
r-hang, line 66: If this feature isn't support for windows yet, we should fail as high up and early as possible as soon as gofips140 flag is set in the sdk with a link to a ticket describing the issue with windows support.

go/tools/builders/fips_test.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+func TestFIPSPackagesFromList(t *testing.T) {
+	// The dedicated FIPS list has one versioned snapshot package per line; blank
+	// lines are ignored and order is preserved.
+	path := filepath.Join(t.TempDir(), "fips_packages.txt")
+	contents := `crypto/internal/fips140/v1.0.0-c2097c7c
+crypto/internal/fips140/v1.0.0-c2097c7c/aes
+crypto/internal/fips140/v1.0.0-c2097c7c/aes/gcm
+
```
r-hang, line 37: what's the purpose of having a blank line here? does this replicate what actually happens?

go/tools/builders/fips_test.go
```diff
@@ -0,0 +1,172 @@
...
+		t.Fatalf("fipsPackagesFromList mismatch:\n got: %v\nwant: %v", got, want)
+	}
+}
+
+// TestGenerateFIPSPackageListScript is an input/output test for
+// generate_fips_package_list.sh: it builds a lightweight lib/fips140 tree (a
+// snapshot zip + alias file) mirroring a Go distribution, runs the script, and
+// checks both emitted files. Skipped when the script or the shell tools it needs
+// are unavailable (e.g. `go test` without Bazel runfiles).
+func TestGenerateFIPSPackageListScript(t *testing.T) {
```
r-hang, line 64: Given how complex this is, i'm not sure a go unit test is the right way to test this. Since the script itself is written in bash, could we use a bash script to test it as well using a `sh_test` rule? I think the first test makes sense, but there are so many skip statements in the second test gives me the sense that the this isn't the right place for the test.

adnanbhanji: Took a look at moving this to sh_test, but I'm hesitant to. The test needs a fixture .zip to feed the script, and with Go we can build that in-memory with archive/zip, no external tools. A bash sh_test would need either a checked-in binary .zip (the repo has zero of those) or the zip CLI at runtime which adds back the kind of dependency/skip the existing test had.

What I did instead was fix the actual problem you flagged - the skips. Removed all of them: Windows is now excluded via target_compatible_with in the BUILD file, and the remaining runfiles checks are hard failures instead of skips. So it's still a Go test but with zero skip statements now.

go/private/context.bzl
adnanbhanji: Good catch - not needed. That was a buildifier-driven load reorder, not related to fips140. Reverted it; the only change in this file now is the GOFIPS140 env line.

go/tools/builders/fips_test.go
adnanbhanji: removed it so fixture matches the actual format

go/tools/builders/fips_test.go
adnanbhanji: Done - added an early fail() in define_sdk_repository_targets that triggers as soon as a GOFIPS140 snapshot is set on a Windows SDK

go/private/rules/generate_fips_package_list.sh
```diff
@@ -0,0 +1,38 @@
...
+# Resolve an alias (e.g. v1.0.0 -> v1.0.0-c2097c7c) when an alias file exists.
+if [ -f "$libdir/$gofips.txt" ]; then
+  ver="$(cat "$libdir/$gofips.txt")"
+else
+  ver="$gofips"
+fi
+
+# Zip entries look like golang.org/fips140@<ver>/fips140/<ver>/<pkg>/<file>.go,
+# which map to the import path crypto/internal/fips140/<ver>/<pkg>.
+unzip -Z1 "$libdir/$ver.zip" \
```
r-hang, line 30: I think we can improve upon this further with a pure skylark implementation, something that looks like. Using a bash script that relies on system commands like unzip will make this less portable.

```
+def _fips_snapshot_packages(ctx, gofips140):
+    lib_dir = "lib/fips140"
+
+    # Resolve an alias (e.g. certified, inprocess, or v1.0.0) to the concrete
+    # version (e.g. v1.0.0-c2097c7c) via lib/fips140/<value>.txt when present.
+    alias_file = ctx.path("{}/{}.txt".format(lib_dir, gofips140))
+    ver = ctx.read(alias_file).strip() if alias_file.exists else gofips140
+
+    zip_path = ctx.path("{}/{}.zip".format(lib_dir, ver))
+    if not zip_path.exists:
+        fail("GOFIPS140 snapshot build requested (gofips140 = '{}'), but the ".format(gofips140) +
+             "snapshot zip {} was not found in the SDK.".format(zip_path))
+
+    # Extract to a throwaway directory so we can enumerate entries, then remove it.
+    # Layout inside the zip:
+    #   golang.org/fips140@<ver>/fips140/<ver>/<pkg...>/<file>.go
+    # which maps to the import path crypto/internal/fips140/<ver>/<pkg...>.
+    tmp = "_fips_pkglist_tmp"
+    ctx.extract(archive = zip_path, output = tmp)
+    root = ctx.path("{}/golang.org/fips140@{}/fips140/{}".format(tmp, ver, ver))
+
+    packages = {}
+    if root.exists:
+        stack = [(root, "crypto/internal/fips140/" + ver)]
+        for _ in range(1000000):
+            if not stack:
+                break
+            d, imp = stack.pop()
+            for entry in d.readdir():
+                if entry.is_dir:
+                    stack.append((entry, imp + "/" + entry.basename))
+                elif entry.basename.endswith(".go") and not entry.basename.endswith("_test.go"):
+                    packages[imp] = None
+
+    ctx.delete(tmp)
+    return sorted(packages.keys())
```

samgurtman:
Is this still actively worked on?

fmeum:
Let me know when you want a full review!

go/private/rules/generate_fips_package_list.sh
blampe: Agree, unzip isn't hermetic here.

genert:
Ping

adnanbhanji:
Hey all, thanks for the pings. I'm going to be working on the pure skylark implementation this week as @r-hang suggested above. @fmeum and others feel free to review it already, I'll actively be working on it.

PR #4633, opened 2026-06-08
Make go/tools/builders absolute paths robust to PWD=/proc/self/cwd.

reltuk:
**What type of PR is this?**

Bug fix

The command invocations and environment variable handling in ./go/tools/builders are careful to construct absolute paths, so that those paths continue to work from new working directories. However, if PWD=/proc/self/cwd, then the computed absolute paths actually change meaning as the working directory changes.

This became acutely relevant with rules_cc 0.2.19, when rule-based cc toolchains that took
`experimental_replace_legacy_action_config_features` picked up a new feature, `sanitize_pwd`. That feature sets PWD=/proc/self/cwd in the environment of archive and link actions. The end result is that those toolchains on that rules_cc version broke with rules_go. Certain builder invocations merge the cc command context, and certain `go` invocations chdir before using the supplied paths.

**Other notes for review**

I wasn't familiar with and didn't investigate the e2e tests, so this PR is lacking one. The requirements on rules_cc 0.2.19 and platform=linux and a rule-based cc toolchain seemed pretty heavy, but maybe this is not so complicated.

I have made a minimal repro repository here: https://github.com/reltuk/sanitize_pwd_repro. I have verified that this branch fixes it. Caveat emptor: the repro repository was created by an LLM–sorry for the verbosity of the README and some of the comments throughout.

The sanitize_pwd feature can be seen in rules_cc here (current tip-of-main): https://github.com/bazelbuild/rules_cc/blob/8edac4e1d8f164e1e5b559598213e57fd74dd8e8/cc/toolchains/args/sanitize_pwd/BUILD

### DIFF
--- a/go/tools/builders/env.go
+++ b/go/tools/builders/env.go
@@ -415,12 +415,23 @@ func abs(path string) string {
 	if strings.HasPrefix(path, "__BAZEL_") {
 		return path
 	}
-
-	if abs, err := filepath.Abs(path); err != nil {
-		return path
-	} else {
-		return abs
+	if filepath.IsAbs(path) {
+		return filepath.Clean(path)
 	}
+	// Instead of filepath.Abs(), we EvalSymlinks on os.Getwd.  On
+	// Linux, PWD=/proc/self/cwd will cause
+	// filepath.Abs("sometool") to return
+	// "/proc/self/cwd/sometool". But that path does not behave
+	// like an absolute path. When a `go` invocation chdirs, that
+	// suddenly points somewhere else.
+	wd, err := os.Getwd()
+	if err != nil {
+		return path
+	}
+	if real, err := filepath.EvalSymlinks(wd); err == nil {
+		wd = real
+	}
+	return filepath.Join(wd, path)
 }
 
 // absArgs applies abs to strings that appear in args. Only paths that are
--- a/go/tools/builders/env_test.go
+++ b/go/tools/builders/env_test.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


go/tools/builders/env.go
```diff
@@ -415,12 +415,23 @@ func abs(path string) string {
...
+	// "/proc/self/cwd/sometool". But that path does not behave
+	// like an absolute path. When a `go` invocation chdirs, that
+	// suddenly points somewhere else.
+	wd, err := os.Getwd()
+	if err != nil {
 		return path
-	} else {
-		return abs
 	}
+	if real, err := filepath.EvalSymlinks(wd); err == nil {
```
fmeum, line 431: `EvalSymlinks` is famous for breaking someone else's use case that you've never considered before, so we've grown weary of it. 

Could you try adding `sanitize_pwd` to https://github.com/bazel-contrib/rules_go/blob/8f974315778d3f75b52087b84c5b3fa331abfeb8/go/private/context.bzl#L137 instead?

reltuk: Makes perfect sense. Adding `sanitize_pwd` to this list fixes the repro.

Is this feature actually useful for reproducibility and/or sanitizing PWD leaking into link output? Like, in the context of Cgo, is it actually nice for the C toolchain to see an env with `PWD=/proc/self/cwd`?

An alternative which lets the feature stay and is probably much safer than `filepath.EvalSymlinks` is to single-hop ReadLink if Getwd == "/proc/self/cwd". Something like:

```go
wd, err := os.Getwd(path)
if err != nil {
    return path
}
if wd == "/proc/self/cwd" {
    if resolved, err = os.ReadLink(wd); err == nil {
        wd = resolved
    }
}
return filepath.Join(wd, resolved)
```

Either way I'm happy to close this PR and open a separate one. Just let me know.

For either case, any guidance on the need for tests in `./tests`? I could add something to `./tests/bcr` if I bump `rules_cc` but whatever rules-based rules_cc toolchain I landed would be somewhat environment dependent and wouldn't have all the machinery of `local_config_cc` (or would be very heavy weight blobs, which doesn't seem like something we would want).

Thank you so much for your help and guidance.

fmeum: It's not needed with clang, but as far as I remember GCC doesn't have all the requires knobs. It does help with older compiler versions too.

I agree that the readlink would be safer, but I still think that disabling the feature is the least likely to cause other problems down the road. Trust me when I say that 10% of all Bazel issues I've ever debugged were related to symlinks 😅

reltuk:
Sounds great! I'll be back soon with a PR that adds `sanitize_pwd` to `_UNSUPPORTED_FEATURES`. Closing this for now :).

PR #4635, opened 2026-06-10
fix: add sanitize_pwd to unsupported cc toolchain features.

reltuk:
**What type of PR is this?**

Bug fix

The recently-added `sanitize_pwd` feature from rules_cc interacts poorly with builders' absolute path calculation. It can cause Cgo compiles to fail as a result.


reltuk:
Previously discussed here: #4633. I have verified that this change fixes the repro repository available at https://github.com/reltuk/sanitize_pwd_repro. Thanks for the elegant suggestion :).

fmeum (approved):

PR #4637, opened 2026-06-18
Preserve common Go settings in go_stdlib_transition

dzbarsky:
## Summary

Author: zbarsky-bot

Start `go_stdlib_transition` from all values in `_reset_transition_keys`; the transition then only filters `//go/config:tags` and disables `//go/private:bootstrap_nogo` and `//command_line_option:collect_code_coverage`. This keeps each Go rule and its `//:stdlib` dependency in the same configuration when resolving the C++ toolchain.

Since #4599 removed `cgo_context_data`, each Go rule resolves the C++ toolchain directly. Resetting `//go/private:request_nogo` made ordinary Go rules and `//:stdlib` resolve configuration-specific C++ runtime archives in different configurations, so `GoLink` could reference an archive that Bazel did not declare. Preserving every `_reset_transition_keys` value prevents `//go/config:static`, `//go/config:debug`, or future settings from recreating the same mismatch.

The regression test verifies the two required `//:stdlib` configurations for an ordinary Go target and `coverdata`. `//:stdlib` has no `_nogo` attribute, so preserving `//go/private:request_nogo` does not run nogo on `//:stdlib`.

## Testing

- `bazel test //tests/core/transition:hermeticity_test`
- `bazel test //tests:buildifier_test`

### DIFF
--- a/go/private/rules/transition.bzl
+++ b/go/private/rules/transition.bzl
@@ -193,15 +193,6 @@ _reset_transition_dict = dict(_common_reset_transition_dict, **{
 
 _reset_transition_keys = sorted(_reset_transition_dict.keys())
 
-_stdlib_keep_keys = sorted([
-    "//go/config:msan",
-    "//go/config:race",
-    "//go/config:pure",
-    "//go/config:linkmode",
-    "//go/config:tags",
-    "//go/config:pgoprofile",
-])
-
 def _go_tool_transition_impl(settings, _attr):
     """Sets most Go settings to default values (use for external Go tools).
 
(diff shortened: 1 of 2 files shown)


scasagrande:
fixed my use case on macos with hermetic-llvm

go/private/rules/transition.bzl
```diff
@@ -200,6 +200,9 @@ _stdlib_keep_keys = sorted([
     "//go/config:linkmode",
     "//go/config:tags",
     "//go/config:pgoprofile",
+    # Preserve //:stdlib's C++ toolchain configuration even though request_nogo
```
fmeum: This comment is a bit weird, it doesn't really explain why we want to have this here.

fmeum: Thinking about this some more, does any difference in the configuration result in a mismatch? What if the user has any other flag transitioned that isn't in this dict?

dzbarsky: It needs to be a difference between the Go rule being compiled and the stdlib. So not any flag, but anything in _common_reset_transition_dict could be a hazard... Looks like "//go/config:static" and "//go/config:debug" Perhaps we should be keeping all keys?

fmeum: Yeah, that makes sense. Could you enforce this in code and update the comment?

go/private/rules/transition.bzl
```diff
@@ -245,17 +236,17 @@ non_go_tool_transition = transition(
 )
 
 def _go_stdlib_transition_impl(settings, _attr):
-    """Sets all Go settings to their default values, except for those affecting the Go SDK.
+    """Preserves common Go settings in the standard library configuration.
```
fmeum: This is a weird one-line description as this *does* reset certain settings.

fmeum (approved):

PR #4638, opened 2026-06-22
Declare GoArchive in rule provider declarations

### LINKED ISSUE #4612: Add `GoArchive` to `rule(provides = [...])` to better support `aspect(required_providers = [...])`
Aspects are able to self-filter using `required_providers`, allowing them to be skipped _before_ their implementation function runs. This optimisation greatly reduces the impact on builds that do not need the provider, as an unused aspect is never configured (required to run the implementation function).

Skipping aspect configuration means any external repositories required to analyse aspect dependencies (e.g. for an executable) do not need to be fetched. It is the difference between fetching Go repositories vs. Go, Java, CC and Rust (etc).

However `required_providers` only considers pre-declared providers (`rule(provides = [...])`).

Various rules already document their always-returned providers;
- `go_binary` -> `GoArchive`
- `go_library` -> `GoInfo` (already pre-declared), `GoArchive`
- `go_test` _does not_ document returned providers, testing showed `GoArchive` is returned.

dzbarsky:
**What type of PR is this?**

Bug fix

`go_binary`, `go_library`, and `go_test` return `GoArchive`, but their rule declarations did not consistently list `GoArchive` in `provides`. Bazel therefore skipped aspects that use `required_providers = [GoArchive]` before running the aspect implementation. Declare `GoArchive` for all three rules and add analysis tests that verify the aspect runs for each rule.

Fixes #4612

**Other notes for review**

Validated with `bazel test --jobs=4 //tests/core/starlark:all` and `bazel test --jobs=4 //tests:buildifier_test`.

### DIFF
--- a/go/private/rules/binary.bzl
+++ b/go/private/rules/binary.bzl
@@ -257,6 +257,7 @@ def _go_binary_kwargs(go_cc_aspects = []):
     return {
         "cfg": go_transition,
         "implementation": _go_binary_impl,
+        "provides": [GoArchive],
         "attrs": {
             "srcs": attr.label_list(
                 allow_files = go_exts + asm_exts + cgo_exts + syso_exts,
--- a/go/private/rules/library.bzl
+++ b/go/private/rules/library.bzl
@@ -31,6 +31,7 @@ load(
 )
 load(
     "//go/private:providers.bzl",
+    "GoArchive",
     "GoInfo",
 )
 load(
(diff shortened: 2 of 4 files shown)


fmeum (approved):
@jayconrod Could you confirm that this wasn't intentional?

jayconrod (approved):
I think it was only omitted because `GoArchive` and other providers were introduced before `provides` and `required_providers` were standard.

PR #4641, opened 2026-06-25
[builder/cgo] escape ldflags for invoking cgo starting Go 1.27

### LINKED ISSUE #4640: Go 1.27: ldflags file written for cgo enabled targets are being expanded before going through flag parsing
In Go 1.27, the cgo tool gains global GCC-compatible @file response file expansion — all arguments starting with @ are expanded (tokenized by whitespace/quoting rules) before flag parsing begins (via the change to src/cmd/internal/objabi/flag.go in [CL 737500](https://go-review.googlesource.com/c/go/+/737500)).

Rules_go passes ldflags to cgo as -ldflags @<tmpfile>, writing the file as a plain space-joined string:
```
-fsanitize-undefined-strip-path-components=0 -lm -lpthread
```
Before Go 1.27: cgo handled @<file> within its own -ldflags flag handler, reading the file content as a single raw string. The ldflags string was passed correctly to the linker.

Go 1.27+: cgo now expands @<file> globally before flag parsing. The file is GCC-tokenized on whitespace, injecting multiple tokens into cgo's own argument list:
```
# original args passed by rules_go
cgo -ldflags @/tmp/cgo-ldflags-xxx.txt -- -some-cflag file.go

# after @file expansion in Go 1.27
cgo -ldflags -fsanitize-undefined-strip-path-components=0 -lm -lpthread -- -some-cflag file.go
```
-ldflags consumes the first token as its value; the remaining tokens (-lm, -lpthread, etc.) are parsed as unknown cgo flags, producing:
```
flag provided but not defined: -fsanitize-undefined-strip-path-components
```
The fix: wrap the file content in double quotes so GCC tokenization produces exactly one token — the full quoted string becomes the value of -ldflags:
```
"-fsanitize-undefined-strip-path-components=0 -lm -lpthread"
```

tchung1118:
**What type of PR is this?**

Bug fix

Go 1.27 changes how the cgo tool expands `@file` response file arguments. Starting in Go 1.27, any argument beginning with @ is expanded before flag parsing using GCC-compatible tokenization — the file content is split on whitespace into multiple tokens.

Rules_go passes cgo its linker flags via -ldflags @<tmpfile>, writing the file as a plain space-joined string (e.g., -fsanitize-undefined-strip-path-components=0 -lm -lpthread). Before Go 1.27, cgo handled @<file> inside its own -ldflags flag handler, treating the file content as a single string. With the new global expansion, the file is tokenized before flag parsing, injecting each whitespace-separated flag as an independent argument into cgo's own flag space. The first token is consumed as the value of -ldflags, and the rest are parsed as unknown cgo flags, producing an error similar to the following:
```
flag provided but not defined: -fsanitize-undefined-strip-path-components
```

This PR fixes the issue by wrapping the ldflags file content in double quotes (with proper escaping of \ and ") when building with Go 1.27+. GCC tokenization then yields exactly one quoted token — the complete ldflags string — which is correctly consumed as the value of -ldflags. A separate build-tag-gated file preserves the original behavior for older Go versions.

Fixes #4640

**Other notes for review**

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -118,7 +118,7 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 			return "", nil, nil, fmt.Errorf("failed to create temporary file for ldflags: %w", err)
 		}
 		defer os.Remove(ldflagsFile.Name())
-		if _, err := ldflagsFile.WriteString(combinedLdFlagsStr); err != nil {
+		if _, err := ldflagsFile.WriteString(formatLdFlagsFileContent(combinedLdFlagsStr)); err != nil {
 			ldflagsFile.Close()
 			return "", nil, nil, fmt.Errorf("failed to write ldflags to temporary file: %w", err)
 		}
@@ -463,6 +463,15 @@ func copyOrLinkFile(inPath, outPath string) error {
 	} else {
 		return linkFile(inPath, outPath)
 	}
+}
+
+func formatLdFlagsFileContent(flags string) string {
+	shouldEscape, _ := onVersionOrHigher(27)
+	if shouldEscape {
+		escaped := strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(flags)
+		return `"` + escaped + `"` + "\n"
+	}
+	return flags
 }
 
 func onVersionOrHigher(version int) (bool, error) {


fmeum (approved):

fmeum (commented):
You need to update the list of builder sources

tchung1118:
> You need to update the list of builder sources

Seems like the build tags are not working on builder sources, so I've switched to checking go version in runtime. Also, CI seems to keep failing due to git checkout failures for some reason.

tchung1118:
CI seems to be passing now.

PR #4645, opened 2026-07-02
compilepkg: skip duplicate cgo compile

sluongng:
Coverage+cgo packages currently spend extra CPU in compilepkg because
nogo needs original cgo-generated Go sources, not the
coverage-instrumented files passed to the compiler. The old path called
cgo2 a second time to produce those sources. That second call also
recompiled every generated and user C-family source, linked the cgo
probe, and ran dynimport even though compilepkg discarded the object
outputs.

Split cgo source generation from the compile/link phase and add a
generate-only path for the nogo copy. Normal cgo compilation still uses
the full path, while the coverage+nogo pass writes generated Go sources
from a separate cgo_nogo work directory and reuses the real
_cgo_imports.go when available.

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -62,164 +62,22 @@ func cgo2(goenv *env, goSrcs, cgoSrcs, cSrcs, cxxSrcs, objcSrcs, objcxxSrcs, sSr
 	}
 	defer cleanup()
 
-	// cgo2 will gather sources into a single temporary directory, since nogo
-	// scanners might want to include or exclude these sources we need to ensure
-	// that a fragment of the path is stable and human friendly enough to be
-	// referenced in nogo configuration.
-	workDir = filepath.Join(workDir, "cgo", packagePath)
-	if err := os.MkdirAll(workDir, 0700); err != nil {
-		return "", nil, nil, err
-	}
-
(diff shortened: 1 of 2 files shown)


sluongng:
Tested this patch on some CGO target in BuildBuddy repo today and got this 

```
   Metric               Baseline     Patched    Improvement
  ━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━━
   Median execution     356.8 ms    324.6 ms           9.0%
  ───────────────────  ──────────  ──────────  ─────────────
   Mean execution       381.6 ms    326.2 ms          14.5%
  ───────────────────  ──────────  ──────────  ─────────────
   p95 execution        388.3 ms    360.8 ms           7.1%
  ───────────────────  ──────────  ──────────  ─────────────
   Mean executor CPU    475.6 ms    434.9 ms           8.6%
```

ran 50 times on baseline and 50 times on patched and patched was faster 48 times

@dzbarsky you might be a good reviewer for cgo stuffs here?

sluongng:
Splitted the PR into 2 smaller commits, one for refactoring and one for re-using code, so that it's a bit easier to review.

dzbarsky (approved):
thanks! splitting made it much clearer

PR #4648, opened 2026-07-02
Support --force_pic for cgo interop

keith:
If you set --force_pic on linux and then link go into a cc_binary, the
C++ code is likely built with pic but the go code was not, leading to
linker failures. Now this default bazel flag is respected.


keith:
technically this isn't 100% correct since the configured cc toolchain could also choose to ignore this flag entirely. so maybe we'd be better off reading the features directly, but that's a lot more complex

fmeum (approved):

PR #4650, opened 2026-07-08
Build Go SDK file attrs in the exec configuration

fionera:
**What type of PR is this?**

Bug fix

The experimental source-bootstrapped SDK (`experimental_bootstrap_go_sdk`) emits the `go` binary, the tool binaries, and the GOROOT `src`/`pkg` layout from a generated bootstrap target. `go_tool_binary` runs the SDK's `go` binary in the exec configuration — the `go` and `tools` attrs of `go_sdk` are already `cfg = "exec"` — so that binary infers `GOROOT` from its own exec-config path.

The remaining SDK file attrs (`root_file`, `libs`, `headers`, `srcs`) had no configuration set, so Bazel supplied them under the target-config path. When those inputs don't sit under the exec-config `GOROOT` the `go` binary derives from its own location, the Go command can't find standard-library packages such as `cmd/pack`, and the bootstrap builder build fails.

This pins `root_file`, `libs`, `headers`, and `srcs` to `cfg = "exec"` so the whole SDK layout lands in one consistent configuration alongside `go` and `tools`. `package_list` is intentionally left unpinned: it is a generated list of standard-library import paths whose contents do not depend on the configuration.

Small bug fix in the experimental bootstrap SDK — no associated issue.

**Other notes for review**

This only changes how the `go_sdk` rule's own inputs are configured. For the normal downloaded SDK these attrs point at static source files, so they resolve to the same artifacts regardless of the transition; the fix matters for the bootstrap SDK, where the layout is generated and would otherwise be produced under the target configuration while `go` runs under the exec one.


fmeum (approved):

go/private/rules/sdk.bzl
```diff
@@ -76,16 +77,19 @@ go_sdk = rule(
             # allow_files is not set to [".a"] because that wouldn't allow
             # for zero files to be present, as is the case in Go 1.20+.
             # See also https://github.com/bazelbuild/bazel/issues/7516
+            cfg = "exec",
```
fmeum: I guess this is technically incorrect (libs need to be built for the target platform) but doesn't matter since no recent Go SDK has any of these. You can either drop this entire attr or revert to target (or leave a comment saying that it shouldn't matter).

fionera: Done!

fmeum (approved):

Issue #2258, opened 2019-10-25
-trimpath not turned off in debug mode


Meai1:
The bazel rules always seem to add the trimpath setting but VSCode and Golang IDE both expect full symbol paths in the binary. 
I thought that this should be solved on the delve debugger side but if they dont want to solve it then I either have to solve it in the build system or in the IDE/IDE plugins. 
I do feel like the build system is the next easiest place to solve this because naturally this is a setting in the go build system anyway, I just dont see where it is available in rules_go via bazel. Looking at the source here, it seems to be set all over the place with no easy switch to turn it off. 

Here is my problem described:
https://github.com/go-delve/delve/issues/1730

arifogel:
I submitted a PR to fix this for VS code: microsoft/vscode-go#2908

davidair:
Would it be possible to introduce a flag that allows to strip "-trimpath" for local developer builds? While reproducible builds are important in the CI/CD pipeline, the ability to debug locally is something that is very important for developers. I had trouble with a number of debuggers (VS Code, GDB, IntelliJ) with my bazel-built binaries (IntelliJ did best but some things still didn't work quite right).

Issue #1708, opened 2018-09-03
Missing external sources when debugging with dlv


prestonvanloon:
To reproduce:
1) Have some bazel project with an external library provided by go_repository.
1) Build with `bazel build -c dbg //program:main`
1) Execute `dlv exec bazel-bin/program/linux_amd64_debug/main`
1) Set a breakpoint in the external library `b github.com/abc/lib.(*Foo).Bar`
1) Enter `c` to continue the execution until the breakpoint is hit
1) List the current source code with `ls` and see `Command failed: open external/github.com/abc/lib/foo.go: no such file or directory` 

Is there any workaround besides not using bazel to build the debug binary?

jayconrod:
It doesn't seem like we're stripping directory prefixes correctly at the moment. I think there are basically four cases here:

* Static files in the main workspace: these should work.
* Generated files in the main workspace: these should work.
* Files in the standard library. These do not work because the paths contain the string `GOROOT`, e.g., `./GOROOT/src/internal/cpu/cpu.go`.
* Files in external repositories. These do not work because their path is relative to the execroot, not the workspace root, e.g., `./external/com_github_pkg_errors/stack.go`

devshorts:
Is there any progress this? This is pretty frustrating

dragonsinth:
I noticed the whole `external/com_github_pkg_errors/stack.go` source path formulation as we started building binaries in bazel and noticed that the stack traces differ now.  Is that the intended source path for external go files over something like `github.com/pkg/errors/stack.go`?

danny-skydio:
Something that works for me:

1. Build with `bazel build -c dbg //program:main`
2. Execute `dlv exec bazel-bin/program/linux_amd64_debug/main`
3. Map the paths using `config substitute-path external/ /path/to/workspace-name/bazel-workspace-name/external`

I put in an absolute path, but that's just out of convenience.
It should also work for mapping GOROOT.

sluongng:
Interesting 🤔 

Perhaps we can document somewhere that user can invoke Delve with [a special config file](https://github.com/go-delve/delve/blob/master/Documentation/cli/README.md)
where `substitute-path` is set to absolute path of `external` and `GOROOT` 🤔

stagnation:
I wrote up a little summary of how I usually debug go programs.
Where the config file or `--init` can be used to debug from multiple projects.
https://meroton.com/docs/tips/debug-buildozer-and-other-bazel-go-projects/

[...]

PR #4651, opened 2026-07-09
compilepkg: record importpath-relative source paths under -trimpath

luzhanning:
**What type of PR is this?**

Feature

Adds an opt-in build setting `--@io_bazel_rules_go//go/config:importpath_trimpath` that records source paths as importpath-relative, the same shape native `go build -trimpath` produces, instead of execroot-relative.

Today the recorded paths are things like `pkg/file.go` for the main repo and `external/gazelle++go_deps+com_github_x_y/file.go` for external deps. Debuggers can't do anything with those: the former is a relative fragment with no anchor, the latter is Bazel-internal naming. In practice this makes remote debugging impossible — if you attach GoLand/Delve to a binary running on a machine that doesn't have a Bazel execroot, every breakpoint fails to resolve. #2258 and #1708 are both this problem, and cockroachdb/cockroach#64379 asked for essentially this feature.

With the setting enabled, compilepkg prepends a `-trimpath` rewrite mapping each package's source directory to its importpath, so the recorded path becomes e.g. `example.com/mod/pkg/file.go`. GoLand resolves those through its module index with no configuration, and dlv needs at most one `substitute-path` rule. The cgo tool gets the same rewrite for its `//line` directives, which as a side effect makes cgo and non-cgo packages record consistent shapes (today cgo paths are trimmed relative to the package dir instead of the execroot).

The reproducibility concern that keeps `-trimpath` always-on (see #2258) still holds: the rewrite prefix is computed inside the builder at runtime, not on the action command line, so cache keys don't change, and the recorded paths are machine-independent. The execroot rewrite from #4435 stays in place as the fallback (Go's `-trimpath` is first-match-wins), so coverage-instrumented copies and generated sources are unaffected. The package source dir is captured before coverage instrumentation redirects sources to workDir copies.

The setting defaults to off because `runfiles.CurrentRepository()` / `CallerRepository()` infer the calling repository by parsing `external/<repo>/` out of `runtime.Caller` paths; with importpath-relative paths they'd report the main repository for every caller. This is documented in `go/modes.rst` along with the new setting.

Addresses #2258 and #1708 for users who opt in.

**Other notes for review**

- `tests/core/go_library:trimpath_test` now covers both modes in both repository layouts (default expectations are unchanged from today).
- External deps are recorded without the `@version` suffix that native `go build -trimpath` uses, since the compile action only knows the importpath. `go_repository` does know the version, so plumbing it through would make the output match the native toolchain exactly — seemed better as a follow-up.
- If there's interest in eventually making this the default, `CurrentRepository` would need a path-independent source of repository identity (e.g. an importpath→repository mapping emitted alongside the repo mapping manifest). Also follow-up material.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -101,6 +101,7 @@ go_config(
     gc_goopts = "//go/config:gc_goopts",
     gc_linkopts = "//go/config:gc_linkopts",
     gotags = "//go/config:tags",
+    importpath_trimpath = "//go/config:importpath_trimpath",
     linkmode = "//go/config:linkmode",
     msan = "//go/config:msan",
     pgoprofile = "//go/config:pgoprofile",
--- a/go/config/BUILD.bazel
+++ b/go/config/BUILD.bazel
@@ -69,6 +69,18 @@ string_flag(
     visibility = ["//visibility:public"],
 )
 
+# When enabled, source paths recorded under -trimpath are rewritten to be
+# importpath-relative (matching native `go build -trimpath`) instead of
+# execroot-relative, so debuggers can resolve them without configuration.
+# Note: runfiles.CurrentRepository()/CallerRepository() infer the caller's
+# repository from the execroot-relative path shape and report the main
+# repository for all callers when this is enabled.
+bool_flag(
+    name = "importpath_trimpath",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
 filegroup(
     name = "all_files",
     testonly = True,
--- a/go/modes.rst
+++ b/go/modes.rst
@@ -97,6 +97,18 @@ or using `Bazel configuration transitions`_.
 | but adds time to the initial build. Leave false unless you want to use       |
 | golangci-lint or another tool that relies on GOPACKAGESDRIVER.               |
 +------------------------+---------------------+-------------------------------+
+| :param:`importpath_trimpath` | :type:`bool`  | :value:`false`                |
++------------------------+---------------------+-------------------------------+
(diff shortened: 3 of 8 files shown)


fmeum:
I generally support this effort, but I'm not a fan of flags that break some use cases while fixing others. How about we try to find a solution for `CurrentRepository` now hat works with the new scheme and then adopt it by default?

Two ideas:
* Gazelle generates a mapping file between repo names and import paths in https://github.com/bazel-contrib/bazel-gazelle/blob/15d1f023b75ced3d89e1dfac9ac9e5e5e3cdd0e5/internal/bzlmod/go_deps.bzl#L876 that could become a runfile of the `runfiles` package.
* Perhaps we could augment the Go compile action to inject the external repo name into each compiled Go package and recover it via reflection.

@jayconrod Do you happen to see a good way to fix this?

fmeum:
A manifest mapping importpath to repo names seems possible: https://github.com/bazel-contrib/rules_go/pull/4652

@luzhanning Would you be interested in picking up that PR? Feel free to reuse any parts if so, I used Claude to stamp this out.

jayconrod:
I agree with @fmeum: I don't like adding a flag for this. We should find a solution that works for all use cases.

Sorry I don't have something more specific to suggest right now. @fmeum's ideas seem reasonable. We could inject a hint to make runfiles work. I'd prefer if we don't rely on `go_deps` though. Ideally, rules_go would not have a hard dependency on Gazelle or go_deps.

PR #4652, opened 2026-07-10
runfiles: implement CurrentRepository via a package repository map

fmeum:
CurrentRepository and CallerRepository currently recover the canonical repository name of the caller by matching the source file path recorded in the binary against the execroot-relative layout (external/<repo>/...). This breaks whenever recorded source paths are rewritten, e.g. with -trimpath or the proposed importpath-relative path scheme.

Instead, derive the caller's link-time package path from the symbol name reported by runtime.CallersFrames, which is unaffected by any source path rewriting, and look it up in a new per-binary file that maps the package path of every linked package to the canonical name of the repository containing its sources:

* The link action writes <binary>.package_repo_map next to the binary from the archives it already receives via -arc (extended with the owning repository name) and the main package, and the file is added to the binary's runfiles.
* The linker injects the runfiles-root-relative path of that file into the runfiles library via -X, but only if the library is actually linked into the binary.
* The runfiles library lazily loads the map and resolves the caller's package path to a repository name, trying successively longer package path candidates to disambiguate package paths that contain dots in their last element (e.g. gopkg.in/yaml.v2). The previous file path based logic is kept as a fallback for binaries built with older rules_go versions or outside of Bazel.

This also makes CurrentRepository work for callers in the main package of a binary in an external repository, which symbol names alone can't provide since main packages are always linked as "main".

### DIFF
--- a/go/private/actions/binary.bzl
+++ b/go/private/actions/binary.bzl
@@ -54,7 +54,7 @@ def emit_binary(
         else:
             extension = goos_to_extension(go.mode.goos)
         executable = go.declare_file(go, path = name, ext = extension)
-    go.link(
+    package_repo_map = go.link(
         go,
         archive = archive,
         test_archives = test_archives,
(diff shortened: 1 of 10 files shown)


luzhanning:
Picked this up in #4653 as suggested — kept your commit unchanged and fixed the CI failures on top (the map is now only emitted for executable link modes, plus two test expectation updates). Feel free to close this one in its favor.

PR #4655, opened 2026-07-11
Add dedicated load files for go_binary(), go_library(), go_test()

### LINKED ISSUE #4654: offer distinct load targets for go_binary / go_library / go_test
All three of these rules are only available directly through def.bzl and it turns out that this makes things extraordinarily complicated when translating to/from environments (lie, say, Google3) in which each has its own .bzl file. It seems to be the trend in other languages to split things up in this useful way so it would be awesome if we could do the same in Go.

ericsalo:
as alternates to def.bzl. This tremendously simplifies migration to and from other environments which expect the same separation and brings Go in line with common practice for other languages (for example, rules_python does this).

**What type of PR is this?**

Feature

It's a nice quality-of-life for amphibian developers.
go_binary() and go_library() and go_test() now each have their own separate .bzl files (in addition to def.bzl).

Fixes #4654

**Other notes for review**

These are all just one-liners which happen to wrap the same functions in go/private/ that are already wrapped by def.bzl, so they are a parallel (if trivial) code path. We could also be slightly more intrusive by defining them to be the new SoT and redirecting def.bzl to point to them instead.

### DIFF
--- a/go/def.bzl
+++ b/go/def.bzl
@@ -14,12 +14,13 @@
 
 """Public definitions for Go rules.
 
-All public Go rules, providers, and other definitions are imported and
-re-exported in this file. This allows the real location of definitions
-to change for easier maintenance.
+This file serves as the primary entry point for loading public Go rules,
+providers, and definitions. Re-exporting them here allows the underlying
+implementations to be maintained and relocated without breaking public APIs.
 
-Definitions outside this file are private unless otherwise noted, and
-may change without notice.
+Definitions outside this file are private and subject to change with the
+exception of (go/go_binary.bzl, go/go_library.bzl, go/go_test.bzl) which are provided as
+a convenience.
 """
 
 load(
--- a/tests/core/entry_points/BUILD.bazel
+++ b/tests/core/entry_points/BUILD.bazel
@@ -1,0 +1,21 @@
+load("//go:go_binary.bzl", "go_binary")
+load("//go:go_library.bzl", "go_library")
+load("//go:go_test.bzl", "go_test")
+
+go_library(
+    name = "entry_lib",
+    srcs = ["entry_lib.go"],
+    importpath = "entry_lib",
+)
+
+go_binary(
+    name = "entry_bin",
+    srcs = ["entry_bin.go"],
+    deps = [":entry_lib"],
+)
+
+go_test(
+    name = "entry_test",
+    srcs = ["entry_test.go"],
+    embed = [":entry_lib"],
+)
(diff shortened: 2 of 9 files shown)


ericsalo:
@fmeum the CI tests appear to be misconfigured?

mkruskal-google:
Do we want a go_proto_library one too?

ericsalo:
> Do we want a go_proto_library one too?

I'm thinking no, or at least not yet; we can always add more later.

go/def.bzl
```diff
@@ -14,12 +14,13 @@
...
-to change for easier maintenance.
+This file serves as the primary entry point for loading public Go rules,
+providers, and definitions. Re-exporting them here allows the underlying
+implementations to be maintained and relocated without breaking public APIs.
 
-Definitions outside this file are private unless otherwise noted, and
-may change without notice.
+Definitions outside this file are private and subject to change with the
+exception of (go/binary.bzl, go/library.bzl, go/test.bzl) which are provided as
+a convenience for hybrid environments.
```
fmeum: It's not so clear what `hybrid` means in this context.

go/library.bzl
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Public entry point for go_library.
+
+This file is provided individually to simplify configuration (e.g., source migration tooling)
```
fmeum: I don't think this needs to be justified. If we do want to add a justification, perhaps saying that this simplifies maintenance and avoids loading unnecessary `.bzl`s is more convincing?

tests/core/entry_points/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+load("//go:binary.bzl", "go_binary")
```
fmeum: The usual ruleset structure requires these files to be called `go_binary.bzl`, not just `binary.bzl`.

go/def.bzl
mkruskal-google: +1, I think we can drop almost all of the justification and explanation here.  Our particular use-case is irrelevant to *most* users of this library, and doesn't really warrant comments

ericsalo: Done.

go/library.bzl
ericsalo: Done.

tests/core/entry_points/BUILD.bazel
ericsalo: Done.

fmeum (approved):

dzbarsky (approved):

PR #4656, opened 2026-07-13
Add tests for cgo ldflags response-file encoding for Go 1.27

darkrift:
This adds tests for the cgo argfile change for GCC compatible file parser in 1.27+

Bonus: fixes a toolchain constraint issue for windows

### DIFF
--- a/go/tools/builders/cgo2.go
+++ b/go/tools/builders/cgo2.go
@@ -468,8 +468,7 @@ func copyOrLinkFile(inPath, outPath string) error {
 func formatLdFlagsFileContent(flags string) string {
 	shouldEscape, _ := onVersionOrHigher(27)
 	if shouldEscape {
-		escaped := strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(flags)
-		return `"` + escaped + `"` + "\n"
+		return encodeResponseFileArg(flags) + "\n"
 	}
 	return flags
 }
--- a/go/private/platforms.bzl
+++ b/go/private/platforms.bzl
@@ -179,8 +179,13 @@ def _generate_platforms():
         ))
         if (goos, goarch) in CGO_GOOS_GOARCH:
             # On Windows, Bazel will pick an MSVC toolchain unless we
-            # specifically request mingw or msys.
-            mingw = ["@bazel_tools//tools/cpp:mingw"] if goos == "windows" else []
+            # specifically request mingw or msys. Bazel's built-in C++
+            # configuration and rules_cc use separate constraint settings, so
+            # include both constraints to support either toolchain source.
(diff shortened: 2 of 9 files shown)


fmeum:
Why is a further change beyond https://github.com/bazel-contrib/rules_go/commit/dd24e55a3427d461031bffcf56fd2e540a54bd65 necessary? I like seeing the tests, but ideally we would have an integration test that exercises this behavior against Go 1.27.

darkrift:
Yeah it must be completely my bad, when I first attempted to build for 1.27rc1, I must have been using an override with a previous commit.

I'll add an integration test so that we have better confidence on this fix.

darkrift:
Should be good now and it also found a real issue with windows

darkrift:
@fmeum ping :)

go/tools/builders/cgo2.go
```diff
@@ -468,8 +468,7 @@ func copyOrLinkFile(inPath, outPath string) error {
 func formatLdFlagsFileContent(flags string) string {
 	shouldEscape, _ := onVersionOrHigher(27)
 	if shouldEscape {
-		escaped := strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(flags)
-		return `"` + escaped + `"` + "\n"
+		return encodeResponseFileArg(flags) + "\n"
```
fmeum, line 533: Die the test fail even before this change? Are you intentionally including this part of the change - the PR name suggests otherwise?

darkrift: I updated the name to mention any kind of tests. I also want to keep this unit test because the new implementation covers more scenarios that golang covers such as `$` and back tick ` `` `

darkrift: As for the test failure, it was a problem with toolchains on windows that needed the change in `platforms.bzl` : https://buildkite.com/bazel/rules-go-golang/builds/8912/list?sid=019f60df-7707-4cb2-be61-816674800001&tab=output

```
--
  | File "C:/b/rmjyx73c/external/rules_go~/go/private/rules/binary.bzl", line 128, column 20, in _go_binary_impl
  | go = go_context(
  | File "C:/b/rmjyx73c/external/rules_go~/go/private/context.bzl", line 605, column 21, in go_context
  | fail("{} has pure explicitly set to off, but no C++ toolchain could be found for its platform".format(ctx.label))
  | Error in fail: @@//:cgo_ldflags has pure explicitly set to off, but no C++ toolchain could be found for its platform
  | ERROR: C:/b/kavzg47b/bazel_testing/bazel_go_test/main/BUILD.bazel:10:10: Analysis of target

```

fmeum (approved):

PR #4660, opened 2026-07-22
upgrade gazelle to 0.52

tyler-french:


tests/bcr/MODULE.bazel
```diff
@@ -18,7 +18,7 @@ local_path_override(
     path = "../..",
 )
 
-bazel_dep(name = "gazelle", version = "0.45.0")
+bazel_dep(name = "gazelle", version = "0.52.2")
```
tyler-french, line 21: are we intentionally keeping this one low?

fmeum:
Do we need to update? With Bzlmod, keeping the dep version means strictly more freedom for users. We can still test with the latest version via an override.

jayconrod:
+1 to keeping the dep at a low version. I've introduced some small breaks with the v2 work, would be good not to force that upgrade yet.

PR #4661, opened 2026-07-23
Simplify the release process

fmeum:
**What type of PR is this?**

Workflows

For regular releases, we use the stock Publish to BCR workflow: push a tag on current master to trigger the flow.

This makes it possible to disable the legacy Publish to BCR app and remove a bunch of code from the releaser, which now only retains the logic to upgrade dependencies.

**Other notes for review**


fmeum:
I'll update the Wiki page after this has been merged.

.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset -o pipefail
+
+# Set by GH actions, see
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+TAG=${GITHUB_REF_NAME}
+# The archive name matches the one used by all previous releases, which is
+# also what .bcr/source.template.json expects.
```
fmeum:
```suggestion
```

jayconrod (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
...
+  push:
+    tags:
+      - "v*.*.*"
+permissions:
+  id-token: write
+  attestations: write
+  contents: write
+jobs:
+  release:
+    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.2.2
```
tyler-french:
```suggestion
    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.7.0
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
...
+    with:
+      release_files: rules_go-*.zip
+      prerelease: false
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
+  publish:
+    needs: release
+    uses: ./.github/workflows/publish.yaml
+    with:
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
```
tyler-french:
```suggestion
      tag_name: ${{ github.ref_name }}
```
This workflow only runs on tag pushes, so `github.ref_name` is always the release tag.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
...
+  id-token: write
+  attestations: write
+  contents: write
+jobs:
+  release:
+    uses: bazel-contrib/.github/.github/workflows/release_ruleset.yaml@v7.2.2
+    with:
+      release_files: rules_go-*.zip
+      prerelease: false
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
```
tyler-french:
```suggestion
      tag_name: ${{ github.ref_name }}
      draft: true
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
...
+  publish:
+    needs: release
+    uses: ./.github/workflows/publish.yaml
+    with:
+      tag_name: ${{ inputs.tag_name || github.ref_name }}
+    secrets:
+      publish_token: ${{ secrets.publish_token || secrets.BCR_PUBLISH_TOKEN }}
```
tyler-french, line 25:
```suggestion
      publish_token: ${{ secrets.publish_token || secrets.BCR_PUBLISH_TOKEN }}
  finalize:
    needs: publish
    runs-on: ubuntu-latest
    steps:
      - uses: softprops/action-gh-release@718ea10b132b3b2eba29c1007bb80653f286566b # v3.0.1
        with:
          tag_name: ${{ github.ref_name }}
```

.github/workflows/release_prep.sh
```diff
@@ -0,0 +1,53 @@
...
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "io_bazel_rules_go",
+    sha256 = "${SHA}",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazel-contrib/rules_go/releases/download/${TAG}/${ARCHIVE}",
+        "https://github.com/bazel-contrib/rules_go/releases/download/${TAG}/${ARCHIVE}",
+    ],
+)
```
tyler-french, line 35:
```suggestion
    urls = [
        "https://github.com/bazel-contrib/rules_go/releases/download/${TAG}/${ARCHIVE}",
    ],
)
```

tyler-french:
Since immutable releases are enabled for this repository, this needs the draft/finalize flow from the current [`rules-template`](https://github.com/bazel-contrib/rules-template/blob/main/.github/workflows/release.yaml). 

Since i thought it would be more helpful: I added a few inline suggestions to update to `release_ruleset@v7.7.0`, create the release with `draft: true`, run the BCR publication, and finalize the release only after that succeeds. This allows the BCR workflow to attach its attestations while ensuring the published release remains immutable. Also to skip mirror.bazel.build since that's not part of the upload flow

tyler-french (approved):

fmeum:
Thanks @tyler-french!

The mirror is only relevant for WORKSPACE, so I think it's fine to drop it. The cloudflare mirror for BCR entries will cover rules_go automatically.

.bcr/config.yml
```diff
@@ -1,3 +0,0 @@
-fixedReleaser:
```
tyler-french, line 1: @fmeum looks like CI doesn't like this change, or wants it to get approved to run CI

fmeum:
I updated the wiki entry.

PR #4666, opened 2026-07-28
go: require Go 1.18 or later

sluongng:
Because maintainers no longer support Go SDKs older than 1.18,
rules_go should not advertise Go 1.14 as its minimum. The old floor
leaves obsolete version branches and tests in paths that every build
must carry.

Raise the minimum and floor SDK selection for directive-less `go.mod`
files at Go 1.18. Update SDK tests and documentation, and remove
pre-1.18 builder, test-main, assembly, embed, and nogo shims while
retaining behavior required by Go 1.18 through Go 1.20.

The Go 1.18 through Go 1.20 `go/types` normalization and pre-Go 1.19
assembler behavior remain version-gated. Building the complete `nogo`
binary with Go 1.18 is separately blocked by the repository's existing
`x/tools` dependency, whose current sources require newer Go APIs and
syntax.


fmeum:
Would we get rid of more code if we assume 1.20 as the baseline? That version is from February 2023, pretty sure we can require that now.

fmeum (approved):

PR #4669, opened 2026-07-28
coverage: emit a real baseline for packages with no tests

### LINKED ISSUE #4668: coverage: packages with no tests report LF:0 instead of a real zero baseline

v0.62.0

n/a

9.1.0

Yes.

macOS/arm64, though nothing here is platform-specific.

### What did you do?

Ran coverage over a `go_library` that no test links:

```python
go_library(
    name = "untested",
    srcs = ["untested.go"],
    importpath = "example.com/untested",
)
```

```console
$ bazel coverage //src:all --combined_report=lcov
```

The package reported at 0%. `go test -cover ./...` has done this natively since
Go 1.22 (golang/go#18909): a package with functions but no test files emits
zero-count blocks and prints `coverage: 0.0% of statements`.

### What did you see instead?

```
SF:src/untested.go
FNF:0
FNH:0
LF:0
LH:0
end_of_record
```

rules_go doesn't pass any baseline coverage files, so Bazel falls back to
`BaselineCoverageAction`, which writes only the path and `end_of_record`. The
zeros are just `LcovPrinter` serializing counters nobody computed.

The catch is that `LF:0` is also the correct output for a file that genuinely
has nothing to cover — constants only, or excluded by build constraints.
Consumers can't tell the two apart, so 0/0 ends up scored as 100%, and a package
with no tests at all scores better than one with partial tests. In our repo that
hid 21 files (~700 lines).

This is bazelbuild/bazel#5716, but that's no longer blocking: Bazel 9 added
`baseline_coverage_files` to `coverage_common.instrumented_files_info`
(bazelbuild/bazel#26133), which is the seam @fmeum suggested there:

> For Go, a separate tool could do this.

Implemented in #4669: it runs `cmd/cover` over the target's srcs to get the real
line set, guarded on `bazel_features` so Bazel 7/8 keep today's behaviour.

loganrosen:
**What type of PR is this?**

> Bug fix

rules_go doesn't pass baseline coverage files, so Bazel falls back to `BaselineCoverageAction`, which writes only the path and `end_of_record`. An untested package lands in the report as `LF:0 / LH:0`, which is also what a file with genuinely nothing to cover looks like. Consumers can't distinguish them, so 0/0 gets scored as 100% and a package with no tests scores better than one with partial tests.

This runs `cmd/cover` over the target's own `srcs`, reads back the block position table it appends, and emits LCOV with every coverable line at count 0, handed to Bazel via `baseline_coverage_files`. For an untested library:

```diff
 SF:src/untested.go
-FNF:0
-FNH:0
-LF:0
+DA:4,0
+DA:5,0
+DA:6,0
+DA:7,0
+LF:4
 LH:0
 end_of_record
```

Fixes #4668

**Other notes for review**

It's 879 lines, but only 85 touch existing files. The rest is one builder action, one `.bzl`, and the tests.

The choices most likely to draw questions:

- The line set comes from `cmd/cover` rather than a separate AST pass, so the baseline agrees with measured output by construction rather than by approximation. `cover -pkgcfg` with `EmitMetaFile` is the mechanism `go test -cover` uses for a package with no test files, which is exactly this situation, and `covdata textfmt` decodes the result into the same text profile `bzltestutil` already consumes. I tried an AST generator first, and against a file whose real coverage run gives `LF:21` it emitted 14 lines, missing `case` clauses, continuation lines of multi-line statements, and closing braces.
- `covdata` is built from the SDK's own sources in the same action that already builds `pack` that way, since Go 1.25 dropped both prebuilt binaries from the distribution to shrink it. The shared Go build cache makes the extra build roughly a second, and a `covdata` compiled from its own SDK decodes whatever that SDK emits.
- This adds an optional `covdata` attribute to `go_toolchain`. Custom toolchain definitions that omit it simply emit no baseline, so nothing breaks for them.
- `Local: true` makes `cmd/cover` record each source under the exec-root-relative name it was given, so no path mapping round trip is needed. It changes only the recorded name, not the block positions.
- An empty meta-data file is the signal that a package has no coverable lines, which is structural rather than reconstructed from the sources. An absent one fails the action, so a future change to `EmitMetaFile` cannot quietly report `LF:0` everywhere.
- LCOV writing is hand-rolled, and the profile pattern is duplicated from `bzltestutil`, because `builder` is a `go_tool_binary` limited to the standard library and `bzltestutil` imports `coverdata`.
- `go_test` is deliberately untouched, since emitting baselines for `_test.go` files would regress results by scoring test files. Embedded libraries are unaffected, as the `go_library` they come from emits its own baseline from its own `srcs`. The gap this leaves is a non-test `.go` file that lives only in a `go_test`'s `srcs`, which gets no baseline. Happy to close that by filtering test files out of the test's own srcs if you'd prefer.
- Sources come from `ctx.files.srcs`, not `go_info.srcs`, to match `source_attributes = ["srcs"]`. `go_info.srcs` would pull in the generated coverage shim and embedded srcs, adding bogus and duplicate records.
- Files with nothing to cover still get an `SF:` record with `LF:0`, so the set of files in the report is unchanged. The zero is now measured rather than assumed.
- cgo files get baselines like any other source. `compilepkg` instruments them before cgo rewrites them, so a measured run reports positions in the unprocessed file, which are the ones this action reads back.
- The action is path mapped under exactly the condition `compilepkg` uses, and skips it for cgo and for targets tagged `local`. The two have to agree, since a generated source under `--experimental_output_paths=strip` would otherwise get a stripped path in the baseline and an unstripped one from the compile action, leaving the zeros in a record measured data never displaces.

This is guarded on `bazel_features.rules.instrumented_files_info_has_baseline_coverage_files`, and no dependency bump is needed since the current 1.36.0 pin already has that flag. On Bazel 7 and 8 the kwarg isn't passed and behaviour is unchanged. The `polyfill_bazel_features.bzl` value is a constant `False`, since Bazel 9 is also the release that removed WORKSPACE support.

Nothing runs on ordinary builds. The action is gated on `go.coverage_enabled` and `go.coverage_instrumented`, so `--instrumentation_filter` is respected too.

The `go_bazel_test` covers an untested library, an untested binary, an untested cgo library, a declaration-only file, a build-constrained file, and that measured coverage isn't displaced. It runs in CI on the Bazel 9 legs, and on Bazel 8 the `bazel_features` gate makes it incompatible so a `//...` pattern skips it. A unit test covers the profile reader. I also ran this across a ~680-file Go monorepo: same set of files in the report, measured coverage byte-identical, and 21 files that had been reporting 100% now report 0%.

### DIFF
--- a/go/private/actions/baseline_coverage.bzl
+++ b/go/private/actions/baseline_coverage.bzl
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
-load("//go/private:common.bzl", "GO_TOOLCHAIN_LABEL")
+load("//go/private:common.bzl", "GO_TOOLCHAIN_LABEL", "SUPPORTS_PATH_MAPPING_REQUIREMENT")
 
 def baseline_coverage_kwargs(go, ctx, sources):
     """Emits a baseline coverage report and returns it as instrumented_files_info kwargs.
@@ -71,5 +71,11 @@ def _emit_baseline_coverage(go, *, sources, out_lcov):
         mnemonic = "GoBaselineCoverage",
         executable = go.toolchain._builder,
         arguments = ["baselinecoverage", args],
+        # The environment carries GOOS, GOARCH and CGO_ENABLED, which decide
+        # which sources the build constraints select. GOROOT is dropped and
+        # passed as an argument by builder_args instead, since path mapping
+        # cannot map environment variable values.
+        env = go.env_for_path_mapping,
+        execution_requirements = SUPPORTS_PATH_MAPPING_REQUIREMENT,
         toolchain = GO_TOOLCHAIN_LABEL,
     )
--- a/go/private/actions/baseline_coverage.bzl
+++ b/go/private/actions/baseline_coverage.bzl
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
-load("//go/private:common.bzl", "GO_TOOLCHAIN_LABEL", "SUPPORTS_PATH_MAPPING_REQUIREMENT")
+load("//go/private:common.bzl", "GO_TOOLCHAIN_LABEL")
+load("//go/private/actions:utils.bzl", "path_mapping_action_settings")
 
 def baseline_coverage_kwargs(go, ctx, sources):
     """Emits a baseline coverage report and returns it as instrumented_files_info kwargs.
@@ -77,16 +78,10 @@ def _emit_baseline_coverage(go, ctx, *, sources, out_lcov):
     # separate record instead of being displaced by measured data. Path mapping
     # rewrites the paths of generated sources, so it can only be enabled where
     # compilepkg also enables it. See the matching condition there.
-    if getattr(ctx.attr, "cgo", False) or "local" in ctx.attr.tags:
-        env = go.env
-        execution_requirements = {}
-    else:
-        # The environment carries GOOS, GOARCH and CGO_ENABLED, which decide
-        # which sources the build constraints select. GOROOT is dropped and
-        # passed as an argument by builder_args instead, since path mapping
-        # cannot map environment variable values.
-        env = go.env_for_path_mapping
-        execution_requirements = SUPPORTS_PATH_MAPPING_REQUIREMENT
+    env, execution_requirements = path_mapping_action_settings(
+        go,
+        getattr(ctx.attr, "cgo", False),
+    )
 
     go.actions.run(
         inputs = depset(go_srcs + [go.toolchain._covdata], transitive = [sdk.headers, sdk.tools]),
--- a/go/private/actions/baseline_coverage.bzl
+++ b/go/private/actions/baseline_coverage.bzl
@@ -44,6 +44,10 @@ def baseline_coverage_kwargs(go, ctx, sources):
         return {}
     if not go.coverage_enabled or not go.coverage_instrumented:
         return {}
+    if not go.toolchain._covdata:
+        # A custom toolchain without a covdata tool cannot decode the coverage
+        # meta-data this action's cover invocation emits.
+        return {}
 
     out_lcov = go.declare_file(go, name = ctx.label.name, ext = ".baseline.lcov")
     _emit_baseline_coverage(go, ctx, sources = sources, out_lcov = out_lcov)
@@ -64,6 +68,8 @@ def _emit_baseline_coverage(go, ctx, *, sources, out_lcov):
 
     args = go.builder_args(go)
     args.add_all(go_srcs, before_each = "-src")
+    args.add("-covdata", go.toolchain._covdata)
+    args.add("-importpath", go.importpath)
     args.add("-o", out_lcov)
 
     # The source paths this action writes to the tracefile have to be the same
@@ -83,7 +89,7 @@ def _emit_baseline_coverage(go, ctx, *, sources, out_lcov):
         execution_requirements = SUPPORTS_PATH_MAPPING_REQUIREMENT
 
     go.actions.run(
-        inputs = depset(go_srcs, transitive = [sdk.headers, sdk.tools]),
+        inputs = depset(go_srcs + [go.toolchain._covdata], transitive = [sdk.headers, sdk.tools]),
         outputs = [out_lcov],
         mnemonic = "GoBaselineCoverage",
         executable = go.toolchain._builder,
--- a/go/private/actions/baseline_coverage.bzl
+++ b/go/private/actions/baseline_coverage.bzl
@@ -1,0 +1,75 @@
+# Copyright 2026 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@io_bazel_rules_go_bazel_features//:features.bzl", "bazel_features")
+load("//go/private:common.bzl", "GO_TOOLCHAIN_LABEL")
+
+def baseline_coverage_kwargs(go, ctx, sources):
+    """Emits a baseline coverage report and returns it as instrumented_files_info kwargs.
+
+    Bazel asks a non-test rule for a "baseline" coverage report so that code no
+    test ever links still appears in the combined coverage report. Absent one,
+    Bazel substitutes a stub that names each source file and nothing else, which
+    its LcovPrinter renders as "LF:0/LH:0". A consumer cannot tell that apart
+    from a file with genuinely nothing to cover, so a wholly untested package
+    reports 100% rather than 0%.
+    See https://github.com/bazelbuild/bazel/issues/5716.
+
+    Returns empty kwargs, leaving Bazel's stub in place, when the running Bazel
+    is too old to accept a baseline, when the build is not collecting coverage,
+    or when --instrumentation_filter excludes this target. Nothing is emitted on
+    an ordinary build.
+
+    Args:
+        go: the Go context.
+        ctx: the rule context.
+        sources: the rule's `srcs`, matching what is declared to
+            coverage_common.instrumented_files_info as source attributes.
+
+    Returns:
+        A dict to splat into coverage_common.instrumented_files_info.
+    """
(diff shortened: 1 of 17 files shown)


go/private/actions/baseline_coverage.bzl
```diff
@@ -0,0 +1,75 @@
...
+    args.add_all(go_srcs, before_each = "-src")
+    args.add("-o", out_lcov)
+
+    go.actions.run(
+        inputs = depset(go_srcs, transitive = [sdk.headers, sdk.tools]),
+        outputs = [out_lcov],
+        mnemonic = "GoBaselineCoverage",
+        executable = go.toolchain._builder,
+        arguments = ["baselinecoverage", args],
+        toolchain = GO_TOOLCHAIN_LABEL,
```
fmeum, line 94: Could you add the `SUPPORTS_PATH_MAPPING` execution requirement (see other actions)?

tests/core/coverage/baseline_coverage_test.go
```diff
@@ -0,0 +1,233 @@
...
+	"strings"
+	"testing"
+
+	"github.com/bazelbuild/rules_go/go/tools/bazel_testing"
+)
+
+func TestMain(m *testing.M) {
+	bazel_testing.TestMain(m, bazel_testing.Args{
+		Main: `
+-- src/BUILD.bazel --
```
fmeum, line 29: Could you add a test that exercises the cgo path?

fmeum (commented):
This is great, thanks for picking this up!

I'm not that familiar with the Go coverage format internals you are using here - @jayconrod if you aren't either, I'm happy to dig into this more.

Somewhat embarrassingly, we don't run tests with Bazel 9, so let us figure out what we need to do to cover this.

go/private/actions/baseline_coverage.bzl
loganrosen: Added. I also had to pass `env = go.env_for_path_mapping`, since the action had no environment at all and `filterAndSplitFiles` reads GOOS, GOARCH and CGO_ENABLED from `go/build.Default`. Build constraints were being evaluated for the host rather than the target when cross compiling.

Verified with `--experimental_output_paths=strip`.

tests/core/coverage/baseline_coverage_test.go
loganrosen: Added, and it turned up that cgo sources should not have been skipped at all, so they now get real baselines.

`compilepkg` instruments the raw cgo sources before cgo rewrites them, so a measured run reports positions in the unprocessed file, which are the ones this action reads back. I confirmed against `go test -cover` on a cgo package that the profile blocks match the `Pos` table.

Skipping was also not the fallback I described. Passing `baseline_coverage_files` replaces Bazel's stub for the whole target, so a skipped cgo file dropped out of the report entirely rather than falling back to it.

jayconrod:
Go made some significant changes to coverage after I left, so I'm afraid I don't know much about how it works these days. I'm unfortunately very limited on review bandwidth this week, so @fmeum if you have time, please dig in without me.

sethvargo:
Hey @fmeum - I'm sure you're swamped with PR review requests, but would you mind giving this another pass please? We're applying custom patches internally, and we think everyone in the community would benefit instead 😄

fmeum:
Is it clear that the legacy cover mode results in the same lines of code etc. as the new coverage backend? I'm asking since the former may be dropped in the future and subtle differences in counts would be pretty confusing.

Do you see a way to base this on the new coverage info instead? This doesn't have to be a blocker, but I would like to understand the cost of not using the new mode better.

fmeum:
Agent has this to say about using the new coverage format directly. If that is actually correct, this looks very appealing:

### `covdata` is a prebuilt SDK tool

The comment in `baseline_coverage.go` says:

> `-pkgcfg` can emit a coverage meta-data file directly (its `EmitMetaFile` field exists for "go test -cover" on a package with no test files, exactly this situation), but decoding that file requires the covdata tool, which the Go SDK does not ship prebuilt -- the go command compiles it on demand into the user's build cache. An action cannot rely on that.

`covdata` ships in the distribution tarball exactly like `cover`:

```
$ ls $(go env GOROOT)/pkg/tool/linux_amd64/     # go1.24.7
addr2line  asm  buildid  cgo  compile  covdata  cover  doc  fix  link
nm  objdump  pack  pprof  preprofile  test2json  trace  vet
```

`cmd/go` resolves it through `base.Tool("covdata")` → `build.ToolDir/covdata`, i.e. `$GOROOT/pkg/tool/<goos>_<goarch>/covdata`, via a plain `os.Stat` (`src/cmd/go/internal/base/tool.go`). There's no on-demand build and no build-cache involvement — it's a distribution tool like `compile` or `link`.

It's also already an input to this very action. `go/private/BUILD.sdk.bazel` defines `tools = glob(["pkg/tool/**", ...])`, and `_emit_baseline_coverage` already passes `transitive = [sdk.headers, sdk.tools]`. `goenv.goTool("covdata", ...)` (`go/tools/builders/env.go`) resolves it identically to the existing `goTool("cover", ...)` call. No new deps, no new action inputs.

Worth knowing too: we already decode new-format meta files in-tree — `go/tools/bzltestutil/bincov/coverage_visitor.go` is a stripped-down `cmd/covdata/dump.go` using `internal/coverage/decodemeta`, for `go_binary` coverage.

### The new-backend version works end to end

```
cover -pkgcfg cfg.json -var X -mode set -outfilelist outfiles.txt <srcs...>
    cfg = {..., "Granularity":"perblock", "Local":true,
           "EmitMetaFile":"<dir>/covmeta.<32 hex chars>"}
covdata textfmt -i=<dir> -o=baseline.txt
```

For a package with no tests that yields:

```
mode: set
untested.go:3.24,5.23 2 0
untested.go:5.23,7.3 1 0
untested.go:8.2,8.14 1 0
...
```

Two subprocesses for the whole package, and the output is already in the exact text format `bzltestutil/lcov.go`'s `convertCoverToLcov` consumes. `Local: true` makes `cmd/cover` emit `f.name` verbatim instead of `<pkgpath>/<base>`, so the paths come out exec-root-relative for free — no `SrcPathMapping` round trip needed.

One edge case: a package with **no functions at all** produces a zero-byte meta file, and `covdata textfmt` then exits 1 with `decoding meta-file ...: EOF`. So you skip covdata for that case and emit `SF:`/`LF:0` records directly. That's structurally unambiguous, which is exactly the distinction `hasFuncBody` currently has to reconstruct by re-parsing the source.

Note this adds no new minimum-Go constraint: `compilepkg` already requires `-pkgcfg` unconditionally (`go/tools/builders/cover.go`).

loganrosen:
Your agent's right that covdata ships, I pulled the same tarball. I was wrong to say it never does. 1.25 dropped it when distpack moved to an allowlist, so our default 1.26.7 SDK doesn't have it and `go tool covdata` would build it on demand there. Comment fixed.

The on-demand build is what we can't do here. The action's inputs are `pkg/tool/**`, so there's no `go` to run, no GOROOT to build against and no writable cache to build into. Decoding the meta file inside `builder` is out too, since `go build` won't let a `go_tool_binary` import `internal/coverage`. `bincov` only manages it by being a `go_tool_library` that rules_go compiles itself.

The line numbers match by construction, since both modes share one AST walk and only branch on where the span gets stored. I diffed a file with case clauses and a multi-line call against `-covermode=set` to be sure and it lined up.

Being the last legacy consumer is a fair thing to raise. The new-format version would be a separate bincov-style tool decoding the meta file itself, which I skipped because bincov already needs a `go1.24` shim for a `cformat` signature change and those APIs clearly move. Glad to build it if you'd rather not keep a foot in the legacy path.

fmeum:
@loganrosen I set Claude on this and it looks like it managed to build `covdata` (similar to the `pack` build we already maintain) and use the new coverage format. Could you take a look at https://github.com/bazel-contrib/rules_go/compare/baseline-coverage-covdata? Feel free to steal ideas and changes from it.

loganrosen:
Took it, thanks. You were right that I'd stopped at the wrong place, looking for a way to reach covdata from inside the action instead of building it once like pack.

One fix on top. `parseCoverProfile` split lines on whitespace, so a source path containing a space failed the action. I borrowed `bzltestutil/lcov.go`'s pattern, which already handles it, and added a unit test.

Your commit is cherry-picked so it's still yours, though a squash would collapse that. Happy to add a `Co-authored-by` trailer, or you can land yours directly and I'll rebase on top.

go/private/actions/baseline_coverage.bzl
```diff
@@ -0,0 +1,100 @@
...
+    args.add("-covdata", go.toolchain._covdata)
+    args.add("-importpath", go.importpath)
+    args.add("-o", out_lcov)
+
+    # The source paths this action writes to the tracefile have to be the same
+    # strings the compile action writes, or the baseline's zeros land in a
+    # separate record instead of being displaced by measured data. Path mapping
+    # rewrites the paths of generated sources, so it can only be enabled where
+    # compilepkg also enables it. See the matching condition there.
+    if getattr(ctx.attr, "cgo", False) or "local" in ctx.attr.tags:
```
fmeum: Could you extract this into a shared helper?

fmeum (approved):
Looks good except for a single comment.

go/private/actions/baseline_coverage.bzl
loganrosen: Done.

PR #4670, opened 2026-07-28
ci: test with Bazel 8 and 9 via a matrix

fmeum:
The main build and test tasks only ran with the default Bazel version from
`.bazelversion`; only the BCR test module was exercised against 8.x and 9.x.
This expands them over the existing `bazel` matrix and makes 9.2.0 the default,
which turned up a fair amount of fallout.

### CI

* Ubuntu, macOS and Windows run the full suite on Bazel 8 and 9. Bazel 7 keeps
  the minimum version task.
* Stardoc's output depends on the Bazel version, so the docs are generated for
  the default version and their diff tests moved to a dedicated task.

### Bazel 9 compatibility

* Load `CcInfo` and `CcToolchainConfigInfo` from rules_cc, return providers
  instead of a struct from a rule implementation, and load the native cc, java
  and shell rules where fixtures use them.
* `//go/toolchain:windows_amd64_cgo` listed both mingw constraint values, which
  became the same value in Bazel 8 and crashes Bazel 9. Pick the one matching
  where `@local_config_cc` comes from.
* `objc_library` needs the Apple CC toolchain to be selected through toolchain
  resolution; the `--*crosstool_top` flags no longer do anything.
* A cross-compiled test can't satisfy the default test toolchain Bazel 9 added,
  and `filegroup` now forwards the runfiles of its srcs.

### Bzlmod

Bazel 9 removed WORKSPACE, so `go_bazel_test` always generates a MODULE.bazel
file and the WORKSPACE template is gone. Targets that only resolved because
WORKSPACE was still enabled on Bazel 7 declare their repositories through
Bzlmod, including the maintainer tools' dependencies and module extensions for
the generated test repositories.

Repository names reach the fixtures differently as a result: `bazel info`
doesn't apply the main repository mapping to labels on the command line, so the
platform in `GO_BAZEL_TEST_BAZELFLAGS` is spelled canonically, and the canonical
name separator differs between Bazel 7 and later, which two tests assert on.

### Toolchains

* Take protoc from the prebuilt toolchains protobuf ships rather than building
  it from source, which is slow and fails on Windows. protobuf is pinned for
  this repository only, so the version users get is unchanged. `//proto:protoc`
  is tagged manual for the same reason.
* Take buildifier from `buildifier_prebuilt`.

### Not covered

* `--incompatible_auto_exec_groups` is disabled: protobuf's protoc authenticity
  check calls `ctx.actions.run_shell` without a toolchain, and proto toolchain
  resolution is the default as of Bazel 9, so the two can't be combined.
* `debian11_zig_cc` runs on Bazel 8: hermetic_cc_toolchain releases that work
  with Bazel 9 require a rules_cc that Bazel 7 cannot load. The zig toolchain
  also can't load shared libraries at runtime, so a few tests are skipped there.
* `rbe_ubuntu2404` runs on Bazel 7: `GoCompilePkg` declares an output directory
  for the cgo-generated sources that stays empty when cgo is enabled on a
  package that doesn't use it, and the remote executor doesn't report empty
  output directories (bazelbuild/bazel#28286).
* A few tests are skipped on Windows: they build protobuf from source, whose
  protoc the mingw toolchain links against a DLL that isn't resolvable when
  Bazel runs it as a build tool, or they assert on the repository directory
  name.


go/tools/bazel_testing/bazel_testing.go
```diff
@@ -451,6 +427,63 @@ func setupWorkspace(args Args, files []string) (dir string, cleanup func() error
...
+	}
+	return goSDKPath
+}
+
+// rulesCCVersion picks the oldest rules_cc that works with the Bazel version
+// under test. Bazel 7 cannot load rules_cc 0.2.14 and later, which introduced
+// the compatibility proxy, while the C++ toolchain of releases before 0.2.18
+// emits a malformed -fuse-ld flag on macOS. Bazel 9 selects a new enough
+// version through its own dependencies either way.
+func rulesCCVersion() string {
```
dzbarsky, line 444: omg

go/tools/builders/stdliblist_test.go
```diff
@@ -4,18 +4,27 @@ import (
...
+}
+
 func Test_stdliblist_noexport(t *testing.T) {
 	testDir := t.TempDir()
 	outJSON := filepath.Join(testDir, "out.json")
 
 	test_args := []string{
 		fmt.Sprintf("-out=%s", outJSON),
-		"-sdk=../go_sdk",
+		fmt.Sprintf("-sdk=../%s", sdkDir()),
```
dzbarsky: why not `"-sdk=../" + sdkDir()` and same above?

go/tools/builders/stdliblist_test.go
```diff
@@ -53,7 +62,7 @@ func Test_stdliblist_export(t *testing.T) {
 	outJSON := filepath.Join(testDir, "out.json")
 	test_args := []string{
 		fmt.Sprintf("-out=%s", outJSON),
-		"-sdk=../go_sdk",
+		fmt.Sprintf("-sdk=../%s", sdkDir()),
```
dzbarsky: probably same simplification everywhere?

tests/buildifier_test.bzl
```diff
@@ -62,9 +64,10 @@ exit $ok
 
     return [DefaultInfo(
         executable = script,
+        # The prebuilt buildifier is a wrapper that needs its own runfiles.
         default_runfiles = ctx.runfiles(
             files = [script, ctx.executable._buildifier] + files,
```
dzbarsky: do we still need the executable here? It won't be coming from the buildifier runfiles below?

.bazelrc
```diff
@@ -20,9 +27,15 @@ build:windows --compiler=mingw-gcc
...
-build:windows --host_platform=@io_bazel_rules_go//go/toolchain:windows_amd64_cgo
-build:windows --platforms=@io_bazel_rules_go//go/toolchain:windows_amd64_cgo
+# Not @io_bazel_rules_go: a module's own repo_name is only an apparent name for
+# other modules, not within the module itself.
+build:windows --host_platform=//go/toolchain:windows_amd64_cgo
+build:windows --platforms=//go/toolchain:windows_amd64_cgo
 build:windows --incompatible_enable_cc_toolchain_resolution
+# Abseil, which protobuf builds against, requires C++17. The mingw toolchain
+# defaults to an older standard.
+build:windows --cxxopt=-std=c++17
```
dzbarsky: should we just set this globally tbh?

MODULE.bazel
```diff
@@ -59,24 +79,86 @@ use_repo(
...
+
+dev_go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps", dev_dependency = True)
+
+# The maintainer tools under go/tools are separate Go modules so that their
+# dependencies don't end up in //:go.mod, which is also used to pick the Go SDK
+# version. go_deps only accepts a single from_file tag per module, so their
+# dependencies are listed here instead; keep them in sync with the respective
+# go.mod and go.sum files.
+#
+# Needed by //go/tools/releaser. This is the pseudo-version of the v6.4.0 tag;
```
dzbarsky, line 93: thanks for finishing these bits, I ran out of steam on bzlmod'ing this one :)

dzbarsky (approved):

jayconrod:
I gave a quick look and didn't see anything concerning, but this is a larger PR than I can really give a competent review for. I trust @dzbarsky's review here, but I'd be happier in general for smaller pieces.

fmeum:
> but I'd be happier in general for smaller pieces.

Yeah, sorry about that, I just didn't see any other way. This was straight up whac-a-mole with any change causing a regression on some other Bazel version. Going forward we should have a much easier time now that we have a proper matrix.

PR #4672, opened 2026-07-29
go: restore go_rule for custom Go rules

### LINKED ISSUE #4662: 0.62.0 breaks `race` mode for any rule calling `go_context()` without declaring the CC toolchain
- 0.61.1: OK in that regard
- 0.62.0: regression

- 0.51.3
- 0.52.2

Not needed to reproduce the primary issue below (pure `rules_go` repro).
Separately, this same regression also breaks Gazelle's own `//language/go:std_package_list` rule when Gazelle is used with `rules_go` 0.62.0 and `race` mode enabled.

- 7.7.0 (as pinned by `rules_go`'s own `.bazelversion`)
- 9.2.0 (as of DataDog/datadog-agent#54068)

Yes.

Linux x86_64 (Ubuntu 24.04)

None needed, the repro below uses `rules_go`'s own default toolchain resolution, no custom CC toolchain or remote execution involved.

### What did you do?
Minimal repro, entirely self-contained within the `rules_go` repo itself:

```sh
git clone https://github.com/bazel-contrib/rules_go.git
cd rules_go
git checkout v0.62.0
bazel test --//go/config:race //tests/extras/gomock/source:client_test
```

The target analyzes and builds successfully, same as on 0.61.1.
`_gomock_source_impl` (in `extras/gomock.bzl`) explicitly calls `go_context(ctx, maybe_needs_cc_toolchain = False)`: it copies a source file and doesn't touch `cgo`/`cc` at all, so it shouldn't be affected by whether a C/C++ toolchain happens to be resolved.

### What did you see instead?
```
ERROR: tests/extras/gomock/source/BUILD.bazel:17:7: in _gomock_source rule //tests/extras/gomock/source:client_mocks:
	File ".../extras/gomock.bzl", line 35, column 20, in _gomock_source_impl
		go = go_context(
	File ".../go/private/context.bzl", line 608, column 22, in go_context
		validate_mode(mode)
	File ".../go/private/mode.bzl", line 65, column 17, in validate_mode
		fail("race instrumentation can't be enabled when cgo is disabled. Check that pure is not set to \"off\" and a C/C++ toolchain is configured.")
Error in fail: race instrumentation can't be enabled when cgo is disabled. Check that pure is not set to "off" and a C/C++ toolchain is configured.
```

Likely cause: #4601 changed `go_context()`'s `cgo_available` computation in `go/private/context.bzl`.
Before:
```python
if has_cc_toolchain_attr:
    cgo_available = cgo_context_info != None
else:
    # Preserve cgo build-mode/tag behavior for rules that won't use the
    # C/C++ toolchain in their own actions.
    cgo_available = not cgo_disabled
```

After:
```python
cgo_available = cgo_context_info != None
```

`cgo_context_info` is only non-`None` when the calling rule has actually resolved `CPP_TOOLCHAIN_TYPE` via its own `toolchains` / `fragments` declaration.
Any rule that calls `go_context()` without declaring that toolchain now gets `cgo_available = False` unconditionally, even when it explicitly opted out of needing `cgo` via `maybe_needs_cc_toolchain = False`, and even when it never asked to build in race mode itself. 
validate_mode()` then fails as soon as the *global* `--//go/config:race` flag is set, regardless of what the individual rule actually needs.

This looks related to #4626 / bazel-contrib/bazel-gazelle#2354 / bazel-contrib/bazel-gazelle#2355, which fixed the equivalent problem for `gazelle_binary` (Go *link* rules) back in 0.61.1.
#4601 (shipped in 0.62.0) removed the fallback those fixes depended on staying safe for everyone else, so it now also hits at least two more call sites that were never updated to declare the toolchain:
- rules_go's own `extras/gomock.bzl` (`_gomock_source`, repro above)
- Gazelle's own `//language/go:std_package_list` rule (`language/go/def.bzl`), hit downstream when building/testing a bzlmod repo with `gazelle` + `rules_go` 0.62.0 and race mode enabled: same traceback shape, same `fail()` site.

It also breaks hand-written custom rules in consuming repos that call `go_context(ctx)` without declaring `CGO_ATTRS` / `CGO_FRAGMENTS` / `CGO_TOOLCHAINS` (from `go/private:context.bzl`).
We hit this in a downstream repo for a custom `cgo -godefs` wrapper rule, fixable on our end once we add those, but it means every custom-rule author who calls `go_context()` needs to audit for this, and (per the discussion on #4626) the requirement isn't currently documented in `go/toolchains.rst`'s "writing new Go rules" section, nor are `CGO_ATTRS` / `CGO_FRAGMENTS` / `CGO_TOOLCHAINS` exposed as a public/documented API for rule authors (they currently live under the **private** `go/private:context.bzl`).

dzbarsky:
## Summary

rules_go v0.62.0 requires rules that compile or link Go code with cgo to declare the optional C++ toolchain, Apple and C++ configuration fragments, and cgo attributes. Add a helper to define such rules ergonomically.

Fixes #4662.


go/private/context.bzl
```diff
@@ -93,6 +93,35 @@ CGO_TOOLCHAINS = [
 ]
 CGO_FRAGMENTS = ["apple", "cpp"]
 
+def go_rule(implementation, attrs = None, fragments = None, toolchains = None, **kwargs):
```
dzbarsky: attrs = {}, fragments = [], toolchains = [], simplify below

fmeum (approved):
Looks great! 

@jayconrod

jayconrod:
I'm unfortunately very limited on review bandwidth this week, so please go ahead without me.

I tend not to like wrappers and prefer minimizing and documenting requirements... but perhaps that's no longer practical. There seem to be a lot of things needed for cgo now.

hunshcn:
Thanks for the quick fix! Could we get a patch release (0.62.1) with this? v0.62.0 breaks downstream repos that enable race mode globally (e.g. gazelle's `std_package_list` fails analysis under `--@rules_go//go/config:race`), so we're holding off on the 0.62 upgrade until a release includes this.

dzbarsky:
I'm currently traveling but will try to kick one off tonight or tomorrow

dzbarsky:
@hunshcn We released 0.63.0 that has this fix.

PR #4675, opened 2026-08-02
go: require Go 1.20 or later

sluongng:
Review feedback on #4666 asked whether a Go 1.20 floor would let
rules_go shed more old-version support. Go 1.18 and Go 1.19 are no
longer supported upstream, and advertising them retains compatibility
branches and tests that current releases do not need.

Raise the minimum SDK and use Go 1.20 when go_sdk.from_file reads a
go.mod or go.work file without a version directive. Refresh the
documentation and SDK-selection fixtures, then remove the pre-Go 1.19
assembler detection, pre-Go 1.20 standard-library branches, and build
tags that disappeared before Go 1.20.

Keep the Go 1.20 nogo version normalization because go/types still
requires it until Go 1.21. Preserve the existing policy of warning,
rather than failing, when an older SDK is explicitly selected. Continue
honoring precompiled standard libraries exposed through GoSDK.libs.


go/private/actions/stdlib.bzl
```diff
@@ -50,10 +49,6 @@ def emit_stdlib(go):
     return [go_info, stdlib]
 
 def _should_use_sdk_stdlib(go):
```
fmeum: Could we drop this entirely? In Go 1.20+, this will always return False due to missing `go.sdk.libs`.

fmeum (approved):

go/private/actions/stdlib.bzl
sluongng: good call.

will send a new version

fmeum (approved):

fmeum:
Thanks, I will merge this after https://github.com/bazel-contrib/rules_go/pull/4670

fmeum:
CI is happy again, merging now

PR #4679, opened 2026-08-04
Treat TestMain as the test entry point only when it takes *testing.M

### LINKED ISSUE #4677: go_test fails to build a package whose TestMain takes *testing.T
### What happened

A package whose only `TestMain` is declared as `func TestMain(t *testing.T)` builds and tests fine under `go test`, but fails to build under `go_test`:

```
bazel-out/.../pkg/testmain.go:132:13: cannot use m (variable of type *testing.M) as *testing.T value in argument to l.TestMain
```

The error points at generated code, so from the user's side it looks like a rules_go bug with no obvious connection to anything in their source.

### Minimal reproduction

```go
// x_test.go
package x

import "testing"

func TestMain(t *testing.T) {
	t.Log("ordinary test, not the entry point")
}
```

```python
go_test(
    name = "x_test",
    size = "small",
    srcs = ["x_test.go"],
)
```

`go test ./...` passes and runs `TestMain` as a test. `bazel test //:x_test` fails to build with the error above.

### Why this is a bug

`go test` decides whether a function named `TestMain` is the package entry point from its **signature**, not from its name. Only `func TestMain(m *testing.M)` is the entry point that wraps the run; a `func TestMain(t *testing.T)` is an ordinary test that merely happens to be named `TestMain`. See [`cmd/go/internal/load/test.go`](https://github.com/golang/go/blob/master/src/cmd/go/internal/load/test.go), the `TestMain` case of `testFuncs.load`:

```go
case name == "TestMain":
	if isTestFunc(n, "T") {
		t.Tests = append(t.Tests, testFunc{pkg, name, "", false})
		continue
	}
```

`genTestMain` keys off the name alone ([generate_test_main.go](https://github.com/bazel-contrib/rules_go/blob/master/go/tools/builders/generate_test_main.go)):

```go
if fn.Name.Name == "TestMain" {
	// TestMain is not, itself, a test
	pkgs[pkg] = true
	cases.TestMain = fmt.Sprintf("%s.%s", pkg, fn.Name.Name)
	continue
}
```

so the `*testing.T` form is registered as the entry point and the generated main calls it with an `*testing.M`.

This matters beyond the toy case: it means a repository whose test suite is green under `go test` cannot be moved onto `go_test` without editing the test sources, which is exactly the property one relies on when migrating an existing project to Bazel. We hit it on a real repository during such a migration.

### Environment

Reproduces on `master` as of today. The code path has been keyed on the name since the file was introduced, so released versions are affected as well.

### Suggested fix

The `TestMain` check sits above the signature checks that are already made for `Test*`/`Benchmark*`/`Fuzz*`, so it has nothing but the name to go on. Moving it below them and taking that branch only when the sole parameter is `*<something>.M` is enough: a `TestMain` with a `*<something>.T` parameter then falls through to the existing `Test*` case and is registered as an ordinary test.

I have that change locally, along with a regression test under `tests/core/go_test` (a package whose only `TestMain` takes a `*testing.T`; it fails to build before the change and asserts the function runs as an ordinary test after it). Happy to open a PR if the approach looks right.

---

Drafted with Claude Code.

H5-O5:
### What's wrong

`go test` decides whether a function named `TestMain` is the package entry
point from its **signature**, not from its name: only `func TestMain(m
*testing.M)` is the entry point that wraps the run. A `func TestMain(t
*testing.T)` is an ordinary test that merely happens to be named `TestMain`
(see [`cmd/go/internal/load/test.go`](https://github.com/golang/go/blob/master/src/cmd/go/internal/load/test.go),
the `TestMain` case of `testFuncs.load`, which checks `isTestFunc(n, "T")`
before treating the function as a test).

`genTestMain` keys off the name alone, so a package containing `func
TestMain(t *testing.T)` has that function registered as the entry point and
the generated main calls it with an `*testing.M`:

```
cannot use m (variable of type *testing.M) as *testing.T value in argument to pkg.TestMain
```

Such a package builds and tests fine under `go test`, but fails to build
under rules_go — so `go_test` is unusable for it. We hit this on a real
repository while moving its `go test` suite onto Bazel: the package built
under `go test` and failed to compile under `go_test`, with an error that
points at generated code rather than at the user's source.

### The fix

The `TestMain` check sat *above* the signature checks that were already
being made for `Test*`/`Benchmark*`/`Fuzz*`, so it had nothing but the name
to go on. Move it below them and take that branch only when the sole
parameter is `*<something>.M`.

A `TestMain` with a `*<something>.T` parameter then falls through to the
`Test*` case and is registered as an ordinary test; any other signature is
skipped, as an unrecognized `Test*` shape already was. Nothing else moves —
the existing checks and the tolerance for an aliased `testing` import are
reused as they are.

### Test

Adds `tests/core/go_test:testmain_signature_test`: a package whose only
`TestMain` takes a `*testing.T`. It fails to build before this change, and
after it, asserts that the function is actually run as an ordinary test.
The `*testing.M` entry point stays covered by the existing
`only_testmain_test`.

### Verification

Both directions were checked with:

```
bazel test //tests/core/go_test:testmain_signature_test \
           //tests/core/go_test:only_testmain_test
```

- **With this change:** both targets `PASSED` — the new fixture builds and
  runs `TestMain(*testing.T)` as an ordinary test, and the existing
  `*testing.M` entry-point case is unaffected.
- **With `go/tools/builders/generate_test_main.go` reverted to its previous
  version and the new fixture kept:** `testmain_signature_test` is `FAILED TO
  BUILD`, with exactly the error this PR fixes —

  ```
  testmain.go:132:13: cannot use m (variable of type *testing.M) as *testing.T value in argument to l.TestMain
  ```

  confirming the fixture is a genuine regression test rather than one that
  would pass either way.

Fixes #4677.

---

Drafted with Claude Code.


jayconrod (approved):
Looks good, thanks!

jayconrod:
CI failures are unrelated to this. I'm seeing them on multiple repos. Reported upstream.

PR #4684, opened 2026-08-12
Preprocess PGO profiles in a separate action

fmeum:
**What type of PR is this?**

Bug fix

rules_go hands the pprof profile to the compiler as `-gcflags=all=-pgoprofile=<raw>`. That bypasses the preprocessing the `go` command performs for its own `-pgo` flag, so every compiler invocation parses the profile and builds the PGO call graph from scratch. `GoStdlib` runs one compiler per CPU inside a single action, and every `GoCompilePkg` action repeats the work, so peak memory scales with the core count of whatever machine the action lands on. A single `GoStdlib` action has been observed peaking at >60 GB.

This PR converts the profile once with `go tool preprofile` in a new `GoPreprofile` action and hands the result to every compile action. 

`go tool preprofile` was added in Go 1.23. Below that the profile is passed through untouched and behaviour is exactly as before.

**Other notes for review**


go/private/context.bzl
```diff
@@ -1057,12 +1063,49 @@ def cgo_context_data_impl(ctx):
         ),
     )
 
+def _preprocess_pgo_profile(ctx, toolchain, profile):
```
jayconrod: Could this go in `go/private/actions` as a new file? Better to keep all the actions in `actions`, and `context.bzl` is already huge.

jayconrod (approved):

dzbarsky (approved):
really nice find, thanks

PR #4691, opened 2026-08-14
link: add build settings to build info

sluongng:
**What type of PR is this?**

Feature

rules_go build information omits the configuration that produced a binary. `go version -m` and `runtime/debug` clients use these settings to distinguish binaries built for different modes, targets, and instrumentation.

Record deterministic settings already known to the link action. Use the Go 1.20 `debug.BuildSetting` and `BuildInfo.String` APIs, filter tags added internally for instrumentation, avoid deriving unknown `GOARM` defaults from the execution platform, and emit ARM64 and RISC-V tuning only for Go 1.23 or later.

This builds on #4690 and is deliberately limited to `BuildInfo.Settings`; main-module identity and VCS settings will be proposed separately.

Part of #3090.

**Other notes for review**

None.

### DIFF
--- a/go/tools/builders/buildinfo.go
+++ b/go/tools/builders/buildinfo.go
@@ -189,8 +189,119 @@ func buildInfoDeps(modules []moduleInfo) []*debug.Module {
 	return deps
 }
 
-func modInfoData(path string, modules []moduleInfo) string {
-	info := &debug.BuildInfo{Path: path, Deps: buildInfoDeps(modules)}
+func buildInfoSettings(buildmode string, buildTags []string, race, msan, cover, go123ArchSettings bool) []debug.BuildSetting {
+	return buildInfoSettingsFromEnv(buildmode, buildTags, race, msan, cover, go123ArchSettings, os.Getenv)
+}
+
+func buildInfoSettingsFromEnv(buildmode string, buildTags []string, race, msan, cover, go123ArchSettings bool, getenv func(string) string) []debug.BuildSetting {
+	if buildmode == "" || buildmode == "normal" {
+		buildmode = "exe"
+	}
+
+	settings := []debug.BuildSetting{
+		{Key: "-buildmode", Value: buildmode},
+		{Key: "-compiler", Value: "gc"},
+	}
+	if cover {
+		settings = append(settings, debug.BuildSetting{Key: "-cover", Value: "true"})
+	}
+	if msan {
+		settings = append(settings, debug.BuildSetting{Key: "-msan", Value: "true"})
+	}
+	if race {
+		settings = append(settings, debug.BuildSetting{Key: "-race", Value: "true"})
+	}
+	if buildTags = buildInfoTags(buildTags, race, msan); len(buildTags) > 0 {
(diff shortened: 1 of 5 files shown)


go/tools/builders/buildinfo.go
```diff
@@ -189,8 +190,117 @@ func buildInfoDeps(modules []moduleInfo) []*debug.Module {
 	return deps
 }
 
-func modInfoData(path string, modules []moduleInfo) string {
-	info := &debug.BuildInfo{Path: path, Deps: buildInfoDeps(modules)}
+func buildInfoSettings(buildmode string, buildTags []string, race, msan, cover, go123ArchSettings bool) []debug.BuildSetting {
+	return buildInfoSettingsFromEnv(buildmode, buildTags, race, msan, cover, runtime.GOARCH, go123ArchSettings, os.Getenv)
```
fmeum: Wouldn't the inclusion of `runtime.GOARCH` make it so that cross-compiled binaries are different from non-cross-compiled binaries? That doesn't seem desirable to me. Does native Go do this?

go/tools/builders/buildinfo.go
```diff
@@ -189,8 +190,117 @@ func buildInfoDeps(modules []moduleInfo) []*debug.Module {
...
+			return append(buildTags[:i], buildTags[i+1:]...)
+		}
+	}
+	return buildTags
+}
+
+func buildInfoArchSetting(goarch, hostGoarch string, go123ArchSettings bool, getenv func(string) string) (string, string) {
+	switch goarch {
+	case "386":
+		return "GO386", firstNonEmpty(getenv("GO386"), "sse2")
```
fmeum, line 260: Could you double-check that we don't inherit `--action_env` or anything non-hermetic here?

fmeum (commented):
The implementation looks reasonable, but in a bit worried that this will unnecessarily hurt caching. Is it really that useful to have these settings recorded in the binary given that they weren't produced by native Go?

@jayconrod I would be interested in your thoughts on this.

jayconrod (approved):
LGTM % @fmeum's comments.

@fmeum You mentioned cacheability. Are you worried about the one-time cost of making this change, or cache invalidation after all other changes? I think any change that causes this info to change will require all packages to be rebuilt anyway.

go/tools/builders/buildinfo.go
sluongng: We use an explicit `env=go.env` and does not enable `use_default_shell_env`, so Bazel does not pass those values to the action.

go/tools/builders/buildinfo.go
sluongng: removed that for now

go/tools/builders/buildinfo.go
fmeum: Is it possible that you forgot to push? I don't see the changes, but they sound good

sluongng: The runtime.GOARCH fix was already pushed in 854333b; this review is on the older 41c7f37. I have now also made use_default_shell_env = False explicit in c3bf5cc.

fmeum (approved):

PR #4694, opened 2026-08-19
link: add main module to build info

sluongng:
**What type of PR is this?**

Feature

rules_go records dependency modules in `BuildInfo.Deps`, but leaves
`BuildInfo.Main` empty even when package metadata identifies the module that
owns an executable. Build-info consumers therefore cannot report the main
module path and version.

This change passes two distinct values to the linker:

- The executable package's import path populates `BuildInfo.Path`.
- The Go PURL in its package metadata populates `BuildInfo.Main`.

Package metadata is kept unconditionally on `GoInfo` and archive data. At link
time, the target metadata is selected as the main module and that module path
is filtered from the candidate dependencies. This prevents a `go_test` binary
from reporting its own module in `BuildInfo.Deps` without a test-only
suppression flag or a duplicate provider field.

The new `go.binary` and `go.link` parameters are appended to preserve
positional compatibility for custom Starlark rules. Tests cover ordinary
binaries, embedded main packages, test binaries, binaries without metadata,
and direct `go.archive` to `go.link` callers.

**Which issue(s) does this PR fix?**

No linked issue.

**Other notes for review**

Build-info stack:

1. #4694 — add main-module build info (this PR)
2. #4695 — add stamped VCS build info

The second draft is cumulative against `master` because both heads are
fork-owned. It depends on this PR and is rebased after this PR changes.

Validation:

- `bazel --nohome_rc test //tests/...` — 467 passed, 1 Darwin-only test skipped
- `bazel test //go/tools/builders:all //tests/core/go_binary:buildinfo_test //tests/core/starlark:package_metadata_test //tests:buildifier_test`
- `bazel build //docs/...`
- `bazel test //docs/...`
- `git diff --check`

### DIFF
--- a/go/private/actions/binary.bzl
+++ b/go/private/actions/binary.bzl
@@ -35,11 +35,17 @@ def emit_binary(
         version_file = None,
         info_file = None,
         executable = None,
-        link_exec_group = None):
+        link_exec_group = None,
+        buildinfo_main_package_path = None,
+        buildinfo_module_metadata = None):
     """See go/toolchains.rst#binary for full documentation."""
 
     if name == "" and executable == None:
         fail("either name or executable must be set")
+    if buildinfo_main_package_path == None:
+        buildinfo_main_package_path = source.importpath
+    if buildinfo_module_metadata == None:
+        buildinfo_module_metadata = getattr(source, "_package_metadata", None)
 
     archive = go.archive(go, source)
     if not executable:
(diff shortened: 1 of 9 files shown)


go/private/actions/archive.bzl
```diff
@@ -238,4 +239,5 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         runfiles = runfiles,
         _headers = headers,
         _package_metadata_files = package_metadata_files,
+        _main_module_package_metadata = getattr(data, "_main_module_package_metadata", None),
```
fmeum: Do we have to emit this on every archive? Based on what it models I would think no.

fmeum (approved):

PR #4695, opened 2026-08-19
link: add stamped VCS build info

sluongng:
Stamped rules_go binaries cannot report their source revision or dirty
state through runtime/debug or go version -m. Users therefore cannot
apply the same provenance checks available for binaries built by cmd/go.
This matters because release verification depends on that provenance.

Map STABLE_VCS, STABLE_VCS_REVISION, STABLE_VCS_TIME, and
STABLE_VCS_MODIFIED workspace values to Go's vcs.* build settings. These
names follow Bazel's stable-status convention while mirroring Go's
build information keys, including the true or false value used by
vcs.modified. Normalize timestamps to UTC.

Keep the filtered stamp line-oriented. Bazel's workspace status contract
forbids keys or values from spanning lines, and the first-space
separator preserves every permitted value character. Cover spaces,
tabs, quotes, and backslashes with a round-trip test.

Own one filtered status action on each configured go_context_data
target and let eligible executable links consume its shared output. The
action graph is:

```text
stable-status.txt
        │
        ▼
one GoVCSStamp action
owned by configured //:go_context_data
        │
        ▼
go_context_data.vcsstamp
        │
        ├── eligible GoLink for binary A
        ├── eligible GoLink for binary B
        └── eligible GoLink for test C
```

This avoids declaring the same filter action for every binary while
keeping the raw stable-status file out of GoLink inputs, so unrelated
status changes do not relink binaries.

Emit VCS settings only when the target and the package metadata
selected for BuildInfo.Main both come from the main repository. Carry
the selected metadata file through embeds and tests, then infer
its provenance from File.owner at the link boundary. Infer target
provenance from the active Go context and treat unknown metadata
owners as untrusted.

Embedding VCS data in tests is intentional when stamping is explicitly
enabled, matching go test -buildvcs=true. Go's default -buildvcs=auto
instead omits VCS metadata from tests. Embedding the current revision
in each eligible stamped go_test binary changes that executable whenever
HEAD changes. Creating a commit, amending a commit message, or rebasing
therefore invalidates cached test results even when the test source did
not change. Compilation outputs remain reusable. Document this cost
and recommend keeping VCS stamping disabled for routine development
while enabling it for release or provenance-verification builds.

Cover shared action inputs, stable refreshes, status value preservation,
go_binary and go_test, direct links, external provenance, and
non-executable link modes.

### Future work

External binaries are intentionally out of scope for this change. A binary from a version-pinned `go_repository` continues to report its module version through `BuildInfo.Main.Version`, but this PR does not emit `vcs.*` settings for external-repository targets. A future change could add repository-specific VCS metadata for commit-pinned external repositories; it must come from that external repository rather than the main workspace's status file.


docs/go/core/defines_and_stamping.md
```diff
@@ -92,3 +92,54 @@ argument on the command line:
...
+When `go_binary` and `go_test` targets are linked with stamping enabled,
+`rules_go` also maps a small stable workspace status convention into Go build
+info. These settings are visible through `runtime/debug.ReadBuildInfo` and
+`go version -m`. They are only emitted when the linked target and its recorded
+main module both come from the main repository; stamped binaries built from
+external repositories, targets without main module metadata, and local wrappers
+around external main modules omit `vcs.*` settings.
+
+***Warning:*** avoid enabling VCS stamping for routine development test runs.
+The current commit hash becomes part of every eligible stamped `go_test`
```
fmeum, line 112: Does Go include provenance for test binaries? I wonder whether we should just disable any kind of build info embedding for Go tests?

docs/go/core/defines_and_stamping.md
```diff
@@ -92,3 +92,54 @@ argument on the command line:
...
+and is derived from the metadata file's owning label at link time.
+
+The workspace status names mirror Go's build info names, with Bazel's
+`STABLE_` prefix and underscores in place of dots.
+
+| Workspace status key | Go build info setting | Notes |
+| --- | --- | --- |
+| `STABLE_VCS` | `vcs` | Required and must be non-empty before any `vcs.*` settings are emitted. |
+| `STABLE_VCS_REVISION` | `vcs.revision` | Optional when non-empty. |
+| `STABLE_VCS_TIME` | `vcs.time` | Optional; must be `RFC3339Nano` and is normalized to UTC. |
```
fmeum:
```suggestion
| `STABLE_VCS_TIME` | `vcs.time` | Optional; must be in `RFC3339Nano` format and is normalized to UTC. |
```

go/tools/builders/buildinfo.go
```diff
@@ -163,6 +171,76 @@ func moduleFromPURL(purl string) (moduleInfo, bool, error) {
...
+	if err != nil {
+		return err
+	}
+
+	entries := buildInfoVCSEntries(stampMap)
+	var builder strings.Builder
+	for _, entry := range entries {
+		builder.WriteString(entry.key)
+		builder.WriteByte(' ')
+		builder.WriteString(entry.value)
```
fmeum, line 212: This won't happen in practice, but can newlines occur in values? Please either leave a comment why not or chose an encoding scheme that clearly handles all characters.

docs/go/core/defines_and_stamping.md
sluongng: I think the default -buildvcs=auto omits it, but -buildvcs=true adds it.
Since all of these are gated behind --stamp anyway, I treated it the same as `-buildvcs=true` explicit opt-in here.

unstamped builds won't be affected.

go/tools/builders/buildinfo.go
sluongng: Added a comment as the workspace status format currently forbids newline char in either key or value.

PR #4701, opened 2026-08-27
Bump golang.org/x/tools to v0.48.0 for Go 1.27 nogo compatibility

### LINKED ISSUE #4681: nogo is incompatible with Go 1.27
I was trying Go `1.27rc2` and ran into what's tracked in https://github.com/dominikh/go-tools/issues/1711 when running with nogo (except that it claimed 4 instead of 3 for the unsupported version. not sure why). Running the build with nogo disabled worked perfectly fine.

Bumping `golang.org/x/tools` to `v0.48.0` resolved the issue.

janantharaj:
**What type of PR is this?**
Bug fix

Currently, when using Go 1.27.0 in rules_go `v0.63.0`, nogo fails with the following error:

```
nogo: nogo: error running analyzers: 2 analyzers skipped due to type-checking error: external/gazelle++go_deps+com_github_aws_aws_sdk_go/internal/sync/singleflight/singleflight.go:9:8: could not import sync (internal error in importing "sync" (cannot decode "sync", export data version 4 is greater than maximum supported version 2); please report an issue)
```

This is caused by a downstream dependency on `x/tools`, which only supports `export data version 4`. Bumping to `v0.48.0` fixes the issue.

**Test**
1. I added a simple `local_path_override` with the changes in this PR to an external repository which depends on rules_go. Previously, I was encountering the error mentioned above, and now nogo successfully passes.
2. `bazel test //tests/core/nogo/...` passes.

Fixes #4681

**Other notes for review**


tests/bcr/MODULE.bazel
```diff
@@ -34,7 +34,7 @@ go_sdk.download(
     patches = [
         "//:test_go_sdk.patch",
     ],
-    version = "1.24.13",
+    version = "1.25.0",
```
janantharaj, line 37: I had to bump this to 1.25.0 to build successfully. This means that `rules_go` clients who wish to use nogo need at least this version of Go. Other `rules_go` features will still work with older versions of Go. Does that seem reasonable?

fmeum: Yes. While we avoid needless dep bumps, this clears the bar. Thanks for figuring this out!

go/private/rules/nogo.bzl
```diff
@@ -45,6 +49,16 @@ def _nogo_impl(ctx):
         ctx,
         maybe_needs_cc_toolchain = maybe_needs_cc_toolchain(ctx.attr, go_infos = ctx.attr.deps),
     )
+
+    # x/tools (nogo's analysis engine) uses APIs added in Go 1.25.
+    version = parse_version(go.sdk.version)
```
fmeum: Let's drop this. Folks can still override `x/tools` and we shouldn't add or own failure mode.

fmeum: Let's drop this. Folks can still override `x/tools` and we shouldn't add or own failure mode.

janantharaj: Thanks, that makes sense! Removed it.

fmeum (approved):

PR #4706, opened 2026-09-02
Apply the static attribute at link time instead of via a transition

fmeum:
**What type of PR is this?**

Bug fix

The `static` attribute of `go_binary` and `go_test` only affects how the binary itself is linked (`-extldflags -static` and the choice of external linking in `emit_link`), but `go_transition` turned it into a setting that every dependency was configured with. Consequences:

* A binary with `static = "on"` compiled a private copy of all of its dependencies, plus nogo runs for all of them.
* Two binaries that only differed in this attribute landed in two configurations.
* The bookkeeping settings `//go/config:static` and `original_static` leaked into every non-Go dependency reached from such a configuration, including the C++ toolchain and its runtime deps.

This drops `static` from the transition and instead reads the attribute in `go_context`, which exposes the effective value as `go.static_link` (the attribute applied on top of the `//go/config:static` flag). `emit_link` uses that instead of `go.mode.static`. `mode` is deliberately left alone since it describes how archives are compiled, and the archive consistency check compares it across dependencies.

Command-line uses of `--@io_bazel_rules_go//go/config:static` are unaffected, and Go tool binaries keep inheriting that value (#4689).

**Other notes for review**

### DIFF
--- a/docs/go/core/rules.md
+++ b/docs/go/core/rules.md
@@ -300,8 +300,8 @@ intermediate rule that allows users to apply these transitions.
 
 The '//go/config:static' and '//go/config:pure' settings are an exception: they are
 inherited from the value set on the command line, as they determine whether the tool
-can run on the execution platform. The 'static' and 'pure' attributes of an enclosing
-rule are still reset.
+can run on the execution platform. The 'pure' attribute of an enclosing rule is still
+reset.
 
 **ATTRIBUTES**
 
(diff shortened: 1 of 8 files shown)


jayconrod (approved):
It doesn't seem like `static` needs to propagate through Go compilation. Would it affect C/C++ dependencies? I guess not.

PR #4709, opened 2026-09-03
link: report only actually-reachable deps in build info

### LINKED ISSUE #4708: runtime/debug.BuildInfo().Deps should report actually imported packages instead of declared BUILD deps
v0.63.0, and `master` branch as of commit f6afe7478c645595839adc211644e6e5474d9fd3.

v0.51.3, and its `master` branch as of commit bazel-contrib/bazel-gazelle@48a1c5337c1c106ae9eedc5e1ae700925a61bc6f.

9.2.0.

Yes.

Linux (Ubuntu 24.04) x86_64, among others.

No.

### What did you do?
Built a `go_binary` (the same applies to `go_test)` that declares a dependency it never actually imports, and inspected the resulting binary's embedded build info at runtime.

`BUILD.bazel`:
```
go_library(
    name = "used",
    srcs = ["used.go"],
    importpath = "example.com/used",
)

go_library(
    name = "unused",
    srcs = ["unused.go"],
    importpath = "example.com/unused",
)

go_binary(
    name = "bin",
    srcs = ["main.go"],
    deps = [
        ":used",
        ":unused",  # declared in BUILD.bazel, never imported by main.go
    ],
)
```

`main.go`:
```go
package main

import (
    "fmt"
    "runtime/debug"

    _ "example.com/used"
)

func main() {
    info, _ := debug.ReadBuildInfo()
    for _, dep := range info.Deps {
        fmt.Println(dep.Path)
    }
}
```

Run `bazel run //:bin` and inspect the printed list.

`example.com/used` in `info.Deps`, and `example.com/unused` absent from it.

Vanilla `go build` computes `runtime/debug.BuildInfo().Deps` from the actual import graph, not from the full dependency graph recorded in `go.mod`: `cmd/go/internal/load/pkg.go`, function `(*Package).setBuildInfo`, starts a breadth-first walk from `p.Internal.Imports`, the packages the binary's own sources actually import, and only adds a module to `Deps` the first time one of its packages is reached by that walk.

A module required by `go.mod` but never imported by anything reachable from main never enters that walk and therefore never appears in `Deps`.
This is the contract `govulncheck` and SBOM/vulnerability-scanning tools rely on when they read `debug.ReadBuildInfo()` at runtime: `Deps` means "modules actually linked into this binary," not "modules this build was allowed to see."
Since `main.go` never imports `example.com/unused`, it should be absent.

### What did you see instead?
Both `example.com/used` and `example.com/unused` in `info.Deps`.

`rules_go` builds this list by walking package_metadata through the full declared `deps=` closure of the Bazel target instead of the import graph, which is a strict superset of what the compiled sources actually import (over-declared deps, deps kept for only some build configurations, a library that declares a dep it doesn't itself end up needing, etc.), so it systematically over-reports.

There is a second, sharper variant of the same root cause with `go_test`: a `go_test` rule compiles multiple archives (the internal test archive, the external test archive, and a synthetic `testmain`) that all share a single Bazel label, so any fix keyed by Bazel label instead of by Go package path conflates them and produces wrong results even after accounting for actual imports.

rdesgroppes:
**What type of PR is this?**
Bug fix

`runtime/debug.BuildInfo().Deps`, built by #4595 (thanks again!), walks the full declared Bazel `deps` graph instead of the packages actually imported.

This diverges from `cmd/go`'s own `(*Package).setBuildInfo` in `cmd/go/internal/load/pkg.go`, which walks only `p.Internal.Imports`.

The mismatch makes `debug.BuildInfo().Deps` over-report dependencies, producing false positives for `govulncheck` and SBOM/vulnerability scanners that trust it to mean packages actually linked into the binary.

Each compiled package now writes an `-imports` manifest listing the `packagePath`s its filtered sources actually import, the same set `checkImports` already computes for the compiler's `importcfg`.

`GoLink` walks this manifest graph by BFS from the main package and filters `package_metadata` down to the genuinely reachable set, instead of trusting the full declared closure.

Reachability is keyed by `packagePath` rather than Bazel label, since a single `go_test` rule compiles multiple archives, internal, external, and a synthetic `testmain`, that all share one label.

`packagePath` is unique per compiled package within a link, matching the identity the linker itself already relies on for deduplication via `depsSeen` in `buildImportcfgFileForLink`.

A `go_binary`'s own `packagePath` is purely cosmetic and not enforced to be unique, so it can collide with a dependency's real one.
The main archive's own imports are therefore passed directly via a new `-main_imports` flag instead of being looked up by that ambiguous key, and its entry is excluded from the general `-imports` list `GoLink` walks.
`TestReadBuildInfoCollidingImportpath` reproduces this collision.

`TestReadBuildInfoDeps` is updated to assert that a declared-but-unimported dependency no longer appears.

`TestGoTestBuildInfoDepsEmbeddedLib` and `TestBuildInfoDepsEmbeddedLib` are added to cover the stub `GoArchive` branch that `_recompile_external_deps` takes for a library that is embedded but not itself imported.

Fixes #4708.

**Other notes for review**
`emit_link` derives the `-package_metadata` and `-imports` flags from `archive.transitive` plus `test_archives`, excluding the main archive's own entry from `-imports` since its `packagePath` can collide with a dependency's.

The link action's `File` inputs come from a dedicated, recursively-built `_buildinfo_link_inputs` field on `GoArchive`, not from `archive.transitive`, since `depset`s support lazy merging of same-typed `depset`s but not lazy transformation into a different element type, so there is no way to extract `_imports`/`_package_metadata` `File`s from `GoArchiveData` elements without an eager pass.

Excluding the main archive from `-imports` uses a `map_each` closure over `archive.data.file`, with `allow_closure = True`.
`_format_imports` compares `File` objects directly (`d.file == main_archive_file`) rather than their `.path` strings, since `Artifact.equals()` compares `execPath` and `root` directly and is unaffected by Bazel's path mapping, unlike a `.path` string, which is only remapped when read from inside a `map_each` callback.

### DIFF
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -57,6 +57,10 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
 
     # store export information for compiling dependent packages separately
     out_export = go.declare_file(go, name = source.name, ext = pre_ext + ".x")
+
+    # store the packages actually imported for determining build-info reachability
+    out_imports = go.declare_file(go, name = source.name, ext = pre_ext + ".imports")
+
     out_cgo_export_h = None  # set if cgo used in c-shared or c-archive mode
 
     nogo = go.nogo
@@ -126,6 +130,7 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
             headers = headers,
             out_lib = out_lib,
             out_export = out_export,
+            out_imports = out_imports,
             out_facts = out_facts,
             out_diagnostics = out_diagnostics,
             out_nogo_validation = out_nogo_validation,
@@ -158,6 +163,7 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
             headers = headers,
             out_lib = out_lib,
             out_export = out_export,
+            out_imports = out_imports,
             out_facts = out_facts,
             out_diagnostics = out_diagnostics,
             out_nogo_validation = out_nogo_validation,
@@ -196,6 +202,7 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         _cxxopts = tuple(source.cxxopts),
         _clinkopts = tuple(source.clinkopts),
         _package_metadata = getattr(source, "_package_metadata", None),
+        _imports = out_imports,
 
         # Information on dependencies
         _dep_labels = tuple([d.data.label for d in direct]),
@@ -222,8 +229,16 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
     cgo_exports = depset(direct = cgo_exports_direct, transitive = [a.cgo_exports for a in direct], order = "preorder")
     package_metadata = getattr(data, "_package_metadata", None)
     package_metadata_files = depset(
-        direct = [package_metadata] if package_metadata else [],
+        direct = [data] if package_metadata else [],
         transitive = [getattr(a, "_package_metadata_files", depset()) for a in direct],
+    )
+    imports_files = depset(
+        direct = [data],
+        transitive = [getattr(a, "_imports_files", depset()) for a in direct],
+    )
+    buildinfo_link_inputs = depset(
+        direct = ([package_metadata] if package_metadata else []) + [out_imports],
+        transitive = [getattr(a, "_buildinfo_link_inputs", depset()) for a in direct],
     )
     return GoArchive(
         source = source,
@@ -238,4 +253,6 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         runfiles = runfiles,
         _headers = headers,
         _package_metadata_files = package_metadata_files,
+        _imports_files = imports_files,
+        _buildinfo_link_inputs = buildinfo_link_inputs,
     )
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -227,6 +227,11 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
     # to the archive only exposes a single header rather than combining all headers.
     cgo_exports_direct = [out_cgo_export_h] if out_cgo_export_h else []
     cgo_exports = depset(direct = cgo_exports_direct, transitive = [a.cgo_exports for a in direct], order = "preorder")
+    package_metadata = getattr(data, "_package_metadata", None)
+    buildinfo_link_inputs = depset(
+        direct = ([package_metadata] if package_metadata else []) + [data._imports],
+        transitive = [getattr(a, "_buildinfo_link_inputs", depset()) for a in direct],
+    )
     return GoArchive(
         source = source,
         data = data,
@@ -239,4 +244,5 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         cgo_exports = cgo_exports,
         runfiles = runfiles,
         _headers = headers,
+        _buildinfo_link_inputs = buildinfo_link_inputs,
     )
--- a/go/private/actions/archive.bzl
+++ b/go/private/actions/archive.bzl
@@ -227,19 +227,6 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
     # to the archive only exposes a single header rather than combining all headers.
     cgo_exports_direct = [out_cgo_export_h] if out_cgo_export_h else []
     cgo_exports = depset(direct = cgo_exports_direct, transitive = [a.cgo_exports for a in direct], order = "preorder")
-    package_metadata = getattr(data, "_package_metadata", None)
-    package_metadata_files = depset(
-        direct = [data] if package_metadata else [],
-        transitive = [getattr(a, "_package_metadata_files", depset()) for a in direct],
-    )
-    imports_files = depset(
-        direct = [data],
-        transitive = [getattr(a, "_imports_files", depset()) for a in direct],
-    )
-    buildinfo_link_inputs = depset(
-        direct = ([package_metadata] if package_metadata else []) + [out_imports],
-        transitive = [getattr(a, "_buildinfo_link_inputs", depset()) for a in direct],
-    )
     return GoArchive(
         source = source,
         data = data,
@@ -252,7 +239,4 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
         cgo_exports = cgo_exports,
         runfiles = runfiles,
         _headers = headers,
-        _package_metadata_files = package_metadata_files,
-        _imports_files = imports_files,
-        _buildinfo_link_inputs = buildinfo_link_inputs,
     )
--- a/go/private/actions/link.bzl
+++ b/go/private/actions/link.bzl
@@ -37,6 +37,12 @@ load(
 def _format_archive(d):
     return "{}={}={}".format(d.label, d.importmap, d.file.path)
 
+def _format_package_metadata(d):
+    return "{}={}".format(d.importmap, d._package_metadata.path)
+
+def _format_imports(d):
+    return "{}={}".format(d.importmap, d._imports.path)
+
 def emit_link(
         go,
         archive = None,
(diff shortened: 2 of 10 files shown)


go/private/actions/archive.bzl
```diff
@@ -222,9 +229,17 @@ def emit_archive(go, source = None, _recompile_suffix = "", recompile_internal_d
...
     package_metadata = getattr(data, "_package_metadata", None)
     package_metadata_files = depset(
-        direct = [package_metadata] if package_metadata else [],
+        direct = [data] if package_metadata else [],
         transitive = [getattr(a, "_package_metadata_files", depset()) for a in direct],
     )
+    imports_files = depset(
+        direct = [data],
+        transitive = [getattr(a, "_imports_files", depset()) for a in direct],
+    )
```
sluongng, line 238: this is similar to `GoArchive.transitive`, perhaps use it instead

go/private/rules/test.bzl
```diff
@@ -763,9 +763,17 @@ def _recompile_external_deps(go, external_go_info, internal_archive, library_lab
...
                 _headers = internal_archive._headers,
                 _package_metadata_files = depset(
-                    direct = [package_metadata] if package_metadata else [],
+                    direct = [arc_data] if package_metadata else [],
                     transitive = [getattr(a, "_package_metadata_files", depset()) for a in deps],
                 ),
+                _imports_files = depset(
+                    direct = [arc_data],
+                    transitive = [getattr(a, "_imports_files", depset()) for a in deps],
+                ),
```
sluongng, line 772: ditto

go/tools/builders/buildinfo.go
```diff
@@ -89,6 +89,67 @@ func validSemverIdentifiers(value string, rejectNumericLeadingZero bool) bool {
...
+		if err != nil {
+			return nil, fmt.Errorf("reading imports manifest %q: %w", path, err)
+		}
+		for _, line := range strings.Split(string(data), "\n") {
+			line = strings.TrimSpace(line)
+			if line != "" {
+				importsByPackage[pkg] = append(importsByPackage[pkg], line)
+			}
+		}
+	}
```
sluongng, line 114: I don't like how we are eagerly loading all the files into memory here.
perhaps we can load them lazily and cache them in-memory as we load?

sluongng (commented):
The problem statement makes sense, especially when you consider sometimes we just include files for a specific in-active build tag into `srcs`. The files will then gen excluded by the go compiler, but all of it's dependencies are still included in deps and thus, added to buildinfo.

I will have to give this PR another read soon (not enough time today). It's unclear to me if we still need the starlark provider information if we just gona propagate the "more accurate" import information through the GoCompilePkg outputs. Perhaps we still do for the version information, but what if we just store the version inside the `.imports` file so that we don't have to store it in skyframe memory? 🤔 

Also wonder if we need to pay attention to CGO deps anyhow. And if we have a nicer way of writing `checkImportsAndBuildCfg` and `checkImports`.

go/private/actions/archive.bzl
rdesgroppes: Right => 2nd commit drops `_package_metadata_files`/`_imports_files`/`_buildinfo_link_inputs` from `GoArchive` and `emit_link` now derives everything from the existing `archive.transitive` plus `test_archives` instead.

go/tools/builders/buildinfo.go
rdesgroppes: => also fixed in the 2nd commit, where `reachableModules` now reads a package's `-imports` manifest only when the BFS actually visits it, appending straight into the queue.

go/private/rules/test.bzl
rdesgroppes: => fixed in 2nd commit too.

go/tools/builders/buildinfo.go
```diff
@@ -89,6 +89,67 @@ func validSemverIdentifiers(value string, rejectNumericLeadingZero bool) bool {
...
+	importsPathByPackage, err := parseKeyedPaths(importsFiles)
+	if err != nil {
+		return nil, err
+	}
+
+	reachable := make(map[string]bool)
+	var queue []string
+	if mainPackage != "" {
+		queue = append(queue, mainPackage)
+	}
```
sluongng: codex found an edge case that may cause collision here

```
a wrapper binary can declare the same importpath as the library it imports:

  go_binary(
      name = "app",
      srcs = ["main.go"],  # Imports example.com/middle.
      importpath = "example.com/middle",
      deps = [":middle"],
  )

  go_library(
      name = "middle",
      srcs = ["middle.go"],  # Imports example.com/leaf.
      importpath = "example.com/middle",
      deps = [":leaf"],
  )

  Assume :leaf carries module metadata for example.com/leaf@v1.0.0. The actual import graph is:

  app's main package → example.com/middle → example.com/leaf

  The binary builds because its main package is compiled as "main", which is distinct from "example.com/middle". But the buggy manifest map uses the binary’s declared importpath, so both
  archives occupy the "example.com/middle" entry. The binary’s manifest replaces the library’s, and traversal never reaches leaf.
```

and the solution seems to be passing the `archive.data._imports` to the builder during emit_link, so that we can read that import manifest instead of using the mainPackage here.

go/private/actions/link.bzl
```diff
@@ -133,18 +141,16 @@ def emit_link(
...
-            for archive_data in test_archives
-            for metadata in [getattr(archive_data, "_package_metadata", None)]
-            if metadata
-        ],
-        transitive = [getattr(d, "_package_metadata_files", depset()) for d in archive.direct],
+    all_archive_data = depset(test_archives, transitive = [archive.transitive])
+    all_archive_data_list = all_archive_data.to_list()
+    buildinfo_link_inputs = depset(
+        [d._package_metadata for d in all_archive_data_list if d._package_metadata] +
+        [d._imports for d in all_archive_data_list],
```
sluongng: I suspect this could be rewritten using depset alone to avoid `.to_list()`

https://bazel.build/extending/depsets#performance

sluongng (commented):
my remaining nits

I think @fmeum  or @dzbarsky could comment more about the starlark performance impact, but the change looks good to me.

I would follow this up with a no-op PR to refactor checkImportsAndBuildCfg code, but it's probably not needed to be part of this PR.

go/tools/builders/buildinfo.go
rdesgroppes: => fixed in 3rd commit: `GoLink` now receives a new `-main_imports` flag with the main archive's own imports manifest, mirroring the existing `-main_module_metadata` flag, and `reachableModules` seeds its BFS from it directly instead of looking up the ambiguous `packagePath` string.

The main archive is also excluded from the general `-imports` list it used to contribute to, since its own entry could otherwise collide with a dependency sharing the same `packagePath`, which the direct-reference fix alone would not have caught.

Added `TestReadBuildInfoCollidingImportpath`, reproducing the reported scenario, verified red without the fix and green with it.

go/private/actions/link.bzl
rdesgroppes: => fixed in 3rd commit too: `_buildinfo_link_inputs` is revived as a `GoArchive` field, but only that one, since it is `File`-typed and cannot be derived from `.transitive`'s `GoArchiveData` elements without an eager pass, unlike `_package_metadata_files`/`_imports_files`, which stay gone.

Built recursively at each `emit_archive`/stub-branch call as before, `emit_link` now merges it in via `transitive=` instead of flattening `all_archive_data` for it.

I managed to replace the remaining `.to_list()` (I still had to filter the main archive out of the `-imports` flag) by a `map_each` closure over `archive.data.file.path`, with `allow_closure = True`, which is here safe because only a small string is captured, not a large analysis-phase object.

go/private/actions/link.bzl
```diff
@@ -133,18 +146,25 @@ def emit_link(
...
+        transitive = [getattr(archive, "_buildinfo_link_inputs", depset())],
     )
 
     builder_args.add_all(arcs, before_each = "-arc", map_each = _format_archive)
-    builder_args.add_all(package_metadata_files, before_each = "-package_metadata")
+    builder_args.add_all(all_archive_data, before_each = "-package_metadata", map_each = _format_package_metadata)
+    builder_args.add_all(
+        all_archive_data,
+        before_each = "-imports",
+        map_each = _format_imports_except(main_archive_path = archive.data.file.path),
```
fmeum: Please use an inline lambda that delegates to a top-level function. That makes the retained objects obvious and thus replaces the comment (which isn't accurate, the string is retained elsewhere).

fmeum (approved):

go/private/actions/link.bzl
rdesgroppes: => addressed in 4th commit: `_format_imports_except`'s closure-returning factory is replaced by a plain top-level `_format_imports(d, main_archive_path)`, called from an inline `lambda d: _format_imports(d, archive.data.file.path)` at the `map_each` call site.

The captured value is now visible directly at the call site instead of inside a separate function definition, so the arguable `allow_closure = True` justification comment is removed rather than reworded.

go/private/actions/link.bzl
```diff
@@ -133,18 +143,25 @@ def emit_link(
...
+        transitive = [getattr(archive, "_buildinfo_link_inputs", depset())],
     )
 
     builder_args.add_all(arcs, before_each = "-arc", map_each = _format_archive)
-    builder_args.add_all(package_metadata_files, before_each = "-package_metadata")
+    builder_args.add_all(all_archive_data, before_each = "-package_metadata", map_each = _format_package_metadata)
+    builder_args.add_all(
+        all_archive_data,
+        before_each = "-imports",
+        map_each = lambda d: _format_imports(d, archive.data.file.path),
```
fmeum: This (subtly) breaks with path mapping. Please pass in `archive.data` and apply `.file.path` in the callback function.

rdesgroppes: @fmeum Wow, today I learned!

Indeed, path mapping **only rewrites** the `.path`/`.dirname`/`.basename` strings a `File` returns **when accessed from inside a `map_each` callback**, via a `PathMapper` Bazel installs on that Starlark thread for the execution phase.

But it never touches the `File` object's own identity: `Artifact.equals()` compares `execPath` and `root` directly, so comparing `File`s with `==` sidesteps the remapping entirely, regardless of when or where the comparison happens, and is arguably the "right" level of abstraction.

Comparing the whole `GoArchiveData` struct would also work, since provider instances compare field-by-field, but `_format_imports` compares just the `File` to avoid the overhead of comparing every other field for no reason.

=> fixed in 5th commit: `_format_imports` now takes `main_archive_file` and checks `d.file == main_archive_file` instead of comparing `.path` strings.

fmeum (approved):

PR #4720, opened 2026-09-22
proto: keep go_proto_library codegen tools on the action's exec platform

ma-oli:
_go_proto_compiler used go_rule, which adds the optional CGO_TOOLCHAINS (C++) toolchain type. The compiler calls go_context(maybe_needs_cc_toolchain = False) and builds no cgo code itself, so it does not need a C++ toolchain, but requiring the type makes the rule's exec-platform resolution follow the cross-compiling C++ toolchain.

When a go_proto_library is built for a non-host target (e.g. GOOS=darwin on a macOS host driving a Linux RBE), that pins the exec-transitioned go-protoc/protoc/plugin tools to the host platform, while the GoProtocGen action (in the constraint-free internal_use_only_go_proto_gen exec group) runs on the first registered exec platform (the Linux RBE worker). The resulting host-built tools then fail to run there with 'Exec format error'.

Declare _go_proto_compiler as a plain rule requiring GO_TOOLCHAIN only (still carrying CGO_ATTRS/CGO_FRAGMENTS that go_context reads) so the tools resolve to the same platform as the action.


fmeum:
It looks like the proper fix would be to resolve https://github.com/bazel-contrib/rules_go/blob/e7c4c5c518faf07b34d0dfb702ab0e721623e650/proto/def.bzl#L225. Do you want to give that a try?

